PR #1, opened 2023-03-28
AST type definition

jjcnn:
The type definition for Miden IR ASTs.

Only `u32` and `bool` are currently allowed as types, but the definition is set up so that `u64` and other integer types can be added later. The cast expression has not yet been added, since there is nothing to cast to yet.

The definition represents an AST in SSA form except for block arguments/phi nodes. These require a liveness analysis, which should be added as a pass once we get to it.

### DIFF
--- a/ir/src/ast.rs
+++ b/ir/src/ast.rs
@@ -1,0 +1,95 @@
+//! The AST definition of Miden IR.
+//!
+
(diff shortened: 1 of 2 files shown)


jjcnn:
Closing, as this is no longer relevant.

PR #9, opened 2023-05-10
Build Pow2 opcode as unary op rather than binary op

jjcnn:
builder.rs now builds Pow2 operations as UnaryOps rather than as BinaryOps.


ir/src/hir/builder.rs
```diff
@@ -505,6 +504,7 @@ pub trait InstBuilder<'f>: InstBuilderBase<'f> {
     binary_int_op!(rotr, Opcode::Rotr);
     unary_numeric_op!(neg, Opcode::Neg);
     unary_numeric_op!(inv, Opcode::Inv);
+    binary_int_op!(pow2, Opcode::Pow2);
```
bitwalker, line 507: Did you mean to make this `unary_int_op!(pow2, Opcode::Pow2)`?

jjcnn: Ah, yes. Moving the code without changing it... Give me a minute.

PR #16, opened 2023-08-04
MASM code emitter

jjcnn:
I'm struggling to figure out how the emitter is supposed to fit in with the other stuff in the `midenc` directory and with `miden-diagnostics`. This draft PR is my current attempt, but I have no idea if it's anywhere close to what we need, and what the exact division of responsibilities is supposed to be between the various files.

For instance, the `Emitter` trait doesn't seem to use file names at all, so presumably the file buffer we need to write to is passed to the emitter from the outside?

### DIFF
--- a/tools/midenc/src/compiler/emitter.rs
+++ b/tools/midenc/src/compiler/emitter.rs
@@ -1,3 +1,4 @@
+use std::{fs::File, io::prelude::*, path::Path};
 use miden_assembly::ast::{self as masm};
 
 /// The emitter for MASM output.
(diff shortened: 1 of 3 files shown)


tools/midenc/src/compiler/emitter.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// Runs the emitter on the AST 
+    ///
+    /// Errors should be reported via the registered error handler,
+    /// Passes should return `Err` to signal that the pass has failed
+    /// and compilation should be aborted
+    fn run(&mut self, input: Self::Input) -> anyhow::Result<Self::Output> {
+        self.writer.write!("{}", input);
+    }
+
+    /// Implementation of Pass::chain
```
bitwalker, line 47: You don't have to implement `chain`, as it has a default implementation for all implementations of `Pass`. The inability to meaningfully chain the pass is implicit in the `Output` type of the pass, as a subsequent pass would necessarily have to have an `Input` type of `()`, which is essentially useless.

Writing the generated MASM to a file probably doesn't need to be a pass - though it could still be useful. We will need to have more control over how/where the MASM is written though, so it's probably better to have a dedicated type for the pass which carries all of the options which control that behavior, e.g. `MasmEmitter` or something like that.

bitwalker (commented):
As requested, I've left some notes about how to approach the implementation for this, let me know if you have any follow up questions!

As an aside, the code I'm merging (likely tomorrow) contains a couple examples of how to write `Display` formatters for code that I hacked together for use in debugging - those may come in handy if you want some examples of prior art of more complex `Display` implementations.

jjcnn:
@bitwalker: I think this is more like it.

The ASTs in the miden-vm crate implement the formatter we need, so in essence the emitter just calls that formatter. We need `MASMAst` to distinguish between programs and modules, though, but I'm not sure where to put it. For the time being I've put it in the emitter file, but it can be moved around.

I wasn't sure exactly how you want to do it wrt. the output file names, but I figure it's something along the lines of what is in `compiler/mod.rs` (the code is commented out because it needs the `artifacts` variable, which is also commented out).

bitwalker (approved):
LGTM, we'll link this up with the codegen pipeline after that gets merged

PR #17, opened 2023-08-23
Wasm -> Miden IR translation

greenhat:
This PR adds `miden-frontend-wasm` crate that translates Wasm to Miden IR.

The public API consists of the single `translate_module` function, which returns Miden IR `Module`. It's integrated into the compiler driver at `compiler/mod.rs`.

The diagnostics handler is passed down the stack and is used to report along with returning errors (via `unsupported_diag` macros).
The parsing and validation of Wasm code are done with `wasmparser` crate.

Unsupported Wasm features are reported as fatal errors. See `module_translator.rs` and `sections_translator.rs` for unsupported tag, and parts of imports section and unsupported types handling. Export and table sections are ignored since rustc almost always emits them.

Wasm module parsing is done in `module_translator.rs` with module sections parsing in `sections_translator.rs`. Wasm types conversion is done in `wasm_types.rs`.

The function code translation is done in `code_translator.rs`, where every Wasm instruction is translated via match.

The Wasm local variables are translated to Miden SSA variables. The SSA form is constructed by `SSABuilder` in `ssa.rs` and is used in `code_translator.rs` (see `Operator::Local*` handling) to translate Wasm locals to SSA variables.

`FunctionBuilderExt` is a wrapper around Miden IR `FunctionBuilder` and `SSABuilder`, providing additional API for dealing with mutable variables and SSA construction. Think of it as `FunctionBuilder` with SSA building capabilities. `FunctionBuilderExt` and `SSABuilder` do not have anything Wasm specific and can be merged into Miden IR `FunctionBuilder`.

The SSA form construction (`ssa.rs`) and control flow ops translation code (`func_translation_state.rs`, parts of `code_translator.rs`) were
extracted from Cranelift's internal Wasm -> CLIF translator [`cranelift-wasm`](https://github.com/bytecodealliance/wasmtime/tree/v11.0.0/cranelift/wasm) with minimal changes.

The Wasm -> Miden IR unit tests are in `module_translator/tests.rs` (module-level, control flow ops) and `code_translator/tests.rs` (per-instruction tests).
The test for the unsupported Wasm ops is in `code_translator/tests_unsupported.rs`.

The Rust -> Wasm -> Miden IR integration tests are in `tests/test_rust_comp.rs`.

What is implemented:
- Wasm locals translation to SSA variables;
- Wasm control flow ops translation;
- Wasm function calls translation;
- Translation for all Wasm spec v1 instructions that have Miden IR counterparts;

### TODO:
- rebase and adopt signed/unsigned integer types (implement signed ops from unsupported list); 
- `select` and `unreachable` ops translation;
- `br_table` translation;
- Globals translation;
- Use Miden's `MemGrow` op for Wasm `memory.grow` (what for `memory.size`?);
- Data section handling;
- translate `i32.wrap_64` to `trunc`;
- Remove all float ops translation and return unsupported error;
- more Rust programs for Rust -> Wasm -> Miden IR tests;

The list of unsupported (so far) Wasm v1 ops is in `code_translator/tests_unsupported.rs`. I'm duplicating it here for the convenience:
```
    CallIndirect { .. },
    // let (mut arg1, mut arg2, cond) = state.pop3();
    // if cond is zero returns arg2, else returns arg1
    Select,
    // Halt the program as it reached the point that should never be executed
    Unreachable,
    /**************************** Branch instructions *********************************/
    // cannot construct, it needs wasmparser's binary reader
    BrTable { targets: .. },
    /****************************** Memory Operators ************************************/
    F32Load { .. },
    F32Store { .. },
    /****************************** Nullary Operators ************************************/
    // Cannot construct since Ieee32 fields are private
    F32Const {
        value: Ieee32(0),
    },
    /****************************** Unary Operators ************************************/
    I32Ctz,
    I32Clz,
    I64Ctz,
    I64Clz,
    I32WrapI64,
    F32Sqrt,
    F64Sqrt,
    F32Ceil,
    F64Ceil,
    F32Floor,
    F64Floor,
    F32Trunc,
    F64Trunc,
    F32Nearest,
    F64Nearest,
    F32Abs,
    F64Abs,
    F32Neg,
    F64Neg,
    F64ConvertI64U,
    F64ConvertI32U,
    F64ConvertI64S,
    F64ConvertI32S,
    F32ConvertI64S,
    F32ConvertI32S,
    F32ConvertI64U,
    F32ConvertI32U,
    F64PromoteF32,
    F32DemoteF64,
    I64TruncF64S,
    I64TruncF32S,
    I32TruncF64S,
    I32TruncF32S,
    I64TruncF64U,
    I64TruncF32U,
    I32TruncF64U,
    I32TruncF32U,
    I64TruncSatF64S,
    I64TruncSatF32S,
    I32TruncSatF64S,
    I32TruncSatF32S,
    I64TruncSatF64U,
    I64TruncSatF32U,
    I32TruncSatF64U,
    I32TruncSatF32U,
    F32ReinterpretI32,
    F64ReinterpretI64,
    I32ReinterpretF32,
    I64ReinterpretF64,
    /****************************** Binary Operators ************************************/
    I32ShrS,
    I64ShrS,
    F32Add,
    F32Sub,
    F32Mul,
    F32Div,
    I32DivS,
    I64DivS,
    I32RemS,
    I64RemS,
    F32Min,
    F32Max,
    F32Copysign,
    F64Copysign,
    F64Add,
    F64Mul,
    F64Div,
    F64Min,
    F64Max,
    /**************************** Comparison Operators **********************************/
    I32LtS,
    I64LtS,
    I32LeS,
    I64LeS,
    I32GtS,
    I64GtS,
    I32GeS,
    I64GeS,
    F32Eq,
    F32Ne,
    F32Gt,
    F32Ge,
    F32Le,
    F32Lt,
    F64Eq,
    F64Ne,
    F64Gt,
    F64Ge,
    F64Le,
    F64Lt,
```

~I'm updating this list as I implement translations, so it's not final.~
I've reviewed all the Wasm v1 spec ops.


greenhat:
@bitwalker What is our plan for Wasm globals? I think I heard that Miden VM might get ops for handling global variables. Alternatively, we could translate globals into memory access ops.

bitwalker:
> What is our plan for Wasm globals? I think I heard that Miden VM might get ops for handling global variables. Alternatively, we could translate globals into memory access ops.

@greenhat I'm going to add globals to the IR, similar to what is provided by Cranelift. However, at runtime there is nothing special about globals, at least for now - they aren't in a different address space, and currently the Miden VM has no notion of globals, which is something I aim to change, but Bobbin and I have only had the briefest discussion about it so far. The compiler currently doesn't (and may never) handle some things traditionally handled by a linker, i.e. separate modules with the same global don't have any way to communicate that the global should be placed at the same address in memory at runtime. That's the main thing we need from Miden VM in order to support globals properly. For now though, we expect programs compiled from Rust to come in a single Wasm module, so we can punt on that a bit, but it's certainly something we'll need eventually.

Globals, and more generally heap memory, are tricky in another sense, because we'll need to think about how we handle the fact that Miden VM is not a byte-addressable architecture, so we either make that part of the IR, or we do some kind of post-processing of load/store ops when translating to MASM. We'll require that globals have a type, much like the rest of the IR to provide us with enough information to do such a transformation. So for now, I'm assuming that we will handle such target-specific details in the translation to MASM, rather than surface it in the IR, but if we find that there are benefits to pushing the responsibility for that to the frontend, then we can certainly discuss that.

greenhat:
> > What is our plan for Wasm globals? I think I heard that Miden VM might get ops for handling global variables. Alternatively, we could translate globals into memory access ops.
> 
> @greenhat I'm going to add globals to the IR, similar to what is provided by Cranelift. However, at runtime there is nothing special about globals, at least for now - they aren't in a different address space, and currently the Miden VM has no notion of globals, which is something I aim to change, but Bobbin and I have only had the briefest discussion about it so far. The compiler currently doesn't (and may never) handle some things traditionally handled by a linker, i.e. separate modules with the same global don't have any way to communicate that the global should be placed at the same address in memory at runtime. That's the main thing we need from Miden VM in order to support globals properly. For now though, we expect programs compiled from Rust to come in a single Wasm module, so we can punt on that a bit, but it's certainly something we'll need eventually.

Makes total sense.  

> Globals, and more generally heap memory, are tricky in another sense, because we'll need to think about how we handle the fact that Miden VM is not a byte-addressable architecture, so we either make that part of the IR, or we do some kind of post-processing of load/store ops when translating to MASM. We'll require that globals have a type, much like the rest of the IR to provide us with enough information to do such a transformation. So for now, I'm assuming that we will handle such target-specific details in the translation to MASM, rather than surface it in the IR, but if we find that there are benefits to pushing the responsibility for that to the frontend, then we can certainly discuss that.

The byte-addressabity issue was on my radar from the start but I've never given it a proper thought. I agree, the translation to MASM seems like the best place for it. I'll parse global definitions from the Wasm module, store them in the Miden module, and pass the global access ops to the IR.

greenhat:
@bitwalker I'm tracking the ops for which I did not find IR counterpart as unsupported in `code_translator/tests_unsupported.rs` and duplicating the list in this PR description for convenience. 

So far, the `Select` and `Unreachable` ops stand out from this list, which comprises mainly float ops that might be ignored for now. We should implement them since there is a high chance of encountering them.

hir/src/module.rs
```diff
@@ -515,6 +515,10 @@ impl<'m> ModuleFunctionBuilder<'m> {
         self.function.id
     }
 
+    pub fn func_builder(&mut self) -> FunctionBuilder {
```
greenhat, line 518: @bitwalker I added `FunctionBuilder` construction here since I heavily use it throughout the SSA form building and op translation.

hir/src/builder.rs
```diff
@@ -652,7 +652,8 @@ pub trait InstBuilder<'f>: InstBuilderBase<'f> {
             addr,
             ty: ty.clone(),
         });
-        into_first_result!(self.build(data, ty, span))
+        let addr_ty = self.data_flow_graph().value_type(addr).clone();
```
greenhat, line 655: @bitwalker I'm passing "full" `Type::Ptr` here because it's expected in `Opcode::results()`.

hir/src/module.rs
bitwalker: That's fine, we may want a trait for this actually, since the `ModuleFunctionBuilder` and `FunctionBuilder` should be basically identical, just used in slightly different circumstances (the former is really just for initial construction of a function as part of a module, the latter is a more general utility for constructing/modifying functions, but they should have essentially the same API).

greenhat: Got it. 
Speaking of "function builders". As I mentioned in the PR description,  I created `FunctionBuilderExt` as a wrapper around Miden IR FunctionBuilder and SSABuilder, providing additional API for dealing with mutable variables and SSA construction. I think of it as `FunctionBuilder` with SSA building capabilities. `FunctionBuilderExt` and `SSABuilder` do not have anything Wasm-specific and can be merged into the unified function-building API.

hir/src/builder.rs
bitwalker: We should probably just fix `results` actually, I don't think there's much benefit in special casing the loads with regard to the controlling type in any case.

As an aside, the way loads are handled is going to change, to distinguish loads of "native" pointers (in a word-addressable address space), and aligned vs unaligned loads of pointers in the default byte-addressable address space. Aligned vs unaligned here refers to whether the address being loaded is aligned to a word or field element boundary (32-byte or 8-byte alignment respectively), in addition to whether the load can be guaranteed to not cross a word boundary. If those properties hold, one can use the aligned loads, which are much more efficient. If not, unaligned loads must be used, which perform extra checks and handle the case of loading across a word boundary.

Those changes fall out from the fact that Miden VMs address space is word-addressable, i.e. each unique address refers to an entire word (4 field elements). There are two ops in MASM which allow addressing the first field element of each word, and I'm working through a proposal to extend those to allow addressing any field element of the word, which would make the Miden VM address space both word- and felt-addressable. In either case though, we have to map our byte-addressable IR to this paradigm, and since we only have the ability to load/store words or field elements, we have to handle cases where a value spans one or both of those. Aligned loads can be made much more efficient because if we know an address is word-aligned, and the type is <= the size of a word (or felt-aligned, <= the size of a felt), we can fetch the value with a single instruction. Unaligned loads require us to figure out whether we can load data from a pair of felts in a single word, or a pair of words, and then shift the values so that they are aligned on the operand stack.

As an additional upcoming restriction, the maximum size of a value that I think we'll allow loads/stores for is <= a single word. We could support more than that, but we can't access more than 16 elements on the operand stack, and using more than a word for temporaries leaves little room for other live values. I would limit it further to single field elements, but there are a number of instructions in Miden that operate on words, so we will necessarily need to permit loading/storing words for use with those instructions. It remains to be seen how common it is for there to be more than 16 slots of the operand stack in use at once.

hir/src/module.rs
bitwalker: Definitely, those would be welcome additions to the builders! A lot of what is there now is just the bare minimum needed to represent the various instructions, as well as enough additional utilities for some of the tests we've written thus far, but there is definitely a lot of room for improvement as you've noted.

hir/src/builder.rs
greenhat: Makes sense. Thanks for the detailed explanation! I saw your proposal earlier today in miden-vm repo.

greenhat:
Continued as #22

PR #21, opened 2023-09-07
feat: implement inline assembly

bitwalker:
This PR expands upon the drafted inline assembly functionality, and in the process adds some useful pieces of functionality that are used during code generation. I've inlined the main commit message below which goes into more detail.

There are two motivations for this currently:

1. The code generator work I'm introducing builds on some of this functionality, e.g. the representation of local variables, the MASM instruction set, and the operand stack data structure.
2. Some higher-level instructions in the IR require more complex sequences of Miden Assembly - to facilitate that, we will define a baked-in `intrinsics` module that exposes procedures written using inline assembly that provide the functionality needed for those instructions. This keeps the resulting MASM clean, while still allowing us to keep those instructions as efficient as possible.

@jjcnn You may notice the changes to the textual output of the IR with regard to inline assembly. You can feel free to omit support for inline assembly entirely from the parser for the time being, as I don't think it is essential at all to support it there for now, however I did provide a short example of what you can expect the syntax to look like in practice, below. Adding support for inline assembly to the parser will be more effort than its worth for now though IMO.

---

This commit fleshes out the implementation of inline assembly that we
had essentially stubbed out prior to this. The following are provided as
part of this implementation:

* A representation of the Miden Assembly instruction set, in crate,
  with the ops we intend to support in the near term. This is used to
  represent the contents of inline assembly, and is also used in the
  upcoming codegen backend as the target IR for Miden Assembly, as it
  is a flatter representation similar to Miden IR. The intent is to also
  implement a tiny evaluator for this subset of Miden Assembly for use
  in tests without having to emit MASM files, and run the Miden VM which
  is much more expensive. It also allows us to track the subset of Miden
  Assembly which we support in the compiler, as well as represent
  experimental changes that have not yet been implemented upstream.
* `MasmBuilder`, which is constructed from an `InstBuilder` via `inline_asm`,
  and plays a similar role to `FunctionBuilder`+`InstBuilder`, but in
  terms of Miden Assembly, in the context of constructing a single inline
  assembly block.
* Improved pretty printing of inline assembly. Now, an inline assembly
  block is printed like so (this block adds the two inputs, and squares
  the result, which is returned as the result of the inline assembly
  block):

  ```hir
  asm (v1, v2) {
      add
      mul.2
  } : felt
  ```
* A generic data structure for use in representing the operand stack,
  and the common operations which are performed on it. This is used to
  validate inline assembly, as well as keep track of values during code
  generation. It will also likely be used for whatever evaluator we
  throw together for use in testing.


hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+    pub op: Opcode,
+    /// Arguments on which the inline assembly can operate
+    ///
+    /// The operand stack will be set up such that the given arguments
+    /// will appear in LIFO order, i.e. the first argument will be on top
+    /// of the stack, and so on.
+    ///
+    /// The inline assembly will be validated so that all other values on
+    /// the operand stack below the given arguments will remain on the stack
+    /// when the inline assembly finishes executing.
```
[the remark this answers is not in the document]
jjcnn: I just noticed that the behavior I describe is indeed allowed, but it's only described in the comments for the builder functions. 

I think it would be a good idea to flesh out these kinds of assumptions either in the comments of `InlineAsm`, or in separate documentation. People will be looking for those descriptions there, not in the builder function comments.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+        let span = self.span;
+        let data = Instruction::InlineAsm(self.asm);
+        self.builder.build(data, ty, span)
+    }
+}
+
+/// Used to construct a single MASM opcode
+pub struct MasmOpBuilder<'a> {
+    asm: &'a mut InlineAsm,
+    stack: &'a mut OperandStack<Type>,
```
jjcnn, line 260: I'm not sure of the purpose of the `stack` field. In particular, I'm not sure why the stack contains `Type` objects. Please add a comment.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+        self.stack.movdnw(idx);
+        self.asm.push(self.ip, MasmOp::Movdnw(idx as u8));
+    }
+
+    /// Pops a boolean element off the stack, and swaps the top two elements
+    /// on the stack if that boolean is true.
+    ///
+    /// Traps if the conditional is not 0 or 1.
+    pub fn cswap(self) {
+        self.stack.drop();
```
jjcnn, line 386: I get that this keeps track of the stack height, but if the stack elements are `Type` objects, then how do you keep track of the elements when the swap is conditional?

AFAICT there is no way to keep track of that, so it should be made clear exactly what information we are keeping track of (is it just the stack height? If so, why are the stack elements of type `Type`?).

The same question applies to all the conditional ops.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+    pub fn if_true(self, then_blk: MasmBlockId, else_blk: MasmBlockId) {
+        self.stack.drop();
+        self.asm.push(self.ip, MasmOp::If(then_blk, else_blk))
+    }
+
+    /// Pops a boolean value from the stack, and executes the given block if it is true,
+    /// otherwise it is skipped. The given block will continue to execute for as long as
+    /// the top value on the stack at the end of the block is true.
+    pub fn while_true(self, body: MasmBlockId) {
+        self.stack.drop();
```
jjcnn, line 549: In general we can't assume that the stack height will be the same at this program point in every iteration of the loop, so simply dropping the conditional value doesn't look correct.

My guess is that you intend to make it a requirement that each program point has a fixed stack height, but that requirement is not specified anywhere, and it puts a limitation on the inline assembly that the programmer can write, so it should be documented somewhere.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+        self.asm.push(self.ip, MasmOp::While(body));
+    }
+
+    /// Repeatedly executes `body`, `n` times.
+    pub fn repeat(self, n: u8, body: MasmBlockId) {
```
jjcnn, line 554: Same comment as above.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+        self.stack.dropn(2);
+        self.asm.push(self.ip, MasmOp::AssertEq);
+    }
+
+    /// Pops the top two words on the stack, and traps if they are not equal.
+    pub fn assert_eqw(self) {
+        self.stack.dropn(8);
+        self.asm.push(self.ip, MasmOp::AssertEq);
+    }
+
```
jjcnn, line 440: There doesn't seem to be a builder function for the `LocAddr` operation.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+    pub fn storew_imm(self, addr: u32) {
+        self.stack.drop();
+        self.asm.push(self.ip, MasmOp::MemStorewImm(addr));
+    }
+
+    /// Pops a boolean value from the stack, and executes the first block if it is true,
+    /// otherwise the second block.
+    pub fn if_true(self, then_blk: MasmBlockId, else_blk: MasmBlockId) {
+        self.stack.drop();
+        self.asm.push(self.ip, MasmOp::If(then_blk, else_blk))
```
jjcnn, line 542: Similar comment to the one for `while_true`: How do you keep track of the stack height at the join point between the branches?

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+        self.asm.push(self.ip, MasmOp::Exec(id));
+    }
+
+    /// Executes the named procedure as a syscall.
+    pub fn syscall(self, id: FunctionIdent) {
+        self.asm.push(self.ip, MasmOp::Syscall(id));
+    }
+
+    /// Pops two field elements from the stack, adds them, and places the result on the stack.
+    pub fn add(self) {
```
jjcnn, line 569: Shouldn't this affect the stack as well?

Same comment for all the binary arithmetic operations.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+    ///
+    /// * `Overflow::Unchecked` - the addition is performed using the `add` op for field elements, which may
+    /// produce a value that is outside of the u32 range, it is the callers responsibility to ensure that the
+    /// resulting value is in range.
+    /// * `Overflow::Checked` - the operation will trap if either operand, or the result, is not a valid u32
+    /// * `Overflow::Wrapping` - computes the result as `(a + b) mod 2^32`, behavior is undefined if either operand
+    /// is not a valid u32
+    /// * `Overflow::Overflowing` - similar to above, the result is computed as `(a + b) mod 2^32`, however a boolean
+    /// is also pushed on the stack after the result, which is 1 if the result of `a + b` overflowed, else 0.
+    ///
```
jjcnn, line 820: This comment is duplicated several times, and really belongs in the `Overflow` type (where I assume it already exists - I haven't checked).

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+    const DEFAULT: Self;
+}
+impl StackElement for Felt {
+    const DEFAULT: Self = Felt::ZERO;
+}
+impl StackElement for Type {
+    const DEFAULT: Self = Type::Felt;
+}
+
+pub trait Stack: std::ops::IndexMut<usize, Output = <Self as Stack>::Element> {
```
jjcnn, line 2047: There's an analysis going on there, but it's not clear to me what the analysis is. Is it just keeping track of the stack height?

jjcnn (changes requested):
This PR ought to be split into at several smaller PRs. At least the following spring to mind: 
- One that introduces locals.
- One that introduces Masm operations and their builders.
- One that introduces the stack simulator/Masm evaluator.
- Possibly a separate one for the actual inline assembly blocks.

jjcnn (commented):
I accidentally pressed 'Submit review' before I had finished my review of the stack simulator, so there's one more review comment.

bitwalker:
Thanks for the quick turnaround! I'm going through your notes now, and should have this turned back around shortly.

hir/src/locals.rs
```diff
@@ -0,0 +1,86 @@
+use std::{alloc::Layout, fmt};
+
+use super::Type;
+
+/// A strongly typed identifier for referencing locals associated with a function
```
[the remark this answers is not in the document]
bitwalker: Locals are implicitly created by use of the `alloca` instruction, which returns the address of a `Local` as a `Value`, which can then be used with the `load` and `store` instructions to read/write it. There is no way to directly create a `Local`, aside from using `alloca`. We don't actually reify a `Local` until code generation though, since it isn't necessary to represent in HIR, however they have a more important role in MASM.

hir/src/asm.rs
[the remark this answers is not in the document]
bitwalker: Inline assembly blocks can support zero or more results, much like function calls. The output of an inline assembly block is given by the result type, which in the text format starts after the trailing `:`. For now though, that's specified using a single `Type` via the builders, so representing zero, one, or more results requires using an appropriate `Type`, e.g. `Type::Unit`, `Type::Felt`, or `Type::Array(Box::new(Type::Felt), n)` for 0, 1, or n-ary results.

> I think it would be a good idea to flesh out these kinds of assumptions either in the comments of InlineAsm, or in separate documentation. People will be looking for those descriptions there, not in the builder function comments.

Could you clarify what assumptions you are referring to here? `InlineAsm` doesn't impose any restrictions on the number or type of results on its own, and what restrictions exist are imposed by the builders (either the `inline_asm` of `InstBuilder`, or by `MasmBuilder`), and their docs reflect those restrictions (unless I missed one somewhere). In general, it is expected that users of the IR (i.e. compiler frontends) are going to work through the builders, and not need to be aware of `InlineAsm` unless they are doing something low-level with the IR, I would be very surprised if the first thing someone looked at was that struct, vs the docs associated with the builders.

I'm hesitant to go too far down the path of separate high-level documentation at this stage, there is still a lot of instability in the IR that remains to be ironed out, and high-level docs are particularly prone to falling out of date, and can be difficult to maintain under a stream of constant change. I'm as keen to get those kind of docs in place as you are, but I want to focus on stabilizing the foundation first, then start building stuff on it.

That said, if you find that there are aspects of the implementation that are undocumented, I would like to make sure that is addressed. I will expand on the `InlineAsm` docs a bit to provide at least a little more overview on it's structure and usage.

hir/src/asm.rs
bitwalker: What is documented here (and the default implementation) is what happens to the operand stack in the VM when that operation is performed. When using a type stack, we're tracking that behavior in the abstract, and we are only primarily concerned with the stack height. I should note that we are only ever pushing `Type::Felt` in this case, so it doesn't actually matter what the conditional or the operands are, as it doesn't change the semantics in the slightest.

If we were to expand the use of type stacks to something more general purpose, then we'd likely override the implementation of certain functions in the trait, to tailor them to the semantics of a type checker (i.e. validating that the operands of an operation like `cswap` are the same). The default implementation simply performs the necessary stack maintenance.

hir/src/asm.rs
bitwalker: Likewise, I'm going to mark this one as resolved since both of these functions are discussed together under `while_true`

hir/src/asm.rs
bitwalker: That's intentional, I don't intend to support accessing locals from inline assembly at this time. If someone wants this, they can use `alloca` to allocate a local in the IR, and then pass the address as an argument to the inline assembly block.

hir/src/asm.rs
bitwalker: Same deal as with `while_true`, with the same general approach to addressing it. Tthis is more generally handled as part of stackification, but when it comes to constructing inline assembly, we probably want to add some additional validation to catch the mistake earlier when defining the assembly, rather than during codegen, just like with the loops.

hir/src/asm.rs
bitwalker: Ah indeed, missed adding the stack manipulation to a bunch of the operators.

hir/src/asm.rs
bitwalker: It doesn't, I can move these docs there though.

hir/src/asm.rs
bitwalker: This trait isn't doing any particular kind of analysis, aside from just generally representing what happens to the operand stack when the stack manipulation ops are applied. It is up to a specific implementation of this trait to define the specific behavior though.

That said, I'll add some documentation to the trait to describe it's default behavior. It is primarily meant to take care of the common plumbing of the operand stack, but some ops, like `cswap` and such, are more sensitive to the specific behavior desired. I may remove the default implementation for those, or move them to the concrete implementation(s), since they are less general.

hir/src/asm.rs
```diff
@@ -0,0 +1,2447 @@
...
+    type Element: StackElement;
+
+    fn storage(&self) -> &Vec<Self::Element>;
+    fn storage_mut(&mut self) -> &mut Vec<Self::Element>;
```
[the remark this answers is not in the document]
bitwalker: Hmm, I suppose we could call them `stack` and `stack_mut`, though that's not really correct semantically speaking. I don't have strong feelings either way.

bitwalker:
@jjcnn For Monday, I've updated the PR with a number of changes in response to your feedback. Aside from splitting up the `asm` module as requested, the most significant is related to the handling of the `exec`, `syscall`, `if.true`, `while.true`, and `repeat` builders. Most of the rest are small tweaks, doc additions/improvements. I've also added a test which demonstrates constructing a function with inline assembly that includes a loop; both to validate the API, and to test that it does what is expected. Each change is split into its own commit to make viewing them atomically easier.

I think it's worth putting a little additional time into polishing the handling of types in the `MasmOpBuilder` so that certain operations preserve type information in some edge cases (i.e. if you use `add_u32` on a `Type::Ptr` value, it turns it into a `Type::U32`, that could potentially cause the operand stack consistency check to fail in some situations). I'd also like to ensure that `cswap` and `cdrop` operations require their operand types to be the same for a similar reason. I'll probably do that Monday, but that won't really change much in this PR.

Let me know if you have any further questions/feedback!

hir/src/asm.rs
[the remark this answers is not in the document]
jjcnn: I'm not talking about fully fledged documentation at this stage. What I'm talking about at this stage is much more low-level, and could be as simple as an unsorted list of design decisions that have been made along the way, and which could eventually be collated into actual user-facing documentation.

The alternative is to remember everything in our heads, and in my experience that always leads to problems later on.

hir/src/asm.rs
jjcnn: I'm taking the liberty of unresolving this conversation to make a point:

Arguments along the lines of "it makes little sense for anyone to write such code" scare me in the context of smart contracts, because the code we're compiling may end up handling billions of dollars worth of assets, and any undefined behavior can potentially be used by malicious actors.

In other words, the space for what code makes sense to write is very broad indeed, and is certainly not limited to code that has well-defined behavior.

hir/src/asm/builder.rs
```diff
@@ -404,24 +404,99 @@ impl<'a> MasmOpBuilder<'a> {
...
+        }
     }
 
-    /// Repeatedly executes `body`, `n` times.
-    pub fn repeat(self, n: u8, body: MasmBlockId) {
-        self.asm.push(self.ip, MasmOp::Repeat(n, body));
+    /// Begins construction of a `repeat` loop, with an iteration count of `n`.
+    ///
+    /// A `repeat` instruction requires no operands on the stack, and will execute the loop body `n` times.
+    ///
```
jjcnn, line 482: Suggested addition: "The abstract state of the operand stack must be consistent when the loop body is entered, as well as when the loop is completed."

hir/src/asm/builder.rs
```diff
@@ -1345,3 +1420,327 @@ impl<'a> MasmOpBuilder<'a> {
...
+    /// In effect, when the call to `while_true` or `repeat` returns, the operand
+    /// stack in the parent builder will look as if the loop instruction has finished
+    /// executing.
+    in_stack: &'a mut OperandStack<Type>,
+    /// This is the operand stack state within the loop.
+    ///
+    /// Upon finalization of the loop instruction, this state is used to validate
+    /// the effect of the loop body on the operand stack. For `repeat`, which is
+    /// unconditionally entered, no special validation is performed. However, for
```
jjcnn, line 1695: Shouldn't `repeat` loops be validated as well? The stack could grow with each iteration, no?

hir/src/asm.rs
bitwalker: > Arguments along the lines of "it makes little sense for anyone to write such code" scare me in the context of smart contracts, because the code we're compiling may end up handling billions of dollars worth of assets, and any undefined behavior can potentially be used by malicious actors.

Let me be more precise. We aren't responsible for someone writing MASM by hand that throws a billion dollars into the equivalent of a crypto incinerator. That is entirely the responsibility of the person writing the code. Inline assembly is by definition an escape hatch, you are being given virtually unlimited power to shoot yourself in the foot however you like. Undefined behavior isn't saying "do this at your own risk", it's a declaration that you _should never do that thing_. Indeed Miden already has a number of instructions for which undefined behavior is specified - it doesn't mean we don't know what happens when, e.g., an unchecked u32 operation overflows past 2^32; it means we don't know how that will affect the rest of the program.

In any case, the issue raised here is moot - the purpose of the loop/if builders that have been added is to validate the inline assembly upon construction, and at least catch if the operand stack was improperly maintained in a way that breaks assumptions that the rest of the compiler makes about the behavior of a given inline assembly block (i.e. the number and types of elements are mismatched across different exit paths in the control flow graph). But we can't make any guarantees about the behavior of the inline assembly itself - it might _look_ like things are in sync, but it may very well be that the author has the same logical value in two different positions on the operand stack along one path vs another. We can't do anything about that.

However, if _we_ introduce undefined behavior, i.e. we lower something with well-defined semantics, such as checked math, to instructions which do not preserve those semantics, and may lead to undefined behavior - that's a different thing altogether. We are responsible for preserving the semantics and behavior that we document, no more, no less. If we document that use of a specific instruction requires certain guarantees on its use, and that failing to uphold those guarantees is undefined behavior, that isn't an issue or a bug, that is expected for a compiler IR.

hir/src/asm/builder.rs
bitwalker: `repeat.n` can't produce dynamic (i.e. not statically analyzable) changes to the operand stack, it is statically expanded to `n` copies of the `repeat` body. In other words, if you make a modification to the stack, we know exactly how many times that modification will occur, so we can model the state of the stack across every iteration, and afterwards.

That said, I agree that some additional validation is needed here:

* We want to ensure that, given _S_ elements on the operand stack upon entry to the loop, that the top _S_ elements on the stack along the loopback edge match those that were on the stack on entry, for all `n` iterations.
* We want to ensure that if each iteration removes elements from the stack, that all `n` iterations can be performed without performing an invalid stack operation.

I'll do that before we merge this.

jjcnn (approved):
Approving under the assumption that the `repeat` issue is fixed.

PR #22, opened 2023-09-09
Wasm -> Miden IR translation

greenhat:
New PR in place of accidentally closed #17.

This PR adds `miden-frontend-wasm` crate that translates Wasm to Miden IR.

The public API consists of the single `translate_module` function, which returns Miden IR `Module`. It's integrated into the compiler driver at `compiler/mod.rs`.

The diagnostics handler is passed down the stack and is used to report along with returning errors (via `unsupported_diag` macros).
The parsing and validation of Wasm code are done with `wasmparser` crate.

Unsupported Wasm features are reported as fatal errors. See `module_translator.rs` and `sections_translator.rs` for unsupported tag, and parts of imports section and unsupported types handling. Export and table sections are ignored since rustc almost always emits them.

Wasm module parsing is done in `module_translator.rs` with module sections parsing in `sections_translator.rs`. Wasm types conversion is done in `wasm_types.rs`.

The function code translation is done in `code_translator.rs`, where every Wasm instruction is translated via match.

The Wasm local variables are translated to Miden SSA variables. The SSA form is constructed by `SSABuilder` in `ssa.rs` and is used in `code_translator.rs` (see `Operator::Local*` handling) to translate Wasm locals to SSA variables.

`FunctionBuilderExt` is a wrapper around Miden IR `FunctionBuilder` and `SSABuilder`, providing additional API for dealing with mutable variables and SSA construction. Think of it as `FunctionBuilder` with SSA building capabilities. `FunctionBuilderExt` and `SSABuilder` do not have anything Wasm specific and can be merged into Miden IR `FunctionBuilder`.

The SSA form construction (`ssa.rs`) and control flow ops translation code (`func_translation_state.rs`, parts of `code_translator.rs`) were
extracted from Cranelift's internal Wasm -> CLIF translator [`cranelift-wasm`](https://github.com/bytecodealliance/wasmtime/tree/v11.0.0/cranelift/wasm) with minimal changes.

The Wasm -> Miden IR unit tests are in `module_translator/tests.rs` (module-level, control flow ops) and `code_translator/tests.rs` (per-instruction tests).
The test for the unsupported Wasm ops is in `code_translator/tests_unsupported.rs`.

The Rust -> Wasm -> Miden IR integration tests are in `tests/test_rust_comp.rs`.

The list of unsupported (so far) Wasm v1 ops is in `code_translator/tests_unsupported.rs`. See the list in `UNSUPPORTED_WASM_V1_OPS`.
I've reviewed all the Wasm v1 spec ops.

### TODO:
- make issue for fixing using `popcnt` for `ctz`, `clz` translation (used to unblock dlmalloc translation);
- dlmalloc test; figure out why hash in mangled function names so unstable or even better - add an option to demangle and remove it;


greenhat:
@bitwalker Following up on https://github.com/0xPolygonMiden/miden-ir/pull/18#issuecomment-1709471866

>I've pushed a commit that expands our type system with signed and unsigned equivalents, all of the standard operators (comparison, arithmetic, etc.) will be compiled based on the semantics of the types involved. I still have to work out what to do about signed integers in general - our only native option in MASM for signed integers are field elements, the support for u32 and u64 is only useful for unsigned ops, as you'd expect. That probably means that we either need to implement our own primitives for representing, say, signed 32-bit integers, or promote all signed 32-bit integers to field elements, but that comes with its own complexities. I suspect for now we'll need some combination of the two, i.e. using field elements for signed integers, with some additional generated code around ops involving them to protect the range of the type in question. I think we can probably kick this can down the road a bit though.

Sounds reasonable. I've tried to implement `div_s/u` and hit a roadblock. Since Wasm only has `i32` and `i64` types, how the [sign is interpreted](https://www.w3.org/TR/wasm-core-1/#sign-interpretation%E2%91%A0) is up to the operation itself (see [`div_s` vs. `div_u`](https://www.w3.org/TR/wasm-core-1/#-hrefop-idiv-smathrmidiv_s_n-i_1-i_2)). I cannot figure out how to translate signed/unsigned integer op to IR using only `Type` variants. I think we might need an `interpret` op (with a better name) that takes a value and returns the same value but with a given target type. So, it can be called before `div_s/u` to interpret the input values as signed/unsigned.
OTOH, for the rest of the ops that don't have signed/unsigned variants, the use of two’s complement for the signed interpretation means that they behave the same regardless of signedness. So if we use `Type::I32` for the Wasm `i32` value in 'add' op, it would not convey the "signedness", i.e., that we should treat a value as a signed integer. So, we end up where for the `add` op, `i32` means nothing with regards to signedness, but for `div` op, it does mean that the value is signed.

bitwalker:
> Since Wasm only has i32 and i64 types, how the sign is interpreted is up to the operation itself (see div_s vs. div_u). I cannot figure out how to translate signed/unsigned integer op to IR using only Type variants. I think we might need an interpret op (with a better name) that takes a value and returns the same value but with a given target type. 

Let's use `cast` for this for now, it basically does exactly this. We can think on whether we want more specific operations for this as we get further along, but for now this should unblock you.

> OTOH, for the rest of the ops that don't have signed/unsigned variants, the use of two’s complement for the signed interpretation means that they behave the same regardless of signedness. 

That's only true for addition/subtraction/multiplication AFAIK, most everything else is going to need to know how to interpret the binary representation in order to have the correct semantics, but perhaps those are precisely the lines along which the distinction is drawn in Wasm (in terms of which ops have _s/_u variants, and which don't).

In any case, I think the implication is that unless an op in Wasm distinguishes signed vs unsigned, we must assume that the semantics are signed - that is after all the purpose of the two's complement representation. Since Wasm uses dedicated ops to interpret a two's complement binary encoding as an plain unsigned encoding when desired, I think the equivalent approach in our IR would be to `cast` to `Type::U32`, and then call the appropriate op, in order to get unsigned semantics. Presumably, unsigned ops are only ever used on values of unsigned type, and vice versa, so I don't think it would actually be necessary to cast back to `Type::I32` afterwards - i.e. the use of a _s/_u op has a type-narrowing effect.

While we could take a similar tack as Wasm, preferring to represent integers in the IR as signed by default, with dedicated unsigned ops; Miden is a bit all over the place with its instruction set: field elements are signed, but have a restricted set of operations, while its u32 instruction set is purely unsigned, with a broader set of operations. We can certainly emulate signed 32-bit integers with what Miden provides, but we're not going to get much out of the box - we'll have to implement special handling for most ops on signed values (e.g. comparisons, right shift, division).

For now at least, let's go with the `cast` approach, and see where that gets us. If that ends up not working well for other reasons, we can revisit this.

greenhat:
> > Since Wasm only has i32 and i64 types, how the sign is interpreted is up to the operation itself (see div_s vs. div_u). I cannot figure out how to translate signed/unsigned integer op to IR using only Type variants. I think we might need an interpret op (with a better name) that takes a value and returns the same value but with a given target type.
> 
> Let's use `cast` for this for now, it basically does exactly this. We can think on whether we want more specific operations for this as we get further along, but for now this should unblock you.

My first thought was to use `cast` but I was turned away from it by the strict wording of `OpCode::Cast` description: "It is not valid to perform a cast on any value other than a field element". I should've checked the `InstBuilder::cast` implementation.

> 
> > OTOH, for the rest of the ops that don't have signed/unsigned variants, the use of two’s complement for the signed interpretation means that they behave the same regardless of signedness.
> 
> That's only true for addition/subtraction/multiplication AFAIK, most everything else is going to need to know how to interpret the binary representation in order to have the correct semantics, but perhaps those are precisely the lines along which the distinction is drawn in Wasm (in terms of which ops have _s/_u variants, and which don't).
> 
> In any case, I think the implication is that unless an op in Wasm distinguishes signed vs unsigned, we must assume that the semantics are signed - that is after all the purpose of the two's complement representation. Since Wasm uses dedicated ops to interpret a two's complement binary encoding as an plain unsigned encoding when desired, I think the equivalent approach in our IR would be to `cast` to `Type::U32`, and then call the appropriate op, in order to get unsigned semantics. Presumably, unsigned ops are only ever used on values of unsigned type, and vice versa, so I don't think it would actually be necessary to cast back to `Type::I32` afterwards - i.e. the use of a _s/_u op has a type-narrowing effect.
> 
> While we could take a similar tack as Wasm, preferring to represent integers in the IR as signed by default, with dedicated unsigned ops; Miden is a bit all over the place with its instruction set: field elements are signed, but have a restricted set of operations, while its u32 instruction set is purely unsigned, with a broader set of operations. We can certainly emulate signed 32-bit integers with what Miden provides, but we're not going to get much out of the box - we'll have to implement special handling for basically all ops on signed values, e.g. not just comparisons, right shift, division; but also simple things like `add`, since it will be up to us to handle signed overflow. I'll do this on an ad-hoc basis as specific signed ops are needed.
> 
> For now at least, let's go with the `cast` approach, and see where that gets us. If that ends up not working well for other reasons, we can revisit this.

Thank you for the detailed explanation. It clears up a lot of things for me. I'll go with the `cast` to `Type::U32` for `*_u` ops.

bitwalker:
> My first thought was to use cast but I was turned away from it by the strict wording of OpCode::Cast description: "It is not valid to perform a cast on any value other than a field element". I should've checked the InstBuilder::cast implementation.

Yeah that documentation needs clarification now that what I'm anticipating for it's usage has been dialed in a bit better. The primary thing I was trying to convey with that is that `cast` is not intended to support casting values of aggregate type (structs, arrays), or other large values. It is basically intended to be the equivalent of LLVM's `bitcast` instruction - in short, it is only intended to support casting between integral types that fit in a single field element, as well as pointer-to-pointer casts. 

It may turn out that definition is too broad for `cast`, and perhaps having more specific casts with dedicated instructions is better, which we do actually have in some form with the `zext`, `sext`, and `trunc` instructions, which are more precise integer conversion instructions, but they don't really represent the notion of signed/unsigned conversion of the same bit size. It may make the most sense to rename `cast` to `ptrcast`, and reserve it for pointer casts only, and then add `as_signed`/`as_unsigned` instructions that specifically handle reinterpreting a signed integer as unsigned and vice versa. We'll need all three kinds of conversions in any case, it really is just going to boil down to how we expose them in the IR. For now though, I think you can just use `cast`, and we can perhaps discuss the overall approach in next week's compiler team meeting.

greenhat:
@bitwalker I pushed the Wasm globals implementation. `global.get` is translated into `InstBuilder::load_symbol`, and `global.set` is translated into `InstBuilder::store` using the address from `InstBuilder::symbol_addr`

https://github.com/greenhat/miden-ir/blob/c6dc663c2d42c37df49544333f72edeaddcff71c/frontend-wasm/src/code_translator.rs#L75-L90

bitwalker:
> I pushed the Wasm globals implementation. `global.get` is translated into `InstBuilder::load_symbol`, and `global.set` is translated into `InstBuilder::store` using the address from `InstBuilder::symbol_addr`

Perfect, that's what I was anticipating, just wasn't 100% sure if Wasm also had the equivalent of relative addressing and operations which use it, sounds like it's much more straightforward, which keeps things nice and simple.

greenhat:
@bitwalker I'm looking at the data segments in the [spec](https://www.w3.org/TR/wasm-core-1/#syntax-data), and it seems like we need to be able to load a byte array into the memory at a predefined address before starting the execution. There could be multiple data segments, each with its offset and size.

Here is how it looks in the wild. The following Rust example:
```rust
#[inline(never)]
#[no_mangle]
pub fn sum_arr(arr: &[u32]) -> u32 {
    arr.iter().sum()
}

#[no_mangle]
pub extern "C" fn __main() -> u32 {
    sum_arr(&[1, 2, 3, 4, 5]) + sum_arr(&[6, 7, 8, 9, 10])
}
```

Compiles into the following Wasm:


```wasm
(module
    (type (;0;) (func (param i32 i32) (result i32)))
    (type (;1;) (func (result i32)))
    (func $sum_arr (;0;) (type 0) (param i32 i32) (result i32)
    (local i32)
    i32.const 0
    local.set 2
    block ;; label = @1
        local.get 1
        i32.eqz
        br_if 0 (;@1;)
        loop ;; label = @2
        local.get 0
        i32.load
        local.get 2
        i32.add
        local.set 2
        local.get 0
        i32.const 4
        i32.add
        local.set 0
        local.get 1
        i32.const -1
        i32.add
        local.tee 1
        br_if 0 (;@2;)
        end
    end
    local.get 2
    )
    (func $__main (;1;) (type 1) (result i32)
    i32.const 1048576
    i32.const 5
    call $sum_arr
    i32.const 1048596
    i32.const 5
    call $sum_arr
    i32.add
    )
    (memory (;0;) 17)
    (global (;1;) i32 i32.const 1048616)
    (global (;2;) i32 i32.const 1048624)
    (export "memory" (memory 0))
    (export "sum_arr" (func $sum_arr))
    (export "__main" (func $__main))
    (data $.rodata (;0;) (i32.const 1048576) "\01\00\00\00\02\00\00\00\03\00\00\00\04\00\00\00\05\00\00\00\06\00\00\00\07\00\00\00\08\00\00\00\09\00\00\00\0a\00\00\00")
)
```

A data segment at the end of the module should be loaded into the memory at the address `1048576` (0x100000). There can be multiple data segments, each with its offset and size, but since the v1 spec doesn't support multiple memories, the memory index (`;0;`) should always be `0`.

bitwalker:
> I'm looking at the data segments in the spec, and it seems like we need to be able to load a byte array into the memory at a predefined address before starting the execution. There could be multiple data segments, each with its offset and size.

It looks like LLVM is predetermining where it's going to place all of the rodata, the main stack, and the start of the heap. They set the `__data_end` and `__heap_base` values as you'd expect; these are the index, in pages, from the start of linear memory, where the end of the rodata segment, and the bottom of the heap are, respectively. The stack grows down, so it's address is set (via `$__stack_pointer`) to the start of the rodata segment, since the stack consumes the first page. So basically, memory is laid out initially as follows:

```
[stack (64k/1 page) | rodata (64k/1 page) | heap (0k/pages)]
                    ^                     ^
                    |                    / \
            $__stack_pointer    __data_end  __heap_base
```
                      

So this is basically the same type of thing I had intended to do with our linker, but LLVM nicely does some of this work for us already. I think we can plan to use globals for this too, but I may add a new variant to represent a section with a specific starting address and size, which we can then take into account when laying out the heap during codegen. The rodata in your example doesn't take up a full page, but they allocate a full page to it anyway, which is different than the default behavior of our globals, but that's easy to accommodate.

greenhat:
@bitwalker I implemented Wasm data section parsing up to the `Module::declare_data_segment` call introduced in #23. I'll add the actual call after #23 is merged, and I rebase on top of it.
Rust's `static mut` ends up in Wasm as `.data` (see [test](https://github.com/0xPolygonMiden/miden-ir/pull/22/commits/c1119e7c58c04b85dea49c34efee3ba564665d47#diff-b3a4ed9b7ee66649d8a60cd205cd33c535f32f4b4b8190e8b447aaa94ac379dbR509) for [this](https://github.com/0xPolygonMiden/miden-ir/pull/22/commits/c1119e7c58c04b85dea49c34efee3ba564665d47#diff-4f3af65fe4bee69b430f30e854186ab099efb07dc380f3d408b6400924357570R9-R23) Rust code) so I'm deducting `readonly` property from the data segment name (`.rodata` substring means it's read-only).

frontend-wasm/src/code_translator.rs
```diff
@@ -119,6 +119,14 @@ pub fn translate_operator(
         Operator::Call { function_index } => {
             translate_call(state, builder, function_index, mod_info, span, diagnostics)?;
         }
+        Operator::CallIndirect {
```
bitwalker, line 122: @greenhat Do you know what in dlmalloc is using indirect calls? I'm guessing there must be some kind of configurable behavior that works by providing a function pointer somewhere, but just want to make sure it's not something critical, or we may have to require the use of `wee_alloc`

greenhat:
@bitwalker I added a test with Rust memory allocations linking to `std` and got `ctz` and `clz` ops in the dlmalloc code and `call_indirect` in `core::fmt` and `std::panicking`. See - https://github.com/0xPolygonMiden/miden-ir/pull/22/commits/4a9e3030e9c323d9ce9ac48c9c0d0504f7313724

frontend-wasm/src/code_translator.rs
greenhat: Yes, I linked to `std` and it's `core::fmt` and `std::panicking`. See https://github.com/0xPolygonMiden/miden-ir/commit/4a9e3030e9c323d9ce9ac48c9c0d0504f7313724

Do we intend to require `no_std` in contracts? If so, it's avoidable. 
Also. I have a naive idea of how to implement `call_indirect`. Since we know the func signature at compile time we can filter all the relevant functions from the table and put them in a generated function where calls are dispatched in runtime via `CondBr`s or `If`s.

greenhat:
@bitwalker I have not found any `shr_s`, `div_s`, or `rem_s` in the dlmalloc wasm code. I did find a couple of `le_s` and `lt_s` though.

greenhat:
@bitwalker I finished the house cleaning, and it's ready for review. Check TODO section in the description for what's left. I can continue working on this PR, or we can create issues for them, and I will tackle them in the new PRs.

tests/rust-wasm-tests/src/dlmalloc.rs
```diff
@@ -0,0 +1,14 @@
+extern crate alloc;
+use alloc::vec::Vec;
+
+#[no_mangle]
+pub fn vec_alloc() {
+    let mut v = Vec::new();
+    if let Err(_) = v.try_reserve(4) {
+        unreachable!();
+    }
+    if let Err(_) = v.push_within_capacity(1) {
```
greenhat, line 10: @bitwalker I finished with no_std + dlmalloc example. `Vec::push` contains code to panic on OOM, which brings `call_indirect` into Wasm. I used unstable `Vec::push_with_capacity` instead.
I also found a `no_global_oom_handling` config option, which causes rustc not to compile methods in `alloc` that panic on OOM (`Vec::push`, `reserve`, etc.) We should use it to throw compilation errors if devs use those methods in their smart contract code. It works only with the unstable `-Z build-std=alloc,core` flag, which makes cargo to build std from the source.

greenhat:
@bitwalker I see #23 is merged. I'd like to rebase this branch on the new main branch. But I think the force push might mess with any comments you might have written as part of the review.

tests/rust-wasm-tests/src/dlmalloc.rs
greenhat: The code looks quite ugly from the smart contract dev point of view. We might prioritize `call_indirect` support to avoid it.

bitwalker: I believe if you set `panic = "abort"` in the `Cargo.toml`, it should remove basically all panic-related code (it simply traps in those cases). This is typically the approach used in Wasm-compiled Rust projects, might be worth a shot to see if that helps.

In general though, I agree we'll want to have some kind of solution for indirect calls, we'll have to revisit when `pcall` is stabilized in Miden.

bitwalker:
@greenhat Go ahead and rebase, I haven't had a chance to get my review completed yet, but can push without messing it up on my end

tests/rust-wasm-tests/src/dlmalloc.rs
greenhat: Thanks for the tip! I did not know about `panic = "abort"`. It seems to be set to "abort" for wasm32 target by default since I did not notice any changes in the wasm code after I set it.
Actually, the `core::fmt` used for message formatting in panic calls caused all the `call_indirect` ops. I found `panic_immediate_abort` std feature in https://github.com/rust-lang/rust/blob/551c7183f8ae46c4b4fd0237acbdc937670bf74a/library/core/src/panicking.rs#L52-L73 and it eliminated all the `core::fmt` code. I switched back to using `Vec::push`.

bitwalker:
@greenhat Can you rebase this on the `bitwalker/codegen` branch? I'd like to merge both, and it'll make things simpler if this branch is based on the latest stuff in the `miden-hir*` crates. I'm going to do my review today regardless, but I'll hold off merging this until #27 is merged, basically treating them as a stacked set of changes.

bitwalker (commented):
Just leaving a few early notes. I'm getting into the `code_translator` modules next, and once done there we should be good to go.

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+            builder.ins().store(ptr, val, span);
+        }
+        /********************************* Stack misc **************************************/
+        Operator::Drop => _ = state.pop1(),
+        Operator::Select => {
+            let (arg1, arg2, cond) = state.pop3();
+            // if cond is not 0, return arg1, else return arg2
+            // https://www.w3.org/TR/wasm-core-1/#-hrefsyntax-instr-parametricmathsfselect%E2%91%A0
+            let zero = builder.ins().i32(0, span);
+            let cond_i1 = builder.ins().neq(cond, zero, span);
```
bitwalker, line 97: *Suggestion*: As much as possible, prefer the `_imm`-suffixed instruction variants for things like this, it makes the IR more canonical by construction, and avoids having to deal with the immediate value during analysis/transformation. I think it tends to be more efficient in the VM as well, but that's less important.

```suggestion
            let cond_i1 = builder.ins().neq_imm(cond, Immediate::U32(0), span);
```

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+            let arg = state.pop1();
+            state.push1(builder.ins().mem_grow(arg, span));
+        }
+        Operator::MemorySize { .. } => {
+            // Return total Miden memory size
+            state.push1(builder.ins().i32(mem_total_pages(), span));
+        }
+        /******************************* Load instructions ***********************************/
+        Operator::I32Load8U { memarg } => {
+            translate_load_zext(I8, I32, memarg, state, builder, span)
```
bitwalker, line 132: *Bug*: You'll need to use the unsigned type, e.g. `U8`, for any zero-extensions like this. The codegen backend will not allow zero-extending a signed type as it will lose the signedness by doing so. You must use casts otherwise.

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+        }
+        Operator::I64Load32U { memarg } => {
+            translate_load_zext(I32, I64, memarg, state, builder, span)
+        }
+        Operator::I32Load { memarg } => translate_load(I32, memarg, state, builder, span),
+        Operator::I64Load { memarg } => translate_load(I64, memarg, state, builder, span),
+        /****************************** Store instructions ***********************************/
+        Operator::I32Store { memarg } => translate_store(I32, memarg, state, builder, span),
+        Operator::I64Store { memarg } => translate_store(I64, memarg, state, builder, span),
+        Operator::I32Store8 { memarg } | Operator::I64Store8 { memarg } => {
```
bitwalker, line 166: This should use `U8` rather than `I8`, since presumably it's intended for storing bytes, not signed integers. Same for the 16 and 32-bit variants below.

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+        }
+        Operator::I64GeS => {
+            let (arg0, arg1) = state.pop2();
+            let val = builder.ins().gte(arg0, arg1, span);
+            state.push1(builder.ins().cast(val, I32, span));
+        }
+        Operator::I32Eqz => {
+            let arg = state.pop1();
+            let imm_zero = builder.ins().i32(0, span);
+            let val = builder.ins().eq(arg, imm_zero, span);
```
bitwalker, line 369:
```suggestion
            let val = builder.ins().eq_imm(arg, Immediate::I32(0), span);
```

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+            state.push1(builder.ins().cast(val, I32, span));
+        }
+        Operator::I64Eqz => {
+            let arg = state.pop1();
+            let imm_zero = builder.ins().i64(0, span);
+            let val = builder.ins().eq(arg, imm_zero, span);
```
bitwalker, line 375:
```suggestion
            let val = builder.ins().eq_imm(arg, Immediate::I64(0), span);
```

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+        }
+        let def_block = {
+            let i = state.control_stack.len() - 1 - (default as usize);
+            let frame = &mut state.control_stack[i];
+            frame.set_branched_to_exit();
+            frame.br_destination()
+        };
+        builder.ins().switch(val, data, def_block, span);
+    } else {
+        // Here we have jump arguments, but Midens's switch op doesn't support them
```
bitwalker, line 450: I'll add support for this, but we can handle that after this is merged

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+        Operator::Call { function_index } => {
+            translate_call(state, builder, function_index, mod_info, span, diagnostics)?;
+        }
+        /******************************* Memory management *********************************/
+        Operator::MemoryGrow { .. } => {
+            let arg = state.pop1();
+            state.push1(builder.ins().mem_grow(arg, span));
+        }
+        Operator::MemorySize { .. } => {
+            // Return total Miden memory size
```
bitwalker, line 127: We should add an instruction for this, similar to `mem_grow`, otherwise it won't match what is returned by `mem_grow`, and that may cause issues. Is `memory_size` actually used anywhere that we've seen yet? IIRC `wee_alloc` and `dlmalloc` both just use `mem_grow`, but I could be misremembering.

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+    state.reachable = false;
+    Ok(())
+}
+
+/// Return the total Miden VM memory size (2^32 addresses * word (4 felts) bytes) in 64KB pages
+const fn mem_total_pages() -> i32 {
+    let bytes_fit_in_felt = 4; // although more than 32 bits can fit into a felt, use 32 bits to be safe
+    let felts_in_word = 4;
+    let total_addresses = u32::MAX as i64;
+    (total_addresses * (felts_in_word * bytes_fit_in_felt) / (64 * 1024)) as i32
```
bitwalker, line 499: *Suggestion*: This whole thing could be a constant, rather than a const function, but it probably will go away as soon as the `memory_size` instruction is added anyway, so your call

```suggestion
    const FELT_BYTES: u64 = 4; // felts are effectively 32 bits
    const WORD_BYTES: u64 = 4 * FELT_BYTES; // 4 felts per word
    const PAGE_SIZE: u64 = 64 * 1024;
    const MEMORY_SIZE: u64 = (u32::MAX as u64 * WORD_BYTES);
    MEMORY_SIZE / PAGE_SIZE
```

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+    ptr_ty: Type,
+    memarg: &MemArg,
+    state: &mut FuncTranslationState,
+    builder: &mut FunctionBuilderExt<'_>,
+    span: SourceSpan,
+) {
+    let (addr_int, val) = state.pop2();
+    let val_ty = builder.func().dfg.value_type(val);
+    let arg = if ptr_ty != *val_ty {
+        builder.ins().trunc(val, ptr_ty.clone(), span)
```
bitwalker, line 554: This will fail if `val_ty` is larger than `ptr_ty`, as truncation must be to a smaller type (it's allowed for the types to be the same size, in which case it is a no-op, but better to cast the pointer type in that case)

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+fn prepare_addr(
+    addr_int: Value,
+    ptr_ty: &Type,
+    memarg: &MemArg,
+    builder: &mut FunctionBuilderExt,
+    span: SourceSpan,
+) -> Value {
+    let full_addr_int = if memarg.offset != 0 {
+        let offset = builder.ins().i32(memarg.offset as i32, span);
+        builder.ins().add(addr_int, offset, span)
```
bitwalker, line 571:
```suggestion
        builder.ins().add_imm(addr_int, Immediate::U32(memarg.offset as u32), span)
```

You'll need to use `u32` for the type of a pointer-as-integer, and likewise for offsets (handling negative offsets as subtracting, rather than adding), since signed 32-bit arithmetic is not implemented yet.

In general I would use `getelementptr` for this kind of thing, however it currently only works on pointers to aggregates, I will probably extend it to support indexing off primitive types as well though to allow offsetting pointers "raw". We could then allow signed offsets, and rely on the backend to work out the most efficient lowering for the effective address

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+) {
+    let cond = state.pop1();
+    let (br_destination, inputs) = translate_br_if_args(relative_depth, state);
+    let next_block = builder.create_block();
+    let then_dest = br_destination;
+    let then_args = inputs;
+    let else_dest = next_block;
+    let else_args = &[];
+    let zero = builder.ins().i32(0, span);
+    let cond_i1 = builder.ins().neq(cond, zero, span);
```
bitwalker, line 662:
```suggestion
    let cond_i1 = builder.ins().neq_imm(cond,Immediate::I32(0), span);
```

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -0,0 +1,1035 @@
...
+    blockty: &wasmparser::BlockType,
+    state: &mut FuncTranslationState,
+    builder: &mut FunctionBuilderExt<'_>,
+    module_info: &ModuleInfo,
+    span: SourceSpan,
+) -> WasmResult<()> {
+    let blockty = BlockType::from_wasm(blockty, module_info)?;
+    let cond = state.pop1();
+    let zero = builder.ins().i32(0, span);
+    let cond_i1 = builder.ins().neq(cond, zero, span);
```
bitwalker, line 823:
```suggestion
    let cond_i1 = builder.ins().neq_imm(cond, Immediate::I32(0), span);
```

bitwalker (changes requested):
All done! In general everything looks great, as expected. I had a few questions/suggestions/nits, but nothing major to speak of.

That said, I'm marking this as requesting changes because I would really like to confirm that how the Miden IR type system is being used is correct. In particular, I32 is being used for most things (in fact, signed types are predominantly used, probably because unsigned types only recently were introduced). The rest of the compiler backend assumes that signed vs unsigned types are used appropriately, as in it doesn't allow one to freely treat signed integers as unsigned and vice versa; you must either use the correct types for the operation, or cast to the needed type, so that the semantics of each type are maintained.

If that presents an issue coming from Wasm in particular, we can perhaps rethink some aspects of this, or at least how we surface things in the IR/builders (e.g. rather than having an `add` instruction, we have an `add_s` vs `add_u`, and let the builders manage the complexities of what to do with the types it is given.

In any case, I mostly just want to hash that out before this gets merged (we can do so on a call even, if that's easier).

All-in-all though, great work on this, I'm excited to get this merged and integrated with the other changes!

bitwalker:
All of the automatic branch deletion shit completely blew this up, sorry, I _think_ because the merge of #27 was done as a rebase, it no longer knows how to merge against `main`, so, unfortunately, you're going to have to rebase against `main` now. It should be straightforward to do locally, since nothing has actually changed since your rebase (unless something got changed during your rebase) but GitHub is all kinds of confused, so we definitely don't want to let it try and merge as-is.

bitwalker:
Thanks @greenhat! Let's catch up on Monday and get this merged ASAP

hir/src/value.rs
```diff
@@ -38,18 +38,28 @@ pub enum ValueData {
         block: Block,
         span: SourceSpan,
     },
+    /// Value is an alias of another value.
+    /// An alias value can't be linked as an instruction result or block parameter. It is used as a
+    /// placeholder when the original instruction or block has been rewritten or modified.
+    Alias {
```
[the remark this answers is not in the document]
greenhat: > Could you elaborate on the semantics of this value type? I gather from comment that it basically acts like a reference to another `Value`, but this potentially impacts a number of places where we have to make sure that we preserve the semantics of the IR, so I want to make sure I fully understand it, and that we update any affected places in the code.
> 
> Namely, I have the following questions:
> 
> 1. When do we expect to observe `Alias` values?

When we search for a var definition in predecessors, we create a block parameter as a sentinel (in `SSABuilder::find_var`). When all the predecessors use a single value to represent our variable, we create an alias to that value and remove the block parameter (in `SSABuilder::finish_predecessor_lookup`) otherwise, we keep the block parameter. This is the only place where we create an alias. 

> 2. What should be done with an `Alias` value if observed during analysis or transformation passes?

It should be resolved to the underlying value (via `DataFlowGraph::resolve_aliases`).

> 
> Regarding my second question, I'm particularly thinking of the block inlining and treeify passes which clone/modify blocks and instructions, and accordingly, the values they define/use. I just fixed a bug I accidentally introduced by shallow cloning instructions which resulted in cloned instructions placed in different blocks accidentally sharing their ValueLists. Based on the description here, I would assume it is a bug to observe an `Alias` value in most situations, but it isn't clear to me when it should be used or not used; observed or not observed.

It's not a bug, but I can see how it might complicate things down the pipeline. I can get rid of the `Alias` and keep the block parameter instead.

hir/src/builder.rs
```diff
@@ -553,6 +553,16 @@ pub trait InstBuilder<'f>: InstBuilderBase<'f> {
         ))
     }
 
+    fn mem_grow(mut self, value: Value, span: SourceSpan) -> Value {
+        require_integer!(self, value, Type::I32);
```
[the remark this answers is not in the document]
greenhat: Done.

frontend-wasm/src/code_translator/mod.rs
greenhat: Done.

hir/src/value.rs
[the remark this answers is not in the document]
bitwalker: Keeping the block parameter is fine, as it will be removed by later transformation passes if it's not needed. But you should also just be able to reference the original `Value`, i.e. the `Alias` value type doesn't sound like it's actually needed, since a `Value` isn't tied to a specific block. As long as the use of a `Value` is dominated by it's definition, you can just use the original `Value` directly (i.e. all control flow paths from the entry block to the block containing the use, flow through the definition, whether that's a block parameter or an instruction result). To illustrate, consider the following IR:

```
pub fn foo(u32, u32) -> u32 {
entry(v0, v1):
    v2 = const.u32 0
    v3 = eq v1, 0
    condbr v3, blk0(v2), blk1

blk0(v4: u32):
    ret v4

blk1:
    v5 = const.u32 2
    v6 = eq v0, 0
    condbr v6, blk2, blk3

blk2:
    br blk3

blk3:
    v7 = div v0, v1
    v8 = mul v7, v5
    br blk4(v8)

blk4(v9: u32):
    ret v9
```

In this function, the use of `v0`, `v1`, and `v5` in `blk3` are safe, because all paths to `blk3` pass through the definition of those values (`v0` and `v1` trivially dominate every block in the function, since they are parameters of the entry block, and `v5` is defined in `blk1`, the immediate dominator of `blk3`.

Note that the block parameter in `blk4` is not needed, but is also perfectly valid, it just obfuscates the origin of `v9`, but it can also be useful in some cases to introduce such parameters, in order to break up the live range of a value.

So coming back to the purpose of `Alias`, it seems like you would intend to use it in `blk3` for the uses of `v0`, `v1`, and `v5`, if I understand correctly. Hopefully my example IR demonstrates why that wouldn't be necessary though. That said, if I'm missing a key difference here, perhaps an example like mine above would help illustrate the use case a bit better for me.

hir/src/module.rs
```diff
@@ -598,6 +602,10 @@ impl<'m> ModuleFunctionBuilder<'m> {
         self.function.id
     }
 
+    pub fn func_builder(&mut self) -> FunctionBuilder {
```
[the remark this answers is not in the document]
greenhat: Good catch! I made a switch to `ModuleFunctionBuilder` and moved `func_builder` to `Function`. I vaguely remember that I tried to use `ModuleFunctionBuilder` initially but something stopped me there.

hir/src/value.rs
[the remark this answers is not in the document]
greenhat: Thank you for the code example and detailed explanation! You are right. I removed `ValueData::Alias`.

frontend-wasm/src/func_translator.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub fn translate_body(
+        &mut self,
+        body: &FunctionBody<'_>,
+        mod_func_builder: &mut ModuleFunctionBuilder,
+        mod_info: &ModuleInfo,
+        diagnostics: &DiagnosticsHandler,
+        func_validator: &mut FuncValidator<impl WasmModuleResources>,
+    ) -> WasmResult<()> {
+        let mut reader = body.get_binary_reader();
+        // log::trace!(
```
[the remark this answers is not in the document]
greenhat: Yep. Removed. Thanks!

frontend-wasm/src/translation_utils.rs
```diff
@@ -0,0 +1,55 @@
...
+        Type::U128 => todo!(),
+        Type::Usize => todo!(),
+        Type::F64 => builder.ins().f64(0.0, SourceSpan::default()),
+        Type::Felt => todo!(),
+        Type::Ptr(_) => panic!("cannot emit zero for pointer type"),
+        Type::Struct(_) => panic!("cannot emit zero for struct type"),
+        Type::Array(_, _) => panic!("cannot emit zero for array type"),
+        Type::Unknown => panic!("cannot emit zero for unknown type"),
+        Type::Unit => panic!("cannot emit zero for unit type"),
+        Type::Never => panic!("cannot emit zero for never type"),
```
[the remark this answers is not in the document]
greenhat: Much better! Done. 
I blame `rust-analyzer` and copilot. :smiley:

frontend-wasm/src/code_translator/mod.rs
greenhat: Sure. Thanks for cathing this! Done.

frontend-wasm/src/code_translator/mod.rs
greenhat: Done.

frontend-wasm/src/code_translator/mod.rs
greenhat: Done.

frontend-wasm/src/code_translator/mod.rs
greenhat: Done.

frontend-wasm/src/code_translator/mod.rs
greenhat: I agree. There is no `memory_size` in `dlmalloc` code. I have not checked `wee_alloc` yet.

frontend-wasm/src/code_translator/mod.rs
greenhat: Done. I left `inttoptr` for now.

frontend-wasm/src/code_translator/mod.rs
greenhat: I'm confused. When called for Wasm `i32_store8` op `val_ty` is `I32` and `ptr_ty` is `U8`. Why will it fail to `trunc` `I32` value to `U8`? Is it because I need to trunc to ` I8` first and then cast to `U8`?

frontend-wasm/src/code_translator/mod.rs
greenhat: Thanks! I like yours more. It's more readable. I took it but left it inside the const function to keep it contained. It's a temporary solution anyway.

frontend-wasm/src/ssa.rs
```diff
@@ -0,0 +1,552 @@
...
+}
+
+/// States for the `use_var`/`predecessors_lookup` state machine.
+enum Call {
+    UseVar(Inst),
+    FinishPredecessorsLookup(Value, Block),
+}
+
+/// Emit instructions to produce a zero value in the given type.
+fn emit_zero(_ty: &Type, //, mut cur: FuncCursor
```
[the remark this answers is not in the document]
greenhat: It's not exactly the same, `translation_utils::emit_zero` inserts at the last position but here we want to insert at the start of the block. It is needed to handle the case when the variable is used, but we cannot find its definition in the predecessors. So, we want to initialize the variable (rather than failing) at the start of the block. We need some sort of function cursor to set an arbitrary insertion point in addition to the builder interface. I could not find one. I would use it in `translation_utils::emit_zero` and call it here.

frontend-wasm/src/ssa.rs
```diff
@@ -0,0 +1,552 @@
...
+    }
+
+    /// Declares a new predecessor for a `Block` and record the branch instruction
+    /// of the predecessor that leads to it.
+    ///
+    /// The precedent `Block` must be filled before added as predecessor.
+    /// Note that you must provide no jump arguments to the branch
+    /// instruction when you create it since `SSABuilder` will fill them for you.
+    ///
+    /// Callers are expected to avoid adding the same predecessor more than once in the case
```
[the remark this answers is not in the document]
greenhat: Good point! Thanks! I added an assertion. I'd rather implement the proposed solution as a separate PR. I made https://github.com/0xPolygonMiden/miden-ir/issues/38 so it does not get lost.

greenhat:
I addressed all the notes. 

Besides that, there are two issues:
* make an issue for fixing using `popcnt` for `ctz`, `clz` translation (used to unblock dlmalloc translation);
* `dlmalloc` compilation test is set to "ignored" because of the different mangled function names on CI. Figure out why the hash in mangled function names is so unstable, or even better - add an option to demangle function names and remove the hash, and use this option in tests.

I'd rather make issues and fix them in the next PR.

frontend-wasm/src/ssa.rs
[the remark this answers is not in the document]
bitwalker: Ah, I see. The `DataFlowGraph` provides `insert_insert` as a primitive, which allows you to specify an `InsertionPoint`, which can be either the start/end of a `Block`, or before/after an `Inst`, but I haven't exposed that in a very convenient way in the `DefaultInstBuilder`. All of the `InstBuilder` implementations actually use this under the covers, so its really just a matter of surfacing it better in one of the builders.

frontend-wasm/src/code_translator/mod.rs
bitwalker: I think what caught my eye here was that there doesn't seem to be anything that ensures `ptr_ty` is smaller than `val_ty` - I thought I had observed one place where that could be the case, but I may have been mistaken on that. In any case, there is an assertion in the builder, so if it ever happens we'll know about it, so I think we can call this one resolved.

bitwalker (approved):
LGTM! Great work on this @greenhat

frontend-wasm/src/code_translator/mod.rs
greenhat: Oh, I see. All store ops in Wasm have `ptr_ty` smaller or equal to `val_ty`. That's why I have not thought about adding such a check.

frontend-wasm/src/ssa.rs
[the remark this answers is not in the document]
greenhat: Got it. I created https://github.com/0xPolygonMiden/miden-ir/issues/40

PR #24, opened 2023-09-21
feat: implement program linker

bitwalker:
This commit introduces [Linker], which is used to take a set of [Module] and perform many of the tasks that a typical system linker would handle. In the Miden world, the linker gets involved at a slightly different stage of compilation however, and does not have the final say in what code will ultimately get executed at runtime (with some exceptions).

Notably, [Linker] is invoked on a set of modules in IR form, rather than in Miden Assembly (MASM) form - in other words, the linker is invoked before modules are lowered to MASM by the codegen backend. This is because one of the critical tasks performed by the linker, is the allocation and layout of linear memory for use by the compiled program. To do this, the linker must determine what set of data segments and globals will end up in the output, validate that there are no conflicts, and that all referenced symbols have corresponding definitions, and then lay out the data segments and globals in memory so that the codegen backend knows what at what address to find a given global, as well as where the start of the heap will fall for use by the program.

The [Linker] is also responsible for garbage collecting unused declarations, but this is dependent on what type of artifact is being produced.

Since MASM is itself an intermediate representation for the Miden VM, the compiler does not actually know whether additional modules will be included in the set provided to the VM at runtime, or even whether all of the emitted modules will be provided to the VM. That said, we've designed the linker to make certain assumptions based on what type of output is being emitted. For example, an executable program is considered "closed", i.e. we assume that no additional code will be
introduced other than what is being linked (aside from the Miden standard library). On the other hand, a library program is considered "open", i.e. we assume that the library will be included as part of a larger executable program that will contain modules we don't necessarily control. Despite that, we do make an assumption that any additonal modules in that situation will play nicely with the memory layout of the library.

---

This is a pre-requisite for subsequent codegen PRs that use the output of the linker to emit Miden Assembly from corresponding Miden IR.

/cc @greenhat - just keeping you in the loop here, as one of the changes here affects your Wasm frontend (very trivially, just a small tweak to the API for declaring global variables).


bobbinth:
Without having looked at the code, one question I wanted to ask: do we have a way to define "external functions" (besides the ones in the Miden standard library). By external functions I mean some functions which the complier will know the signature but not the implementation.

Basically, I wonder if the compiler will be able to output something like:
```
use.my_library::my_module

being
    ...
    exec.my_module::some_procedure
    ...
end
```
Where `my_library` is written in MASM but the compiler knows all procedure signatures in that library.

bobbinth:
I agree that going with metadata/signature file is a better approach. Let's create an issue for this, and we can discuss there file format etc.

hir/src/globals.rs
```diff
@@ -164,26 +189,26 @@ impl GlobalVariableTable {
...
-            // intrinsics available for them
-            let offset = hp.align_offset(layout.align());
-            unsafe {
-                hp = hp.add(offset).add(layout.size());
+            let align = layout.align();
+            let mut gv_size = layout.size();
+            let align_offset = size % align;
+            if align_offset != 0 {
+                gv_size += align - align_offset;
             }
```
jjcnn, line 200: The same calculation is replicated in `offset_of`. It could probably be a helper function.

hir/src/program/linker.rs
```diff
@@ -0,0 +1,594 @@
...
+
+    /// If an executable is being linked, discover unused functions and garbage collect them.
+    ///
+    /// Once a function has been identified as dead and is collected, any transitive items it
+    /// references may also be collected if they are orphaned as a result of the collection.
+    ///
+    /// If a library is being linked, this function is a no-op, as it is not known what will
+    /// be needed at runtime once used in the context of a program.
+    fn garbage_collect(&mut self) {
+        // TODO:
```
jjcnn, line 520: Is this on your immediate TODO list? If not, then we should create an issue for it.

jjcnn (approved):
LGTM. Just a couple of minor comments, which you can deal with as you see fit.

hir/src/builder.rs
```diff
@@ -703,6 +799,121 @@ pub trait InstBuilder<'f>: InstBuilderBase<'f> {
...
+                        ty
+                    );
+                    let element_size = element_ty.size_in_bytes();
+                    let min_align = element_ty.min_alignment();
+                    let align_offset = element_size % min_align;
+                    let padded_element_size = if align_offset != 0 {
+                        element_size + (min_align - align_offset)
+                    } else {
+                        element_size
+                    };
```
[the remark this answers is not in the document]
bitwalker: It's slightly different, since what we're doing here is calculating the size up to the element being selected. I do think there would be some benefit to having some helpers for some of these alignment calculations though, since they occur in a few places.

hir/src/builder.rs
```diff
@@ -703,6 +799,121 @@ pub trait InstBuilder<'f>: InstBuilderBase<'f> {
...
+                    ty = element_ty;
+                    match index {
+                        0 => continue,
+                        1 => {
+                            offset += padded_element_size;
+                        }
+                        n => {
+                            offset += padded_element_size * n;
+                        }
```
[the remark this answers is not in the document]
bitwalker: Ah indeed, this was refactored a bit, and the `match` is no longer necessary.

hir/src/program/linker.rs
bitwalker: Good suggestion, I wasn't sure if I was going to get to it while I was working on the linker, but I think I'm going to defer this to a later time, it's not necessary for now.

Issue #32, opened 2023-10-16
Implement support for indirect calls via dynexec/procref


bitwalker:
Now that 0xPolygonMiden/miden-vm#1078 is merged (which introduces `dynexec`), the groundwork for supporting indirect calls in Miden Assembly is there. Before we can make use of it though, we need the proposed `procref` instruction, which would push the hash of a specified function name on the stack. This is done prior to `dynexec`, which requires that hash in order to execute the indirect call. We need `procref` because we don't know the hash of the function until after compilation, which `procref` solves by relying on the Miden assembler to expand it into `push.HASH` as it knows the hashes of all the functions.

Once that is implemented, we can make use of it to implement indirect calls in the IR:

```[tasklist]
### Tasks
- [ ] Add a new `call.indirect` opcode, to be paired with the `PrimOp` instruction type
- [ ] Add a new `Type` variant, `Type::Function(Box<FunctionType>)` which will be used in conjunction with `Type::Ptr` to represent function pointers
- [ ] Add a new data segment to `Program` which represents every function in the compiled program as a table of hashes. Initialize this table at runtime using `procref`
- [ ] https://github.com/0xPolygonMiden/compiler/issues/133
- [ ] Lower indirect calls to MASM by treating the function pointer operand as an index into the function table, load the callee hash, and then dispatch via `dynexec`. Indexing beyond the bounds of the table should trap.
```

bobbinth:
It might also be good to wait for https://github.com/0xPolygonMiden/miden-vm/issues/1091 - but, of course, not necessary.

bitwalker:
Agreed, I think it is beneficial to wait for those changes, though we can at least lay the groundwork for indirect calls without supporting them in codegen. If it turns out that we reach a point where we have `procref` but not the changes proposed in that issue, we can experiment with what's there now, but I don't anticipate doing that unless indirect calls turn out to be more important than initially thought when compiling from Rust.

greenhat:
@bitwalker While translating the basic wallet Wasm component example [PR](https://github.com/0xPolygonMiden/compiler/pull/131) I found a case where table is used as a parameter for module instantiation and I want to make sure we capture these semantics in the IR. 

[...]

PR #35, opened 2023-10-17
feat: implement conversion of masm ir to 'real' masm

bitwalker:
As the PR title states, this enables conversion from our internal MASM IR to the "real" MASM representation provided by the `miden-assembly` crate. We can then emit to disk using those APIs, compile via the assembler, or run with the Miden VM.

NOTE: This PR omits a couple things pending some other work:

1. Emission of MASL libraries is blocked by 0xPolygonMiden/miden-vm#1108
2. The initialization of root context memory (i.e. writing data segments to the heap) is not yet implemented. We'll probably implement something here in the near future, but I'm punting a bit while I work on some other things
3. The function table that we'll want to write to memory to support `dynexec` is on hold until #32 is implemented

### DIFF
--- a/codegen/masm/src/masm/program.rs
+++ b/codegen/masm/src/masm/program.rs
@@ -1,16 +1,23 @@
-use std::path::Path;
+use std::{collections::BTreeMap, path::Path};
 
-use miden_hir::{self as hir, DataSegmentTable, FunctionIdent};
+use miden_hir::{self as hir, DataSegmentTable, FunctionIdent, Ident};
 
 use super::*;
 
+/// A [Program] represents a complete set of modules which are intended to
+/// be shipped together as an artifact, either as an executable, or as a library
+/// to be integrated into a larger executable.
 #[derive(Default)]
 pub struct Program {
+    /// The set of modules which belong to this program
     pub modules: Vec<Module>,
+    /// The function identifier for the program entrypoint, if this is an executable module
     pub entrypoint: Option<FunctionIdent>,
+    /// The data segment table for this program
     pub segments: DataSegmentTable,
 }
 impl Program {
+    /// Create a new, empty [Program]
     pub fn new() -> Self {
         Self::default()
     }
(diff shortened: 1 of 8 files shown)


codegen/masm/src/masm/program.rs
```diff
@@ -22,6 +29,66 @@ impl Program {
...
+                }
+                .to_string(),
+            )
+            .expect("invalid entrypoint function name");
+            let imported =
+                BTreeMap::from([(entry_import.alias.to_string(), entry_module_path.clone())]);
+            let invoked = BTreeMap::from([(entry_id, (entry_name, entry_module_path))]);
+            let imports = ModuleImports::new(imported, invoked);
+
+            // TODO: Write data segments, initialize function table
```
greenhat, line 82: Would an issue be a better fit for these tasks?

hir/src/asm/isa.rs
```diff
@@ -616,6 +617,378 @@ pub enum MasmOp {
...
+    pub fn into_node(
+        self,
+        _codemap: &miden_diagnostics::CodeMap,
+        imports: &super::ModuleImportInfo,
+        local_ids: &FxHashMap<FunctionIdent, u16>,
+        proc_ids: &FxHashMap<FunctionIdent, miden_assembly::ProcedureId>,
+    ) -> SmallVec<[miden_assembly::ast::Node; 2]> {
+        use miden_assembly::ast::{Instruction, Node};
+        let node = match self {
+            Self::Padw => Node::Instruction(Instruction::PadW),
```
greenhat, line 630: Making `node` to be `Instruction` rather than `Node` would allow us to remove `Node::Instruction(...)` and reduce the case statement to `Self::Padw => Instruction::PadW,`. If `Instruction::*` can be imported, this would even reduce the case to `Self::Padw => PadW,`.

greenhat (approved):
Looking great! Only a few minor notes.

hir/src/asm/isa.rs
bitwalker: I had a similar thought (about handling the `Node`-wrapping at the end of the block), but was already quite some way through the implementation so didn't bother to change it. I can do that before merging though.

As for importing the enum variants directly, I don't like doing that, and prefer to alias the enum type itself if it is too verbose, e.g. `use miden_assembly::ast::Instruction as Inst`, but in this case it's not really that much noise.

PR #37, opened 2023-10-18
feat: implement codegen of i32 ops

bitwalker:
This PR adds support for lowering IR instructions that have i32 operands to Miden Assembly. Since these are the most likely to be needed by Wasm-compiled modules, we wanted to get these implemented ASAP.

There are three main pieces here:

1. The hand-written MASM modules in the `intrinsics` namespace, e.g. `intrinsics::i32`
2. The parsing of MASM files into our internal MASM representation so that we can use them with the emulator. For this we lean on the parsing facilities provided by the `miden-assembly` crate
3. The implementation of the i32 variants for relevant instructions in `OpEmitter` in the `miden-codegen-masm` crate

While we could construct IR representing the `intrinsics` modules, it's far too verbose to be easy to maintain that way, hence the need for 1 and 2. We also want to be able to test modules compiled with uses of those intrinsics, hence the need for 2.


codegen/masm/src/stackify/emit/binary.rs
```diff
@@ -490,6 +534,7 @@ impl<'a> OpEmitter<'a> {
                 self.emit(Op::Div);
             }
             Type::U32 | Type::U16 | Type::U8 => self.unchecked_div_u32(),
+            Type::I32 => self.checked_div_i32(),
```
greenhat, line 537: Is it temporary until we have an unchecked division for i32 implemented? OTOH, `intrinsics::i32::checked_div` does not check for an overflow. I'm confused.

codegen/masm/src/stackify/emit/int32.rs
```diff
@@ -704,11 +861,28 @@ impl<'a> OpEmitter<'a> {
         self.emit(Op::U32CheckedMin);
     }
 
+    /// Pops two i32 values off the stack, `b` and `a`, and puts the result of `min(a, b)` on the stack
+    ///
+    /// This operation is checked, if the operands or result are not valid i32, execution traps.
+    pub fn min_i32(&mut self) {
+        self.emit(Op::Exec("intrinsics::i32::min".parse().unwrap()));
```
greenhat, line 868: I don't see `min` and `max` definitions in `i32.masm`.

greenhat (changes requested):
Looking good! Please check my questions for a few ops.

codegen/masm/intrinsics/i32.masm
```diff
@@ -0,0 +1,321 @@
...
+    eq.0             # [a_msb == 0, b_msb, b, a]
+    swap.1 eq.0      # [b_msb == 0, a_msb == 0, b, a]
+    swap.1 dup.1 neq # [a_msb != b_msb, b_msb == 0, b, a]
+
+    # if a_msb != b_msb, then a > b (if a_msb == 0), or a < b (if a_msb == 1)
+    if.true # [b_msb == 0, b, a]
+        movdn.2 drop drop         # [b_msb == 0]
+        push.1 push.NEG1          # [-1, 1, b_msb == 0]
+        swap.2                    # [b_msb == 0, 1, -1]
+        cdrop                     # [1 or -1]
```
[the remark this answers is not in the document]
bitwalker: I think the issue is actually just that the `1` and `-1` are swapped. The comments here show the contents of the stack after applying the instructions on that line. Since we know in this branch of the containing `if.true` that `a_msb != b_msb`, we know that `b_msb == 0` is equivalent to `a_msb == 1`.

bitwalker:
Sorry @greenhat, this was supposed to be a draft PR still, since I hadn't finished adding tests yet. I've changed the status temporarily until that gets done.

bitwalker:
@greenhat This PR is a bit of a mess now, because there are a mix of changes in here related to loading intrinsics from disk, and that got tangled up with the changes to the driver and compiler. I'd like to merge this as-is (with the broken i32 intrinsics included), and open a separate PR with the fixes for the intrinsics, otherwise this is going to get messy fast. Once this PR gets merged, I'll open the follow-up PR with those fixes, but I'm going to use this branch as my base for those changes for now.

greenhat:
> NOTE: The sole failing test is the dlmalloc expect test because the compiled Wasm for dlmalloc doesn't match what's in the expect test. I don't know why this is failing in CI but not locally, even running the same toolchain, but maybe this is something fixed in one of your PRs?

It's weird that it does not fail locally. I got it disabled in my branch because IIRC it was failing down the pipeline anyway. Let's disable it for now.

As a side note, I disabled the `signed_arith` test as well since local env details leaked into `rodata` (local path and newlines).

greenhat (approved):

greenhat:
Approved.

PR #42, opened 2023-10-19
feat: implement clz/ctz instructions

### LINKED ISSUE #39: Implement clz/ctz instructions
These functions count the number of leading/trailing zeroes in the binary representation of an integral type, and are often used when using certain integer types as bitsets. In particular, these are used in the allocators that Rust programs use when compiled to Wasm, so this will be a blocker for compiling any Rust/Wasm module that uses the `alloc` crate (i.e. anything that uses `Box`, `Vec`, etc.). See #22 

This is the tracking issue for the implementation of these instructions, which has already been started, but was on hold while other tasks were being completed.

bitwalker:
This PR builds on changes in #37, and adds implementations of `clz` and `ctz` for integral types up to 32 bits.

In this PR, a couple of small tweaks/fixes as well:

* Add support for `loc_store` and `loc_storew`, which we use in one of the new intrinsics
* Fix up and unify formatting of MASM opcodes
* Automatically detect (and add to default link set) all intrinsic modules we define under `codegen/masm/intrinsics`
* Remove `rust-wasm-tests` from workspace members, since it is compiled and run with a separate build profile via `frontend-wasm`'s test suite. When added as a workspace member the profile `rust-wasm-tests` defines is ignored, so it's not going to do what we expect. We'll catch any issues with that crate when CI runs our tests anyway.

Closes #39


bitwalker:
NOTE: I'm adding some tests for this, as found some issues (not with the intrinsics, but somewhere in the emulator I think when testing the behavior, something to do with how memory and the operand stack interoperate, I'll fix it tomorrow when I have a moment)

greenhat:
I made #44 for Wasm translation.

greenhat (commented):
Looking good!
I could not check the MASM code of the intrinsics in my head due to its complexity. So having tests for them would be nice, but I'm recalling from the discord thread that these ops will be implemented in VM soon, so I guess we can afford to skip the tests for now.

I saw your note but missed the revert of the status to draft. I'll do another round when its ready.

PR #49, opened 2023-10-26
Rust to MASM compilation and semantic tests

### LINKED ISSUE #47: Rust to MASM compilation and semantic tests
In a separate crate (under midenc umbrella), establish a test suite for a set of Rust programs, that does the following:

* Compile a Rust program to MASM, checking against the expected Wasm, IR, and MASM code (compilation test). 
* Run the Rust program (native binary) against the compiled MASM program and check that on the same inputs they produce the same outputs (semantic test).
* Use property-based testing to get good coverage for the program inputs.

greenhat:
Close #47 

The idea for `CompilerTest` is to mix and match sources (Rust, Wasm, IR) and expected compiler output (IR, MASM) to test the whole pipeline(Rust->Wasm->IR->MASM) or any part of it (Rust->IR, Wasm->IR, IR->MASM, etc). 

The compiled MASM code is executed on the emulator and VM, and the results are compared with the native Rust code. The inputs are generated(fuzzed) via `proptest`. 

The Rust source for instructions test is generated in the runtime and compiled as with rustc. With macros, every test for an instruction is defined on one line. 
~~I added only a dozen instructions so far, mainly because not all instructions are implemented yet for the whole pipeline. I plan to add more instructions tests in the subsequent PRs.~~ 
**EDIT:** I've gone through all the IR ops and implemented almost all that have Rust counterparts and that fit the individual instruction type of test.
See https://github.com/0xPolygonMiden/miden-ir/blob/858528a1ce3dc0422b740bad8faca0f701a64869/tests/integration/src/rust_masm_tests/instructions.rs

The Rust source code for app tests is split into crates. For the included Fibonacci sequence app, in `rust-apps/fib` crate is the actual code of the app, and the `rust-apps-wasm/fib` crate is a wrapper for Wasm that calls the code from the first crate. In the semantic test, to get the Rust native (expected) results, the code from the first crate is called directly. This ensures that we use the same code for the semantic test that is compiled to Wasm. See https://github.com/0xPolygonMiden/miden-ir/blob/5ba0305f55b3c8565db5533434d3c35e9cd33a97/tests/integration/src/rust_masm_tests/apps.rs


greenhat:
@bitwalker, Could please take a look at the overall API design and crate layout? It's all stubbed out for now and I'd like to start implementing it.

Cargo.toml
```diff
@@ -10,7 +10,8 @@ members = [
   "hir-type",
   "tools/*",
   "frontend-wasm",
-  "tests/rust-wasm-tests"
+  "compiler-tests-rust-source",
```
bitwalker, line 13: *Nit*: I'd prefer to move these crates to the `tests` subdirectory, and give them simpler names, e.g. `tests/rust` and `tests/integration`; where the first is for testing compilation from Rust sources, and the latter is for integration tests of the compiler itself.

I'm open to suggestions as well, but would rather err on the side of broad strokes than getting too specific.

compiler-tests/src/comp_test.rs
```diff
@@ -0,0 +1,39 @@
+/// Compile to different stages (e.g. Wasm, IR, MASM) and compare the results against expected output
+pub struct CompTest {}
```
bitwalker, line 2: *Nit*: May as well call this `CompilerTest`, it's only 4 more characters.

compiler-tests/src/rust_masm_tests/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+use crate::execute_vm;
+use crate::CompTest;
+
+#[test]
+fn u32_div() {
+    // Test expected compilation artifacts
+    let mut test = CompTest::new();
+    test.rust_source("compiler-tests-rust-source", "div_u_app")
+        .expect_wasm(expect_file!["./expected/div_u.wat"])
+        .expect_ir(expect_file!["./expected/div_u.mir"])
```
bitwalker, line 18: *Question*: Is the expectation that we emit all of these artifacts for every program? I think we'll in general want to assert specific things about the output of each test program, rather than just expecting whatever the compiler emits (as that will be subject to regular change due to optimizations and such). 

For a subset of core fundamental programs I think it's fine to emit everything and be very precise about what each artifact should look like, much like the `lit` tests are designed to do for the IR, but for most programs I would expect us to be testing the output of running the program via the emulator or the VM, rather than caring about the contents of the `.wat`, `.hir`, or `.masm`. At the end of the day, we want to validate that executing the code does what it is supposed to do, and throw enough programs at it that we have confidence that we're not just testing the happy paths.

Speaking of `lit`/`filecheck`, I would like to make sure it is part of the integration test picture. It is primarily designed for testing transformations, but it can also be used to test changes in representation (e.g. `.hir` -> `.masm`). We have the framework for it set up, but its just waiting on `midenc` to be finished up (which I should have done by Monday).

bitwalker (approved):
Looks good overall, just have a few notes/questions in my review.

Cargo.toml
greenhat: Ha! Initially, I put them in `tests/compiler` and `tests/compiler-rust-source` but then I thought that since we use a "flat" folder structure for crates (e.g. `hir-*`), I should do the same for tests. 
I'll do `tests/integration` and `tests/rust`.

compiler-tests/src/rust_masm_tests/mod.rs
greenhat: > _Question_: Is the expectation that we emit all of these artifacts for every program? I think we'll in general want to assert specific things about the output of each test program, rather than just expecting whatever the compiler emits (as that will be subject to regular change due to optimizations and such).

Oh, no, this is the example of a "check everything" scenario (I should've made more). We can implement all sorts of checks in `CompilerTest`, like `contains`, `match_regex`, and whatnot, even non-text checks using IR nodes/trees. The source is not limited to Rust and could be loaded from Wasm or IR to test only part of the pipeline. In a test, we can pick only those checks that are interested or even none at all. The builder pattern was the wrong choice here; we don't need to shove it all into one line. Here is what it should look like:

```rust
test.rust_source(...);
test.expect_ir(expect_file!["./expected/file.mir"]);
``` 

or 


```rust
test.rust_source(...);
test.expect_ir_contains("...");
test.expect_masm_contains("...");
``` 

or skipping the Rust and starting from Wasm source:


```rust
test.wasm_source(...);
test.expect_ir_contains("...");
``` 

or even starting from IR:


```rust
test.ir_source(...);
test.expect_masm_contains("...");
```

> For a subset of core fundamental programs I think it's fine to emit everything and be very precise about what each artifact should look like, much like the `lit` tests are designed to do for the IR, but for most programs I would expect us to be testing the output of running the program via the emulator or the VM, rather than caring about the contents of the `.wat`, `.hir`, or `.masm`. At the end of the day, we want to validate that executing the code does what it is supposed to do, and throw enough programs at it that we have confidence that we're not just testing the happy paths.

I agree. Most of the programs will not have any checks of the Wasm, IR, or MASM code. I have high hopes on https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-smith to fuzz the hell out of it.

> 
> Speaking of `lit`/`filecheck`, I would like to make sure it is part of the integration test picture. It is primarily designed for testing transformations, but it can also be used to test changes in representation (e.g. `.hir` -> `.masm`). We have the framework for it set up, but its just waiting on `midenc` to be finished up (which I should have done by Monday).

Of course! I'm looking forward to start using it.

Cargo.toml
bitwalker: In general the structure is flat like you expected, but I think of `midenc` and the `hir*` crates as the primary focus of the repo, and thus keep those at the root (though `midenc` needs to be moved to the root now that the repo is renamed). On the other hand, the test crates, and supporting tools like `filecheck` are auxiliary, so nesting those in folders like `test` and `tools` keeps the focus on the "public" face of the compiler, and avoids too much clutter in the root due to incidental things like what tools we're building as part of our infrastructure and that sort of thing.

With multi-crate projects like this, there are a number of different ways to organize them, and they all have their merits (e.g. `rust-lang/rust`, `cranelift`), so I'm not too fixated on this specific structure, but at least now you know my rationale 😄

compiler-tests/src/rust_masm_tests/mod.rs
bitwalker: Got it, that sounds good to me! Definitely interested to see what we can do with `wasm-smith` and `proptest` to generate programs for testing!

greenhat:
@bitwalker Without #54, we cannot run our emulator under the`proptest`. I tried a couple of ideas(reference, and then Rc) to make a quick fix but with no luck. 
We should be able to run "real" masm in VM under `proptest`. But with `Program::to_program_ast` broken (missing local procedures), I cannot test it. I hope for the changes in your PR to help with that. Meanwhile, I'm looking into `Module::to_module_ast` to figure out if I can whip up some workaround.

bitwalker:
Sounds good, I'll get things sorted ASAP then

bitwalker:
I think once #37 gets merged, we can get this merged on top of that, but I think you have some fixes here above and beyond what's in that PR, so I want to see how the diff looks once #37 is in before we merge this.

greenhat:
> I think once #37 gets merged, we can get this merged on top of that, but I think you have some fixes here above and beyond what's in that PR, so I want to see how the diff looks once #37 is in before we merge this.

Sounds good! I will rebase it on top of #37 in the meantime.

tests/integration/expected/and_u32.masm
```diff
@@ -0,0 +1,611 @@
...
+          exec.realign_dw
+        end
+      end
+    end
+  end
+
+end
+
+export.entrypoint
+  u32.and
```
greenhat, line 605: Since i32 and mem intrinsics modules are always linked by `MasmCompiler::compile` we end up with a lot of dead code. Shouldn't we link them only when those functions are used?

bitwalker: It's a bit tricky, because they aren't called in a normal sense like regular functions, they are the result of lowering instructions in the IR to MASM. In order to detect their use, we have to check if any of the instructions backed by intrinsics are used. We can certainly be cleverer about this in the future, but in the short term it seemed likely that we would need both intrinsics modules in practice, so felt like a low priority to me. In the near future though, mainly when emitting programs, I would like to do a proper dead code elimination pass to strip out any unused/dead code.

greenhat: Makes sense. For some reason, I thought we would not want to run transformation passes on MASM, but yeah, DCE pass on MASM might be easier than trying to avoid generating a dead code.

bitwalker: It's easier to run transformations on the IR, but due to things like this, it becomes necessary sometimes to run transformations on the emitted MASM as well. The peephole optimization pass I've mentioned before is another example - since lowering from the IR isn't 100% perfect, we can further optimize the emitted MASM to improve the output of codegen without changing the semantics of the program. We are necessarily limited in what kinds of transformations we can do safely, but there is still plenty of room for optimization even at the MASM level.

codegen/masm/src/masm/program.rs
```diff
@@ -162,14 +164,8 @@ impl Program {
...
                 masm::ProcedureId::from_name(entry.function.as_str(), &entry_module_path);
-            let entry_name = masm::ProcedureName::try_from(
-                FunctionIdent {
-                    module: Ident::with_empty_span(entry_import.alias),
-                    ..entry
-                }
-                .to_string(),
-            )
-            .expect("invalid entrypoint function name");
+            let entry_name = masm::ProcedureName::try_from(entry.function.as_str())
```
bitwalker, line 167: **Question:** I'm not sure how this could work - we can't reference functions (other than locals) without the parent module name. Specifically the name associated with a module when it is `use`'d. For example, functions imported via `use intrinsics::i32` must be called like `i32::icmp`; while functions imported via `use intrinsics::i32->int32` would need to be called as `int32::icmp`, we can't just call `icmp` anywhere unless we're literally in the `intrinsics::i32` module.

codegen/masm/src/masm/program.rs
```diff
@@ -250,6 +246,15 @@ impl From<&hir::Program> for Program {
     }
 }
 
+impl fmt::Display for Program {
```
bitwalker, line 249: I think this is fine as a short-term thing, but this is missing the code that will go in the top-level `begin` block of a `Program` (which we're not currently representing here, but will be added very soon, since it is where we will need to place code for initializing memory and things of that nature). We'll also want to be careful that we don't treat this as a canonical representation of a `Program` in textual form, since Miden Assembly doesn't support multiple modules in a single file. I'm assuming we're using this for test assertions though, so for that purpose I think this is fine, just be aware that this format is likely to change, so if we use it in a bunch of test assertions, we'll have to update all of them when that happens (which may well be unavoidable until I get the `filecheck`/`lit` tooling updated).

codegen/masm/src/masm/program.rs
greenhat: I believe I made this change because `ProcedureName::try_from` does not accept `::` in the name. I was not sure if it was a bug so I made this workaround but forgot to document it.

codegen/masm/src/stackify/emitter.rs
```diff
@@ -721,7 +721,7 @@ impl<'a> MasmEmitter<'a> {
         };
         let pos = stack
             .find(&value)
-            .expect("value not found on operand stack");
+            .expect(format!("value {} not found on operand stack", value).as_str());
```
bitwalker, line 724: I wonder if it would be more convenient for us to modify `find` to raise a pretty error like this if the value isn't found. I should check and see if we ever handle the `None` case with anything other than a panic. We don't need to do that here, but if you have thoughts on it, let me know.

greenhat: Sure, that sounds great to me.

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -191,7 +192,10 @@ pub fn translate_operator(
         }
         Operator::I64ExtendI32U => {
             let val = state.pop1();
-            state.push1(builder.ins().zext(val, I64, span));
+            let u32_val = builder.ins().cast(val, U32, span);
```
bitwalker, line 195: I really think we're going to need to reconsider the impedance mismatch between how Wasm represents integral types and how the IR does it, or at least provide an additional set of APIs that are more akin to `transmute` than conversion. Right now, the code we emit when casting signed -> unsigned, or larger bitwidth to smaller must check that the value fits in the target type. However, due to the way Wasm represents things, this results in a lot of redundant code that simply doesn't need to exist, and it isn't code we can easily remove later (we might be able to do some of it with peephole optimization, but it will be fragile to the distance between the conversion in, and the conversion out of a given type).

We should probably discuss this on a call, because it'll be a lot to cover, but here's some ideas we could explore:

1. Modify the IR type system to be closer to Wasm, this has some downsides for non-Wasm frontends, but would at least make the code we generate from Wasm a lot more optimal
2. Modify the IR type system to have signless integral types, and/or provide a set of instructions that interpret the value of an integral type based on whether the instruction is a signed or unsigned variant (e.g. in addition to `add` which operates on `Type::I32` or `Type::U32` according to the type, we'd also have `add_s` and `add_u` which treat the bits of a `Type::I32` or `Type::U32` as signed or unsigned depending on the instruction rather than the value type). The type returned from such an instruction would always be the same as the input type.
3. Rather than changing the type system or instruction set, we would add a new instruction for transmuting between types. Transmutation would perform no range checks, and would do the bare minimum necessary to treat a value of a given source type as the desired target type. For example, `transmute i32 : u32` would result in no code being emitted during codegen.

Personally, I'm leaning towards 2. While 1 would be an option, it's a more significant change, and weakens the type system, which might be fine, but is a tradeoff. 3 feels like an incomplete hack really, but is probably the simplest option to help with the poor codegen output. The problem is that sign/zero-extension is also affected by this, so we'd either need to weaken the guarantees of those instructions, or make a theoretical `transmute` handle extension as well, so you could do things like transmute between `i32` and `i64`.

Thoughts?

codegen/masm/src/masm/program.rs
greenhat: Good catch on the missing entrypoint. Yes, it's for the test assertions only. I don't see how this implementation can be limited to tests, but I can probably switch it to `Debug` trait to emphasize its "internaliness".

frontend-wasm/src/module_translator.rs
```diff
@@ -10,16 +10,47 @@ use crate::sections_translator::{
...
     wasm: &[u8],
-    _config: &WasmTranslationConfig,
+    config: &WasmTranslationConfig,
     diagnostics: &DiagnosticsHandler,
 ) -> WasmResult<Module> {
+    translate_module_inner(wasm, config, diagnostics).map(|res| res.module)
+}
+
+/// Translate a sequence of bytes forming a valid Wasm binary into Miden IR program
+pub fn translate_program(
```
bitwalker, line 27: FYI: I'm planning to add attributes to HIR entities like `Program`, `Module`, and `Function`, so that we can, for example, decorate a `Function` with an attribute that indicates that it is the entrypoint for whatever `Program` it gets linked into.

I think that will make this function unnecessary, though there may be other reasons to have it for testing and such. In all other situations, this should be handled by `midenc-compile` though I think.

frontend-wasm/tests/expected/add.hir
```diff
@@ -0,0 +1,27 @@
+module noname
```
bitwalker, line 1: I think we'll want to replace these tests with `filecheck`/`lit` once that tooling is ready, since it will be a lot less fragile to changes, and we can be more precise about what we actually care about in the code that is emitted. I'm generally wary of tests like this which break if even the slightest change to the format/etc. is made. That said, having a few such tests is still useful.

Luckily it should be relatively painless to replace these tests when my changes to `filecheck` are done (hopefully next week), but something to keep in mind as you are writing more tests.

codegen/masm/src/masm/program.rs
bitwalker: Ah, I think it's the `ProcedureId` which is qualified with the module, the `ProcedureName` is unqualified. I think you're right about the change here, I'll have to confirm whether we're emitting the correct `ProcedureId` though.

tests/integration/src/compiler_test.rs
```diff
@@ -0,0 +1,379 @@
...
+        let session = default_session();
+
+        let entrypoint = FunctionIdent {
+            module: Ident::new(Symbol::intern("noname"), SourceSpan::default()),
+            function: Ident::new(
+                Symbol::intern(entrypoint.to_string()),
+                SourceSpan::default(),
+            ),
+        };
+        let hir_module = wasm_to_ir(&wasm_bytes, &session);
```
bitwalker, line 111: I think we're going to want to use the interface provided by `midenc-compile` here (and extend it as needed). Otherwise we risk testing something other than what will be generated by the compiler in practice.

We don't need to make that change in this PR, particularly if there are things missing that make that awkward currently, but if you could make a note of the things you'd like to have from the `midenc-compile` interface, I'll make sure to get those implemented ASAP and then update these test harnesses.

tests/integration/src/compiler_test.rs
```diff
@@ -0,0 +1,379 @@
...
+        eprintln!("{}", String::from_utf8_lossy(&output.stderr));
+        panic!("Rust to Wasm compilation failed!");
+    }
+    let wasm = fs::read(&output_file).unwrap();
+    fs::remove_file(&input_file).unwrap();
+    fs::remove_file(&output_file).unwrap();
+    return wasm;
+}
+
+fn wasm_to_wat(wasm_bytes: &Vec<u8>) -> String {
```
bitwalker, line 298: Just out of curiosity, how easy would it be for us to add a frontend for the `.wat` format? Definitely not a priority, but might be nice to have. 

I am going to take this code though and add it to `midenc-compile` to implement support for `wat` as an output format. I had been looking for that, but thought it would be more involved than this, so this is great!

tests/integration/src/felt_conversion.rs
```diff
@@ -0,0 +1,117 @@
...
+
+impl From<i64> for TestFelt {
+    fn from(t: i64) -> Self {
+        Self((t as u64).into())
+    }
+}
+
+// Reverse TestFelt to Rust types conversion
+
+impl From<TestFelt> for bool {
```
bitwalker, line 65: Most of these should use `TryFrom`, rather than `From`, since many of these conversions aren't safe. We for sure won't be able to get them merged into `winterfell` with `From` so its better if we stick to what we can get upstreamed (plus, since this is used in our test suite, we avoid accidentally truncating a value and thinking a result is correct when it isn't).

tests/integration/src/felt_conversion.rs
```diff
@@ -0,0 +1,117 @@
+use miden_core::Felt;
+use miden_core::StarkField;
+
+/// Wrapper around `Felt` that implements `From` for a bunch of types that are missing in Winterfell
+/// Should be moved to Winterfell eventually
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TestFelt(pub Felt);
```
bitwalker, line 7: So we'll have to use `TryFrom` for the signed types, since technically we can't represent negative values in a `Felt`. Obviously in practice we're encoding negative values by casting to unsigned, but I doubt the `winterfell` folks are going to be ok with merging conversions like that.

Unfortunately, that means that for signed types, we're going to simply be forced to either maintain a wrapper like this, or use `Felt::new(some_i32 as u32 as u64)` to handle representing signed types in general.

bitwalker:
Looks good in general! Just a few notes/questions, but you can go ahead and merge this once you get back to me on those (and once you've cherry-picked the commits from #63 so we can effectively merge both PRs at once).

frontend-wasm/src/code_translator/mod.rs
greenhat: My least favorite option is 1. I find the second option quite confusing and somewhat close to the first option, so I'm leaning toward the third option for now. Let's discuss it on a call.

frontend-wasm/src/module_translator.rs
greenhat: I only introduced it to pass the Wasm's start function as an entrypoint, and I did not like how it came out. It turned out kind of useless anyway since we can only get limited `main` function there. I'm going to remove it.

frontend-wasm/tests/expected/add.hir
greenhat: Sounds good.
Yes, those are pretty fragile, but the ease of re-generation kinda offsets it. As our formatting stabilizes, I expect them to break only when we change the compiler code that affects the output. I value their fragility when reviewing PRs, though. I want to see how the changes to the compiler affect the output.
Besides their primary function (golden tests, regressions, etc.) I'm using them to peek under the hood at the different stages of the compilation pipeline and to know what code the semantic tests are executing. It's the sort of thing you put in debug prints to figure out what is happening when things are going south. But in this case, it's already available. I also often open wat, hir, and masm files side by side to follow the code changes through the compilation process. The above probably only works on relatively small code sizes, which is true for most smart contracts. As for their precision, instead of dumping the whole module, they can be narrowed to function/block with extra filtering on top.
With all that said, I'm eager to try the lit tests and see how they shape my workflow.

tests/integration/src/compiler_test.rs
greenhat: I 100% agree. I wanted to use `midenc` here initially, but I could not figure out how to get out IR and MASM without the disk and parsing text files. But, since the parsing of MASM by the Miden VM is a test of our emitter, and I'm using the disk for Rust compilation anyway, I should at least get the MASM from `midenc`. The IR part of the tests might be going away soon in the light of the upcoming `lit` tests, so I leave them as is for now if I could not figure out an easy way of getting the IR from the `midenc`.
I'd like to do all of the above in a separate PR if you don't mind since this PR is getting big, and I have other PRs stacked on top of it.

tests/integration/src/compiler_test.rs
greenhat: The easiest way would be to use https://crates.io/crates/wat to convert `wat` to binary `wasm` and use the existing frontend.

tests/integration/src/felt_conversion.rs
greenhat: I'd stick with this wrapper and don't try to merge them into the `winterfell` since we need a uniform way of going from Rust types to `Felt` and back for all integral types we want to use in `instruction` tests. See https://github.com/0xPolygonMiden/miden-ir/blob/4496a697eb98798ed41ff53a87f51e0ae1bcf336/tests/integration/src/rust_masm_tests/instructions.rs#L90-L151 . Macros are written to be type-agnostic and work for any passed type. Any supported Rust type is handled the same way in the macros.
The doc string is misleading since I wrote it before discovering the above requirement.

tests/integration/src/felt_conversion.rs
greenhat: Please take a look at my comment above.

frontend-wasm/tests/expected/add.hir
bitwalker: Makes sense! I'm certainly in favor of having a mix of tests, including a set like this. We can discuss further once there is some prior art in the repo for `filecheck` so see if that changes anything from your perspective. Having the compiler executable and debugger will provide a lot of the utility you seek as well, particularly once I add some more options for debugging specific passes by emitting only changes/modifications to stdout (if you are familiar with MLIR, it does this, and it's one of the more useful things they've built into the tooling there).

I'd like to have something more akin to MLIR's conversion infrastructure as well, but that's something I'll explore down the road when I have time to clean up/refactor the IR a bit.

tests/integration/src/compiler_test.rs
bitwalker: > I'd like to do all of the above in a separate PR if you don't mind since this PR is getting big, and I have other PRs stacked on top of it.

For sure! No problem :)

tests/integration/src/compiler_test.rs
bitwalker: Makes sense, I'll keep that in mind for the future

tests/integration/src/felt_conversion.rs
bitwalker: That's fine, might be worth removing that comment in the source then. I guess some of the conversions might still be worth upstreaming, but we don't need to make any effort to do that IMO.

frontend-wasm/tests/expected/add.hir
greenhat: Great! I've tried to port a simplified MLIR's conversion infra to pliron. I especially loved how the legality of ops kind of drove the conversion process.

PR #50, opened 2023-10-30
feat: implement compiler driver, update midenc

bitwalker:
This PR implements most of the compiler driver (the part of `midenc` which actually coordinates and configures compilation, amongst other tasks.

**NOTE:** This is currently mostly complete, but there are a few small things I'm finalizing/cleaning up, namely the handling of custom pass pipelines which is of importance in our testing infrastructure. There are also a few things I've stubbed out, but which are not actually implemented yet, e.g. the `debug`, `exec`, and `run` commands.

@greenhat I'm opening this up as a draft so you can start review and raise any questions you have, just be aware there are a few things that may change slightly before this is finalized.

I plan to have this finished up by EOD, Oct 30.


greenhat (approved):
Looking great! No questions so far.

PR #53, opened 2023-10-31
fix: unreachable raises assertion during stackification

### LINKED ISSUE #52: Unreachable inst causes `unhandled terminator in non-terminator context` error in `stackify` pass
Error:
```
thread 'rust_masm_tests::apps::fib' panicked at codegen/masm/src/stackify/pass.rs:1389:9:
unhandled terminator in non-terminator context: PrimOp(PrimOp { op: Unreachable, args: EntityList { index: 0, unused: PhantomData<miden_hir::value::Value> } })
```

When the Rust code `9999/n` compiles to the following Wasm:
```wat
(module
  (func $fib (;0;) (type 0) (param i32) (result i32)
    block ;; label = @1
      local.get 0
      i32.eqz
      br_if 0 (;@1;)
      i32.const 99999
      local.get 0
      i32.div_u
      return
    end
    unreachable
    unreachable
  )
  (memory (;0;) 16)
  (global (;1;) i32 i32.const 1048576)
  (global (;2;) i32 i32.const 1048576)
  (export "memory" (memory 0))
  (export "fib" (func $fib))
)
```

Which in turn translates to the following IR:
```
module noname

const $0 = 0x00100000;

pub fn fib(i32) -> i32 {
block0(v0: i32):
    v2 = eq v0, 0 : i1;
    v3 = cast v2 : i32;
    v4 = neq v3, 0 : i1;
    condbr v4, block2, block3;

block1(v1: i32):

block2:
    unreachable ;

block3:
    v5 = const.i32 99999 : i32;
    v6 = cast v5 : u32;
    v7 = cast v0 : u32;
    v8 = div.checked v6, v7 : u32;
    v9 = cast v8 : i32;
    ret v9;
}
```

The `stackify` pass fails with the error above.

bitwalker:
This closes #52, and ensures the same issue won't happen again by removing the (no longer necessary) assertion that triggered it in the first place.

### DIFF
--- a/codegen/masm/src/stackify/emitter.rs
+++ b/codegen/masm/src/stackify/emitter.rs
@@ -1342,11 +1342,6 @@ impl<'a> MasmEmitter<'a> {
 
     /// Emit code for a non-terminator instruction, which consumes and produces values on the operand stack
     fn emit_op(&mut self, inst: hir::Inst, ix: &hir::Instruction, stack: &mut OperandStack) {
-        assert!(
-            !ix.opcode().is_terminator(),
-            "unhandled terminator in non-terminator context: {:?}",
-            ix
-        );
         match ix {
             Instruction::GlobalValue(op) => self.emit_global_value(inst, op, stack),
             Instruction::UnaryOpImm(op) => self.emit_unary_imm_op(inst, op, stack),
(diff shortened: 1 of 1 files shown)


greenhat (approved):
LGTM! Thank you for a quick fix!

PR #58, opened 2023-11-07
Always create the destination branch argument for the sentinel value when looking for the value definition in the predecessors

### LINKED ISSUE #40: Implement emitting zero at the start of the block in `frontend_wasm::ssa::emit_zero`
Extracted from https://github.com/0xPolygonMiden/miden-ir/pull/22#discussion_r1365584520

>The DataFlowGraph provides insert_inst as a primitive, which allows you to specify an InsertionPoint, which can be either the start/end of a Block, or before/after an Inst, but I haven't exposed that in a very convenient way in the DefaultInstBuilder. All of the InstBuilder implementations actually use this under the covers, so its really just a matter of surfacing it better in one of the builders.

### LINKED ISSUE #51: Missing definition for SSA value (v6)
```rust
#[no_mangle]
pub fn fib(n: u32) -> u32 {
    let mut a = 2;
    while a < n {
        a = a * 3;
    }
    a
}
```

Compiles to the following Wasm:

```wat
(module
  (func $fib (;0;) (type 0) (param i32) (result i32)
    (local i32)
    i32.const 2
    local.set 1
    loop (result i32) ;; label = @1
      block ;; label = @2
        local.get 1
        local.get 0
        i32.lt_u
        br_if 0 (;@2;)
        local.get 1
        return
      end
      local.get 1
      i32.const 3
      i32.mul
      local.set 1
      br 0 (;@1;)
    end
  )
  (memory (;0;) 16)
  (global (;1;) i32 i32.const 1048576)
  (global (;2;) i32 i32.const 1048576)
  (export "memory" (memory 0))
  (export "fib" (func $fib))
)
```

Which translates to the following IR:

```
module noname

const $0 = 0x00100000;

pub fn fib(i32) -> i32 {
block0(v0: i32):
    v2 = const.i32 0 : i32;
    v3 = const.i32 2 : i32;
    br block2(v3);

block1(v1: i32):

block2(v5: i32):
    v7 = cast v5 : u32;
    v8 = cast v6 : u32;
    v9 = lt v7, v8 : i1;
    v10 = cast v9 : i32;
    v11 = neq v10, 0 : i1;
    condbr v11, block4, block5;

block3(v4: i32):

block4:
    v12 = const.i32 3 : i32;
    v13 = mul.wrapping v5, v12 : i32;
    br block2(v13);

block5:
    ret v5;
}
```

v6 is undefined.

greenhat:
Close #51

Close #40 since the only use case for `emit_zero` at the start of the block was in the code removed by this PR.

**This PR is ready for review but is built on top and is intended to be merged after #49.**

This removes the remnants of the previously removed `ValueData::Alias` variant. It does not matter if we find the definition in the predecessors since we cannot swap all sentinel value uses with it. So, this fix always makes the sentinel value into a block argument.


tests/integration/expected/fib.hir
```diff
@@ -8,13 +8,29 @@ global external @gv2 : i32 = $0 { id = 2 };
...
+block2(v3: i32):
+    v4 = const.i32 3 : i32;
+    v5 = cast v3 : u32;
+    v6 = cast v4 : u32;
+    v7 = gt v5, v6 : i1;
+    v8 = cast v7 : i32;
+    v9 = neq v8, 0 : i1;
+    condbr v9, block4, block5;
+
+block3(v2: i32):
```
bitwalker, line 24: Is there any way to avoid emitting these useless blocks? They don't seem to be referenced by anything, and they contain no instructions, which seem like something we should be able to catch quite easily. Right now IR like this will fail our validator pass because we assert that there are no empty blocks (dead blocks are fine, but empty ones are technically invalid, especially if they are present in the CFG).

bitwalker (approved):

bitwalker:
Once the conflict is resolved, we can get this merged

tests/integration/expected/fib.hir
greenhat: I don't see an easy way of doing this in the frontend.
The `block1` is the "exit" block created right after the entry block. We branch there on the `end` op that represents the end of the function. In this code example, we always exit the function via the "early" `return` op. 
The `block3` is the "next" block after the loop block created on the "loop" op. We branch to it when we exit the loop "normally" by not branching on the next iteration. In this code example, we always branch on the next iteration and never exit the loop "normally" but only via `return`.

Can we use a special pass to eliminate such blocks that the frontend would call before returning an IR? OTOH, a DCE pass should be able to eliminate them.

bitwalker: One trivial way to handle this, if a more precise method isn't viable, is to do a simple linear scan of all blocks in the function at the end of translating Wasm to the IR, and remove any blocks which are empty. Since it appears that in the Wasm translation, an empty block only appears when it is unreachable, it should be safe to remove such blocks without computing the CFG/checking for reachability. 

Empty blocks are not valid in the IR, so writing a generic pass to remove such blocks doesn't make much sense, but even if we were to do so, such a pass would require a CFG analysis to determine whether that block is reachable or not (because a reachable empty block is definitely an error). That's more expensive to compute, and if we can avoid that overhead, all the better. In practice, the `InlineBlocks` pass does this kind of thing already, but all of our analyses/passes assume that the IR has basic validity properties upheld (blocks aren't empty, and have a valid terminator, and that uses of SSA values are dominated by their definition).

greenhat: Good idea! I was way too fixated on the strictness of the "empty blocks not being a successor to any block" rule. I made https://github.com/0xPolygonMiden/compiler/issues/63 and will do it in my next PR if you don't mind.

bitwalker: Sounds good!

bitwalker:
Something got messed up here, probably due to merging #37, but GitHub thinks this branch contains commits that are now in `main` (just with different commit hashes). If you want, just cherry-pick your commits from this branch on the #49 branch, and close this PR, and then you can just merge #49 with everything, since both PRs are approved and you were going to merge this into that one anyway.

greenhat:
I'd rather avoid messing with #49, so I made a new  #66 in place of this one.

PR #62, opened 2023-11-16
test/fix: add tests for, and address bugs with i32 icmp

bitwalker:
NOTE: I'm still adding more tests, looking for more bugs, so this shouldn't be merged yet, but can be reviewed. As I add to it, I'll do so in new commits to aid in re-review.

### DIFF
--- a/codegen/masm/intrinsics/i32.masm
+++ b/codegen/masm/intrinsics/i32.masm
@@ -21,19 +21,24 @@ export.overflowing_add # [b, a]
     # is `a` signed?
     dup.2 exec.is_signed # [is_a_signed, is_b_signed, b, a]
 
+    # do both operands have the same sign?
+    #
+    # NOTE: We refer to `is_b_signed` as `is_signed` after this,
+    # because if `is_same_sign` is true, then `is_signed` reflects
+    # whether both `a` and `b` are signed. If `is_same_sign` is false,
+    # then overflow is not possible, and the value of `is_b_signed`
+    # will have no effect on the result
+    dup.1 eq # [is_same_sign, is_signed, b, a]
+
     # compute result
     movup.3 movup.3 u32wrapping_add
 
     # is `result` signed?
-    dup.0 exec.is_signed # [is_result_signed, result, is_a_signed, is_b_signed]
+    dup.0 exec.is_signed # [is_result_signed, result, is_same_sign, is_signed]
 
-    # if `b` and `result` have the same sign, which is different than the sign of `a`,
-    # then overflow has occurred. In all other cases, the result is valid
-    movup.3         # [is_b_signed, is_result_signed, result, is_a_signed]
-    dup.1 eq       # [do_b_and_result_have_same_sign, is_result_signed, result, is_a_signed]
-    swap.1 movup.3 # [is_a_signed, is_result_signed, do_b_and_result_have_same_sign, result]
-    neq             # [do_a_and_result_have_different_sign, do_b_and_result_have_same_sign, result]
-    and             # [overflowed, result]
+    # if both operands have the same sign, and the result differs, overflow has occurred
+    movup.3 neq # [signs_differ, result, is_same_sign]
+    movup.2 and # [overflowed, result]
 end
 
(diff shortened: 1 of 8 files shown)


bitwalker:
@greenhat This PR will contain any fixes/tests for the i32 intrinsics. It's based on the branch from #37, so that will need to be merged first.

codegen/masm/intrinsics/i32.masm
```diff
@@ -21,19 +21,18 @@ export.overflowing_add # [b, a]
     # is `a` signed?
     dup.2 exec.is_signed # [is_a_signed, is_b_signed, b, a]
 
+    # do both operands have the same sign?
+    dup.1 eq # [is_same_sign, is_signed, b, a]
```
greenhat, line 25:
```suggestion
    dup.1 eq # [is_same_sign, is_b_signed, b, a]
```

codegen/masm/intrinsics/i32.masm
```diff
@@ -21,19 +21,18 @@ export.overflowing_add # [b, a]
...
+
     # compute result
     movup.3 movup.3 u32wrapping_add
 
     # is `result` signed?
-    dup.0 exec.is_signed # [is_result_signed, result, is_a_signed, is_b_signed]
+    dup.0 exec.is_signed # [is_result_signed, result, is_same_sign, is_signed]
```
greenhat, line 31:
```suggestion
    dup.0 exec.is_signed # [is_result_signed, result, is_same_sign, is_b_signed]
```

codegen/masm/src/tests.rs
```diff
@@ -413,3 +434,470 @@ fn stackify_sum_matrix() {
...
+            )
+            .freeze(),
+        )
+        .expect("failed to load intrinsics::i32");
+
+    let zero = Felt::new(0i32 as u32 as u64);
+    let neg_one = Felt::new(-1i32 as u32 as u64);
+    let one = Felt::new(1i32 as u32 as u64);
+    let max = Felt::new(i32::MAX as u32 as u64);
+    let min = Felt::new(i32::MIN as u32 as u64);
```
greenhat, line 457: As a side note, `proptest` should be a good fit for these kinds of tests. Its strategy for integer types covers such edge cases.

greenhat (approved):
LGTM!

codegen/masm/intrinsics/i32.masm
bitwalker: I wanted to make it clear that the purpose of that value after computing `is_same_sign` is to indicate whether `a` and `b` are signed (when `is_same_sign` is true). When `is_same_sign` is false, it's irrelevant whether `is_signed` refers to `a` or `b` specifically, since when the signs differ, overflow is impossible.

That said, if you found it confusing, I'm happy to commit this suggestion. What do you think?

codegen/masm/src/tests.rs
bitwalker: For sure, I mainly wanted to get a few sanity tests implemented so I can unblock you on these intrinsics, but I plan to revisit many of these tests to do property-based testing :)

codegen/masm/intrinsics/i32.masm
greenhat: Well, after the explanation, it makes sense. Although it still requires more cognitive overhead, I'm ok with it.

bitwalker: I'll rewrite the comments here to make it clearer upon first reading - I'd rather not risk it be confusing.

bitwalker:
@greenhat This PR is ready to merge. I've rewritten the intrinsics tests with `proptest`, as you suggested, and we now have coverage for the entire `intrinsics::i32` module tests-wise (a few functions which are basically simple wrappers are not explicitly tested, but their interesting behavior _is_ tested, e.g. `wrapping_add` and `checked_add` are wrappers for `overflowing_add` which do something appropriate with the `overflowed` flag). 

Once we get this merged, we should be good to go to start working with Wasm modules that have signed `i32` ops!

PR #65, opened 2023-11-17
feat: add attributes to the ir

bitwalker:
Attributes provide a way to decorate IR entities with metadata that can be used to guide compilation. This is implemented in the parser as well, but only for functions at this time (since no other entities carry attributes yet).

In this commit, the only interesting use we care about is identifying the entrypoint for a program, which can be done by decorating a Function with the `#[entrypoint]` attribute. Alternatively, one can manually specify the entrypoint function via Program, assuming it has the appropriate linkage and calling convention, but the default will be to search for the `#[entrypoint]` attribute.

This commit also reworks some of the MASM IR structures to account for how the entrypoint is tracked, as well as where we store the code for the top-level code block for a program (i.e. where global memory init will take place, as well as the code responsible for invoking the user-controllable entrypoint).

---

@greenhat If the tests pass, this PR can be merged, otherwise I'll fix any issues with it later today, but I wanted to get it up ASAP. This is the mechanism by which we'll control what functions are marked as the entrypoint, and it also modifies the way modules/programs get emitted, so it probably affects the test output to some degree. Eventually we'll use attributes to control other things (like marking a function as (no)inline, (no)optimize, etc.).


bitwalker:
@greenhat This is ready to merge once you're satisfied there aren't any obvious issues. I restructured things a bit in the MASM IR, and my next PR (which will be stacked on this one) will update the `miden-assembly` crate and integrate the changes there, as well as some things I kicked down the road until I got that done.

hir/src/module.rs
```diff
@@ -353,6 +353,17 @@ impl Module {
         self.globals.find(name).map(|gv| self.globals.get(gv))
     }
 
+    /// Find the first function in this module marked with the `entrypoint` attribute
+    pub fn entrypoint(&self) -> Option<FunctionIdent> {
```
greenhat, line 357: Should we forbid the module to add a function with an entrypoint attribute if the module already has one? To avoid multiple functions with an entrypoint attribute. I see the downside of needing an extra step to swap an entrypoint function, but other than that, it seems fine. Although it might be hard to enforce this on transformation passes.

greenhat (approved):
Looks great! I have only one question.

hir/src/module.rs
bitwalker: I think we can make this a job for the linker, since the same problem exists if there are entrypoints declared in multiple modules. Additionally, `Program` allows one to manually override what function is the entrypoint, and in that situation, we don't need to raise an error if there are multiple functions with the `entrypoint` attribute, since we're being told which function to use. So if we have the linker handle this, we can keep the check in one place, and also be able to handle the case where an entrypoint was specified manually.

We could add a module validation rule though that raises a warning if there are multiple entrypoints in a module, since even if it doesn't end up being an error due to specifying an entrypoint, it doesn't make sense to define multiple entrypoints in a single module anyway (unless we add something like Rust's `cfg` attributes, and you have multiple entrypoints that are mutually exclusive based on configuration).

PR #68, opened 2023-11-18
chore: update miden-assembly to next

bitwalker:
**NOTE:** This is intended to be merged after #65.

This updates our `miden-assembly` dependency, and handles all of the changes to the instruction set, as well as cleans up a few places that were hacky because of missing APIs in `miden-assembly`.

There are still a couple changes to make once this is merged (fleshing out support for indirect calls and assertions with error codes; creating `.masl` files when requested), but I'll do that separately to keep this PR smaller.

### DIFF
--- a/hir/src/asm/builder.rs
+++ b/hir/src/asm/builder.rs
@@ -262,23 +262,35 @@ impl<'a> MasmOpBuilder<'a> {
     }
 
     /// Pops the top element on the stack, and traps if that element is != 1.
-    pub fn assert(mut self) {
-        self.build(self.ip, MasmOp::Assert);
+    pub fn assert(mut self, error_code: Option<u32>) {
+        let op = error_code
+            .map(MasmOp::AssertWithError)
+            .unwrap_or(MasmOp::Assert);
+        self.build(self.ip, op);
     }
 
     /// Pops the top element on the stack, and traps if that element is != 0.
-    pub fn assertz(mut self) {
-        self.build(self.ip, MasmOp::Assertz);
+    pub fn assertz(mut self, error_code: Option<u32>) {
+        let op = error_code
+            .map(MasmOp::AssertzWithError)
+            .unwrap_or(MasmOp::Assertz);
+        self.build(self.ip, op);
     }
 
     /// Pops the top two elements on the stack, and traps if they are not equal.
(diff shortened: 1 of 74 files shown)


hir/src/asm/builder.rs
```diff
@@ -1865,30 +1939,24 @@ fn apply_op_stack_effects(
...
-        | MasmOp::U32CheckedDivMod
-        | MasmOp::U32CheckedShl
-        | MasmOp::U32CheckedShr
-        | MasmOp::U32CheckedRotl
-        | MasmOp::U32CheckedRotr
-        | MasmOp::U32CheckedMin
-        | MasmOp::U32CheckedMax
+        MasmOp::U32Lt | MasmOp::U32Gt | MasmOp::U32Lte | MasmOp::U32Gte => {
+            let rty = stack.pop().expect("operand stack is empty");
+            let lty = stack.pop().expect("operand stack is empty");
```
greenhat, line 1944:
```suggestion
            let rty = stack.pop().expect("failed to pop right operand: stack is empty");
            let lty = stack.pop().expect("failed to pop left operand: stack is empty");
```

hir/src/asm/isa.rs
```diff
@@ -225,6 +233,14 @@ pub enum MasmOp {
     MemStorew,
     /// Same as above, but the address is given as an immediate
     MemStorewImm(u32),
+    /// TODO
```
greenhat, line 236:
```suggestion
    /// Read two sequential words from memory starting at `a`, overwriting the first two words on the stack,
    /// and advancing `a` to the next address following the two that were loaded
    /// [C, B, A, a] <- [*a, *(a + 1), A, a + 2]
```

hir/src/asm/isa.rs
```diff
@@ -225,6 +233,14 @@ pub enum MasmOp {
...
+    MemStream,
+    /// TODO
```
greenhat, line 238:
```suggestion
    /// Pops the next two words from the advice stack, overwrites the
    /// top of the operand stack with them, and also writes these words
    /// into memory at `a` and `a + 1`
    ///
    /// [C, B, A, a] <- [*a, *(a + 1), A, a + 2]
```

greenhat (approved):
Looking great!

PR #79, opened 2023-12-11
Miden SDK proof-of-concept using Wasm component model

### LINKED ISSUE #72: PoC account code compilation (WIT version)

greenhat:
Close #72 

This PR adds a WIT implementation of Miden SDK and a basic wallet test using the Wasm component model.

### Marking `call`-able functions (account type) 
I chose to mark the whole Wasm component as having `call`-able functions. The benefits vs. marking individual functions as `call`-able are:
- Data segments contain the data for all functions in the component, so we have lower chances of loading unwanted data;
- Marking the individual functions can be done only via an attribute in the doc string, which is more error-prone than marking the whole component. Plus, I'm pretty sure we'll want to treat the account code specially, so we'll have to mark the whole component anyway.
- The whole component is marked via exporting a "marker" interface (see `account` interface defined in `miden.wit`(SDK) and used in `basic-wallet.wit`). Alternatively, we can use metadata in the Cargo.toml where will define some properties of the **Miden** component anyway.

Wasm components are limited to one exported world, so we cannot have both `call`-able and `exec`-able functions in the same component with this approach. That is the drawback of this approach.
The helper function are expected to be linked as a regular Rust crate to avoid the CanonicalABI overhead when called.

###  Wasm `resource` does not play well with `call`, so we should not use it
I found a problem that I believe prevents us from using the `resource` for our types. As an experiment, I made `Asset` a `resource` and since the implementation is opaque ([see](https://github.com/0xPolygonMiden/compiler/blob/f8c10f412265b9e814ef249dfa4a22e06a0bacbd/sdk/src/lib.rs)) and not on the WIT "level" ([see](https://github.com/0xPolygonMiden/compiler/blob/f8c10f412265b9e814ef249dfa4a22e06a0bacbd/sdk/wit/miden.wit#L93)) there is no knowledge of the size of the `Asset` instance in the compiled Wasm component. The `resource` handle is a pointer to the instance. Thus making it impossible to pass the `resource` value via a `call`. 
As I understood from the generated code, the `resource` is passed as a pointer to the instance, and since its methods are implemented in the Wasm component where the `resource` instance "resides," they can access its memory directly no matter from what component they were called.

### Use "library" Wasm component for note instead of a "command"
When using "command" Wasm component the [runtime](https://github.com/rust-lang/rust/blob/master/library/std/src/rt.rs) is quite sizable and pulls in arg parsing and thread-related setup that we don't need. So I decided to use a "library" Wasm component instead that is implementing a `note` interface ([see](https://github.com/0xPolygonMiden/miden-ir/blob/fa776b6a82cda2616ffb65017a130c7dcc130279/tests/rust-apps-wasm/basic-wallet/p2id-note/src/lib.rs#L14-L26)).


greenhat:
@bitwalker Following our discussion, I removed the component with the account's helper functions.

sdk/wit/miden.wit
```diff
@@ -0,0 +1,132 @@
...
+    /// as the faucet_id is included in the description of the non-fungible asset and this is guaranteed
+    /// to be different as per the faucet creation logic. Collision resistance for non-fungible assets
+    /// issued by the same faucet is ~2^95.
+    variant asset {
+        fungible(fungible-asset),
+        non-fungible(non-fungible-asset),
+    }
+
+    /// Inputs of the currently executed note (4 words = 16 felts)
+    type note-inputs = tuple<felt, felt, felt, felt, felt, felt, felt, felt, felt, felt, felt, felt, felt, felt, felt, felt>;
```
bitwalker, line 94: The inputs are variably-sized, i.e. there are only as many items in the inputs as there are actual inputs. I would make this a list/array of `felt`, with a note that there are never more than 16 inputs provided.

sdk/Cargo.toml
bitwalker, line 1: This crate feels like it belongs under `tests`, or maybe `examples` or something (though it's not really the latter either). The "real" SDK isn't going to live in this repo, so this is basically just a place to sketch out its design currently, and in an effort to keep our repo in some kind of sane hierarchy, I think we should treat this crate as a test crate for now. I'm open to alternatives though, let me know what you think

bitwalker (changes requested):
A couple small notes, but I think this looks good so far!

We discussed the following on our compiler call today, but just to document the decisions somewhere relevant:

1.) The need to avoid `resource` at the moment is due to missing features in the VM, for which we have an outstanding proposal that may or may not enable the use of `resource` in the future - but for the time being it can't be used.

2.) As you noted, all public exports of a component should be `call`ed, and we formalized that decision today (though it isn't final if we decide to revisit that in the future for some reason).

sdk/wit/miden.wit
greenhat: Agree. I modeled it way too closely to the existing kernel MASM API. Fixed in https://github.com/0xPolygonMiden/compiler/pull/79/commits/c0d3aec94ce4665bb7594a4279d875e56c1a509d

sdk/Cargo.toml
greenhat: Good point! I agree. I put it there when I experimented with the Rust-based approach, although I felt it was not so fitting. Then, I just got used to the place. :) Moved to `tests/rust-apps-wasm/sdk` in https://github.com/0xPolygonMiden/compiler/pull/79/commits/e46205782a55458fff1f30d08474f5d9857afa2b

bitwalker (approved):
LGTM!

Issue #80, opened 2023-12-18
Wasm component model (CM) support in Wasm frontend (tracking)


greenhat:
## Implementation details

Translate the Wasm CM `world` into the IR `Module`. Since a non-trivial Wasm component has a complex nested architecture (see [basic-wallet example](https://github.com/0xPolygonMiden/miden-ir/blob/8a602ca7fcd2b792c6baf3fbe05dc45d89650e8d/tests/integration/expected/sdk_basic_wallet/basic_wallet.wat#L1-L6379) that has 3 components, 4 component instances, 3 core modules and 5 core module instances) it makes sense for the frontend to encapsulate all this complexity and inline everything to produce a single IR `Module`.
The Wasm CM `export` and `import` top-level directives + [canonical definitions](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#canonical-abi) (`canon`) to be translated and stored in the IR `Module` for every imported(external) and exported IR module function.

Here is how the IR `Type` variants for `list` and `variant` might look like:

```rust
pub enum Type {
    ...
    /// An enum variant type
    Variant(VariantType),
    /// A list of values of the same type
    List(Box<Type>),
}

/// Variants are close to Rust `enum` declarations where a value is one of many
/// cases and each case has a unique name and an optional payload associated
/// with it.
#[derive(Clone, Hash, Eq, PartialEq, Debug)]
pub struct VariantType {
    /// The representation to use for this type
    pub(crate) repr: TypeRepr,
    /// The computed size of this struct
    pub(crate) size: u32,
    /// The list of cases that this variant can take.
    pub cases: Box<[VariantCase]>,
}

/// One case of a `variant` type which contains the name of the variant as well
/// as the payload.
#[derive(Clone, Hash, Eq, PartialEq, Debug)]
pub struct VariantCase {
    /// Name of the variant, unique amongst all cases in a variant.
    pub name: String,
    /// Optional type associated with this payload.
    pub ty: Option<Type>,
}

```

## Implementation plan

To ease the review process and get the feedback early, I plan to split the Wasm CM support implementation into the following stages/PR sets:

```[tasklist]
### Tasks
- [ ] https://github.com/0xPolygonMiden/compiler/issues/88
- [ ] https://github.com/0xPolygonMiden/compiler/issues/89
- [ ] https://github.com/0xPolygonMiden/compiler/issues/111
- [ ] https://github.com/0xPolygonMiden/compiler/issues/90
```

[...]

PR #82, opened 2023-12-20
[1/3] Wasm CM support in frontend-wasm. Preparation.

greenhat:
#80 

Since a Wasm component module (CM) support in the frontend expected to add a significant amount of code, I'd like to split it in the multiple PRs to ease the review process.
This is the first PR in the series of stacked PRs. I prefixed the PR title [1/x] to highlight that.
This PR is based on #79 and should be merged after it.

This PR only moves the existing module-related frontend code into the separate top-level module `module`. No Wasm CM related code is added in this PR.


bitwalker (approved):
LGTM

PR #84, opened 2023-12-27
[2/3] Wasm CM support in frontend-wasm. Integration tests preparation (lazy IR compilation)

greenhat:
#80

This PR is the second in the stacked PRs and should be merged after #82 

This PR splits Wasm and IR compilations and unifies the compilation pipeline, making all tests go through the same IR passes.


frontend-wasm/Cargo.toml
```diff
@@ -20,7 +20,7 @@ thiserror.workspace = true
 smallvec.workspace = true
 log.workspace = true
 anyhow.workspace = true
-wasmparser = "0.107"
```
bitwalker, line 23: Any reason to keep this pinned to the build version vs minor? The former would've still installed 0.118.1 with a `cargo update` even though it mentions an older version. I don't have strong feelings either way, but I tend to pin to the minor version unless stability is a problem with a given crate.

bitwalker (approved):

frontend-wasm/Cargo.toml
greenhat: Minor would be enough here. I was unhappy with the breaking change below, so I was extra cautious.

PR #86, opened 2023-12-30
fix: improper handling of inlined blocks in inline-blocks transform

### LINKED ISSUE #85: `InlineBlocks` pass crashes with `(signal: 11, SIGSEGV: invalid memory reference)`
Discovered in #84 in `test_rust_comp::rust_array()` test.

Un-ignore the test to reproduce.

The IR passes are defined in https://github.com/0xPolygonMiden/compiler/blob/430ef78692174b28de18830e36ffaad1a91462e2/tests/integration/src/compiler_test.rs#L495-L504. The crash goes away if `InlineBlocks` pass is omitted(commented out).

I've also reproduced it with rustc 1.74.1 and the latest nightly. I suspect the `unsafe` code that operates the `intrusive_collections`.

FWIW, here the Wasm:
```wasm
(module
  (type (;0;) (func (param i32 i32) (result i32)))
  (type (;1;) (func (result i32)))
  (func $sum_arr (;0;) (type 0) (param i32 i32) (result i32)
    (local i32)
    i32.const 0
    local.set 2
    block ;; label = @1
      local.get 1
      i32.eqz
      br_if 0 (;@1;)
      loop ;; label = @2
        local.get 0
        i32.load
        local.get 2
        i32.add
        local.set 2
        local.get 0
        i32.const 4
        i32.add
        local.set 0
        local.get 1
        i32.const -1
        i32.add
        local.tee 1
        br_if 0 (;@2;)
      end
    end
    local.get 2
  )
  (func $__main (;1;) (type 1) (result i32)
    i32.const 1048576
    i32.const 5
    call $sum_arr
    i32.const 1048596
    i32.const 5
    call $sum_arr
    i32.add
  )
  (memory (;0;) 17)
  (global (;1;) i32 i32.const 1048616)
  (global (;2;) i32 i32.const 1048624)
  (export "memory" (memory 0))
  (export "sum_arr" (func $sum_arr))
  (export "__main" (func $__main))
  (data $.rodata (;0;) (i32.const 1048576) "\01\00\00\00\02\00\00\00\03\00\00\00\04\00\00\00\05\00\00\00\06\00\00\00\07\00\00\00\08\00\00\00\09\00\00\00\0a\00\00\00")
)
```

bitwalker:
This patches the inline-blocks pass to ensure that inlined blocks are not visited multiple times by the inliner - once inlined, these blocks are treated as not inlineable from that point forward, allowing us to visit successor blocks with the inliner normally without considering the inlined block itself.

Additionally, the following issues are addressed:

* We did not raise an error if attempting to remove a node from its OrderedArenaMap if it had already been removed. These nodes remain accessible via the map so that outstanding references are still usable, but are treated in all other respects as if they are not contained in the map - this was not being handled properly by `remove`
* The `link` field of `BlockData` was superfluous and has been for some time, but was missed because it is a public field of a public struct, and there are places where we were incorrectly referencing this field as if it was in use. In fact, the `link` field we care about is the one associated with the `LayoutNode` used by `OrderedArenaMap`, and the `BlockData` one was never linked into anything, resulting in some places incorrectly treating blocks as detached when they were in fact attached.

Closes #85

I'll merge #83 after this is merged

### DIFF
--- a/frontend-wasm/src/function_builder_ext.rs
+++ b/frontend-wasm/src/function_builder_ext.rs
@@ -436,7 +436,7 @@ pub struct FuncInstBuilderExt<'a, 'b: 'a, 'c, 'd: 'c> {
 }
 impl<'a, 'b, 'c, 'd> FuncInstBuilderExt<'a, 'b, 'c, 'd> {
     fn new(builder: &'a mut FunctionBuilderExt<'b, 'c, 'd>, block: Block) -> Self {
-        assert!(builder.data_flow_graph().is_block_inserted(block));
+        assert!(builder.data_flow_graph().is_block_linked(block));
         Self {
             builder,
             ip: InsertionPoint::after(ProgramPoint::Block(block)),
--- a/hir-analysis/src/validation/block.rs
+++ b/hir-analysis/src/validation/block.rs
@@ -137,7 +137,7 @@ impl<'a> Rule<BlockData> for BlockValidator<'a> {
         diagnostics: &DiagnosticsHandler,
     ) -> Result<(), ValidationError> {
         // Ignore blocks which are not attached to the function body
-        if !block_data.link.is_linked() {
+        if !self.dfg.is_block_linked(block_data.id) {
             return Ok(());
         }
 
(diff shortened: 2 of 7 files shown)


greenhat (approved):
Looking good!

greenhat (approved):
Looking good!

PR #91, opened 2024-01-15
docs: add guides for compiling rust->masm

bitwalker:
This expands the documentation, and adds some useful guides for compiling "manually" from Rust to Miden Assembly, in two parts. I also fixed a couple small things discovered while testing.

There are still a number of things to document and expand on, but wanted to get this pushed ASAP. Additionally, I want to turn the example crate from the guides into a crate under `examples` and add it to our test suite, I'll do that in a separate PR though.

NOTE: I found some additional bugs/tweaks/etc that I have not documented yet while putting this together, I'll do that shortly.


bitwalker:
@greenhat I needed to update our rust-toolchain.toml file and the toolchain used in CI, since recent nightlies have broken some of your recent expect tests due to Wasm codegen changes. I moved us to the nightly release corresponding to the 1.75 stable release.

Once 1.76 is released on stable, we should be able to move back to stable for regular development anyway, but for now we'll have to keep tabs on nightly releases manually to see when things change on us.

Makefile.toml
```diff
@@ -146,40 +143,53 @@ run_task = [
...
+run_task = [
+  { name = ["install-midenc"] },
+]
+
 [tasks.check]
 category = "Build"
 description = "Runs cargo check on the workspace"
-command = "rustup"
-args = ["run", "${CARGO_MAKE_TOOLCHAIN}", "cargo", "check", "${@}"]
+command = "argo"
```
greenhat, line 156:
```suggestion
command = "cargo"
```

docs/src/guides/rust_to_wasm.md
```diff
@@ -0,0 +1,176 @@
+# Compiling Rust To WebAssembly
+
+This chapter will walk you through compiling a Rust crate to a WebAssembly (Wasm) module
+in binary (i.e. `.wasm`) form. The Miden compiler has a frontend which can take such
+modules and compile them on to Miden Assembly, which will be covered in the next chaper.
```
greenhat, line 5:
```suggestion
modules and compile them on to Miden Assembly, which will be covered in the next chapter.
```

docs/src/guides/rust_to_wasm.md
```diff
@@ -0,0 +1,176 @@
...
+[lib]
+# Build this crate as a self-contained, C-style dynamic library
+# This is required to emit the proper Wasm module type
+crate-type = ["cdylib"]
+
+[dependencies]
+# Use a tiny allocator in place of the default one, if we want
+# to make use of types in the `alloc` crate, e.g. String. We
+# don't need that now, but its good information to have in hand.
+#wee_alloc = "0.4"
```
greenhat, line 46: Judging by the issues https://github.com/rustwasm/wee_alloc/issues the project has critical bugs and does not seem to be maintained anymore.

greenhat (approved):
LGTM! Just a note on wee_alloc.

docs/src/guides/rust_to_wasm.md
bitwalker: It's not a great allocator, but it _is_ extremely small, the issues people have with it stem from the fact that it doesn't do any complex freelist management, so it "leaks" memory, but for a great many Wasm applications this is perfectly suitable since they are not generally long-running applications.

That said, I will add a note about the tradeoffs with using it, just so people realize the potential downsides - but I would still recommend it in a variety of single-shot style programs. One could even use a stupid-simple bump allocator, which would be even smaller, but useful for a number of use cases as well. I could see it being beneficial for Miden to write a custom allocator that makes better tradeoffs for a ZK machine in the future, but something like dlmalloc is overkill for the kinds of simple applications people are going to be targeting Miden with.

PR #96, opened 2024-01-23
`cargo update`, update Miden VM to the latest commit in the `next` branch

greenhat:


greenhat:
@bitwalker Should we switch to `rev` key to depend on a specific commit? 
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

bitwalker:
I've been waiting on the 0.8.0 release to switch back to stable releases, but I'm fine with pinning to a specific commit in the near term.

bitwalker (approved):

PR #98, opened 2024-01-24
feat: set up lit infrastructure

bitwalker:
This commit simply wires up the test suite, follow on PRs will introduce the actual tests for the suite. Until that is done, test-lit has been removed from the default test command to avoid annoying developers until the suite is stable.

@greenhat I added an example test here for you to look at, but it currently is busted due to a parser issue. The suite is temporarily disabled while I get a few more tests written up and all of them passing, but this just makes sure the infra is in place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 target/
 bin/midenc
-bin/filecheck
+bin/litcheck
+.crates.toml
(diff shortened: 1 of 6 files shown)


greenhat (approved):
LGTM!

PR #119, opened 2024-02-01
ABI mismatch resolution design and low-level(core) types to Miden SDK 

### LINKED ISSUE #113: Low-level tx kernel API types in Miden SDK
Introduce low-level (`core`?) types and functions so that their signatures will be close to the existing transaction kernel API.

Introduce the conversion functions between low-level and high-level types.

Native Miden VM `felt` support is extracted into #118

### LINKED ISSUE #114: Check transaction kernel API function signatures for compatibility with compiler generated wrappers
Write the signatures of the transaction kernel functions.
Figure out how to generate a wrapper from/to the Canonical ABI.

Check transaction kernel API function for incompatibility with compiler generated wrappers.

greenhat:
Close #113 
Close #114

This PR:
* Adds a low-level (core) types for the Miden SDK so that Wasm signatures of the functions be closer to tx kernel ad-hoc ABI. 
* Introduce conversion functions in the Miden SDK between core types and high-level types;
* Adds all tx kernel API functions in Miden SDK (`miden.wit`);
* Adds the analysis and a solution - adapter function generation. See `docs/canonabi-adhocabi-mismatch.md`


greenhat:
@bitwalker
The meat of this PR is in:
- `docs/canonabi-adhocabi-mismatch.md`, where I describe the adapter function generation and one case where we need to change the tx kernel API (`tx::get_assets`). I intend to make an issue out of it after the merge.
- `miden.wit` with all the tx kernel API functions added;

@bobbinth, Could you please take a look at my suggestion for `tx::get_assets` changes? Your feedback is very much appreciated.

bobbinth:
> @bobbinth, Could you please take a look at my suggestion for `tx::get_assets` changes? Your feedback is very much appreciated.

If I understood correctly, the proposal is basically as follows:

First, change the `get_assets` procedure to work like so:
```
#! Inputs:  [dest_ptr, num_assets]
#! Outputs: []
#!
#! Panics if num_assets is not equal to the number of assets in the note.
export.get_assets
```
Then, add a new procedure to retrieve the number of assets. This procedure could look like so:
```
#! Inputs:  []
#! Outputs: [num_assets]
export.get_num_assets
```
If this is correct, I think this could work, but a couple of thoughts:

First, this does introduce a little bit of inefficiency (since under the hood this will be making 2 syscalls instead of 1). This is probably not a big deal in this context, but I do wonder if we should implement these modifications as additional procedures (probably slightly differently names) and keep the current procedure as is.

One important thing to note: when `get_assets` writes assets into memory, "pads" the number of words to the next multiple of 2. So, for example, if there are 5 assets in the note, it will actually write 6 words, and the last word would be `[ZERO; 4]`. This is because under the hood we use [pipe_double_words_to_memory](https://github.com/0xPolygonMiden/miden-vm/blob/main/stdlib/asm/mem.masm#L65) procedure which is quite a bit more efficient than the [alternative](https://github.com/0xPolygonMiden/miden-vm/blob/main/stdlib/asm/mem.masm#L89).

Another thing that might be interesting here: the total number of assets a note can carry is currently limited to 256 (and we are considering reducing this a bit - e.g., maybe to 64). So, another potential approach may be:

1. The caller pre-allocates memory of a list of 256 words.
2. Then makes a call to `get_assets` using the pointer to the list as the argument.
3. `get_assets` writes assets into the list returning the total number of assets it has written.

The signature of MASM procedure would look like so:
```
#! Inputs:  [dest_ptr]
#! Outputs: [num_assets]
export.get_assets
```

If I were to write an equivalent interface in Rust it would look something like this:
```Rust
/// Returns the number of assets written into the `target` array.
fn get_assets(target: &mut [Word; 256]) -> u8
```
Is this something we can express using canonical ABI?

One convenient aspect of this approach is that it automatically takes care of the issue of double vs. single words I mentioned above.

bobbinth:
One other question: how are we thinking of exposing this to the end user? I think neither of the mentioned approaches are particularly "user-friendly." I'd imagine that something like this would be much more intuitive:
```Rust
get_assets() -> &[Asset]
```
But not sure how difficult this would be.

greenhat:
Thanks for the quick feedback! 

> > @bobbinth, Could you please take a look at my suggestion for `tx::get_assets` changes? Your feedback is very much appreciated.
> 
> If I understood correctly, the proposal is basically as follows:
> 
> First, change the `get_assets` procedure to work like so:
> 
> ```
> #! Inputs:  [dest_ptr, num_assets]
> #! Outputs: []
> #!
> #! Panics if num_assets is not equal to the number of assets in the note.
> export.get_assets
> ```
> 
> Then, add a new procedure to retrieve the number of assets. This procedure could look like so:
> 
> ```
> #! Inputs:  []
> #! Outputs: [num_assets]
> export.get_num_assets
> ```

Yes, this exactly what I meant. I should've added MASM definitions besides WIT.

> 
> If this is correct, I think this could work, but a couple of thoughts:
> 
> First, this does introduce a little bit of inefficiency (since under the hood this will be making 2 syscalls instead of 1). This is probably not a big deal in this context, but I do wonder if we should implement these modifications as additional procedures (probably slightly differently names) and keep the current procedure as is.
> 
> One important thing to note: when `get_assets` writes assets into memory, it "pads" the number of words to the next multiple of 2. So, for example, if there are 5 assets in the note, it will actually write 6 words, and the last word would be `[ZERO; 4]`. This is because under the hood we use [pipe_double_words_to_memory](https://github.com/0xPolygonMiden/miden-vm/blob/main/stdlib/asm/mem.masm#L65) procedure which is quite a bit more efficient than the [alternative](https://github.com/0xPolygonMiden/miden-vm/blob/main/stdlib/asm/mem.masm#L89).
> 

I see. We need to take it into account. But since this adapter function will be written manually, so not that big of a deal. 

> Another thing that might be interesting here: the total number of assets a note can carry is currently limited to 256 (and we are considering reducing this a bit - e.g., maybe to 64). So, another potential approach may be:
> 
> 1. The caller pre-allocates memory for a list of 256 words.
> 2. Then makes a call to `get_assets` using the pointer to the list as the argument.
> 3. `get_assets` writes assets into the list returning the total number of assets it has written.

I like this idea even more! :) I think we can splurge on memory to save some cycles.

> 
> The signature of MASM procedure would look like so:
> 
> ```
> #! Inputs:  [dest_ptr]
> #! Outputs: [num_assets]
> export.get_assets
> ```
> 
> If I were to write an equivalent interface in Rust it would look something like this:
> 
> ```rust
> /// Returns the number of assets written into the `target` array.
> fn get_assets(target: &mut [Word; 256]) -> u8
> ```
> 
> Is this something we can express using canonical ABI?

No really. In WIT, there is no fixed-size array. Only dynamic `list` and static `tuple`. `tuple` of that size might not work, I need to test to be sure. But even if works, it will be ugly and hard to use(number of actual items, etc.). In the case of `list`, we will allocate an array in our manually written adapter function to 256 items, pass it to tx kernel call and copy actual items to a new array, and dealloc the first one. So, the user-facing WIT signature will be `get-asset: func() -> list<core-asset>` with Rust binding `fn get_assets() -> Vec<CoreAsset>`.

> 
> One convenient aspect of this approach is that it automatically takes care of the issue of double vs. single words I mentioned above.

Yes, this approach is much better.

greenhat:
> One other question: how are we thinking of exposing this to the end user? I think neither of the mentioned approaches are particularly "user-friendly." I'd imagine that something like this would be much more intuitive:
> 
> ```rust
> get_assets() -> &[Asset]
> ```
> 
> But not sure how difficult this would be.

Due to the "shared nothing" nature of Wasm CM, there is no way to pass a reference (resources might be an alternative, but they have another purpose). So the best fit in WIT would be `fn get-assets() -> Vec<CoreAsset>`.

bitwalker:
One thought here: I think we should distinguish between functions that must exist across a "component" boundary, i.e. that must be called with `call` or `syscall` using the Canonical ABI; versus higher-level, user-friendly wrapper functions that present a more natural interface typical in the source language (i.e. Rust).

For example, we need a primitive like `fn get_assets(target: &mut [MaybeUninit<Word>; 256]) -> u8` to `call`/`syscall`; but we could implement/generate a `fn get_assets() -> [Word; 256]` function in Rust to provide the more user-friendly API for hand-written Rust code.

The more pressing issue is that there is no way to represent pointers _at all_ in the Wasm Canonical ABI (i.e. cross-component function calls). Even if we could though, they would be useless, as we cannot share memory across `call`/`syscall` boundaries (which is the exact reason why the Canonical ABI doesn't have a representation for pointers). It's not even clear to me how `tx::get_assets` can work at all across contexts. Can you shed some light on that @bobbinth, or am I missing something obvious?

It seems to me, that for something like `get_assets`, the actual "protocol" would be something like this:

```rust
/// This function stages the assets in the advice provider, and returns the asset count
///
/// NOTE: "Masm" here doesn't exist, but implies a Miden Assembly procedure
extern "Masm" fn prepare_assets() -> u8;

/// This function pipes `count` words from the advice provider to memory at `dest`
///
/// NOTE: "miden-intrinsic" doesn't exist, but implies a Miden Assembly instruction or compiler intrinsic
extern "miden-intrinsic" fn advice_pipe_to_memory(dest: *mut Word, count: usize);

/// This is the actual user-facing `get_assets` function
///
/// It returns a boxed slice since the actual allocation is of unknown extent
/// until `prepare_assets` is called. This also ensures that the allocated
/// memory is tracked as a normal heap allocation. The box itself can be
/// "leaked" if the assets should live for the duration of the program, which
/// would allow it to be passed around as a reference with static lifetime.
pub fn get_assets() -> Box<[Word]> {
    unsafe {
        let count = prepare_assets() as usize;
        let mut assets = Box::<[Word]>::new_uninit_slice(count);
        advice_pipe_to_memory(assets.as_mut_ptr(), count);
        assets.assume_init()
    }
}
```

I'm not sure there is any other practical way to handle passing data of this size across contexts. Obviously within a "component", we can use the standard approach of passing a pointer to the callee, which then writes data to that address, but that won't work for things like `get_assets` AFAIK.

bitwalker:
I just realized that the pointer-returning `get_assets` function here is actually a wrapper function in miden-base, not the actual kernel function. Such functions are notionally part of the _calling_ component context, not the callee context, which means we _cannot_ represent them in the component model.

What we would actually want to represent in the component model are the actual functions which sit on the boundary, i.e. the underlying kernel functions in this case; but more generally any function that sits on the other side of a context switch. The Canonical ABI enforces the shared-nothing semantics of the component model, so that must be kept in mind when describing interfaces in WIT (and when talking about "components" more generally).

So something like `get_assets` must be implemented in a Rust crate that provides the "porcelain" (to borrow Git parlance) for Miden Base, and _that_ crate might reference the underlying component that provides Canonical ABI bindings for the kernel itself. We can then emit C ABI compatible functions from the Rust crate to expose the high-level bindings to any language with a C-compatible FFI. Since many languages have tools akin to `bindgen`, this provides maximal reuse, and minimal maintenance throughout the toolchain.

I don't think binding against the handwritten Miden Assembly wrappers is necessarily the right approach in all circumstances, personally, versus writing those wrappers in Rust against a kernel component, like shown in my snippet above for `get_assets`. There will likely be situations where it makes sense to do so, for purposes of optimization, but the more we do in MASM wrappers, the more constraints we impose on higher-level languages. That doesn't mean we have to eschew generating trampolines (i.e. adapting handwritten MASM procedures to the Canonical ABI using special compiler support) entirely, or even the majority of the time - but I think we should try to put as much control in the hands of the source language as possible, to avoid introducing unnecessary abstractions that must be unwound in some circumstances. 

As an example of that, let's say we have a MASM-native `get_assets` function like `fn get_assets() -> [Word; 256]`, but we want to pass the assets around, so we're going to wrap it in a function like `fn get_boxed_assets() -> Box<[Word]>`. In rust this is going to result in multiple redundant (and wasteful) allocations:

1.) The call to `get_assets` will allocate a `[Word; 256]` in its stack frame, and pass the pointer to `get_assets`. There is no way to allocate only as much memory as needed for the number of actual assets with the given signature, so we always allocate 256 words. The Rust stack frame is implemented using heap memory, so this allocates 256 words in linear memory.
2.) We then want to place this in a `Box`, so we must allocate an additional 256 words of memory, and then copy the contents of the array on the stack, into the box.

Even if `get_assets` took a pointer, and returned the number of assets written, we would still have to allocate the full 256 words for the lifetime of the `Box`. The only way to avoid wasting the space, is to have a two-phase protocol where you request the necessary capacity first, then call `get_assets` with a sufficiently-sized pointer. I suppose the question then becomes, how many potential sharp corners like this are there overall? If it's a small, closed set, then perhaps its not important to expose the kernel itself as a component (though we still don't escape the fact that none of these type of functions can be exposed in the component model, they have to be pulled in as regular Rust dependencies as part of a larger component).

Anyway, wanted to document my thoughts here.

bitwalker (commented):
As mentioned in the main PR discussion thread, I'm wondering if we're actually going down the right path here with trying to represent kernel wrapper functions in the component model. In particular, those wrappers are meant to be called in the original caller's context, but the component model semantics dictate/imply that the compiler-generated wrappers will actually be in a transient "passthrough" context, since all cross-component calls must to be executed in a context associated with that component, to enforce the shared-nothing semantics of a component instance (naively, one might do this with `call` across the board, but this would be incorrect, as I will talk about further down). It is definitely incorrect to use `exec`'s semantics with any cross-component call, for a host of reasons.

Without some type of "native" component-like entity in the VM itself, our hands are tied with how broadly we can apply the Wasm component model, since some of the features/constraints are aided/enforced by the component runtime (the Wasm host, basically). For our purposes, we really need two things in the VM to make more general use of Wasm components, and realize the promise of the model:

1. Support for "out of order" context switches. This is a low-level constraint issue that we have to solve in order to do more interesting things with contexts in general. Essentially, a `call` always instantiates a new context, and `syscall` always switches to the root context; there is no way to call back to a context earlier in the "context stack" beyond that. This is the core reason we can't support resources in the Wasm component model, because it relies on being able to call back to the component context which instantiated the resource to do anything with them.

2. Support for persistent contexts. Currently, only the root context is truly persistent, while all other contexts are transient, i.e. they only live as long as the call which instantiated them. Another way to think about this is "context graph" instead of "context stack". Persistent contexts would allow a context to be instantiated once, and reused throughout the lifecycle of a program. In general, this is necessary to have any kind of support for sharing (via some form of encapsulation) across contexts. As a simple example, it might be convenient/powerful to have note scripts obtain a capability object that it uses for the duration of its execution to perform various actions granted by its capabilities. Naturally, the only way to make such a thing, is for the context which creates it to remain instantiated for the lifetime of the object, so that it can be called upon to perform actions on behalf of whoever holds it.  As an even more fundamental example, consider a component that uses things such as `OnceCell` or `lazy_static!`, it starts to become apparent that repeatedly instantiating such a component is wasteful. Obviously there are still cases where one needs to guarantee that a call happens in a fresh context, but that isn't precluded by persistence.

3. Support for context-owned resources, with the same basic semantics as Wasm component model resources. When a context is instantiated, one of the bits of data associated with that context is a table of resource handles that have been created (initially empty). The code running in that context may request that resources be created/destroyed, operations which modify that table. A context would be kept alive until all of its resources are freed, or until the top-level program stops executing (either normally, or due to a trap). Calling any function on a resource would switch to the context which created it. Resources would then form the foundation of a capability object model used by code such as the transaction kernel which require guarantees around access control (amongst other things) that are tricky to implement correctly (and with sufficient generality) any other way (especially with the other features mentioned above present).

Without 3, we almost certainly cannot call any function which wraps the transaction kernel; because the component exposing the wrapper will be in a different context than the original caller, which is a problem because the kernel currently "authenticates" actions based on the context it is called from. This was, IIRC, a necessary choice due to the `caller` instruction being the only way to get anything remotely trustworthy/concrete about the calling context. There are various restrictions imposed as a result of this authentication model though. Resources free us from this though, as they are entirely opaque outside the owning context, cannot be manufactured from thin air (in the same sense that one might manufacture a pointer to arbitrary memory from an integer), and once instantiated can only be manipulated by the owning context itself. The nice thing about this is it is a perfect transport for capabilities, which are not only more powerful, but much less fragile. For example, you could split off a single, limited capability (e.g. read-only access to some sensitive data), and hand it off to a not-fully-trusted entity to use to perform some useful work on your behalf, without giving away the keys to the kingdom, so to speak. Doing such a thing _at all_ today isn't possible, but is essential for composability in the ecosystem (IMO).

Without 1 and 2, resources could only be represented in a limited fashion, i.e. only the root context could make use of them. This would be difficult to handle in the Wasm frontend, since it would require us to try and catch when resources are used by any non-kernel code, but could probably be hardcoded. The bigger issue though, is that without 1 and 2, the overhead imposed by context switches is enormous for functions which don't need a fresh component instance (i.e. functions for which `exec` or `syscall` would be used). It's also very likely that using `call` for all component exports would cause soundness/correctness violations. For example, a component might be written in such a way that it assumes that an expression such as `f(g())`, evaluated in a different context than the one which exports those functions, will allow `f` to observe side effects of `g`, but that would not be possible if all cross-component calls are invoked with a new context (as dictated by `call`).

TL;DR: We must restrict the use of Wasm components, until the features mentioned above are implemented in sum or part, to functions which are explicitly designed to be executed in a fresh context, i.e. by `call` or `syscall`. We risk violating the semantics of the model otherwise, and not only will that break reasonable code in non-obvious ways, but it will also discard one of the big benefits of the component model to begin with - cross-language compatibility.

I think I will put the remainder of my review on hold until we discuss this in more detail with @bobbinth , and come up with a concrete plan of attack across the board. It's a tough place we find ourselves in, because I think it is self-evident that the features listed above (and by extension the parts of the component model that we'd like to integrate) would be a massive boon in a variety of areas - but it is not yet clear to me if we have a path to implementing them or not. If we don't, we're going to be in a bind, because so much is affected by whether they form part of the foundation we build on:

* Fundamentally, the component model semantics assume/dictate certain things about the "statefulness" of component instances. If we break with those semantics, we will run into trouble with the kinds of code emitted by compilers targeting the Wasm component model (and which know nothing about Miden).
* Support for components expressed in different languages, but used together, fundamentally depends on both the shared-nothing architecture, and the Canonical ABI; any deviance in semantics will sacrifice this support.
* If we are forced to elide various bits of functionality from component exports due to our inability to enforce the correct semantics for all functions, then languages other than Rust which wish to compile to Miden via the Wasm component model will be forced to reimplement our Rust-specific code in their language, at minimum. For some languages, it might not be possible to implement solutions possible in Rust, or without special support by `midenc`. A major promise of the component model is cross-language compatibility.
* If we were to abandon the Wasm component model, we still need something basically equivalent, in order to ensure that a Miden smart contract written in Rust can be called by another contract written in another language, with all that entails. The problem is that other than the Wasm component model, there isn't anything even remotely close in terms of maturity and tooling - a major advantage of being able to build on top of it.
* Even more problematic, without the Wasm component model support in Rust, we would find ourselves back at square one in figuring out how to ensure that cross-context calls don't make assumptions about shared memory; how to represent such calls at all in Rust, etc. This is basically true of any language built on a shared-by-default memory model.
* Is the current authentication/authorization model used by the transaction kernel actually workable outside of handwritten Miden Assembly? If we don't have some mechanism by which to implement a more robust security model, and it turns out that due to the code emitted by compilers like Rust, invoking kernel procedures happens too deep in the call stack (I seem to recall there being a pretty hard limit here), what's our workaround?

Anyway, I've gone on waaaay too long here, but wanted to get my thoughts on this written down somewhere while they're fresh.

greenhat:
@bitwalker I rebased and cleaned up the todos. Please, make another round.

bitwalker (approved):

PR #125, opened 2024-02-08
Update `cargo-component` to v0.7.0, use `wee_alloc` in Wasm components and Miden SDK tests.

### LINKED ISSUE #107: Update cargo-component to v0.7
Following #106 

The bindings generation approach was changed. See https://github.com/bytecodealliance/cargo-component/pull/212 for details.

greenhat:
Close #107 

**This PR is stacked on top of #100 and should be merged after it**

In this PR:

The panic handler and global allocator is now set in the `lib.rs` file of each component crate. Before, the `cargo-component-bindings` was not built correctly for `no_std` and was linking the stdlib allocator.

The Rust bindings for the WIT are now generated in the `src` directory of the component crate, instead of the `target/.../bindings` directory.

The Wasm `bulk-memory` proposal is enabled via `RUSTFLAGS` so that `memcpy` function import is replaced with Wasm `memory.copy` op.


tests/rust-apps-wasm/add-comp/src/lib.rs
```diff
@@ -1,6 +1,14 @@
 #![no_std]
 
-cargo_component_bindings::generate!();
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+#[panic_handler]
+fn my_panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
```
bitwalker, line 8: I think you end up with even better codegen if you use `core::intrinsics::abort()` here, though that comes with the need for `#![feature(core_intrinsics)]`

bitwalker (approved):
Looks good! Just one note about improving the code generated from the Wasm examples (i.e. reducing panics down to a single `unreachable` instruction), but feel free to take it or leave it.

PR #131, opened 2024-02-20
[2/x] Translate basic wallet Wasm component

### LINKED ISSUE #109: Translate basic-wallet Wasm component

greenhat:
Close #109 

**This PR is stacked on top of #127 and should be merged after it**

This PR implements missing pieces in the Wasm CM translation needed for basic wallet example translation.

Wasm tables translation is not yet implemented, along with the `call_indirect` op translation (ignored) and are extracted into #133.

This PR adds:
- `miden_hir::Type::Tuple` to represent Wasm CM tuple types;
- Wasm module instantiation arguments translation;;
- `FuncEnvironment` for `call` op translation;
- The bulk of the component translation is extracted into `ComponentTranslator` to make the overall algorithm and steps more clear and concise;

`Type::split` implementation for `Tuple` is temporarily plugged with `todo!`. Let me know if you want me to implement it.


hir-type/src/lib.rs
```diff
@@ -65,6 +65,8 @@ pub enum Type {
     Struct(StructType),
     /// A vector of fixed size
     Array(Box<Type>, usize),
+    /// A tuple of values of the given types
```
bitwalker, line 68: Do we actually need a `Tuple` type? A tuple is basically just an anonymous struct, there is no meaningful difference between them. I'd rather avoid introducing additional type-level abstractions if we can avoid it, since it requires handling that type in various places throughout the compiler, and we don't really want to do that unless it is actually semantically significant.

If there is a specific reason why `StructType` is insufficient, let me know, but for now I'm going to leave it out of the rebased changes and use structs instead.

greenhat: Probably, not. I cannot find any reason why it cannot be represented as a struct.

bitwalker (approved):

PR #135, opened 2024-02-29
Remove `build-std` cargo option and switch to `core::intrinsics::abort()` in panic_handler

### LINKED ISSUE #124: use `core::intrinsics::abort()` in panic handler and ditch `build-std` option (compiles `std`, `alloc`, `core`)
See the Rust -> MASM compilation example at https://github.com/0xPolygonMiden/compiler/pull/91 for panic handler.

Remove `build-std` option in `CompilerTest`.

greenhat:
**This PR is stacked on top of #134 and should be merged after it**

Close #124

### DIFF
--- a/frontend-wasm/tests/rust_source/add.rs
+++ b/frontend-wasm/tests/rust_source/add.rs
@@ -1,9 +1,14 @@
 #![no_std]
 #![no_main]
+// This allows us to abort if the panic handler is invoked, but
+// it is gated behind a perma-unstable nightly feature
+#![feature(core_intrinsics)]
+// Disable the warning triggered by the use of the `core_intrinsics` feature
+#![allow(internal_features)]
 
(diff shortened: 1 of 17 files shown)


greenhat:
@bitwalker I switched to using `core::intrinsics::abort()` in panic handler from `build-std` cargo option and hit a problem. `Vec`[indexing](https://github.com/0xPolygonMiden/compiler/blob/860ec1c66c38f561139e198142e32fd131dfa381/tests/rust-apps-wasm/sdk/p2id-note/src/lib.rs#L31) brings its bounds checking with text formatting(`core::fmt`, etc.), which is unaffected by `core::intrinsics::abort()`. Switching to `Vec::get()` with unwrapping fares slightly better, but still brings some unnecessary text formatting. I have not dive deep into it yet, just thought you might have an idea.

Issue #118, opened 2024-01-31
Native Miden VM `felt` and `word` support (tracking)


greenhat:
```[tasklist]
### Tasks
- [ ] https://github.com/0xPolygonMiden/compiler/issues/116
- [ ] https://github.com/0xPolygonMiden/compiler/issues/117
- [ ] https://github.com/0xPolygonMiden/compiler/issues/144
```

greenhat:
@bitwalker @bobbinth While working on #138 and modeling `Felt` as `Felt(f64)`, I found a problem, to which I cannot find a solution. Since we treat arithmetic f64 ops as felt ops in Wasm to IR translation, I'd like to ensure that only our SDK code is allowed to use f64 type in Rust code. So far I did not find any way to do it. Otherwise, we can end up with a situation where either a user's code or even worse, a dependency uses f64, and it will be silently translated to felt ops. These errors will be extremely hard to debug.

I wonder if we should reconsider using f64 for felt. My backup plan is to use `Felt(u64)` and define all arithmetic ops as extern functions that end up as Wasm imports that will be translated to felt ops. With this approach, we can dim f64 type illegal on Wasm to IR translation and display a proper user error.

It should look like:

```rust
#[repr(transparent)]
pub struct Felt(u64);
 
extern "C" {
    #[link_name = "miden_compiler_intrinsics_felt_add"]
    fn felt_add(a: Felt, b: Felt) -> Felt;
}

impl Add for Felt {
    type Output = Self;

    #[inline(always)]
    fn add(self, other: Self) -> Self {
        unsafe { felt_add(self, other) }
    }
}
```

greenhat:
Following up our call. Use `Felt(f64)` but with intrinsics (extern functions) as ops and not allow any native float point operations in the translated Wasm code.

PR #140, opened 2024-03-05
[1/x] Miden SDK as plain Rust dependencies

### LINKED ISSUE #138: Miden SDK as a plain Rust dependencies
### Goals

Following https://discord.com/channels/893151281848406016/1148602585431953450/1207070272478380032

Implement Miden stdlib and base(tx kernel API) as plain Rust dependencies, using extern stubs in the Rust crate that use a naming convention recognized by the Miden compiler to look up the signature for that function in its builtin set of "well known" functions and do the proper lowering.

### Implementation details

The Rust API of tx kernel should closely follow the WIT interface signatures explored in #119 to allow later switching to the WIT interface for stdlib and tx kernel without braking the user code too much. Also, the adapter function generation techniques drafted in #119 (see in [docs](https://github.com/0xPolygonMiden/compiler/blob/main/docs/canonabi-adhocabi-mismatch.md#adapter-function-code-generation)) can be used for lowering the extern stubs.

The base types `Felt` and `Word` (described in #118) along with other types needed in these crates (`Asset`, `Recipient`, etc.) will be later used in re-mapping using `wit-bindgen` (to be implemented in #116) in the WIT interface used in the account and note scripts.

The MAST root hashes of the lowered functions should be encoded in the function names (extern stubs) to avoid the need for hard coding them in the Miden compiler.

```[tasklist]
### Tasks
- [ ] Create `miden-stdlib` and `miden-tx-kernel` crates using the extern stubs;
- [ ] Lower the `miden-stdlib` and `miden-tx-kernel` functions to MASM;
```

greenhat:
Ref #138
Ref #118 

This PR adds:
- `miden-sdk` crate with a small subset of tx kernel API functions, one for each transformation strategy, with partial transformation of the parameters and results on the Rust side (bindings);
- infra in the `frontend-wasm` crate to recognize and transform parameters and results for function calls for these well-known functions during the Wasm translation;

The native `Felt` type is not implemented in this PR and will be added in the next PR.


greenhat:
@bitwalker I got to the point where I need to figure out where to put the Wasm module import (Miden SDK extern function) in the IR. 

### How to represent a Miden SDK extern function Wasm import in IR? 

Component-level import vs. module-level import.

### As a component-level import
Pros (vs. module-level import):
* Concentrating lowering/lifting logic in one place; One can view the generation of the adapter function from C ABI to the kernel tx ad-hoc ABI as a lowering/lifting of sorts.
* When we switch to WIT-based Miden SDK in the future, the lowering/lifting logic's place will remain unchanged, only the logic itself will change.

Cons (vs. module-level import):
* Introduces a different kind of component import in the IR that will require special handling;
* Wasm component import is `call`ed, while Miden SDK import is `exec`ed; Although one could argue that it's a lowering/lifting implementation detail.
* Requires generating a synthetic component to wrap the Wasm module since the Rust compilation produces only a Wasm module; Although, we might need to do this anyway for tables (see [comment](https://github.com/0xPolygonMiden/compiler/issues/32#issuecomment-1970655897)) so this point might actually be moved to pros.

I want to go with component-level import, and I'm starting to draft the code for it. What do you think?

bitwalker:
@greenhat Now that I've finished rebasing/merging the other PRs, since you are still working on this one, you'll need to create a new temporary branch off of `main`, then cherry-pick your commits made for this PR to that new branch (not the other commits that GitHub _thinks_ are now part of this PR, so a simple `git rebase main` isn't an option here), then force push, i.e. 

1. `git checkout main && git pull origin main && git checkout -b temp`
2. Cherry-pick c564e5b and everything after it to `temp`
3. While still in `temp` branch: `git branch -D greenhat/i138-rust-miden-sdk && git checkout -b greenhat/i138-rust-miden-sdk && git push --force origin greenhat/i138-rust-miden-sdk && git branch -D temp`. 

You probably already know the process, but just to make sure everything is painless as possible, that's what I would do. There will be conflicts due to the formatter changes, but most things should be straightforward.

As mentioned in one of my review comments on the other PRs: the `Tuple` type was discarded in favor of `Struct` types, since there is no semantic difference between them, so if you run into any conflicts due to that, just replace any uses of `Tuple` with `Struct` and you should be good to go.

Happy to answer any questions about any of the new changes as well, just let me know.

greenhat:
Sure, will do. Thanks for the detailed steps and thanks for taking care of my previous PRs.

hir/src/component/mod.rs
```diff
@@ -86,11 +84,14 @@ impl formatter::PrettyPrint for ComponentImport {
         };
         let name = match self {
             ComponentImport::CanonAbiImport(import) => import.interface_function.to_string(),
-            ComponentImport::MidenAbiImport(import) => import.function_id.to_string(),
+            ComponentImport::MidenAbiImport(_import) => "()".to_string(),
```
bitwalker, line 87: Shouldn't this also be `import.interface_function.to_string()`? Or is there not a meaningful name in that field yet for this import type.

hir/src/component/mod.rs
```diff
@@ -86,11 +84,14 @@ impl formatter::PrettyPrint for ComponentImport {
...
+        };
+        let import = match self {
+            ComponentImport::CanonAbiImport(_) => const_text("import"),
+            ComponentImport::MidenAbiImport(_) => const_text("import_miden"),
```
bitwalker, line 91: I wonder if it might not be better to do like we do with function calling conventions, and use a marker attribute for the ABI of an import, i.e. `(import (abi canon)` or `(import (abi miden)`, and so on.

tests/integration/expected/rust_sdk_account_test/miden_sdk_account_test.hir
```diff
@@ -1,8 +1,8 @@
...
-    (lower (import #get_id<0x0000000000000000000000000000000000000000000000000000000000000000> (digest 0x0000000000000000000000000000000000000000000000000000000000000000) (type fn() -> i32)) (#miden:tx_kernel/account #get_id<0x0000000000000000000000000000000000000000000000000000000000000000>)
-    (lower (import #get_inputs<0x0000000000000000000000000000000000000000000000000000000000000000> (digest 0x0000000000000000000000000000000000000000000000000000000000000000) (type fn(i32) -> i32, i32)) (#miden:tx_kernel/note #get_inputs<0x0000000000000000000000000000000000000000000000000000000000000000>)
+    (lower (import_miden () (digest 0x0000000000000000000000000000000000000000000000000000000000000000) (type fn(i32, i32, i32, i32) -> i32, i32, i32, i32)) (#miden:tx_kernel/account #add_asset<0x0000000000000000000000000000000000000000000000000000000000000000>)
```
bitwalker: Looks like I missed two things when porting over the textual format:

1. The interface function type is not formatted using the proper type syntax, it should be `(type (func (param i32) (param i32) .. (result i32))`
2. If the digest is in the imported name, we probably don't need to duplicate it as a separate attribute/field of the import, and instead just expect to parse it out of the name (or vice versa, drop it from the name and expect to get it from the `digest` attribute)

Anyway, heads up on that, we'll need to fix that to ensure the text format is round-trippable

hir/src/component/mod.rs
greenhat: Well, there is no any different public "outer" name (like the `WIT` one in CM) besides the one I already have in Wasm core module import (e.g. `miden:tx_kernel/note.get_inputs`) so I decided to omit it altogether and even removed the field for it from `MidenAbiImport`. Do you think we should keep it and print it here?

hir/src/component/mod.rs
greenhat: Nice! Done. I do not like `import_miden`. I just put it there as a sort of placeholder to differentiate between the types of imports.

tests/integration/expected/rust_sdk_account_test/miden_sdk_account_test.hir
greenhat: 1. I missed this. Done. 
2. Good catch! Yep, I already parsing it only once and storing it in `digest`. So I removed the digest from the function name in https://github.com/0xPolygonMiden/compiler/pull/140/commits/9176975dfc8b8b048f9d5cc28228bc0882b34951

sdk/tx-kernel/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[link_name = "get_id<0x0000000000000000000000000000000000000000000000000000000000000000>"]
+    fn extern_account_get_id() -> AccountId;
+    #[link_name = "add_asset<0x0000000000000000000000000000000000000000000000000000000000000000>"]
+    fn extern_account_add_asset(_: Felt, _: Felt, _: Felt, _: Felt, _: i32);
+}
+
+#[link(wasm_import_module = "miden:tx_kernel/note")]
+extern "C" {
+    #[link_name = "get_inputs<0x0000000000000000000000000000000000000000000000000000000000000000>"]
+    fn extern_note_get_inputs(ptr: i32) -> i32;
```
bitwalker, line 37: The `ptr` value here can be `*mut u8`. It will have the correct type when lowered to Wasm (i.e. the Wasm `i32` type). We should always try to use the most accurate type possible that will have the correct C ABI representation to match what is expected by the callee (rather than trying to represent how we translate the C ABI types to Miden types here).

sdk/tx-kernel/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    // #! Writes the inputs of the currently execute note into memory starting at the specified address.
+    // #!
+    // #! Inputs: [dest_ptr]
+    // #! Outputs: [num_inputs, dest_ptr]
+    // #!
+    // #! - dest_ptr is the memory address to write the inputs.
+    unsafe {
+        struct RetArea([u64; MAX_INPUTS]);
+        let mut ret_area = ::core::mem::MaybeUninit::<RetArea>::uninit();
+        let ptr = ret_area.as_mut_ptr() as i32;
```
bitwalker, line 62: Taking into account my other comment, the cast here goes away.

sdk/tx-kernel/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+pub fn get_id() -> AccountId {
+    unsafe { extern_account_get_id() }
+}
+
+// Temporary use u64 instead of Felt until https://github.com/0xPolygonMiden/compiler/issues/118#issuecomment-1978388977 is resolved
+
+const MAX_INPUTS: usize = 256;
+
+#[inline(always)]
+pub fn get_inputs() -> Vec<u64> {
```
bitwalker, line 50: The `Felt` type is defined as i32 below, but then we're returning `Vec<u64>` here, shouldn't this be `Vec<Felt>`?

sdk/tx-kernel/src/lib.rs
bitwalker: Also, this API can return `usize`, since we know that `usize` is equivalent to `u32` for Miden. That simplifies some of the other code that works with this API, i.e. no redundant casts.

sdk/tx-kernel/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+        // and return the number of inputs
+        Vec::from_raw_parts(ptr as *mut u64, num_inputs as usize, num_inputs as usize)
+    }
+}
+
+// Temporary until Felt(f64) is implemented
+pub type Felt = i32;
+
+pub struct CoreAsset {
+    pub inner: (Felt, Felt, Felt, Felt),
```
bitwalker, line 74: Any reason to use a tuple here rather than `Word`, where `Word` is `pub type Word = [Felt; 4];`?

bitwalker: I should also clarify: ABI-wise, `(Felt, Felt, Felt, Felt)` is identical to `[Felt; 4]`, so using one is essentially a synonym of the other here.

sdk/tx-kernel/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+        // Compiler generated adapter function will drop the returned dest_ptr
+        // and return the number of inputs
+        Vec::from_raw_parts(ptr as *mut u64, num_inputs as usize, num_inputs as usize)
+    }
+}
+
+// Temporary until Felt(f64) is implemented
+pub type Felt = i32;
+
+pub struct CoreAsset {
```
bitwalker, line 73: I'd tag this with `#[repr(transparent)]`

sdk/tx-kernel/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+pub fn add_assets(asset: CoreAsset) -> CoreAsset {
+    unsafe {
+        #[repr(align(8))]
+        struct RetArea([u8; 32]);
+        let mut ret_area = ::core::mem::MaybeUninit::<RetArea>::uninit();
+        let CoreAsset { inner: inner0 } = asset;
+        let (t1_0, t1_1, t1_2, t1_3) = inner0;
+        let ptr6 = ret_area.as_mut_ptr() as i32;
+        extern_account_add_asset(t1_0, t1_1, t1_2, t1_3, ptr6);
+        let l7 = *((ptr6 + 0) as *const i64);
```
bitwalker, line 86: I'd suggest reworking this function to simplify it:

```rust
// CoreAsset (with repr(transparent)) has the same size/layout as its sole field, which in turn is equivalent to an array of 4 Felt
// Here, we ensure that the memory area the Miden procedure will write to has the correct size/alignment
let mut result = MaybeUninit::<CoreAsset>::uninit();

extern_account_add_asset(
    asset.inner.0,
    asset.inner.1,
    asset.inner.2,
    result.as_mut_ptr(),
);

// This is safe because CoreAsset has the same size/layout as [Felt; 4]
result.assume_init()
```

sdk/tx-kernel/src/lib.rs
```diff
@@ -0,0 +1,94 @@
+#![no_std]
+
+#[panic_handler]
```
bitwalker, line 3: There will be an issue here if these libraries all define a panic handler and global allocator. They are supposed to only be defined in the top-level executable/dylib, and nowhere else in the dependency tree.

So depending on whether the crate is being compiled standalone as a dylib, or as a dependency in rlib form, we need these items, unfortunately there isn't a config flag/feature just for that.

I'm not sure exactly what the best approach is here long term, but I thought in the near term, these crates were going to be pure Rust dependencies, so we don't need to compile them as dylibs on their own, and all of this machinery can be provided by the top-level account/note script crate.

sdk/types/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+            Ok(Felt(value as f64))
+        }
+    }
+
+    pub fn as_u64(self) -> u64 {
+        self.0 as u64
+    }
+}
+
+impl From<Felt> for u64 {
```
bitwalker, line 36: We should add `From<u8|u16|u32> for Felt`, as well as `From<usize> for Felt` (with a `#[cfg(target_pointer_width = "32")]` on it), since those are all trivially convertible to Felt, and likely to be common conversions.

Adding `TryFrom` for all of the signed versions of those values is probably worthwhile as well, but less important I think.

sdk/types/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    #[inline(always)]
+    fn add(self, other: Self) -> Self {
+        // unsafe { felt_add(self, other) }
+        Felt(self.0 + other.0)
+    }
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct Word(Felt, Felt, Felt, Felt);
```
bitwalker, line 59: I think we should just use `pub type Word = [Felt; 4];` for this. There are no special semantics to `Word`, and no operations that we are planning to expose via `Word` directly, so making a newtype for it doesn't seem useful IMO. Especially since most other code in Miden that deals with words assumes that `Word` and `[Felt; 4]` are synonymous.

bitwalker: I should also note that even in the compiler, it is assumed that a word is equivalent to an array of 4 felts.

hir-type/src/lib.rs
```diff
@@ -662,6 +675,46 @@ impl fmt::Display for LiftedFunctionType {
     }
 }
 
+/// The Miden function type as it is defined in the MASM source (tx kernel, stdlib)
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MidenAbiFunctionType {
```
bitwalker, line 680: I'd _really_ like to avoid adding yet another copy of `FunctionType` in the IR. Especially since in this case, `MidenAbiFunctionType` conveys no new semantics. The more uniform the IR, the better - we should only add new types for things when there is a strong rationale for doing so.

I think this makes it evident to me that we should extend `FunctionType` with the notion of ABI, and remove this and `LiftedFunctionType`. Conveying the ABI is basically the only reason for `LiftedFunctionType` to exist at all, and `MidenAbiFunctionType` makes it clear we need it for more than just a single ABI.

I think we should add a new `Abi` enum to `hir-type`, with the following values:

* `Wasm`, i.e. the type signature of a function expressed in terms of the _Canonical ABI_ of the Wasm Component Model. This indicates that additional lowering/lifting code is required between caller and callee. It also dictates the calling convention for the callee.
* `Canonical`, i.e. the type signature of a function in canonical form. This is the default ABI unless otherwise specified. Canonical in this context means that no special lowering is required between caller and callee - all the compiler needs to deal with are the details of the specific calling convention.
* `Miden`, i.e. the type signature of a procedure in Miden Assembly. Similarly to `Wasm`, this ABI indicates that the additional lifting/lowering code is required between caller and callee, and dictates the calling convention for the callee.

Once defined, add an `abi` field to `FunctionType`, and add the following to the `FunctionType` API:

```rust
pub fn new_with_abi(abi: Abi, ...) -> Self;
pub fn with_abi(mut self, abi: Abi) -> Self;
pub fn abi(&self) -> Abi;
```

Then we can get rid of the duplicate `FunctionType` clones, and use a single type to represent all these various use cases.

Thoughts?

frontend-wasm/src/miden_abi/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+        RpoDigest::try_from(digest).map_err(|e| e.to_string())?,
+    ))
+}
+
+pub fn is_miden_sdk_module(module_id: Symbol) -> bool {
+    tx_kernel::types().contains_key(module_id.as_str())
+}
+
+/// Get the target Miden ABI tx kernel function type for the given module and function id
+#[inline(always)]
```
bitwalker, line 30: Probably not worth inlining this call, since all of the panic infra will be here, there's a good chance LLVM won't even _try_ to inline it. Even if it does though, it will reduce the opportunity to inline more important things.

If you think it's worth it though, I'd maybe just change it to `#[inline]`, to reflect that it's just a hint, rather than trying to force it. Ultimately, both rustc and LLVM are free to ignore either one, but that could change at any time in the future.

sdk/tx-kernel/src/lib.rs
greenhat: Thanks! Done in https://github.com/0xPolygonMiden/compiler/pull/140/commits/7b21de4a54a1f79834946685a9cf8a95254473c0

sdk/tx-kernel/src/lib.rs
greenhat: Yep. Done in https://github.com/0xPolygonMiden/compiler/pull/140/commits/7b21de4a54a1f79834946685a9cf8a95254473c0

sdk/tx-kernel/src/lib.rs
greenhat: Yes. For some weird reason, I temporarily switched `Felt` to be `i32`(instead of `i64`) until proper felts are implemented in the next PR, but then forgot to bring it back. Fixed in https://github.com/0xPolygonMiden/compiler/pull/140/commits/7b21de4a54a1f79834946685a9cf8a95254473c0

sdk/tx-kernel/src/lib.rs
greenhat: I think it was a PTSD :) from spending too much time with WIT before (it does not have static arrays). Fixed to array in https://github.com/0xPolygonMiden/compiler/pull/140/commits/7b21de4a54a1f79834946685a9cf8a95254473c0

sdk/tx-kernel/src/lib.rs
greenhat: Done in https://github.com/0xPolygonMiden/compiler/pull/140/commits/7b21de4a54a1f79834946685a9cf8a95254473c0

sdk/tx-kernel/src/lib.rs
greenhat: Great tip! I grabbed the code from the generated WIT bindings and got so used to it that I forgot to make it nice. Done in https://github.com/0xPolygonMiden/compiler/pull/140/commits/7b21de4a54a1f79834946685a9cf8a95254473c0

sdk/types/src/lib.rs
greenhat: Agree. Done in https://github.com/0xPolygonMiden/compiler/pull/140/commits/7b21de4a54a1f79834946685a9cf8a95254473c0

hir-type/src/lib.rs
greenhat: I agree. I might've carried away a bit with creating types for every ABI.
Done in https://github.com/0xPolygonMiden/compiler/pull/140/commits/dd340f0408baa1c2b17423e87f455017eb8da5a0 
I made `FunctionType::new` to set `Abi::Canonical` and added `new_wasm` and `new_miden` for ergonomics.

frontend-wasm/src/miden_abi/mod.rs
greenhat: Good catch! Thanks. I have no reason to inline it. I probably added it automatically after spending some time in the bindings crate, where almost everything is unlined. Fixed.

sdk/tx-kernel/src/lib.rs
greenhat: Good point! Thanks! Done.

sdk/types/src/lib.rs
greenhat: Good catch! I added it in the next PR (where the proper `Felt` type is implemented) in https://github.com/0xPolygonMiden/compiler/pull/151/commits/bd3b5e732224252cfe97956363abb30a0b0e5c18

greenhat:
@bitwalker Thank you for the detailed review notes! I addressed them all. Please, make another round.

sdk/tx-kernel/src/lib.rs
bitwalker: Ah right, forgot WIT doesn't have fixed-size arrays (yet) 😅

sdk/tx-kernel/src/lib.rs
bitwalker: No worries 🙂. I have to imagine working through the WIT internals has definitely left its scars 😆

hir-type/src/lib.rs
bitwalker: No worries at all! I understand the inclination to do so, especially since the exact shape and semantics of these things has been up in the air - easier to work with new types that don't require modifying a bunch of downstream code. I think up until now, it felt like maybe a new type for the Component Model bits made sense to both of us, but this sort of cemented the idea that we were just fundamentally missing something in `FunctionType`. I think that's an ideal way to evolve our APIs anyway 🙂

frontend-wasm/src/miden_abi/mod.rs
bitwalker: Yeah, I use `#[inline]` and `#[inline(always)]` _a lot_, mostly because we're frequently working on pretty low-level stuff where it can make a noticeable benefit. A lot of people don't realize that without those annotations, Rust won't do much inlining at all on its own, instead it relies entirely on LLVM to figure things out.

That said, I think there are probably quite a few places where I've used them inappropriately, so I'm probably not always setting the best example here 😅

frontend-wasm/src/miden_abi/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use miden_core::crypto::hash::RpoDigest;
+use miden_hir::{FunctionType, Symbol};
+
+/// Parse the stable import function name and the hex encoded digest from the function name
+pub fn parse_import_function_digest(import_name: &str) -> Result<(String, RpoDigest), String> {
+    // parse the hex encoded digest from the function name in the angle brackets
+    // and the function name (before the angle brackets) example:
+    // "miden:tx_kernel/note.get_inputs"
+    let mut parts = import_name.split('<');
```
bitwalker, line 12: Perhaps we need to use `rsplit` here so that we don't run into issues with function names that have brackets in them (i.e. due to generics)? Maybe that's not a practical issue here, but just something that came to mind while glancing through this function.

frontend-wasm/src/miden_abi/transform.rs
```diff
@@ -0,0 +1,101 @@
...
+    match get_transform_strategy(func_id.function.as_symbol().as_str()) {
+        ListReturn => list_return(func_id, args, builder, span, diagnostics),
+        ReturnViaPointer => return_via_pointer(func_id, args, builder, span, diagnostics),
+        NoTransform => no_transform(func_id, args, builder, span, diagnostics),
+    }
+}
+
+/// No transformation needed
+#[inline(always)]
+pub fn no_transform(
```
bitwalker, line 45: Just as a heads up, we should make sure that any function type which would be handled as `no_transform`, uses `Abi::Canonical` so that it goes through the normal lowering process. For the other Abi variants, we either want to transform them such that they are `Canonical`, or that they require no further transformation aside from direct translation to Miden Assembly.

To be clear, haven't checked to see if that's already the case anyway, but wanted to raise it just in case that conflicts with what your operating assumptions are.

frontend-wasm/src/miden_abi/transform.rs
```diff
@@ -0,0 +1,101 @@
...
+    span: SourceSpan,
+    _diagnostics: &DiagnosticsHandler,
+) -> Vec<Value> {
+    let call = builder.ins().call(func_id, args, span);
+    let results = builder.inst_results(call);
+    results.to_vec()
+}
+
+/// The Miden ABI function returns a length and a pointer and we only want the length
+pub fn list_return(
```
bitwalker, line 58: I think this is basically a variant of `no_transform` right? Discarding an unused return value is totally fine with regular functions.

Again, if we're not emitting a wrapper function or something in these cases, feel free to ignore, just flagging it in case the assumption was that we needed to do something different for these functions.

I will definitely make a note though to double check how we're handling multi-result functions elsewhere in the compiler, just in case this is necessary due to lack of support later in the pipeline.

frontend-wasm/src/miden_abi/transform.rs
```diff
@@ -0,0 +1,101 @@
...
+) -> Vec<Value> {
+    // Omit the last argument (pointer)
+    let args_wo_pointer = &args[0..args.len() - 1];
+    let call = builder.ins().call(func_id, args_wo_pointer, span);
+    let results = builder.inst_results(call).to_vec();
+    let ptr_arg = *args.last().unwrap();
+    let ptr_arg_ty = builder.data_flow_graph().value_type(ptr_arg).clone();
+    assert_eq!(ptr_arg_ty, I32);
+    let ptr_u32 = builder.ins().cast(ptr_arg, U32, span);
+    for (idx, value) in results.iter().enumerate() {
```
bitwalker, line 88: *Suggestion:* This is probably a good use case for the `memcpy` builder, if you think it is appropriate here

frontend-wasm/src/module/mod.rs
```diff
@@ -7,19 +7,22 @@ use std::{borrow::Cow, collections::BTreeMap, ops::Range};
...
 use crate::{component::SignatureIndex, error::WasmResult, unsupported_diag};
 
 pub mod build_ir;
-pub mod func_env;
 pub mod func_translation_state;
 pub mod func_translator;
 pub mod function_builder_ext;
 pub mod instance;
 pub mod module_env;
+pub mod module_tratnslation_state;
```
bitwalker, line 25: *Typo:* A sneaky one! Almost didn't notice it haha

bitwalker (approved):
Looks good! Just a couple last questions/notes, but I'll sign off and let you respond/handle those as you see fit.

I'll try to get to your next PR tomorrow, Friday at the latest!

frontend-wasm/src/miden_abi/mod.rs
greenhat: `import_name` is the string that we put in `link_name` annotation at
https://github.com/0xPolygonMiden/compiler/blob/cfabce971c3357f2feb3d559406be3ea97890468/sdk/tx-kernel/src/lib.rs#L20-L26 and not a real function name. I believe we're safe as long as we control `link_name` content.

frontend-wasm/src/miden_abi/transform.rs
greenhat: Good point! Actually, since all the Miden ABI transformation happens in the frontend, I'll use `Abi::Canonical` for all the function type definitions because all of them should have this ABI after the processing regardless whether an actual transformation was conducted. This means that `Abi::Miden` is not used, so it can be deleted. Keep in mind that `Abi` is not included in [`Signature`](https://github.com/0xPolygonMiden/compiler/blob/cfabce971c3357f2feb3d559406be3ea97890468/hir/src/function.rs#L160-L175). Besides that, I removed `(abi XXXXX)` in `ComponentImport` textual form and printed it as part of `FunctionType` in the textual form. 
All the aforementioned changes are made in the #156 due to their invasive nature to the two subsequent PRs that makes rebasing a PITA. See commit https://github.com/0xPolygonMiden/compiler/pull/155/commits/82870e39cb379f04e3a5a5d370e9734377b633f7

frontend-wasm/src/miden_abi/transform.rs
greenhat: I'm not sure, I'm following. `list_return` discards the second return value, i.e. transforming the result to adopt it to the expected Wasm stack state where only one return value is expected. So in my mental model it's not `no_transform`. 
To clarify, there are no wrapper functions generated during the transformation process. The transformation code is generated at the call site.

frontend-wasm/src/miden_abi/transform.rs
greenhat: We're storing stack values in memory, I'm not sure how `memcpy` can help here. If we could only use `mem_storew` VM op analog here, that would be great. I see that most of the cases would be storing `Word` or its multiples.

frontend-wasm/src/module/mod.rs
greenhat: Oh, indeed, a sneaky one! Thanks! Fixed.

frontend-wasm/src/miden_abi/transform.rs
bitwalker: Ok, you can disregard this then, my note was based on the premise that we were emitting wrapper functions to adapt the ABI differences, but if we're just generating code at each call site, then you're already doing what I was suggesting

frontend-wasm/src/miden_abi/transform.rs
bitwalker: Ah, right, I've got to remember that we can't treat the stack like normal memory in most cases (with the exceptions of locals), so `memcpy` wouldn't work anyway.

I can surface some of the word-oriented ops in the instruction set, but we'll have to figure out ways to make those ergonomic/safe to use (for the time being, I'd probably just restrict the operand type to be literally a word, but in theory anything with a compatible layout could use those ops for efficiency).

bitwalker:
LGTM!

PR #147, opened 2024-03-11
chore: various housekeeping items

bitwalker:
* Update to a current nightly toolchain, we were running behind and want to stay a bit ahead of the current stable release. Some changes were triggered as a result.
* Add `rustfmt` configuration as well as an `.editorconfig`, then apply it to all crates except the `frontend-wasm` crate to avoid generating a bunch of conflicts with @greenhat's work. You should run it on the crate once we merge some of your pending PRs just to get that out of the way, and then we'll have a consistent formatting style throughout the compiler.
* Remove the old `filecheck` tool, since the new one is out-of-tree


greenhat (approved):
LGTM!

PR #149, opened 2024-03-14
feat: rewrite hir text format and parser

bitwalker:
This PR builds on #147, so it should be merged after that PR.

This PR replaces the old ad-hoc text format, which was brittle, incomplete, and overly complicated, with a new Wasm-like s-expr format. It introduces a proper formatter as well, and ensures we can roundtrip modules through the IR and back. The only thing not round-trippable at this point are global value expressions, and that's just because I haven't got time yet.

Additionally, a variety of issues were fixed:

* A whole bunch of things that were missing in the parser, but unnoticed because of our lack of round-trip testing.
* Mismatch in how constants were formatted and how they were parsed. It is not assumed little-endian in memory, big-endian when rendered, and is properly marshaled that way now.
* Parsing diagnostics were not properly emitted by `midenc` during compilation
* Support for `.wat` files was added to the `midenc` frontend
* Our first set of "real" lit tests, which I put together by taking them out of the `frontend-wasm` crate and moving them into the `tests/lit` directory, see there for more details.

We should merge this and #147 before any of the other in-flight PRs, because the updated Rust toolchain breaks all of the tests due to codegen changes in `rustc` (again). So this updates the tests for two different essentially irrelevant sets of changes in one go.

NOTE: The new formatter config has caused some churn in some of the modified files of the `frontend-wasm` and integration test crates. I've made notes in the relevant commits what changed, but just be aware you'll have to dig through some noise unfortunately :(. Once we get those crates fully formatted, we should stop seeing the churn thankfully.


hir/src/formatter/mod.rs
```diff
@@ -0,0 +1,167 @@
+mod document;
+mod print;
+
+#[cfg(not(feature = "std"))]
+use alloc::string::String;
+use core::fmt;
+
+pub use self::document::{concat, const_text, display, flatten, indent, nl, text, Document};
+
+pub trait PrettyPrint {
```
greenhat, line 10: Neat! When I was adding `Component` syntax, I thought that it`d be great to try strict Wadler's pretty printer here. And then, I saw you used it in the assembler PR and figured that it will be here as well. Great!

greenhat (approved):
Looking great!

Issue #95, opened 2024-01-19
In-place definition of the cargo projects in integration tests 


greenhat:
### Motivation
Defining the cargo project in the filesystem for a test requires extra activation energy and makes reading the test hard.

### Suggested solution
Employ defining the whole cargo project in the test code using the builder pattern. See the cargo test suite - https://github.com/rust-lang/cargo/blob/57b75970e022e8519fe82cc38a7aed4862f67089/tests/testsuite/check.rs
An example:
```rust
    let foo = project()
        .file(
            "Cargo.toml",
            r#"
                [package]
                name = "foo"
                version = "0.0.1"
                authors = []

                [dependencies.bar]
                path = "../bar"
            "#,
        )
        .file(
            "src/main.rs",
            "extern crate bar; fn main() { ::bar::baz(42); }",
        )
        .build();
    let _bar = project()
        .at("bar")
        .file("Cargo.toml", &basic_manifest("bar", "0.1.0"))
        .file("src/lib.rs", "pub fn baz() {}")
        .build();

    foo.cargo("check")
        .with_status(101)
        .with_stderr_contains("[..]this function takes 0[..]")
        .run();
```

bitwalker:
I like it, would definitely help reduce clutter for sure!

PR #151, opened 2024-03-19
[2/x] Native Miden VM felt type support as `Felt(f64)`

### LINKED ISSUE #144: Native Miden VM `Felt`, `Word` types and ops
To expose native Miden VM felts in Rust, implement a Felt(f64) type and
implement the following:

- failable constructor from u64;
- arithmetic ops recognizable in the compiler;

`Felt` an `Word` types will be re-mapped in WIT by #116

greenhat:
Close #144

**This PR is stacked on top of #140 and should be merged after it**

This PR introduces:
- `miden-prelude` crate for intrinsics and the standard library bindings re-exported in `miden-sdk` crate for Miden rollup.
- `Felt(f64)` type representing the native felt in Miden VM;
- Rust arithmetic, comparison operations and functions implemented via intrinsics;
- `assert*` functions implemented via intrinsics; 
- `felt!` macro for making `Felt` from integer constants that checks the validity at compile time. Less verbose than `Felt::new().unwrap()` and moves the check to compile time.
- Starts guides for developing Miden programs in Rust.

**Note:**
- When a structure containing `Felt` is returned via a pointer, Wasm `f64.load` and `f64.store` are used and are translated as felt ops. This leaves a tiny chance that either a user's code or its dependencies might contain these ops and only these ops (we blow up on all other float ops) for floats, and we treat them as felts, but I cannot imagine this as a real-world case so we can probably ignore it.
- I wrote only a smoke test in https://github.com/0xPolygonMiden/compiler/blob/1c056f2f191df4fe7d1eaec2ffc3d2fc182e3e9d/tests/rust-apps-wasm/rust-sdk/account-test/src/lib.rs#L24-L47 and extracted the proper test suite into #154 because it needs #95 implemented first.


sdk/prelude/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "miden-sdk-types"
-description = "Miden SDK types"
+name = "miden-prelude"
```
bitwalker, line 2: I think we probably need a better name for this crate, as `miden-prelude` is super non-descript. More generally we should probably make sure this (and other crates which are part of the Rust bindings to Miden SDK code written in Miden Assembly) uses some kind of convention. Something like `miden-bindgen-<category>`, where `<category>` describes what bindings are provided by the crate.

We aren't currently generating the bindings, so `bindgen` might seem a bit misleading, but we may be able to do so in the future, and for now it conveys a decent idea of what the crate is about. We could probably go even more specific, but I'm not sure how verbose we really want to be.

Not a huge priority though, just something I wanted to note.

sdk/prelude/src/intrinsics/felt.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    #[link_name = "assert_eq"]
+    fn extern_assert_eq(a: Felt, b: Felt);
+}
+
+/// Creates a `Felt` from an integer constant checking that it is within the valid range at compile
+/// time.
+#[macro_export]
+macro_rules! felt {
+    ($value:expr) => {{
```
bitwalker, line 67: If you make this a `($value:literal)` clause, you can avoid the need to do the `const VALUE..` hack entirely.

bitwalker (approved):
A couple notes, but looks fine for now. We'll probably end up reworking this a bit depending on how we resolve the felt repr issue.

sdk/prelude/Cargo.toml
greenhat: Good point! I was coming to this name from the user perspective. I want them to import only one crate whether they target Miden rollup or pure Miden VM. The rollup users would use the `miden-sdk` crate that re-exports `miden-prelude` crate and `miden-sdk-tx-kernel` crate. For the users that target pure Miden VM (without rollup) I want to provide intrinsics and stdlib in one crate and I initially wanted to use the name `miden-stdlib`, but it's already taken by one of the VM crates (see https://crates.io/crates/miden-stdlib) so the closest name to intrinsics+stdlib I could think of was `miden-prelude`. 

If I understand your idea, we will end up with the following crates:
`miden-bindgen-intrinsics` - for the bindings to the intrinsics functions
`miden-bindgen-stdlib` - for the bindings to the stdlib functions
`miden-bindgen-tx-kernel` - for the bindings to the tx_kernel functions

We can re-export all of them in the `miden-sdk` crate for the rollup users and in `miden-?`(what name if not `prelude`) crate for the pure Miden VM users. What do you think?

sdk/prelude/src/intrinsics/felt.rs
greenhat: Great hint! Thanks! Done.

PR #177, opened 2024-04-30
[1/2] stdlib blake3 call MASM compilation test

### LINKED ISSUE #160: Test suite for Miden SDK (`tx_kernel` API) ABI transformation
### Goal

The goal is to test both the ABI transformation strategies implemented in the [frontend](https://github.com/0xPolygonMiden/compiler/blob/6466b64e0434d1a77e4379e3035a5aaa9bd4c356/frontend-wasm/src/miden_abi/transform.rs#L48-L113) and the Rust bindings written for every Miden SDK `tx-kernel` API function in the `miden-tx-kernel` crate (see [`get_inputs`](https://github.com/0xPolygonMiden/compiler/blob/6466b64e0434d1a77e4379e3035a5aaa9bd4c356/sdk/tx-kernel/src/lib.rs#L41-L61) for example). 

### Implementation details

We can generate the `tx_kernel` API function stubs in the IR that will assert certain facts about the inputs and return mocked data, and then run the compiled code in the Miden VM. This approach has a drawback that it would not be able to use property-based testing, since the inputs and outputs are hard-coded in the stubs.
```[tasklist]
### Tasks
- [x] MASM compilation is passing
- [ ] https://github.com/0xPolygonMiden/compiler/issues/223
- [x] MASM assembly is passing
- [x] Generate Rust code to test the values received from the `tx_kernel::get_inputs` call
- [ ] https://github.com/0xPolygonMiden/compiler/issues/225
- [ ] https://github.com/0xPolygonMiden/compiler/issues/230
```

### LINKED ISSUE #161: Test suite for Miden stdlib ABI transformation
### Goal

The goal is to test both the ABI transformation strategies implemented in the [frontend](https://github.com/0xPolygonMiden/compiler/blob/6466b64e0434d1a77e4379e3035a5aaa9bd4c356/frontend-wasm/src/miden_abi/transform.rs#L48-L113) and the Rust bindings written for every Miden stdlib function in the `miden-prelude` crate (see [mem](https://github.com/0xPolygonMiden/compiler/blob/6466b64e0434d1a77e4379e3035a5aaa9bd4c356/sdk/prelude/src/stdlib/mem.rs#L76-L113) for example). 

### Implementation details

Since the Miden stdlib doesn't have dependencies, we can link it and call the Miden stdlib functions using their Rust bindings providing the inputs and assessing the outputs in the property-based tests running the compiled code in the Miden VM.
Alternatively, we can generate the Miden stdlib function stubs in the IR (like for `tx_kernel` API in Miden SDK) that will assert certain facts about the inputs and return mocked data and then run the compiled code in the Miden VM. This approach has a drawback that it would not be able to use property-based testing, since the inputs and outputs are hard-coded in the stubs.

greenhat:
This PR adds a compilation test for calling the blake3 hash function from the stdlib.


greenhat:
> Do you have some repro steps I can follow to replicate locally? I'll need to poke at it a bit. Definitely odd that a controlling loop is set, but the transition (current_block -> target_block) is going from a block that ostensibly doesn't belong to a loop, to one that does - in that scenario, there should be no controlling loop (otherwise `current_block` would belong to it), but there are a few edge cases due to how the controlling loop state is managed, and either the assertion is overly restrictive, or a more fine-grained condition needs to be checked.

`cargo test blake3` will run the test that causes the issue.
> 
> You can try disabling the assertion temporarily to see how far that gets you while I investigate, but that will only help you if the assertion is overly restrictive.

After disabling the assertion, I hit the addressable stack limit(16) with index 16.

bitwalker:
> After disabling the assertion, I hit the addressable stack limit(16) with index 16.

Awesome 🤦, I was hoping we'd have a bit more time before hitting that wall, but I guess at least now we have an easily reproducible real-world program that hits it. I'll add this to my TODO list

bitwalker (approved):

PR #179, opened 2024-05-04
[0/x] Fix: add `inttoptr` after emitting the address of a global variable

greenhat:
Otherwise, the subsequent `store`/`load` will fail (expecting a `Ptr`).

### DIFF
--- a/codegen/masm/src/codegen/emitter.rs
+++ b/codegen/masm/src/codegen/emitter.rs
@@ -432,7 +432,7 @@ impl<'b, 'f: 'b> BlockEmitter<'b, 'f> {
             hir::GlobalValueData::IAddImm { .. } | hir::GlobalValueData::Symbol { .. } => {
                 let mut emitter = self.inst_emitter(inst_info.inst);
                 emitter.stack_mut().push(addr);
-                emitter.inttoptr(&Type::Ptr(Type::Unknown.into()));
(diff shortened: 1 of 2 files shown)


codegen/masm/src/codegen/emitter.rs
```diff
@@ -431,6 +432,7 @@ impl<'b, 'f: 'b> BlockEmitter<'b, 'f> {
             hir::GlobalValueData::IAddImm { .. } | hir::GlobalValueData::Symbol { .. } => {
                 let mut emitter = self.inst_emitter(inst_info.inst);
                 emitter.stack_mut().push(addr);
+                emitter.inttoptr(&Type::Ptr(Type::Unknown.into()));
```
greenhat, line 435: @bitwalker I tried to find the type of the global var, but seems like we're missing it in the `GlobalValueData::Symbol`. 
BTW, why do we return `Ptr(i8)` at https://github.com/0xPolygonMiden/compiler/blob/cc3f74072b2eff7f48b8e31386f8ed4e7820baaa/hir/src/dataflow.rs#L155 for `Symbol`?

bitwalker (changes requested):
Just one adjustment, but otherwise looks good!

codegen/masm/src/codegen/emitter.rs
greenhat: Got it. Done.

bitwalker (approved):

PR #181, opened 2024-05-06
chore: handle assembler refactoring changes

### LINKED ISSUE #33: Implement support for assertion error codes
The various assertion instructions, e.g. `assert` in Miden Assembly now support constant integral error codes. We should add support for that in the IR.

### LINKED ISSUE #57: Missing Miden stdlib import in the `Module` when emitting stdlib's function calls in `stackify` pass
To reproduce uncomment `test_int_op!(and, &, u64);` test in `instructions.rs` in #49 

For `1u64 & 1u64` Rust code is compiled into the following IR:
```
module noname

const $0 = 0x00100000;

pub fn entrypoint(i64, i64) -> i64 {
block0(v0: i64, v1: i64):
    v3 = band v1, v0 : i64;
    br block1(v3);

block1(v2: i64):
    ret v2;
}
```

Which translates into the following MASM:
```
export.entrypoint
  exec.checked_and

end
begin
    exec.entrypoint
end
```

Which Miden `Assembler` cannot compile.

### LINKED ISSUE #59: Compiling an app target to MASM is broken due to `__main` is an invalid name for a procedure in the Miden Assembler
A `main` function in Rust cannot have arguments, and we need them to pass inputs to a MASM program in app targets of the cargo projects that we want to compile to MASM. 
The workaround is to use `extern "C" fn __main(...) -> ... {}` to have to pass inputs and get the output from the VM. However, the Miden Assembly deems `__main` as the invalid name for a procedure.

bitwalker:
This modifies various bits of the compiler that are tightly integrated with the Miden VM assembler components now that 0xPolygonMiden/miden-vm#1277 is merged. Notably:

* The instruction set for MASM now is a 1:1 match, though for the time being we still have it duplicated in the compiler so that we can use the same patterns for analysis/building/translating as we do the IR. I plan to revisit that in the future, to try and unify them, but for now it is not important enough to tackle yet
* Support for assertion error codes is added for `assert` and `assertz` instructions (fixes #33)
* We no longer need explicit imports for standard library functions (fixes #57)
* The rules for procedure names/identifiers have been loosened (fixes #59)
* Support for some basic advice provider ops is added to the emulator
* Support for MASL libraries is **removed**, for the time being we only support emitting MASM text files, until the MAST changes are fully landed in the assembler, at which point we will emit MAST by default
* We now use the formatting utilities from `miden-core` for pretty printing AST/IR
* `miden_codegen_masm::masm::Program` is now more or less a container of modules + the global data segment table, it no longer contains the executable "main" function, that is now generated as part of a `Module` of executable type. I'm not super happy with how these structures are organized/used in the codegen backend right now, but until things stabilize a bit, I don't want to try and revisit it. I think longer term we'll want to have a more unified set of core libraries used between the compiler and assembler (or perhaps the assembler becomes more integrated into the compiler), and since most of the awkwardness here is a result of duplication/impedance mismatch, I think those sorts of changes would resolve the vast majority of my complaints.

Next up are the felt changes


greenhat (approved):
Looking great!

greenhat:
Do you want me to rebase my stack of PRs on top of this PR, or should I wait for the next one?

PR #182, opened 2024-05-07
feat: implement stores for felts/machine-width values

bitwalker:
@greenhat This PR implements the most likely needed operations for stores involving felts or machine-width integer values. I haven't yet implemented 64-bit integers and larger, or aggregate types, but I'm opening this now so you have something you can work with while those get implemented.

NOTE: I have not really tested these implementations, but they are based on the same principles as the `load` ops. Definitely need testing here, but again, mostly looking to get the ABI testing unblocked, if there are bugs here, we can sort those out as they come until we have time to do more thorough testing.

### DIFF
--- a/codegen/masm/intrinsics/mem.masm
+++ b/codegen/masm/intrinsics/mem.masm
@@ -331,3 +331,234 @@ export.load_dw # [waddr, index, offset]
         end
     end
 end
+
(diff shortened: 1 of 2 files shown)


greenhat (approved):
Looking good! Nothing caught my eye when I ran the MASM code in my brain :) .

PR #183, opened 2024-05-09
chore: run clippy on CI, fix all clippy warnings

### LINKED ISSUE #70: Check for clippy warnings on CI
It's hard to use a clippy when there are a lot (65) of warnings in the whole workspace. I suggest we add a CI job that will fail on clippy warnings. Although I find almost all of the warnings useful we can disable the ones we don't want.

greenhat:
Close #70


greenhat:
I copied clippy arguments from the miden-vm crate's CI.

bitwalker (approved):
LGTM

PR #186, opened 2024-05-13
[DO NO MERGE] Basic wallet account code compilation to MASM

greenhat:
This PR is intended to prepare the presentation for Bali.


greenhat:
The Rust code https://github.com/0xPolygonMiden/compiler/blob/c14c70b076b17ddb083f23f7e5562908537b914c/tests/rust-apps-wasm/rust-sdk/account-test/src/lib.rs#L1-L27

compiles to IR - https://github.com/0xPolygonMiden/compiler/blob/c14c70b076b17ddb083f23f7e5562908537b914c/tests/integration/expected/rust_sdk_account_test/miden_sdk_account_test.hir#L1-L280

But the MASM compilation fails with:
```
thread 'rust_masm_tests::rust_sdk::account' panicked at codegen/masm/src/codegen/opt/operands/tactics/linear.rs:38:17:
assertion failed: b.value.is_alias()
```
Commenting the assertion did not get me far.

To reproduce run:
```
cargo test account
```

greenhat:
Fibonacci.

Rust https://github.com/0xPolygonMiden/compiler/blob/c14c70b076b17ddb083f23f7e5562908537b914c/tests/rust-apps/fib/src/lib.rs#L1-L13

HIR https://github.com/0xPolygonMiden/compiler/blob/c14c70b076b17ddb083f23f7e5562908537b914c/tests/integration/expected/fib.hir#L1-L40

MASM
https://github.com/0xPolygonMiden/compiler/blob/c14c70b076b17ddb083f23f7e5562908537b914c/tests/integration/expected/fib.masm#L1-L43

bitwalker:
@greenhat I've addressed a variety of bugs that were blocking compilation to MASM, but I'm able to compile the account code in the test you put together to MASM.

**NOTE:** There are a couple of other issues with codegen that need to be addressed, but we can at least put together the code artifacts for a demo for now. To sum up the major items:

* Currently load/store and side-effecting operations can be reordered around each other, which is incorrect. We can fix this by introducing control dependencies in the treegraph scheduler, which is already implemented, what isn't implemented is assigning control dependencies to the various side-effecting operations, as well as ensuring that certain operations can not be reordered relative to one another (for now, we should preserve the original load/store ordering, and ensure that their ordering with respect to call sites is preserved).
* Operand stack overflow
* Read-only data segments
* Finish up felt representation changes

greenhat:
Do you want to merge this PR, so your fixes got merged, or did you cherry-pick them into your branch(es) already? I can restore the original code for the account test along with this one in this branch.

bitwalker:
@greenhat You can go ahead and merge this PR, probably easier that way

PR #187, opened 2024-05-21
fix: basic wallet account compilation

bitwalker:
This PR applies some fixes that were causing the basic wallet account code from #186 to hit asserts in the backend. Once merged, the only remaining blockers are related to felts, stores of complex types, load/store reordering, and operand stack overflow handling.


greenhat (approved):
LGTM!

PR #193, opened 2024-05-28
[1/x] `tx_kernel::get_inputs` MASM compilation test

greenhat:
**This PR is stacked on the #198 and should be merged after it** 

This PR adds a MASM compilation test for `tx_kernel::get_inputs` call


hir-transform/src/treeify.rs
```diff
@@ -327,15 +328,25 @@ impl RewritePass for Treeify {
...
+                // this subtree later in this loop
+                // see https://github.com/0xPolygonMiden/compiler/issues/194
+                detach_q.push_back(b);
 
                 // Mark the control flow graph as modified
                 changed = true;
             }
         }
 
+        for b in detach_q {
```
bitwalker, line 346: Won't this cause us to hit the same bug, just at the end of treeification? I think it isn't evident here because you've also silenced the assertion that was previously raised, by ignoring when a detach of an already-detached block is attempted in `DataFlowGraph::detach_block`.

I think at a minimum it will be necessary to track which blocks we've already detached, and if we reach a subtree rooted at a block which we've already detached, then we skip `detach_tree` entirely. That has the same effect as your change to `DataFlowGraph::detach_block`, but is explicitly localized to just this specific operation where we know we might attempt to detach a block multiple times.

greenhat: > Won't this cause us to hit the same bug, just at the end of treeification? I think it isn't evident here because you've also silenced the assertion that was previously raised, by ignoring when a detach of an already-detached block is attempted in `DataFlowGraph::detach_block`.

Yes, you're right, it just silences the repeated detachment. I did this to resolve another issue with inserting a block after the detached block in the treefication (copying) process.

greenhat:
@bitwalker After dealing with repeated tree detach (moved to the end of the treeify pass) I hit an issue where the loop header block (block2) was treeified for every predecessor except the entry block(block0) in `#_ZN9wee_alloc15alloc_first_fit17h5cf1f963ec6ad59fE` function because we're explicitly skipping treefication in our outer loop but the loop header block treefied later when It's reached through other predecessors. In the entry block, the original successor is referenced, which was detached during the treeification. This blows up in the inlining pass when the detached original loop header block (block2) is inlined into the entry block (block0). 

To reproduce run `cargo test get_inputs`.

greenhat:
I cleaned this PR to contain only the `tx_kernel::get_inputs` compilation. It's ready for review and merge.

bitwalker:
@greenhat You can remove the changes to the treeify pass here, and rebase on main, then we should be able to merge this.

greenhat:
> @greenhat You can remove the changes to the treeify pass here, and rebase on main, then we should be able to merge this.

Done.

bitwalker (approved):

PR #198, opened 2024-06-02
fix: refactor treeify pass to address issues with complex control flow

bitwalker:
This PR fixes known issues with the treeify pass that caused compilation of complex control flow to fail. See the commit message for more specifics.

We do need more tests of this pass, but for now it is run by default in all compiler pipelines, so I would expect to see tests failing for unexpected reasons if this (and the original pass for that matter) are doing something incorrect.

@greenhat After fixing this pass, I was able to get compilation of the `get_inputs` test to reach a point where it fails due to the missing implementation of `memory.grow`, so I think it has also addressed issues you thought were in the inliner pass. We'll have to implement that instruction so that we can run that test the rest of the way, but that's where things are at after this PR is merged.

### DIFF
--- a/hir-transform/src/treeify.rs
+++ b/hir-transform/src/treeify.rs
@@ -1,33 +1,62 @@
-use std::{collections::VecDeque, rc::Rc};
+use std::{
+    collections::{BTreeMap, VecDeque},
+    rc::Rc,
+};
 
 use midenc_hir::{
     self as hir,
     pass::{AnalysisManager, RewritePass, RewriteResult},
     Block as BlockId, Value as ValueId, *,
 };
-use midenc_hir_analysis::{BlockPredecessor, ControlFlowGraph, DominatorTree, LoopAnalysis};
+use midenc_hir_analysis::{BlockPredecessor, ControlFlowGraph, DominatorTree, Loop, LoopAnalysis};
 use midenc_session::Session;
-use rustc_hash::FxHashSet;
+use smallvec::{smallvec, SmallVec};
 
 use crate::adt::ScopedMap;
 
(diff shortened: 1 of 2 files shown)


hir-transform/src/treeify.rs
```diff
@@ -273,63 +276,111 @@ impl RewritePass for Treeify {
...
+        // For each block B, treeify B IFF it has multiple predecessors, where for each
+        // predecessor P, P dominates B. Treeifying B involves creating a copy of B and
+        // the subgraph of the CFG rooted at B, for each P.
+        //
+        // Loops are handled with a bit more subtlety, in that a loop header may have
+        // multiple predecessors, but only one which dominates the loop header - such
+        // loops are not treeified. However, a loop header with multiple dominating
+        // predecessors _is_ a candidate for treeification, but the loop is only cloned
+        // for each of the dominating predecessors, not all predecessors.
+        //
```
greenhat, line 338: I'm not sure I understand this explanation of the difference in handling the loop headers. I mean, looking at the code the only difference I see is a decision to not treeify the loop header when the loop header dominates the root (down below):
```rust
                    if block_infos.is_loop_header(b).is_some()
                        && block_infos.rpo_cmp(b, root).is_lt()
                    {
                        continue;
                    }
```

greenhat (approved):
Looking great! Only one question regarding the comments.

hir-transform/src/treeify.rs
bitwalker: I'll expand on the commentary a bit, a few parts of it are a tad outdated after reviewing them

PR #202, opened 2024-06-04
[3/x] Always duplicate stack operands in CopyAll stack operand scheduling tactic

greenhat:
Close #200

**This PR is stacked on top of #199 and should be merged after it**

Issue #201, opened 2024-06-04
Unexpected controlling loop assertion in `BlockEmitter::target_controlling_loop` for `wee_alloc::alloc_first_fit` function in `tx_kernel::get_inputs` test


greenhat:
To reproduce, run `cargo test get_inputs` in #199 branch

The assertion fails in the `wee_alloc::alloc_first_fit` function (the same function required new `treeify` pass) code with:
```
thread 'rust_masm_tests::abi_transform::tx_kernel::test_get_inputs' panicked at codegen/masm/src/codegen/emitter.rs:1008:17:
assertion `left == right` failed: unexpected controlling loop: block9, parent: Some(block2)
  left: Some(loop1)
 right: None
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2d24fe591f30386d6d5fc2bb941c78d7266bf10f/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/2d24fe591f30386d6d5fc2bb941c78d7266bf10f/library/core/src/panicking.rs:72:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/2d24fe591f30386d6d5fc2bb941c78d7266bf10f/library/core/src/panicking.rs:298:5
   4: midenc_codegen_masm::codegen::emitter::BlockEmitter::target_controlling_loop
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/codegen/emitter.rs:1008:17
   5: midenc_codegen_masm::codegen::emitter::BlockEmitter::emit_cond_br
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/codegen/emitter.rs:394:40
   6: midenc_codegen_masm::codegen::emitter::BlockEmitter::emit_inst
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/codegen/emitter.rs:218:44
   7: midenc_codegen_masm::codegen::emitter::BlockEmitter::emit
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/codegen/emitter.rs:185:48
   8: midenc_codegen_masm::codegen::emitter::FunctionEmitter::emit
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/codegen/emitter.rs:136:21
[...]

PR #203, opened 2024-06-04
[4/x] fix: change the pointer type to `i32` in MASM function signature for `tx_kernel::get_inputs`

greenhat:
**This PR is stacked on top of #202 and should be merged after it**

This PR fixes the last obstacle (pointer type) to generate MASM for `tx_kernel::get_inputs` call.

Temporary comment out the assertion for controlling loop in `emitter.rs` until #201 is addressed.

### DIFF
--- a/tests/integration/expected/abi_transform_tx_kernel_get_inputs.masm
+++ b/tests/integration/expected/abi_transform_tx_kernel_get_inputs.masm
@@ -3,7 +3,7 @@ mod abi_transform_tx_kernel_get_inputs
 use.miden:tx_kernel/note
 
 export.entrypoint
-    exec."_ZN19miden_sdk_tx_kernel10get_inputs17h6e5762b00f2d3838E"
+    exec."miden_sdk_tx_kernel::get_inputs"
 end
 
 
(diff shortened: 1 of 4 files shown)


sdk/tx-kernel/src/externs.rs
```diff
@@ -15,7 +15,7 @@ extern "C" {
 #[link(wasm_import_module = "miden:tx_kernel/note")]
 extern "C" {
     #[link_name = "get_inputs<0x0000000000000000000000000000000000000000000000000000000000000000>"]
-    pub fn extern_note_get_inputs(ptr: *mut Felt) -> usize;
+    pub fn extern_note_get_inputs(ptr: i32) -> usize;
```
bitwalker, line 18: Do you know why the pointer type is causing an issue here? We should try to figure that out, since I'm surprised that this ends up being irrelevant (as both pointers and 32-bit integers should basically have the same representation when compiled to MASM). I'm guessing we must be doing something wrong during codegen if this is causing an error.

greenhat: Oh, my bad. The issue was the `Felt` instead of `i32` at https://github.com/0xPolygonMiden/compiler/blob/e661c17f4e15ed2a0ae8554cfb599a1c1d50778a/frontend-wasm/src/miden_abi/tx_kernel/note.rs#L11
Here, I changed it to `i32` for some weird reason, probably I was tired.  `*mut Felt` works just fine, I restored it.

greenhat: Good catch! Thanks!

bitwalker (approved):

PR #204, opened 2024-06-06
[5/x] Replace unsupported characters in module and function names in Wasm imports

greenhat:
**This PR is stacked on top of #203 and should be merged after that**

This PR introduces the following replacements in module and function names:
https://github.com/0xPolygonMiden/compiler/blob/eccef313c11cbefac9ff74f9b8edfb23d0ad616d/frontend-wasm/src/translation_utils.rs#L149-L154


greenhat:
No need to sanitize the ident's, raw idents should be used.

Issue #208, opened 2024-06-13
Namespace of the compiled MASM module for the Rust code


greenhat:
The compiler produces compilation artifacts in a form of one MASM file per module using module ID for the file name. The Rust code resides in a module with an ID corresponding to the crate name. If we load the MASM module without specifying the namespace, it gets assigned a `#anon` namespace. But I cannot figure out the way to call the procedure from the module with `#anon` namespace. 
I've tried to call the procedure without importing the module assuming the `#anon` namespace is the default one, but it doesn't work.  I've tried the `use` directive with the `#anon` namespace, but it doesn't work either. 
At the end I went with a workaround by assigning a synthetic namespace `user_ns` to the module, but the DX is not ideal.

greenhat:
Procedures from anonymously parsed modules should be callable after https://github.com/0xPolygonMiden/miden-vm/pull/1363 is merged.

bitwalker:
To elaborate a bit on how we should address this once 1363 is merged:

* We need to ensure that we use absolute paths for all procedure references, and elide emitting imports (unless we have a specific reason to use relative paths and imports on a case-by-case basis). The syntax for this is the usual path syntax, but prefixed with `::`, e.g. `::foo::bar`, and can be used anywhere that you specify paths in MASM.
* The assembler parser has been modified so that when an absolute path with a single component is parsed, it treats it as implicitly in the `#anon` namespace, e.g. `exec.::foo` is treated as a procedure defined in an anonymous module. More precisely, in a procedure context like that, it parses it as a `ProcedurePath::Absolute`, whose `path` field is a `LibraryPath` with `LibraryNamespace::Anon` and no additional path components, and whose `name` field is simply the identifier `foo`. If used in a module path context, it is essentially the same, except the resulting `LibraryPath` contains `foo` as the sole path component.
* Raw identifiers can be used anywhere that a procedure name is expected, so the following is valid syntax `exec.::"miden:tx-kernel/get-inputs"`, and would refer to a procedure named `miden:tx-kernel/get-inputs` in the anonymous module.
* When emitting a program, where the program is all in a single anonymous module, we can combine the two points above to reference all procedures in the program absolutely.

[...]

PR #209, opened 2024-06-13
[6/x] Build the VM `Program` from the compiled MASM in integration tests

greenhat:
**This PR is stacked on the #205 and should be merged after it**

Includes the namespace for Rust module workaround explained in #208


tests/integration/src/compiler_test.rs
```diff
@@ -584,37 +571,101 @@ impl CompilerTest {
...
+        todo!()
+    }
+    src
+}
+
+// Generate the MASM program source code (call the entrypoint function)
+fn masm_prog_source(user_ns_name: &str, entrypoint: FunctionIdent) -> String {
+    let module_name = entrypoint.module.as_str();
+    let function_name = entrypoint.function.as_str();
+    format!(
```
bitwalker, line 647: Just as an aside, we won't need to explicitly import modules like this once 0xPolygonMiden/miden-vm#1363 is merged, we can switch to using fully-qualified procedure paths, e.g. `exec.::path::to::module::function` (note the leading `::`)

bitwalker (approved):

PR #210, opened 2024-06-13
[7/x] Use the MASM module paths for the tx kernel module names

greenhat:
**This PR is stacked on the #209 and should be merged after that**

This PR changes the `wasm_import_module` names from the Wasm CM inspired to the MASM module paths that most likely will be used when assembling the tx kernel modules for the VM (taken from the `miden-base`).


frontend-wasm/src/miden_abi/tx_kernel/account.rs
```diff
@@ -17,6 +17,6 @@ pub(crate) fn signatures() -> ModuleFunctionTypeMap {
         FunctionType::new([Felt, Felt, Felt, Felt], [Felt, Felt, Felt, Felt]),
     );
     account.insert(GET_ID, FunctionType::new([], [Felt]));
-    m.insert("miden:tx_kernel/account", account);
+    m.insert("miden::account", account);
```
bitwalker, line 20: **Question:** Won't the `foo:bar/baz`-style identifiers be the standard for Wasm-compiled components? If our packaging approach is based on Wasm components/WIT, we probably should make sure that we handle such identifiers correctly, rather than working around it on a case-by-case basis such as seems to be the case here.

If instead these aren't the identifiers we'd expect to see from a typical component, then feel free to disregard - I just want to make sure I know whether we're making this change due to an issue in the compiler, or for some other reason.

greenhat: > **Question:** Won't the `foo:bar/baz`-style identifiers be the standard for Wasm-compiled components? If our packaging approach is based on Wasm components/WIT, we probably should make sure that we handle such identifiers correctly, rather than working around it on a case-by-case basis such as seems to be the case here.
> 

Yes, here is an example in P2ID note script - https://github.com/0xPolygonMiden/compiler/blob/f255cfd5da3e27c35b784271dd7ed5ac2b6b3770/tests/integration/expected/wit_sdk_basic_wallet/basic_wallet_p2id_note.wat#L68

> If instead these aren't the identifiers we'd expect to see from a typical component, then feel free to disregard - I just want to make sure I know whether we're making this change due to an issue in the compiler, or for some other reason.

In this case, the `miden::account` is the module ID for the handwritten tx kernel MASM module. I suppose, when we use Wasm CM for tx kernel, we're going to re-work this.

bitwalker (approved):

PR #212, opened 2024-06-17
[8/x] Avoid generating empty blocks for `if.true` op in MASM

### LINKED ISSUE #207: Fill the empty `if.true else [op] end` workaround
As a temporary workaround for #206 fill the empty `if [..] else` with a `nop`-like op sequence (i.e. `push.1 drop`).

greenhat:
Close #207 

**This PR is stacked on the #210 and should be merged after it**

This PR adds a workaround to avoid generating empty `if.true` blocks in MASM when the IR block consists of the single `ret` instruction and the codegen does not generate any MASM ops for it (return value is already on the stack, etc.).

### DIFF
--- a/codegen/masm/src/codegen/emitter.rs
+++ b/codegen/masm/src/codegen/emitter.rs
@@ -178,6 +178,34 @@ impl<'b, 'f: 'b> BlockEmitter<'b, 'f> {
         // up these unused values is pushed into the successor on entry.
         self.drop_unused_operands();
 
+        // Workaround for the https://github.com/0xPolygonMiden/compiler/issues/207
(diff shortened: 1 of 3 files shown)


bitwalker (changes requested):
Due to 0xPolygonMiden/miden-vm#1360, I don't think we'll ever merge this. Let's keep it around though until that PR gets merged, then we can replace this with a PR that updates our dependency on `miden-assembly` to pick up the upstream changes.

Marked this as "request changes" to ensure we don't accidentally merge it.

greenhat:
Closing in favor of #221

Issue #214, opened 2024-06-18
Fuzz the compiler with a non-trivial code which is not overly complex


greenhat:
### Motivation
To fill the gap between manually written Wasm (too simple) and Rust programs using the malloc (too complex to debug) in our tests. We need something in the middle to cover a non-trivial control flow and memory ops (byte vs. word memory addressing), etc.

### Implementation details
Generate the Wasm code with `wasm-smith`, compile it to MASM and run them both, comparing the runtime state (stack, memory) afterward.

greenhat:
There is also `Wasmlike` paper at https://news.ycombinator.com/item?id=36992222 to generate the semantically valid Wasm programs.

PR #219, opened 2024-06-21
Semantic tests for Wasm code generated with `wasm-smith`

greenhat:
Ref #214 

Plan:
- Generate a one-function module with `wasm-smith`;
- Compile it to MASM;
- Execute them Wasmtime vs. Miden VM and check the runtime state (stack, memory);


bitwalker:
Is this something we want to pick up again, now that we do have some tests that execute Wasm (via `wasmi`)?

greenhat:
> Is this something we want to pick up again, now that we do have some tests that execute Wasm (via `wasmi`)?

I don't know. We have a fuzza(agent) that generates Rust code. I don't think wasm-smith produces code that is shaped exactly as rustc produces. I think we shouldn't bother spending time supporting wasm-smith generated code.

PR #220, opened 2024-06-24
[11/x] Use`FileName::as_str`for source file name to avoid enclosing virtual filenames in brackets

greenhat:
**This PR is stacked on the #218 and should be merged after it**

Use `FileName::as_str` for source name in Wasm translation to avoid `<name>` (enclosed in brackets) which is produced by `FileName::to_string` since the brackets are illegal in MASM module ids.

### DIFF
--- a/midenc-compile/src/stages/parse.rs
+++ b/midenc-compile/src/stages/parse.rs
@@ -44,7 +44,7 @@ impl Stage for ParseStage {
                     input,
                     session,
                     &WasmTranslationConfig {
-                        source_name: name.to_string().into(),
+                        source_name: name.as_str().unwrap().to_string().into(),
                         ..Default::default()
                     },
                 ),
@@ -52,7 +52,7 @@ impl Stage for ParseStage {
                     input,
                     session,
                     &WasmTranslationConfig {
-                        source_name: name.to_string().into(),
+                        source_name: name.as_str().unwrap().to_string().into(),
                         ..Default::default()
                     },
                 ),


bitwalker (approved):

bitwalker:
Having thought about this some more, I think actually we should tackle this issue a bit differently. See my comment in 0xPolygonMiden/miden-diagnostics#2. I'm going to leave this marked approved for now, so we can merge it as a temporary workaround, but the fundamental issue here requires a more appropriate fix I think. We can address it properly in another PR, or in this one, whichever you think is more appropriate.

greenhat:
> Having thought about this some more, I think actually we should tackle this issue a bit differently. See my comment [here](https://github.com/0xPolygonMiden/compiler/issues/208#issuecomment-2187579686). I'm going to leave this marked approved for now, so we can merge it as a temporary workaround, but the fundamental issue here requires a more appropriate fix I think. We can address it properly in another PR, or in this one, whichever you think is more appropriate.

I'm not sure if I follow. I'm using "unbracketed" name here to avoid `ConvertHirToMasm` failing at https://github.com/0xPolygonMiden/compiler/blob/8b8117fd95c04b3e3f8b4016754bf0c1fdbad1d6/codegen/masm/src/convert.rs#L117
I suppose we can leave brackets but treat every module name there as raw ids. Is that the idea?

bitwalker:
Right, my bad, I got this one mixed in with other issues related to module/procedure paths

PR #221, opened 2024-06-25
[8/x] Update the Miden VM deps  with  empty `if.true` support

greenhat:
**This PR is stacked on the #210 and should be merged after it**

This PR updates Miden VM crates to include empty `if.true` support introduced in 
https://github.com/0xPolygonMiden/miden-vm/pull/1360

### DIFF
--- a/hir/src/asm/isa.rs
+++ b/hir/src/asm/isa.rs
@@ -824,6 +824,8 @@ pub enum MasmOp {
     Emit(u32),
     /// Emit a trace event with the given code
     Trace(u32),
+    /// No operation
+    Nop,
 }
 
 macro_rules! unwrap_imm {
@@ -1099,7 +1101,8 @@ impl MasmOp {
             | Self::DebugFrame
             | Self::DebugFrameAt(_)
             | Self::DebugFrameRange(..)
-            | Self::Breakpoint => 0,
+            | Self::Breakpoint
+            | Self::Nop => 0,
         }
     }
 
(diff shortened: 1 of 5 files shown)


hir/src/asm/isa.rs
```diff
@@ -1404,6 +1407,18 @@ impl MasmOp {
             Instruction::Debug(DebugOptions::LocalInterval(start, end)) => {
                 Self::DebugFrameRange(unwrap_u16!(start), unwrap_u16!(end))
             }
+            Instruction::Nop => Self::Nop,
```
bitwalker, line 1410: I'm fine with `Nop` being at the end, since it is new, but lets group the other ops together with their non-immediate versions so they are easy to find.

hir/src/asm/isa.rs
```diff
@@ -1404,6 +1407,18 @@ impl MasmOp {
...
+            Instruction::LtImm(imm) => Self::LtImm(unwrap_imm!(imm)),
+            Instruction::LteImm(imm) => Self::LteImm(unwrap_imm!(imm)),
+            Instruction::GtImm(imm) => Self::GtImm(unwrap_imm!(imm)),
+            Instruction::GteImm(imm) => Self::GteImm(unwrap_imm!(imm)),
+            Instruction::U32LtImm(_) => todo!(),
```
bitwalker, line 1415: This should more or less work like the other immediate u32 ops right? Any reason to make these `todo!` here?

bitwalker (approved):
A few small nits, otherwise looks good

hir/src/asm/isa.rs
greenhat: Done in https://github.com/0xPolygonMiden/compiler/pull/221/commits/13af81c24e7ed03175fd59cd52bc64c1d2c10631

hir/src/asm/isa.rs
greenhat: Done in https://github.com/0xPolygonMiden/compiler/pull/221/commits/13af81c24e7ed03175fd59cd52bc64c1d2c10631

PR #222, opened 2024-06-26
[1/x] Update VM with the absolute paths support

### LINKED ISSUE #211: Assembler fails to parse emitted `mem::intrinsics::*` calls
After I drafted the solution for #207 the next error popped up. 
Emitted `mem::intrinsics::*` calls fail to be parsed by the assembler with the error:
```
  x invalid syntax
    ,-[user_ns::abi_transform_tx_kernel_get_inputs:47:25]
 46 |     u32div.16
 47 |     exec.intrinsics::mem::load_sw
    :                         ^|
    :                          `-- found a :: here
 48 |     push.2
    `----
  help: expected primitive opcode (e.g. "add"), or "end", or control flow
        opcode (e.g. "if.true")
```

I think we should import them with `use.intrinsics::mem` first and then call with `exec.mem::load_sw`.

greenhat:
Close #211

### DIFF
--- a/codegen/masm/src/emulator/mod.rs
+++ b/codegen/masm/src/emulator/mod.rs
@@ -6,7 +6,9 @@ mod functions;
 use std::{cell::RefCell, cmp, rc::Rc, sync::Arc};
 
 use miden_assembly::{ast::ProcedureName, LibraryNamespace};
-use midenc_hir::{assert_matches, Felt, FieldElement, FunctionIdent, Ident, OperandStack, Stack};
+use midenc_hir::{
+    assert_matches, Felt, FieldElement, FunctionIdent, Ident, OperandStack, Stack, Symbol,
+};
 use rustc_hash::{FxHashMap, FxHashSet};
 
 use self::functions::{Activation, Stub};
@@ -1547,6 +1549,19 @@ impl Emulator {
                     return Ok(EmulatorEvent::EnterLoop(body_blk));
                 }
                 Op::Exec(callee) => {
+                    // remove the `::` prefix (absolute path) if any from the
+                    // callee.module
+                    let callee = if callee.module.as_str().starts_with("::") {
+                        FunctionIdent {
+                            module: Ident::with_empty_span(Symbol::intern(
+                                &callee.module.as_str()[2..],
+                            )),
+                            function: callee.function,
+                        }
+                    } else {
+                        callee
+                    };
+                    dbg!(&callee);
                     let fun = self
                         .functions
                         .get(&callee)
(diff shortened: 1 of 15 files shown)


codegen/masm/src/emulator/mod.rs
```diff
@@ -1547,6 +1549,19 @@ impl Emulator {
                     return Ok(EmulatorEvent::EnterLoop(body_blk));
                 }
                 Op::Exec(callee) => {
+                    // remove the `::` prefix (absolute path) if any from the
```
greenhat, line 1552: @bitwalker I've tried to find anything resembling the invocation infra from the VM in the emulator to properly implement absolute paths but did not find any. I put a quick and dirty workaround stripping the `::` prefix.

bitwalker:
FYI - the absolute paths PR was merged, so we can switch away from my branch to `next`

codegen/masm/src/emulator/mod.rs
bitwalker: We may need to extend our representation a bit, in order to allow importing MASM modules from source. In particular, we may want to have a variant of `FunctionIdent` called `Callee` used in our MASM representation, like so:

```rust
pub enum Callee {
    MastRoot(RpoDigest),
    Path {
        module: Option<Ident>,
        function: Ident,
        absolute: bool,
    }
}
```

When translating HIR `FunctionIdent` to `Callee`, it would become  a path where the `module` would always be set, and `absolute` would always be `true` (since function identifiers in HIR are always absolute).

When translating MASM `InvocationTarget` to `Callee`, we'd convert `InvocationTarget::AbsolutePath` as an absolute callee, `InvocationTarget::ProcedurePath` as a relative path, `InvocationTarget::ProcedureName` as a relative path, and `InvocationTarget::MastRoot`, as a MAST root. Similar translations could be supported for other `InvocationTarget`-like procedure identifiers, but the MASM opcodes only use `InvocationTarget`.

The emulator then would need to resolve relative paths at runtime, but could treat absolute paths and MAST roots as, well, absolute references. The emulator currently does not track MAST roots at all, so that would be needed to have full compatibility, but would be a pretty simple thing to add.

greenhat: Thank you for the thorough explanation. Do you want me to implement it in this PR? I'd rather make an issue out of it and tackle it in a new PR.

bitwalker:
@greenhat Is this ready to be merged now?

greenhat:
Yes. I was waiting for your replay in https://github.com/0xPolygonMiden/compiler/pull/222#discussion_r1664197859 above regarding the emulator.

codegen/masm/src/emulator/mod.rs
bitwalker: @greenhat That's fine by me

PR #224, opened 2024-06-27
[2/x] `CompilerTest` stubs generation for imported functions in MASM program

### LINKED ISSUE #223: Generate a stub code for the import function while assembling the MASM
To test the ABI transformation, we need a tx-kernel API function to exist. Since calling the genuine function requires a context which is expensive to set up, we can generate a function and hard-code an arguments check and return values.

greenhat:
Close #223 

**This PR is stacked on the #222 and should be merged after it**

### DIFF
--- a/tests/integration/expected/abi_transform_tx_kernel_get_inputs_4.hir
+++ b/tests/integration/expected/abi_transform_tx_kernel_get_inputs_4.hir
@@ -1,6 +1,6 @@
 (component 
     ;; Modules
-    (module #abi_transform_tx_kernel_get_inputs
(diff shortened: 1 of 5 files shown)


greenhat:
@bitwalker The MASM assembly (with generated stub for tx kernel `get_inputs`) is done. Now, onto the runtime errors!
The first hit is:
```
failed to execute program on VM: FailedAssertion { clk: 17, err_code: 0, err_msg: None }
```
Which is close to the program start and can be easily located. :)

bitwalker (approved):

PR #226, opened 2024-07-02
[3/x] Remove synthetic MASM namespace workaround for the module compiled from Rust

greenhat:
Which was added as a workaround in #220 for the issue #208.

**This PR is stacked on the #224 and should be merged after it**


bitwalker (approved):

bitwalker:
@greenhat If this is ready it can be merged

Issue #225, opened 2024-07-01
Failed assertion while executing `get_inputs` ABI transformation test (`clk=17`) on VM


greenhat:
Error:
```
FailedAssertion { clk: 17, err_code: 0, err_msg: None }
```

Somewhere in the following code:
```
...
export."miden_tx_kernel_sys::get_inputs"
    mem_load.0x00000000
    push.16
    u32wrapping_sub
    dup.0
    dup.0
    push.2147483648
    u32and
    eq.2147483648
    assertz
    add.8
    u32assert
    dup.0
    u32mod.16
    dup.0
    u32mod.4
    swap.1
    u32div.4
    movup.2
    u32div.16
    exec.::intrinsics::mem::load_sw
    dup.1
    dup.0
    push.2147483648
    u32and
    eq.2147483648
    assertz
    add.12
    u32assert
...
```

To pinpoint the failing op I'm exploring the following options:
- Pull out a better error message from the VM;
- Execute the VM via the iterator and print the VM state after every op.

greenhat:
@bitwalker I confirmed this is caused by https://github.com/0xPolygonMiden/compiler/issues/174

bitwalker:
This is fixed by #242

Issue #174, opened 2024-04-18
`OpEmitter::assert_unsigned_int32` blows up for `u32` case if values have high(32nd) bit set


greenhat:
Discovered in #173.

The Rust `a >= b` (both `u32`) translates into the following IR:
https://github.com/0xPolygonMiden/compiler/blob/663b146b45a027770392f7ea76f7144cef34beea/tests/integration/expected/ge_u32.hir#L12-L17

From which the following MASM is generated:
https://github.com/0xPolygonMiden/compiler/blob/663b146b45a027770392f7ea76f7144cef34beea/tests/integration/expected/ge_u32.masm#L659-L670

The `assertz` fails if `a` or `b` have high(32nd) bit set, which is a valid u32 value.

**EDIT**: Can be reproduced by the commented tests mentioning this issue in their comment.
After the fix, these tests should be enabled.

greenhat:
@bitwalker I hit it in #225. As a temporary workaround I suggest we disable it and emit nothing when casting to u32.

bitwalker:
Just to confirm a detail here: the bit being set is the _32nd_ bit? Or the 31st? 2^32 is _not_ representable as a 32 bit integer, 2^32-1 is the maximum value that will fit in 32 bits.

greenhat:
From the `get_inputs` test the value under cast is 4294967280 which is a valid u32 value with the 32nd bit set. But we're expecting the 32nd bit to be zero for u32 at https://github.com/0xPolygonMiden/compiler/blob/59044947746ba79919faabb0c8f8ea52d6006842/codegen/masm/src/codegen/emit/int32.rs#L86

bitwalker:
Fixed in #242. The issue is that the wrong types of casts were being used - and more generally, we lacked an appropriate cast operation for reinterpreting bits as different types. 242 implements a new `bitcast` op which does just that, and is now used in all the places where a bitcast is appropriate (i.e. we don't care about preserving the value of the source type).

PR #232, opened 2024-07-04
[5/x] Disable emitting assert for u32 cast

greenhat:
**This PR is stacked on the #227 and should be merged after it**

This PR temporarily disables emitting assertion on u32 cast to get past #225 that is caused by #174.

Do not merge this PR if #174 is already fixed.

### DIFF
--- a/codegen/masm/src/codegen/emit/int32.rs
+++ b/codegen/masm/src/codegen/emit/int32.rs
@@ -114,8 +114,9 @@ impl<'a> OpEmitter<'a> {
     /// See `is_signed` for semantics and stack effects of the signedness check.
     #[inline]
     pub fn assert_unsigned_int32(&mut self) {
-        self.is_signed_int32();
-        self.emit(Op::Assertz);
+        // TODO: temporarily disabled until https://github.com/0xPolygonMiden/compiler/issues/174 is fixed
+        // self.is_signed_int32();
+        // self.emit(Op::Assertz);
     }
 
     /// Emits code to assert that a 32-bit value on the operand stack is equal to the given constant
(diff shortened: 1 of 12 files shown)


codegen/masm/src/codegen/emit/int32.rs
```diff
@@ -114,8 +114,9 @@ impl<'a> OpEmitter<'a> {
     /// See `is_signed` for semantics and stack effects of the signedness check.
     #[inline]
     pub fn assert_unsigned_int32(&mut self) {
```
bitwalker, line 116: I've looked into this issue, and the fix is to remove the call to `assert_unsigned_int32` from `cast` when casting from i32 to u32, from u32 to i32, and from i32 to u64; and replace it with a call to a new helper called `assert_i32`, see the implementation of `assert_i64` in `emit/int64.rs` for how to implement it.

The problem is that the `assert_unsigned_int32` helper is intended to assert that the sign bit of a 32-bit value is unset, _not_ to assert that the value is non-negative, which is how we are using it in `cast`.

All the other uses of `assert_unsigned_int32` are correct

bitwalker (changes requested):
I've had a chance to look into this issue, and found the problem. I'd suggest we implement the correct fix in this PR, rather than closing this one and opening a new one, but I'll leave that up to you.

bitwalker:
This can be closed in favor of #242 once it is merged

greenhat:
Closing in favor of #242

codegen/masm/src/codegen/emit/int32.rs
greenhat: I have no found `assert_i32` in #242 so I made #250

PR #233, opened 2024-07-04
[6/x] Add traced execution option on the VM in the integration tests

greenhat:
**This PR is stacked on #232 and should be merged after it**

### DIFF
--- a/tests/integration/src/exec_vm.rs
+++ b/tests/integration/src/exec_vm.rs
@@ -1,10 +1,10 @@
 use miden_core::{Program, StackInputs};
-use miden_processor::{DefaultHost, ExecutionOptions};
+use miden_processor::{DefaultHost, ExecutionError, ExecutionOptions};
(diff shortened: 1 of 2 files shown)


tests/integration/src/exec_vm.rs
```diff
@@ -1,10 +1,12 @@
+#![allow(unused)]
```
bitwalker, line 1: *Nit:* Is this actually needed? I prefer to only disable warnings module-wide if it is especially noisy and unhelpful; otherwise it's usually easier to just decorate specific items with this attribute instead.

bitwalker (approved):

tests/integration/src/exec_vm.rs
greenhat: Not really. I agree that decorating a specific item is a better approach.

greenhat: Done.

PR #236, opened 2024-07-05
[7/x] Fix memory cleanup in `Emulator::stop`

greenhat:
**This PR is stacked on the #233 and should be merged after it**

### DIFF
--- a/codegen/masm/src/emulator/mod.rs
+++ b/codegen/masm/src/emulator/mod.rs
@@ -570,7 +570,7 @@ impl Emulator {
     pub fn stop(&mut self) {
         self.callstack.clear();
         self.stack.clear();
-        self.memory.clear();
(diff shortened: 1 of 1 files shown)


codegen/masm/src/emulator/mod.rs
```diff
@@ -572,7 +572,7 @@ impl Emulator {
     pub fn stop(&mut self) {
         self.callstack.clear();
         self.stack.clear();
-        self.memory.clear();
+        self.memory = vec![Self::EMPTY_WORD; self.memory.len()];
```
bitwalker, line 575:
```suggestion
        self.memory.fill(Self::EMPTY_WORD);
```

greenhat: Good catch! In #237, I reworked it completely (it only cleans "dirty" addresses).

PR #237, opened 2024-07-05
[8/x] Add `Emulator::print_trace` option to print the current stack and memory on every step

greenhat:
**This PR is stacked on the #236 and should be merged after it**

### DIFF
--- a/codegen/masm/src/emulator/memory.rs
+++ b/codegen/masm/src/emulator/memory.rs
@@ -1,0 +1,59 @@
+use std::{
+    fmt::Debug,
+    ops::{Index, IndexMut},
+};
+
+use miden_core::FieldElement;
+use midenc_hir::Felt;
+use rustc_hash::FxHashSet;
+
+const EMPTY_WORD: [Felt; 4] = [Felt::ZERO; 4];
+
+pub struct Memory {
+    memory: Vec<[Felt; 4]>,
+    set_memory_addrs: FxHashSet<usize>,
+}
+
+impl Memory {
+    pub fn new(memory_size: usize) -> Self {
+        Self {
+            memory: vec![EMPTY_WORD; memory_size],
+            set_memory_addrs: Default::default(),
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.memory.len()
+    }
+
+    pub fn reset(&mut self) {
+        for addr in self.set_memory_addrs.iter() {
+            self.memory[*addr] = EMPTY_WORD;
+        }
+        self.set_memory_addrs = Default::default();
+    }
+}
+
+impl Index<usize> for Memory {
+    type Output = [Felt; 4];
+
+    fn index(&self, index: usize) -> &Self::Output {
+        &self.memory[index]
+    }
+}
+
+impl IndexMut<usize> for Memory {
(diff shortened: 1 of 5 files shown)


codegen/masm/src/emulator/memory.rs
```diff
@@ -0,0 +1,59 @@
...
+
+use miden_core::FieldElement;
+use midenc_hir::Felt;
+use rustc_hash::FxHashSet;
+
+const EMPTY_WORD: [Felt; 4] = [Felt::ZERO; 4];
+
+pub struct Memory {
+    memory: Vec<[Felt; 4]>,
+    set_memory_addrs: FxHashSet<usize>,
```
bitwalker, line 14: *Suggestion:* I would make this a `BTreeSet`, so that you can print the memory addresses in order, plus its probably faster than a hashset in this case, since memory addresses are totally ordered, and highly likely to be written to in order from least to greatest

codegen/masm/src/emulator/mod.rs
```diff
@@ -1242,6 +1255,11 @@ impl Emulator {
         // control flow effect occurred to reach it
         let ix_with_op = state.next();
         if let Some(ix_with_op) = ix_with_op {
+            if self.print_trace {
+                eprintln!("stk: {}", self.stack.debug().pretty());
+                eprintln!("mem: {:?}", self.memory);
```
bitwalker, line 1260: **Suggestion:** Might be nicer to put memory first, followed by the operand stack and current op, since that orders them from least significant to most significant (and memory is likely to be very noisy in larger programs with rodata and such, which might push visibility of the operand stack off screen).

hir/src/asm/stack.rs
```diff
@@ -458,6 +458,19 @@ impl<'a, E: StackElement, T: ?Sized + Stack<Element = E>> DebugStack<'a, T> {
         self.limit = Some(n);
         self
     }
+
+    pub fn pretty(&self) -> String {
```
bitwalker, line 462: I think we should make this a `Display` impl, and base it on the existing `Debug` impl. The only difference between this and `Debug` seems to be how we format each stack element, which is certainly useful, but I'd rather use `Display` for it, since it is far more ergonomic than building strings directly, and you get `to_string` for free.

bitwalker (changes requested):
Overall looks good! One minor suggestion about ordering of items in the output that you can take or leave based on your preference; and two small changes that I think we should make prior to merging.

codegen/masm/src/emulator/memory.rs
greenhat: Good catch! I missed it since I only dealt with a couple of addresses, but the order is important.
Done in https://github.com/0xPolygonMiden/compiler/pull/237/commits/8ca444b1a4e62a4faf11005216554dc58cac6a1a

hir/src/asm/stack.rs
greenhat: Good point! Done in https://github.com/0xPolygonMiden/compiler/pull/237/commits/8ca444b1a4e62a4faf11005216554dc58cac6a1a

bitwalker (approved):
LGTM!

Issue #239, opened 2024-07-08
Assembler cannot resolve `::std::math::u64::checked_lte` 


greenhat:
While working on #230 the assembler returned the following error:
 
```
VM assembly error:
  x assembly failed
  help: see diagnostics for details

Error:   x undefined procedure
    ,-[abi_transform_tx_kernel_get_inputs_4:67:32]
 66 |         push.4294967295.1
 67 |         exec.::std::math::u64::checked_lte
    :                                ^^^^^|^^^^^
    :                                     `-- unable to resolve this reference to its definition
 68 |         assert
    `----
Error:   x name resolution cannot proceed
    ,-[./asm/math/u64.masm:33:6]
 32 | #! The input values are assumed to be represented using 32 bit limbs, but this is not checked.
 33 | #! Stack transition looks as follows:
    :      ^^^^^|^^^^^
    :           `-- this name cannot be resolved
 34 | #! [b_hi, b_lo, a_hi, a_lo, ...] -> [c_hi, c_lo, ...], where c = (a + b) % 2^64
    `----
```

greenhat:
@bitwalker The checked variants were removed in https://github.com/0xPolygonMiden/miden-vm/pull/1142
Do we want to keep using the checked variants as intrinsics or switch to unchecked variants?

bitwalker:
@greenhat We should replace the use of `checked_*` with the unchecked variant, but prefix the call with a call to  `std::math::u64::u32assert4` to check the inputs as expected.

For arithmetic operations, we should do the same, but additionally use the overflowing variants for the actual op (rather than unchecked), and assert that no overflow occurs.

bitwalker:
Fixed in #242

Issue #228, opened 2024-07-02
Emulator fails to execute code `eq_*64` that VM executes fine


greenhat:
On the following MASM:
```
# mod test_rust_23a27398cbfb7447cb610e7b877565f4e43d633b55f98504051b93381d40740f

export.entrypoint
    movdn.3
    movdn.3
    movup.2
    eq
    movdn.3
    eq
    and
end


begin
    exec.::test_rust_23a27398cbfb7447cb610e7b877565f4e43d633b55f98504051b93381d40740f::entrypoint  
end
```

With error:
```
thread 'rust_masm_tests::instructions::eq_i64' panicked at /Users/dzadorozhnyi/src/miden-ir/hir/src/asm/stack.rs:284:9:
invalid operand stack index (3), only 2 elements are available
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2d24fe591f30386d6d5fc2bb941c78d7266bf10f/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/2d24fe591f30386d6d5fc2bb941c78d7266bf10f/library/core/src/panicking.rs:72:14
   2: midenc_hir::asm::stack::Stack::movdn
             at /Users/dzadorozhnyi/src/miden-ir/hir/src/asm/stack.rs:284:9
   3: midenc_codegen_masm::emulator::Emulator::run_once
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/emulator/mod.rs:1296:21
   4: midenc_codegen_masm::emulator::Emulator::step
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/emulator/mod.rs:1076:27
   5: midenc_codegen_masm::emulator::Emulator::run
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/emulator/mod.rs:1168:25
   6: midenc_codegen_masm::emulator::Emulator::invoke_function
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/emulator/mod.rs:676:17
   7: midenc_codegen_masm::emulator::Emulator::invoke
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/emulator/mod.rs:627:46
   8: midenc_codegen_masm::emulator::Emulator::start
             at /Users/dzadorozhnyi/src/miden-ir/codegen/masm/src/emulator/mod.rs:520:21
[...]

Issue #169, opened 2024-04-10
Miden and Rust disagree on the data layout for felt/word


bitwalker:
While discussing an issue that @greenhat encountered while working on the Wasm frontend, it became apparent that we missed a hole in the design for lowering Rust code to Miden (and more generally, any language like Rust, which is designed for a byte-addressable memory model).

Specifically, we made the following design choices:

* Miden VM should be viewed as a 32-bit machine from the perspective of Rust. This is for multiple reasons:
  - Miden's notion of a memory address is a u32 value
  - Miden's instruction set only provides efficient implementations of key operations in terms of u32 values; e.g. bit manipulation functions like shifts/logical operators, load/store ops, and two's complement arithmetic functions.
* Translate byte-addressable memory models to Miden's word-addressable memory as follows: Miden's memory is "word"-addressable (where "word" is 4 field elements), but can be viewed as felt-addressable if you pair a word address with an element index. It can further be viewed as byte-addressable if you constrain each addressable felt to the u32 value range, which allows us to perform bit masking/shifting ops to extract specific bits - pairing a word address with an element index, and a byte offset allows you to reference a specific byte anywhere in Miden's memory.
* Designate the `f64` type as a sentinel type to indicate that a given value should be treated as a field element when lowered to Miden Assembly. There is no support for floating point values in Miden, so their presence in any Rust code we compile is treated as an error, except in specific narrow circumstances. This allows us to recognize when it is safe to translate certain instrinsics to felt-native ops on the value when lowering to Miden Assembly.

These choices were made to ensure that code compiled from Rust to Miden would be maximally efficient, considering the semantic differences between the targets Rust was designed to support, and Miden. Furthermore, it does not require us to perform unsafe transformations on rustc-generated code in order to work on Miden, a major win. Most importantly however, it allows us to take advantage of the fact that these design choices allow Miden to be viewed as a variant `wasm32` target (currently, using either `wasm32-unknown-unknown`, or `wasm32-wasi`). We gain an enormous amount of tooling and support as a result.

## The Problem

[...]

Issue #74, opened 2023-11-22
Failing instruction tests in the integration suite


greenhat:
### 64-bit ops are not implemented in the codegen

```rust
// test_bool_op!(ge, >=, u64);
// test_bool_op!(ge, >=, i64);
// test_bool_op!(gt, >, u64);
// test_bool_op!(gt, >, i64);
// test_bool_op!(le, <=, u64);
// test_bool_op!(le, <=, i64);
// test_bool_op!(lt, <, u64);
// test_bool_op!(lt, <, i64);
// test_int_op!(add, +, u64);
// test_int_op!(add, +, i64);
// test_int_op!(sub, -, u64);
// test_int_op!(sub, -, i64);
// test_int_op!(mul, *, u64);
// test_int_op!(mul, *, i64);
// test_int_op!(div, /, u64);
// test_int_op!(div, /, i64);
// test_int_op!(rem, %, u64);
// test_int_op!(rem, %, i64);
// test_unary_op!(neg, -, u64);
// test_unary_op!(neg, -, i64);
// test_unary_op!(not, !, u64);
// test_unary_op!(not, !, i64);
// test_int_op!(shl, <<, u64);
// test_int_op!(shl, <<, i64);
// test_int_op!(shr, >>, u64);
// test_int_op!(shr, >>, i64);
// test_unary_op!(neg, -, i64);
```

### MASM compilation error (missing import for intrinsic)

Might be fixed by #67

```rust
//
// Comparison ops
//
// test_bool_op!(ge, >=, i32);
// test_bool_op!(ge, >=, i16);
// test_bool_op!(ge, >=, i8);
//
// test_bool_op!(gt, >, i32);
// test_bool_op!(gt, >, i16);
// test_bool_op!(gt, >, i8);
//
// test_bool_op!(le, <=, i32);
// test_bool_op!(le, <=, i16);
// test_bool_op!(le, <=, i8);
//
// test_bool_op!(lt, <, i32);
// test_bool_op!(lt, <, i16);
// test_bool_op!(lt, <, i8);
//
// Arithmetic ops
//
// test_int_op!(mul, *, u32);
// test_int_op!(mul, *, u16);
// test_int_op!(mul, *, u8);
// test_int_op!(mul, *, i32);
// test_int_op!(mul, *, i16);
// test_int_op!(mul, *, i8);
//
// Bitwise ops
//
// test_int_op!(shr, >>, i8);
// test_int_op!(shr, >>, i16);
[...]

PR #242, opened 2024-07-12
fix: implement solution for felt representation mismatch between rust and miden

bitwalker:
This PR implements fixes for the following items:

- #169
- #239 
- #228 
- Most of the issues in #74 (a few of the 64-bit ops may still be unimplemented, but most are now)

See the commit messages for specific changes, but in general:

- Felt is now represented as f32 in Rust
- Felt is now 32-bits in the HIR type system
- Conversions to/from Felt have been rewritten accordingly
- Lowerings for 64-bit ops have been reviewed/tweaked, and in some cases replaced with stdlib calls. There were bugs found during this process that may be related to other ongoing issues being diagnosed
- Lowerings for u128 have been implemented (only for a limited subset of ops currently)
- The codegen crate now links in the Miden stdlib when using `MasmCompiler`, other areas of the codebase that do not test via that interface can make use of the same primitives it uses, but I have not done that here.

@greenhat One issue that will come up once this is merged (temporarily) is that the `cargo-miden` tests will fail until we update the `rust-templates` repo to point to this commit. The templates refer to a version of the compiler that uses the 64-bit felt representation, which will cause an error. I tested locally by manually specifying my local `rust-templates` clone, where I used a path dependency to refer to the version of the compiler I was working on, so this PR has a working test suite - it just won't work in CI until we make the changes I just described. Unfortunately I don't know of any way around it in this case.

We should probably merge this before our other PRs - I suspect some of the issues I found and fixed will help in your PR stack with testing and what not, and the spills PR needs these changes as my tests depend on the 32-bit felt layout.


greenhat:
> We should probably merge this before our other PRs - I suspect some of the issues I found and fixed will help in your PR stack with testing and what not, and the spills PR needs these changes as my tests depend on the 32-bit felt layout.

Sounds good! I'm still in the middle of the spills PR review. I could switch to this one and then resume spills PR review.

bitwalker:
@greenhat I'd prioritize this one first, but if you are in the middle of a tricky part of the spills PR don't feel like you have to switch in the middle of that - just whenever you find a convenient time to context switch.

greenhat:
> @greenhat One issue that will come up once this is merged (temporarily) is that the `cargo-miden` tests will fail until we update the `rust-templates` repo to point to this commit. The templates refer to a version of the compiler that uses the 64-bit felt representation, which will cause an error. I tested locally by manually specifying my local `rust-templates` clone, where I used a path dependency to refer to the version of the compiler I was working on, so this PR has a working test suite - it just won't work in CI until we make the changes I just described. Unfortunately I don't know of any way around it in this case.

What I did in such cases:
1. Make a PR in `rust-templates` repo and use a last commit from this PR as a dependency;
2. Use a commit from step 1 PR in cargo extension in this PR for fetching `rust-templates`;
3. Update the commit hashes if merging these PRs renders some commits unavailable.

It's a bit of extra work, but it keeps CI green.

frontend-wasm/src/code_translator/mod.rs
```diff
@@ -327,10 +327,6 @@ pub fn translate_operator(
             // https://www.w3.org/TR/wasm-core-1/#op-isub
             state.push1(builder.ins().sub_wrapping(arg1, arg2, span));
         }
-        Operator::F64Sub => {
```
greenhat, line 330: Good catch! Thanks!

bitwalker:
@greenhat I'll merge this PR once CI is green, but if you are curious, take a look at the latest commit to see how I set up usage of the new rust-templates options

PR #244, opened 2024-07-14
feat(codegen): teach instruction scheduler about side effects

### LINKED ISSUE #191: Ensure that load/store ops are not reordered with respect to each other in treegraph
Currently, we do not attach control dependencies to side-effecting operations in the computation of the dependency graph, which is used to derive the treegraph from which we schedule the actual Miden Assembly instructions. The result is that things like load and store instructions contained in different sub-trees of the treegraph can be reordered with respect to one another. This is not an issue when the memory referenced by those instructions is disjoint, but in general is an issue, and applies to other side-effecting instructions as well, including function calls.

The dependency graph already supports the notion of "control" dependencies, which are edges that instead of representing a data dependency, represent an explicit ordering of instructions such that control must pass through the predecessor instruction before reaching the successor. We actually use this already to ensure that side-effecting operations with no/unused results are always executed before the terminator instruction of that block, by attaching a control dependency from the terminator to the side-effecting operation, such that the latter must always be scheduled before the former.

We simply need to ensure control dependencies for other instruction relationships are added to the dependency graph.

bitwalker:
Our instruction scheduler is based on a block-local data dependency graph that is decomposed into a structure we call a treegraph. The topological ordering of nodes in the treegraph, determines the order in which expression trees in a block will be scheduled.

This necessarily led to the need for some notion of control dependencies: first to ensure that the block terminator was always the last thing scheduled; then to handle instructions with no results that had side effects, as our dead-code elimination pass would remove nodes from the dependency graph which appeared dead, and control dependencies ensured that they were both kept live, and scheduled late in the block.

One thing we didn't handle however, was implicit control dependencies, manifested in the relationship between instructions with side effects - loads and stores, calls to side-effectful functions, etc. If there was no path between them in the dependency graph, they could be reordered relative to one another, which can cause all kinds of weird behavior in the final program. Consider moving a write to memory after a read which expects to observe that write - not good.

This commit refines our existing control dependency handling a bit, by ensuring that within a block, memory reads and writes form implicit control dependencies on other instructions which could affect the state of the world it sees. Memory reads implicitly depend on any writes which appear earlier in the block, ensuring that the writes can't be moved past the read. Memory writes implicitly depend on both previous writes _and_ reads - we don't want to reorder stores relative to each other, and we don't want to move a store above a load, potentially affecting the value observed by that load.

We don't yet have good tests to really determine how drastic of an affect this has on the scheduler, but the one expect test affected by this change (and included in this commit), demonstrates that it does more faithfully preserve the original program order in HIR. For now these extra edges in the dependency graph are placed very conservatively, which reduces the opportunities for the scheduler to place instructions as optimally as possible. We'll need to introduce some static analyses that will let us reason more precisely about the relationships between certain loads/stores, and to determine if a function call actually has memory effects (or other side effects, such as the advice provider); these would let us remove some edges, and bring back optimization opportunities.

Closes #191 

---

@greenhat I've based this on my branch in #241, so we'll want to merge that one first.

### DIFF
--- a/codegen/masm/src/codegen/scheduler.rs
+++ b/codegen/masm/src/codegen/scheduler.rs
@@ -1150,18 +1150,25 @@ fn build_dependency_graph(
     graph
 }
 
-/// Discover any instructions in the given block that have no predecessors, but that must be
-/// scheduled anyway, i.e. due to side effects - and make the block terminator dependent on them to
-/// ensure that they are scheduled.
+/// This function performs two primary tasks:
+///
+/// 1. Ensure that there are edges in the graph between instructions that must not be reordered
+/// past each other during scheduling. An obvious example is loads and stores: you might have two
+/// independent expression trees that read and write to the same memory location - but if the order
+/// in which the corresponding loads and stores are scheduled changes, it can change the behavior
+/// of the program. Other examples include function calls with side effects and inline assembly.
+///
+/// 2. Ensure that even if an instruction has no predecessors, it still gets scheduled if it has
+/// side effects. This can happen if there are no other effectful instructions in the same block.
+/// We add an edge from the block terminator to these instructions, which will guarantee that
+/// they are executed before leaving the block.
 ///
 /// We call these instruction->instruction dependencies "control dependencies", since control flow
-/// in the block depends on them being executed first. In a way these dependencies are control-flow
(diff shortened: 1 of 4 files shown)


greenhat (approved):
Looking good!

PR #258, opened 2024-07-24
chore: add nightly rustc requirement to cargo-miden README.md

greenhat:


tools/cargo-miden/README.md
```diff
@@ -4,6 +4,12 @@ This crate provides a cargo extension that allows you to compile Rust code to Mi
 
 ## Installation
 
+In order to install(build) the extension, you need to have the nightly Rust toolchain installed:
+
+```bash
+rustup toolchain install nightly-2024-03-10
```
bitwalker, line 10: If building from the root of the compiler repo, this is already implied, as the toolchain in `rust-toolchain.toml` is used. It would probably be better to add one in the `tools/cargo-miden` directory if we expect people to build `cargo-miden` from that directory instead, that way we don't need to hardcode a specific nightly version in the README (which will change on a regular basis).

However, if you are expecting users to run `cargo install cargo-miden`, then just installing the toolchain isn't enough, you have to specify what toolchain to use when running `cargo install`, i.e. `cargo +nightly-2024-03-10 install cargo-miden`, otherwise `cargo install` will use whatever the users default toolchain is (typically stable).

bitwalker (approved):
I'm approving this so that I'm not blocking you from getting this merged, but the comment I left has some suggested changes (what route you pick is up to you).

tools/cargo-miden/README.md
greenhat: Oh, you're right! I forgot about specifying the toolchain in `cargo install`. Thanks! Fixed.
I've written this readme for those who want to install the extension without checking out the repo, so I don't think putting `rust-toolchain.toml` in `tools/cargo-miden` is necessary.

bitwalker (approved):

PR #260, opened 2024-07-26
feat: implement rodata initialization

bitwalker:
**NOTE:** This PR is a work in progress, and blends multiple issues together as I work through implementation and testing of both rodata init, as well as some related issues in our test suite.

A rough summary of what this contains:

* Refactor the compiler driver to support multiple inputs
* Refactor the compiler pipeline to support a mix of MASM and HIR inputs
* Refactor the compiler pipeline to always output a `midenc_codegen_masm::Program`
* Move responsibility for assembly of a `miden_core::Program` into `midenc_codegen_masm::Program::assemble`
* Implement support for generating two different types of extra setup code for executable programs:
  - Data segment initialization on program start, before invoking the entrypoint
  - Test harness, which allows initializing memory of the program from the advice stack on startup. This is primarily intended to allow initializing shadow stack memory such that a Rust-compiled function can be called as if the caller is itself a Rust function, which is useful for testing, and necessary for testing non-immediate inputs.
* Implement a richer Miden VM test executor, which allows initializing advice inputs in addition to the operand stack. It also captures much more information about the program under test, and can dump a variety of useful information when errors occur.
* Make use of the fact that we parse DWARF debug info when parsing Wasm modules, to allow reifying actual source spans during translation to HIR. This depends on having DWARF debug info in the Wasm module, as well as the source paths contained in the Wasm, actually mapping to a file on disk, so that we can read it into memory in order to compute byte offsets (WIP)
* Fix the blake3 and get_inputs tests in the integration test suite (WIP, trying to get debug info working so that it is easier to determine where an assertion is being raised when the test fails; would also like to test that the test harness itself is set up properly).

This PR will remain a draft until the blake3 and get_inputs tests are working, at which point I will try to rework the commits in this PR into more granular changes. At this point, there are debugging statements and other cruft present throughout the code, and the test suite may be broken due to some of the changes not being handled yet.


codegen/masm/src/masm/program.rs
```diff
@@ -41,43 +81,204 @@ impl Program {
...
+            block.push(Op::MemStoreImm(word_addr));
+        } else {
+            block.push(Op::Drop);
+        }
+    }
+
+    /// Emit the sequence of instructions necessary to consume rodata from the advice stack and
+    /// populate the global heap with the data segments of this program, verifying that the
+    /// commitments match.
+    fn emit_data_segment_initialization(&self, block: &mut Block) {
```
greenhat, line 179: I was naively thinking of writing a rodata bytes directly into memory with `MemStoreImm`. Could you please elaborate on why go through the advice stack? I suspect it's an obvious reason, but I fail to see it.

hir/src/asm/isa.rs
```diff
@@ -1781,17 +1771,25 @@ impl MasmOp {
             Self::U32Clo => Instruction::U32Clo,
             Self::U32Cto => Instruction::U32Cto,
             Self::U32Lt => Instruction::U32Lt,
-            Self::U32LtImm(imm) => Instruction::U32LtImm(imm.into()),
+            Self::U32LtImm(imm) => return smallvec![Instruction::PushU32(imm), Instruction::U32Lt],
```
greenhat, line 1774: Just a heads up. I removed immediates variant of these ops in my VM update PR https://github.com/0xPolygonMiden/compiler/pull/261

sdk/stdlib-sys/src/stdlib/crypto/hashes.rs
```diff
@@ -74,163 +77,100 @@ extern "C" {
...
-    ),
+    extern_hash_1to1: unsafe extern "C" fn(u32, u32, u32, u32, u32, u32, u32, u32, *mut u8),
 ) -> [u8; 32] {
-    let mut felts_input = [Felt::from_u64_unchecked(0); 8];
-    for i in 0..8 {
-        felts_input[i] = Felt::from_u64_unchecked(u32::from_le_bytes(
-            input[i * 4..(i + 1) * 4].try_into().unwrap(),
-        ) as u64);
-    }
+    let input = unsafe { core::mem::transmute::<_, [u32; 8]>(input) };
```
greenhat, line 106: Nice!

tests/integration/src/rust_masm_tests/abi_transform/tx_kernel.rs
```diff
@@ -50,7 +46,11 @@ fn test_get_inputs(test_name: &str, expected_inputs: Vec<Felt>) {
     test.expect_masm(expect_file![format!("../../../expected/{artifact_name}.masm")]);
 
     let vm_program = test.vm_masm_program();
-    // let vm_out = execute_vm_tracing(&vm_program, &[]).unwrap();
+
+    let exec = MidenExecutor::new(vec![]);
+    let trace = exec.execute(&vm_program);
+    let vm_out = trace.into_outputs();
+    dbg!(&vm_out);
```
greenhat, line 53: I suppose the plan is to use `trace.read_from_rust_memory` to compare with expected inputs, right? Just like in the `test_blake3_hash` test.

hir/src/asm/isa.rs
bitwalker: I'd prefer to keep them in our IR, even though they are gone from the underlying MASM AST, because it is easier to canonicalize to that form when lowering from HIR, and then do peephole optimization based on that canonical form. We don't do that yet, but will start doing so pretty soon.

tests/integration/src/rust_masm_tests/abi_transform/tx_kernel.rs
bitwalker: Yeah exactly, we'll get some return values on the operand stack, but will need to use `read_from_rust_memory` to validate the contents of the returned vector.

hir/src/asm/isa.rs
greenhat: Oh, I meant I removed the VM's counterparts, not ours. The changes in my PR are exactly as you made here. Sorry for the confusion.

bitwalker: Oh I see, my bad!

tests/integration/src/compiler_test.rs
```diff
@@ -81,196 +50,517 @@ impl HirArtifact {
...
+/// Secondly, we need to prepare the [midenc_session::Session] object for the compiler. This is
+/// where we specify inputs, and various bits of configuration that are important to the test, or
+/// which are needed in order to obtain useful diagnostic output. This phase requires us to
+/// construct the base configuration here, but make it possible to extend/alter in each specific
+/// test.
+///
+/// Lastly, we must run the test, and in order to do this, we must know where our inputs and outputs
+/// are, so that we can fetch files/data/etc. as needed; know the names of things to be called, and
+/// more.
+pub struct CompilerTestBuilder {
```
greenhat, line 194: I love the new design! `CompilerTest` started small but quickly grew out of control. I was not happy with it for a long time but did not find a time to properly re-design it. Thank you!

greenhat (commented):
Hey, I think `intrinsics::mem::store_dw` silently doing nothing might cause troubles. See https://github.com/0xPolygonMiden/compiler/blob/401845ecb9e9bbe907470165cf3db99abd3d588d/codegen/masm/intrinsics/mem.masm#L571-L575
There are a few of them in both `blake3` and `get_inputs` tests and I'm not so sure anymore that we're not hitting them on the happy path.
I suggest to crash(via assert) there until I have it implemented.

bitwalker:
@greenhat Yeah I've been suspicious that we're hitting it and not aware of it, and unfortunately the way I stubbed out that intrinsic doesn't help! Good call on making it assert, I'll give it a unique error code too, so we'll know for sure if we hit it.

I've got the last debug info PR for the VM in review now, so I'm going to get back to this today and try to get it wrapped up now that I can actually see what's going on.

tests/integration/src/compiler_test.rs
bitwalker: For how quickly and haphazardly we had to grow the test suite, I think your initial implementation has held up pretty well! It's made it through several changes/refactorings without too much pain in the process, which is all you really can ask for. The nice thing is that by the time I got around to the changes in this PR, the existing code made it very clear what the patterns were, so we have you to thank for doing the hard work of finding them first 😃

frontend-wasm/src/miden_abi/transform.rs
```diff
@@ -15,20 +19,40 @@ enum TransformStrategy {
...
-        _ => panic!("No transform strategy found for function {}", function_id),
+fn get_transform_strategy(module_id: &str, function_id: &str) -> TransformStrategy {
+    #[allow(clippy::single_match)]
+    match module_id {
+        "std::mem" => match function_id {
+            stdlib::mem::PIPE_WORDS_TO_MEMORY => return TransformStrategy::ReturnViaPointer,
+            stdlib::mem::PIPE_DOUBLE_WORDS_TO_MEMORY => return TransformStrategy::ReturnViaPointer,
+            _ => (),
+        },
+        "std::crypto::hashes::blake3" => match function_id {
```
greenhat, line 30: `"std::crypto::hashes::blake3"` string literal is also used in `stdlib::crypto::hashes` module. It'd be better to define in one place.

greenhat (approved):
Great job! Just a couple of minor notes.

greenhat:
I'm taking a closer look into the failing tests now. From the get-go, I see that some expected tests are failing, so `UPDATE_EXPECT=1` should take care of those issues, and we will see what's failing next.

greenhat:
#271 is ready to merge

frontend-wasm/src/module/module_env.rs
```diff
@@ -513,33 +533,32 @@ impl<'a, 'data> ModuleEnvironment<'a, 'data> {
...
                 let ty = convert_valtype(ty);
                 locals.push((cnt, ty));
             }
-            self.result.debuginfo.wasm_file.funcs.push(FunctionMetadata {
+            self.result.wasm_file.funcs.push(FunctionMetadata {
                 locals: locals.into_boxed_slice(),
                 params: sig.params().into(),
             });
         }
-        body.allow_memarg64(false);
```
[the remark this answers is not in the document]
bitwalker: This is removed because `wasmparser` moved control over this to `WasmFeatures`, this function no longer exists.

codegen/masm/src/masm/program.rs
bitwalker: This function actually applies for more than testing, it is also how we initialize global memory on program start. The reason for using the advice stack is because it is vastly more efficient via `pipe_double_words_to_memory`, and because large amounts of memory would require many instructions to encode, making the program size much larger. By providing it via the advice stack, the actual initialization routine is compact, and we can load arbitrary amounts of data without affecting the program size.

The downside is that we have to orchestrate getting this data into the advice stack somehow, and there is no automation for this currently. My hope is that we solve this with future improvements to the VM.

PR #271, opened 2024-08-06
Fix `not an executable` error and `rust_sdk::account` test in integration tests

greenhat:


tests/integration/src/compiler_test.rs
```diff
@@ -998,8 +999,14 @@ impl CompilerTest {
         let output = midenc_compile::compile_to_memory(self.session.clone())
             .unwrap()
             .expect("no codegen outputs produced, was linking disabled?");
-        let midenc_codegen_masm::MasmArtifact::Executable(program) = output else {
-            panic!("expected executable output, got library");
+        let program = match output {
+            midenc_codegen_masm::MasmArtifact::Executable(program) => program,
+            midenc_codegen_masm::MasmArtifact::Library(lib) => {
+                let Some(entrypoint) = self.entrypoint else {
```
bitwalker, line 1005: Since we have an entrypoint, what we can do instead is ensure that we add `--entrypoint ID --exe` to the set of `midenc_flags`; this will tell the linker to use that function as the entrypoint, and to link an executable, rather than a library. I actually meant to do that at one point, but looks like I forgot. In any case, that would be a more general approach that leans on the driver to do this for us (and will help us catch if for some reason the driver is producing libraries even when we specify an entrypoint).

I do think we should keep your addition of `Program::from_lib` though, that's still useful.

We should probably make the driver set `--exe` if `--entrypoint` is given, for the ergonomics, but that's a small issue.

codegen/masm/src/masm/program.rs
```diff
@@ -76,6 +76,14 @@ impl Program {
         })
     }
 
+    /// Create a new [Program] from a [Library] and an entrypoint
+    pub fn from_lib(entrypoint: FunctionIdent, library: Library) -> Self {
```
bitwalker, line 80:
```suggestion
    pub fn from_library(entrypoint: FunctionIdent, library: Library) -> Self {
```

bitwalker (changes requested):
Looks good, however there is actually a mechanism we can use to do this more generally via the `CompilerTestBuilder` and driver flags, so despite this being a perfectly acceptable fix, I think it is better to also test that part of the driver infrastructure by leaning on it for this.

tests/integration/src/compiler_test.rs
bitwalker: I've actually done this on my branch, if you want to just open a PR to add the `Program::from_lib` changes

greenhat:
I'm closing this since most of these tests are already fixed in the next branch and in #274

PR #275, opened 2024-08-08
Fix `build_new_project_from_template` cargo extension test

greenhat:
Add `miden::*` tx kernel modules to the `Linker::allow_missing` list.
Add stdlib linking in the cargo extension.
Fix session output path generation.


midenc-session/src/outputs.rs
```diff
@@ -144,7 +144,8 @@ impl OutputFiles {
             .unwrap_or_else(|| OutputFile::Real(self.output_path(ty)));
         if let OutputFile::Real(ref mut path) = output {
             if let Some(name) = name {
-                path.set_file_name(name);
+                *path = path.join(name);
```
bitwalker, line 147: There's another bug here actually - we should set the extension on the path only when one of the following is true:

1. ) It is intermediate output type (e.g. HIR)
2. ) The path is _not_ sourced from `self.out_file`

When the path is sourced from `self.out_file`, we want to use that _exact_ path for the final output of the compiler (i.e. the output type is MAST - in the future it will probably be MASL for this, but for now it is MAST). This is because the user specifically requested that path be used for the output.

The other catch, is that for intermediate output types, we should only use `self.out_file` to derive a path when `self.out_dir` is _not_ set (which should be the case already, but we should verify that is the case). In other words, the intermediate outputs should go directly in `self.out_dir`, named after the session (or `name`, if set), with the output type-specific extension - while `self.out_file` should only be used for the final output. The use of `self.out_file` to derive `self.out_dir` should be done when `OutputTypes` is constructed probably, rather than here, but regardless, that's the semantics I want for the driver.

Putting this in terms of the CLI itself:

```
# Place all outputs in the current directory with file_stem `foo`
midenc compile --name foo out.wasm

# Place all outputs in `target/` with file_stem `foo`
midenc compile --output-dir target/ foo.wasm

# Place all outputs in `out` with file stem `foo`, except for MAST
# which will be written to `mast/out.mast`
midenc compile --output-dir out/ -o mast/out.mast foo.wasm

# Place all outputs in `out` with file stem `foo`, except for MAST
# which will be written to `out/out.mast`
midenc compile -o mast/out.mast foo.wasm
```

We actually have both `--output-dir` and `--target-dir`, one for actual outputs, and one for compiler intermediates (i.e. what `rustc` puts in `target/`), but we're not using `--target-dir` correctly currently. So to be more precise we should probably revisit all of it in one go.

I guess let's leave this for now, and I'll revisit all of the output orchestration later.

tools/cargo-miden/src/build.rs
```diff
@@ -36,7 +36,7 @@ pub fn build_masm(
     }]);
     let project_type = if is_bin { "--exe" } else { "--lib" };
     let source_manager = Arc::new(DefaultSourceManager::default());
-    let options = midenc_compile::CompilerOptions::parse_options(&[project_type])
+    let options = midenc_compile::CompilerOptions::parse_options(&[project_type, "-l", "std"])
```
bitwalker, line 39: I'm planning to make `-lstd` implicit, but this is a good workaround for the moment.

bitwalker (approved):

bitwalker:
Manually merged in b71259c

midenc-session/src/outputs.rs
greenhat: Thanks for the detailed explanation! The examples help a lot!

bitwalker: You bet! Now that I've refactored the output handling, things work more or less like so:

* `--target-dir` sets the directory in which all intermediate outputs are placed, it defaults to `$CWD/target/midenc`
* `--output-dir` set the directory in which to place final outputs, for now this is basically overlapping with `--output-file`, just one specifies a directory, the other a specific file; but we may end up with multiple outputs once we start producing packages, in which case the distinction makes more sense
* `--output-file` sets the file path for whatever compiler output is most appropriate (just like `--output-dir`, this only applies to assembled outputs), for now this is the `.masl` file containing the `miden_assembly::Library` or `miden_core::Program`
* `--emit=OUTPUT_TYPE[=PATH][,...]` tells the compiler to emit `OUTPUT_TYPE`, and `PATH` behaves like `--output-dir`/`--output-file` for the given output type. If `PATH` is relative, then `PATH` is relative to the working directory of the compiler. If `PATH` is not specified, then it defaults to a compiler-chosen name in `--target-dir` (for intermediates), or the effective `--output-dir`. The compiler-chosen name is based on the name of the item being emitted, otherwise it uses the session name, which is derived from the input file stem (or an explicit `--name`).

So, recapping my previous examples:

```
# Place MAST output in the current directory with file_stem `foo`
midenc compile --name foo out.wasm

# Place MAST output in `mast/out.mast`, overriding the explicit `--name` of `foo` 
# and implicit name of `bar` that would have been used to derive the file stem
midenc compile --name foo -o mast/out.mast bar.wasm

# Place MASM outputs in `target/`, and MAST output in the current directory with file_stem `foo`
midenc compile --emit=masm,mast --target-dir target/ --output-dir ./ foo.wasm

# Place all intermediate outputs in `target`, except MASM outputs are in `$CWD/masm`, and the MAST
# will be written to `mast/out.mast`. Note that `--output-dir` is effectively ignored.
midenc compile --emit=ast,hir,mast,masm=$CWD/masm  --target-dir target/ \
    --output-dir out/ -o mast/out.mast foo.wasm

# Run the compiler just to obtain MASM output, do not assemble MAST, and place it in `masm`
midenc compile --emit=masm --target-dir masm/ foo.wasm
```

I plan to add an `--emit=all` option as well, but for now you have to be explicit. I think it would also be good to have `--output-dir` override `--target-dir` when the only output being emitted is an intermediate output type, like in the last example above, as in that case the last intermediate type to be produced by the compiler sort of _is_ the output for that session.

This might be overly flexible too, I'm sort of accounting for the expectation that we'll have more voluminous output in the future, where controlling where compiler-internal artifacts and user-requested artifacts are placed is desirable, but it may never get to that point. If I was going to get rid of any option here, it would be `--target-dir`, and just use `--output-dir` for that instead; or alternatively, have `--target-dir` be the same as `--output-dir` unless explicitly given. In any case, if you run into anything that is a pain point, let me know.

greenhat: I agree. BTW, your comments here deserve to be in the docs. It'd be a shame to bury them in this PR.

Issue #246, opened 2024-07-15
Test release process by publishing compiler and Miden SDK crates


greenhat:
```[tasklist]
### Tasks (according to CONTRIBUTING.md)
- [x] Run `release-plz update` in the repo root folder to update the compiler crates versions and generate changelogs.
- [x] Run `release-plz update` in the `sdk` folder to update the SDK crates versions and generate changelogs.
- [x] Create a release PR naming the branch with the `release-plz-` suffix (its important to use this suffix to trigger the crate publishing on CI in step 4).
- [x] Review the changes in the release PR, commit edits if needed and merge it into the main branch.
- [x] The CI will automatically run `release-plz release` after the release PR is merged to publish the new versions to crates.io.
```

greenhat:
@bitwalker Well, the first attempt to publish https://github.com/0xPolygonMiden/compiler/pull/249#pullrequestreview-2186336971 is failed:
```
error: failed to publish to registry at https://crates.io/
    
    Caused by:
      the remote server responded with an error (status 400 Bad Request): The following category slugs are not currently supported on crates.io: Compilers
    
      See https://crates.io/category_slugs for a list of supported slugs.
```

I've run the `--dry-run` (I should've done it earlier 🤦)  and besides that, we're going to need to specify the `version` along the `path` for all the deps:
```
  2024-07-19T13:34:37.502097Z ERROR  failed to release package

Caused by:
    failed to publish midenc-session:     Updating crates.io index
    error: all dependencies must have a version specified when publishing.
    dependency `midenc-hir-symbol` does not specify a version
    Note: The published dependency will use the version from crates.io,
    the `path` specification will be removed from the dependency declaration.
```

I'm making a new release PR with these fixes.

greenhat:
The latest publishing attempt - https://github.com/0xPolygonMiden/compiler/actions/runs/10043779504/job/27757299114
Issues:
- `miden-assembly` does not have `version` set;
- git tags are pushed despite the `git_tag_enable=false` in the `release-plz.toml`.

greenhat:
We cannot set the `version` for VM crates in our deps since we depend on the unreleased version of the VM (v.0.10.0) on the following features in particular:
- https://github.com/0xPolygonMiden/miden-vm/pull/1363
- https://github.com/0xPolygonMiden/miden-vm/pull/1360

[...]

PR #272, opened 2024-08-06
wip: alpha release

bitwalker:
This PR supercedes #260, and introduces a `next` branch on which our usual development will take place once merged.

@greenhat This is the same changes as #260, just rebased on `main` to pick up the few PRs of yours we merged already. I'd fork from `next` for all subsequent PRs.

Once this is merged, it's because the requirements for the alpha release have been met (i.e. the memory init issues have been worked out, and CI is green), hence the naming of this PR.

Until merged, stack PRs on top of this one, and we can follow our usual process that way.

PR #276, opened 2024-08-08
Release v0.0.1 (take 4): Initial release of the compiler and SDK crates

greenhat:
Ref #246 

Well, the release PR against the main branch is not going to work since the VM v0.10 has MSRV 1.80 and some breaking API changes. Let's test the publishing from the next branch.

In this PR, I set the release CI job to run on merges to the `next` branch. We'll revert this after publish testing is complete.

After this PR is merged, the release CI job will publish all the unpublished crates.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ name: release-plz
 on:
   push:
     branches:
-      - main
+      - next # TODO: switch to main after testing the publish workflow
 
 env:
   CARGO_MAKE_TOOLCHAIN: nightly-2024-05-07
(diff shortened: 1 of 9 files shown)


bitwalker (approved):

bitwalker:
If I understand correctly, we should merge this once #272 is green? Or should we merge it now, and it will release once we have a successful build on that branch? Feel free to merge this whenever ready if it's the latter.

greenhat:
Since it is testing of the publishing process and not the actual public release, I thought the green test suite is not required, and we can merge it now. On the second thought, maybe publishing with a red test suite is not a good idea.
The CI job should publish the crates once it merged. Since we are not running(BTW, should we?) the tests in the `release` job, green `cargo build` is the only hard requirement.

PR #292, opened 2024-08-30
chore: fix clippy warnings in tests

greenhat:


bitwalker:
LGTM - any idea why CI isn't catching these? I'm assuming we don't clippy with `--tests` or something

bitwalker (approved):

PR #302, opened 2024-09-05
chore: fix mkdocs warnings, move cargo-miden README to docs

greenhat:
In this PR:

- I tried to exclude the `README.md` via the https://github.com/apenwarr/mkdocs-exclude plugin with no luck, so I just changed its extension to `README.txt`
- Move canonabi file to appendices and add it in the nav;
- Create `usage/cargo-miden.md` and move the `/tools/cargo-miden/README.md` content there.

We should probably add a CI job of building the book and check for warnings.


docs/appendix/canonabi-adhocabi-mismatch.md
bitwalker, line 1: I think we should probably move this to a top-level `design` or `internal_docs` directory, due to how deep in the technical weeds this document is, but it's fine for now.

bitwalker (approved):
LGTM!

docs/appendix/canonabi-adhocabi-mismatch.md
greenhat: I agree.  I noticed that it does not fit in the existing structure, but I tried to find an existing folder for it temporarily.

PR #305, opened 2024-09-06
docs: revisit/update documentation

bitwalker:
This PR contains all the updates I made while reviewing the docs, and expanding in various places. I discovered a few small bugs in the process as well.

One thing that was obvious pretty quickly, is that the `miden` CLI needs _a lot_ of love, `miden run` is totally broken at the moment for anything involving MAST. I was unable to test that part of one of the guides due to this. The good news is, those things should be relatively easy to fix, but something we'll need to be aware of.

I think we also need to expand our docs on getting started with the Cargo extension, we have the basics, but it would be nice to have a more robust walkthrough/tutorial type thing. Might need to wait until we have support for accounts/note scripts though, we'll see.

Once this is merged, we should release 0.0.6 (to pick up the fixes contained here), and merge `next` into `main` so that the docs can be published to production.


docs/appendix/known-limitations.md
```diff
@@ -0,0 +1,255 @@
...
+look like in the language, but it is not possible to actually implement either of those in Rust
+today. The reasons for this are covered in depth in the tracking issue linked above, but to
+briefly summarize, the primary issue has to do with the fact that Rust programs are compiled
+for a "shared-everything" environment, i.e. you can pass references to memory from caller to
+callee, write to caller memory from the callee, etc. In Miden however, contexts are "shared-nothing"
+units of isolation, and thus cross-context operations, such as performing a `call` from a note script
+to a method on an account, are not compatible with the usual calling conventions used by Rust and
+LLVM.
+
+The solution to this relies on compiling the Rust code for the `wasm32-wasip2` target, which emits
```
greenhat, line 197: Do you mean `wasm32-wasip1` previously known as `wasm32-wasi`?
```suggestion
The solution to this relies on compiling the Rust code for the `wasm32-wasip1` target, which emits
```

greenhat (approved):
Looking great! Just a typo.

docs/appendix/known-limitations.md
bitwalker: No, I actually meant `wasm32-wasip2`, which is the Rust target that compiles for the Wasm component model. The `wasm32-wasip1` target is the renamed `wasm32-wasi` target (which is now deprecated), and corresponds to the `preview1` WASI support, while `wasm32-wasip2` is `preview2`, which contains the transition to the Component Model.

Issue #308, opened 2024-09-06
Use Miden SDK as a Rust lib instead of Wasm component


greenhat:
```[tasklist]
### Tasks
- [ ] #116
- [ ] #117
```

Issue #90, opened 2024-01-11
Wasm CM support in cargo extension


greenhat:
### New project command 
Should create a Wasm CM Rust project, so the template should be updated:
- add WIT file
- set Wasm CM metadata in Cargo.toml

### Build project command
Start by calling the `cargo_component` to build the project and switch to use it as a library when our needs grow. We probably would like to avoid forking it but rather building on top of it.

**EDIT**(denys): I believe building on top of the `cargo-component` is no longer an option given that we are going to add plenty of customizations touching Cargo.toml (miden packaging), Wasm component building, etc. The only feasible solution I see is to copy the relevant parts of the `cargo-component` into `cargo-miden`.

greenhat:
@bitwalker Do we want the extension to keep the support for building a non-Wasm CM projects as it does today along with Wasm CM? I mean, our whole toolchain will be Wasm CM oriented. Going Wasm CM only for the extension will simplify things a lot, but we would not be able to make a release until almost everything in the Beta1 milestone is implemented.

PR #313, opened 2024-09-10
chore: add `CompilerTestInputType::CargoMiden` and use `cargo-miden` build it

### LINKED ISSUE #312: Use `cargo-miden` to compile Rust code in `CompilerTest`
Before I start working on #308 I'd like to prepare our test infra and add a `CompileTestInputType::CargoMiden` variant that would call `cargo-miden` library to compile Rust code to Wasm.

greenhat:
Close #312

The `CompilerInputTestType::CargoComponent` will be removed, and all relevant tests will be switched to the `CargoMiden` variant after the #90 and #308 are implemented.

### DIFF
--- a/tests/integration/Cargo.toml
+++ b/tests/integration/Cargo.toml
@@ -31,6 +31,7 @@ midenc-codegen-masm.workspace = true
 midenc-session.workspace = true
 midenc-compile.workspace = true
 midenc-debug.workspace = true
+cargo-miden.workspace = true
 wasmprinter = "0.2.80"
 proptest.workspace = true
 sha2 = "0.10"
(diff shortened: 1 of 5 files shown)


tests/integration/Cargo.toml
```diff
@@ -31,6 +31,7 @@ midenc-codegen-masm.workspace = true
 midenc-session.workspace = true
 midenc-compile.workspace = true
 midenc-debug.workspace = true
+cargo-miden = { path = "../../tools/cargo-miden" }
```
bitwalker, line 34: *Suggestion:* We may as well add `cargo-miden` to the workspace `Cargo.toml`, since it's going to get picked up as a dependency at the workspace level anyway due to its presence here.

bitwalker (approved):
LGTM - just one suggestion, but a small one

tests/integration/Cargo.toml
greenhat: Done.

PR #315, opened 2024-09-10
Docs/copy edit 1

kmurphypolygon:
- This PR does a quick scan and update for style guide.
- Please note only one # (H1) per page.
- Also added back the requirements.txt and run script for easy spin up of docs.
- Fixed a link.


requirements.txt
bobbinth, line 1: Quick comment: we usually put this file as well as `run.sh` into `scripts` directory and name them `serve-doc-site.sh` and `docs_requirements.txt` (e.g., see [here](https://github.com/0xPolygonMiden/miden-base/tree/main/scripts).

kmurphypolygon: Hi there, I noticed you didn't have them in this repo, so I just dropped in what we normally use in the KL repo and others without rewriting anything. Ideally, we should have a common process across all doc sites so as not to confuse tech writers but feel free to customize this as you wish and if you do customize please add some instructions on how you prefer to spin up the site :)

bitwalker: FYI: These files are already present in the `docs` directory, and built using `cargo make book` or `cargo make serve-book` depending on whether you want to run it live or not.

docs/appendix/known-limitations.md
```diff
@@ -124,7 +123,7 @@ will be expanding the SDK rapidly over the next few weeks and months, but for th
 you encounter a missing API that you need, let us know, so we can ensure it is prioritized above
 APIs which are lesser used.
 
-### Rust/Miden FFI (Foreign Function Interface) and Interop
+### Rust/Miden FFI (foreign function interface) and interop
```
bitwalker, line 126: Shouldn't "foreign function interface" remain capitalized here, to emphasize the correlation to the initials of the "FFI" acronym?

.gitignore
```diff
@@ -3,3 +3,15 @@ bin/midenc
 bin/litcheck
 .crates.toml
 .crates2.json
+
+# Docs platform ignores
```
bitwalker, line 7: *Note:* Our docs build in the `target` directory, which is already ignored, and our `docs` directory has a dedicated `.gitignore` for things within that directory, so most of these directives will never be matched.

That said, I'm fine with adding these here anyway (mostly the IDE-specific directives, since we should have those anyway), just in case someone accidentally builds the docs in the project root locally - but just wanted to clarify where you can expect these things to actually be output in this repo.

bitwalker (changes requested):
LGTM! However, I'd like to remove the `run.sh` and `requirements.txt` in favor of the ones that already exist in the repo under the `docs` directory before merging. I can make that change, as well as make sure our `docs/README.txt` explains how to build and run the documentation using either of the two methods I mentioned in my other comment - unless you'd prefer to do so, just let me know!

requirements.txt
kmurphypolygon: ok, cool ... that's great, please go ahead and remove the files I dropped in, no problem

docs/appendix/known-limitations.md
kmurphypolygon: Not usually: https://en.wikipedia.org/wiki/Foreign_function_interface

.gitignore
kmurphypolygon: Yes, whenever I spin it up locally and make changes, I'm prompted to push a whole load of unnecessary stuff ... so that's why I added these like we have in the Miden base repo.

kmurphypolygon:
Just a quick note please @bitwalker .. we have an error on main docs 

```The following pages exist in the docs directory, but are not included in the "nav" configuration:
             - miden/miden-compiler/appendix/known-limitations.md```

This file will remain live and searchable even if there is no menu link to it ... so I'm not sure it that's what you need? Or perhaps you need to add a menu link?

bitwalker (approved):

bitwalker:
@kmurphypolygon I've added the missing nav item for that page, build instructions for the docs, and the tweaks I suggested, in the latest commit. I'm going to go ahead and merge this, but if you spot any other issues or have other suggested changes, let me know!

PR #324, opened 2024-09-16
doc: add `--release` option to `cargo miden build` command

greenhat:


bitwalker (approved):
Not sure how to handle deploying just docs fixes, which have to be merged to `main` rather than `next`, but I was kinda hoping `main` would reflect the most recent stable release.

I guess for now, let's merge this to `next`, but we should come up with a process for managing these changes separate from the rest of our development work. Probably the way to go is to just patch `main` and then merge `main` into `next`, since `main` will always have a common base with `next`

PR #326, opened 2024-09-17
chore: remove `miden-assembly` dependency from `sdk/base-sys` for `bindings` feature

greenhat:
Use `miden-assembly` dependency only when building MASL library.


bitwalker (approved):
Nice, that might speed up compilation a bit

PR #329, opened 2024-09-19
[1/x] feature(cargo-miden): support building Wasm component from a Cargo project

### LINKED ISSUE #90: Wasm CM support in cargo extension

### LINKED ISSUE #117: Type/interface remapping support in `cargo-component`

### LINKED ISSUE #308: Use Miden SDK as a Rust lib instead of Wasm component

### LINKED ISSUE #334: Failed to build a Wasm component from the Wasm core module with "non-component" imports
**TLDR;** Using core Wasm module imports for our intrinsics and Miden SDK functions is not playing well with the Wasm component model.

### What

When building a Wasm component from a Wasm core module compiled from the `basic-wallet` with Rust SDK example, the build fails with the following error:
```
Caused by:
    0: module was not valid
    1: module requires an import interface named `miden::account`
```

### Why

The reason for this error is that the Wasm core module contains imports that `wit-component`(used by `cargo-component` to build the Wasm component) fails to recognize as component imports, i.e. imports that are defined in the WIT file. In this case, it's `miden::account::add_asset` tx kernel function from the Miden SDK. Effectively, all our intrinsics and Miden SDK functions are triggering this error.

This happens at https://github.com/bytecodealliance/wasm-tools/blob/51ade657f6d6874b1bdfb87a47299dbe7e61d5c2/crates/wit-component/src/validation.rs#L411-L491, where the module name of the core module import is expected to be found in the interface imports in the WIT file.

### How to reproduce
In https://github.com/0xPolygonMiden/compiler/pull/329 branch run:
```bash
RUST_LOG=debug cargo test rust_sdk_basic
```
To check out the Wasm core module, use `wasm-tools print PATH_TO_CORE_WASM` with the path reported in logs at `componentizing WebAssembly module ...`

### How to fix

The most promising approach IMO would be to not use Wasm core module imports for our intrinsics and Miden SDK functions. The reason why `wit-component` treats them as errors is fair, and I think it'd be difficult if at all possible to alter that behavior in the upstream.

So instead of defining our Miden SDK functions and intrinsics as `extern` https://github.com/0xPolygonMiden/compiler/blob/a876d8b134bc118e1944b5a72747cf2c13a235a1/sdk/base-sys/src/bindings/tx/externs.rs#L5-L13
we could define them as functions with `unreachable!()` bodies and recognize them by function names and signatures in the frontend:
```rust
#[inline(never)]
#[no_mangle]
fn miden_account_add_asset(_: Felt, _: Felt, _: Felt, _: Felt, ptr: *mut CoreAsset) {
    unreachable!()
}
```

Although, `#[inline(never)]` is only a hint to the compiler, with a side effect in the body we should be safe from inlining.

greenhat:
Close #90 
Close #117 
Close #308 
Close #334 

This PR adds support to the `cargo-miden` to use `cargo-component` library to build Wasm component Rust projects. The existing `cargo-miden` ability to build core Wasm module Rust projects is preserved.

The Miden SDK and intrinsics are pulled through the Wasm CM as an imported interface (see `miden-core-import.wit`). The Miden SDK types are defined in WIT (see `miden.wit`) and remapped to the `miden-sdk` crate types at `cargo-component` library invocation in `cargo-miden` along with the skip list for function bindings generation at
https://github.com/0xPolygonMiden/compiler/blob/fe28b1cffef6a15b02f8e86fadce9c14263281cb/tools/cargo-miden/src/lib.rs#L175-L203 

Then, when translating Wasm component the original(linkable) MASM module and function names are recovered at the core Wasm module and the component level  by https://github.com/0xPolygonMiden/compiler/blob/fe28b1cffef6a15b02f8e86fadce9c14263281cb/frontend-wasm/src/miden_abi/mod.rs#L61-L111

See `test/rust-apps-wasm/rust-sdk/basic-wallet/` test project for an example and `rust_sdk_basic_wallet()` test.

To not grow this PR any bigger, #341 and #342 are created to fix the introduced quick-and-dirty ways for function generation skipping and MASM module and function names recovery.


bitwalker:
Just wanted to follow up and let you know I've started reviewing, but haven't finished just yet, I'll try to get my review wrapped up today (Tuesday).

greenhat:
I forgot to address the provisioning of the WIT interface for the Miden SDK. I made https://github.com/0xPolygonMiden/compiler/issues/343 for it.

bitwalker (approved):

PR #332, opened 2024-09-23
wip: hir refactoring

bitwalker:
TBD


docs/design/ir.md
```diff
@@ -0,0 +1,839 @@
...
+a variable in the source language is assigned conditionally, then in the IR, there will be a block
+with a parameter corresponding to the value of that variable after it is assigned. All uses after
+that point, would refer to that block parameter, rather than the value from a specific branch.
+Similarly, loop-carried variables, such as an iteration count, are typically manifested as block
+parameters of the block corresponding to the loop header.
+2. Operation results. The most common way in which values are introduced.
+
+Values have _uses_ corresponding to operands or successor arguments (special operands which are used
+to satisfy successor block parameters). As a result, values also have _users_, corresponding to the
+specific operation and operand forming a _use.
```
greenhat, line 255:
```suggestion
specific operation and operand forming a _use_.
```

hir2/src/patterns/applicator.rs
```diff
@@ -0,0 +1,179 @@
...
+            for pattern in patterns {
+                callback(Rc::clone(pattern));
+            }
+        }
+        for pattern in self.rewrite_patterns_set.any_op_patterns() {
+            callback(Rc::clone(pattern));
+        }
+    }
+
+    pub fn match_and_rewrite<A, F, S>(
```
greenhat, line 77: This looks astonishing! Looking forward to `RewritePattern` impl example.

hir2/src/ir/traits.rs
```diff
@@ -0,0 +1,209 @@
...
+                        .with_primary_label(op.span(), format!("incorrect number of operands, expected 2, got {}", op.num_operands()))
+                        .with_help("this operator implements 'BinaryOp', which requires it to have exactly two operands")
+                        .into_report()
+                )
+            }
+        }
+    }
+}
+
+derive! {
```
greenhat, line 133: The verification subsystem is fascinating! Great job! There is a lot of macros and type-level magic going on here. I must admit that after reading all the extensive documentation, I still don't fully understand how it works. I'm fine with this level of complexity, but looking at the `derive!` macro usage, I'm not sure that macro's benefits outweigh the complexity. I mean this `derive!` call expands to the following code:

```rust
#[doc = r" Op's regions have no arguments"]
pub trait NoRegionArguments {}
impl<T: crate::Op + NoRegionArguments> crate::Verify<dyn NoRegionArguments> for T {
    #[inline]
    fn verify(&self, context: &crate::Context) -> Result<(), crate::Report> {
        <crate::Operation as crate::Verify<dyn NoRegionArguments>>::verify(
            self.as_operation(),
            context,
        )
    }
}
impl crate::Verify<dyn NoRegionArguments> for crate::Operation {
    fn should_verify(&self, _context: &crate::Context) -> bool {
        self.implements::<dyn NoRegionArguments>()
    }

    fn verify(&self, context: &crate::Context) -> Result<(), crate::Report> {
        #[inline]
        fn no_region_arguments(op: &Operation, context: &Context) -> Result<(), Report> {
            for region in op.regions().iter() {
                if region.entry().has_arguments() {
                    return Err(context
                        .session
                        .diagnostics
                        .diagnostic(Severity::Error)
                        .with_message("invalid operation")
                        .with_primary_label(
                            op.span(),
                            "this operation does not permit regions with arguments, but one was \
                             found",
                        )
                        .into_report());
                }
            }
            Ok(())
        }
        no_region_arguments(self, context)?;
        Ok(())
    }
}
```

Which is not that larger than the macro call itself. If part of it can be generated by an attribute macros it's not that hard to write the rest of it by hand (`Verify::verify` only?). I mean, the benefits of `derive!` macro might not outweigh the drawbacks (remembering the syntax, writing code inside a macro call, etc.).

hir2/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+#![feature(debug_closure_helpers)]
+#![feature(trait_alias)]
+#![feature(is_none_or)]
+#![feature(try_trait_v2)]
+#![feature(try_trait_v2_residual)]
+#![feature(tuple_trait)]
+#![feature(fn_traits)]
+#![feature(unboxed_closures)]
+#![feature(const_type_id)]
+#![feature(exact_size_is_empty)]
```
greenhat, line 19: There goes my hope to someday ditch the nightly. :)

hir-macros/src/lib.rs
```diff
@@ -25,6 +26,71 @@ pub fn derive_spanned(input: proc_macro::TokenStream) -> proc_macro::TokenStream
...
+/// pub struct If {
+///     #[operand]
+///     condition: Bool,
+///     #[region]
+///     then_region: RegionRef,
+///     #[region]
+///     else_region: RegionRef,
+/// }
+#[proc_macro_attribute]
+pub fn operation(
```
greenhat, line 62: I really dig the `operation` proc macro attribute. I expanded it in a few ops. So much pretty complex boilerplate is generated. Awesome!

hir2/src/ir/operation/name.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub(super) fn downcast_ref<T: Op>(&self, ptr: *const ()) -> Option<&T> {
+        if self.is::<T>() {
+            Some(unsafe { self.downcast_ref_unchecked(ptr) })
+        } else {
+            None
+        }
+    }
+
+    #[inline(always)]
+    unsafe fn downcast_ref_unchecked<T: Op>(&self, ptr: *const ()) -> &T {
```
greenhat, line 91: I'm a fan of op traits implementation! TIL about `DynMetadata`. So cool!

greenhat (approved):
It looks super cool! I'm looking forward to using it!

hir2/src/patterns/applicator.rs
bitwalker: I've gotten virtually of the rewrite infra built now, but the last kink I'm working out is how to approach the way the `Listener` classes are used in MLIR. In effect, the way it is used in MLIR relies on mutable aliases, which is obviously a no go in Rust, and because of that (and I suspect somewhat as an artifact of the design they chose), the class hierarchy and interactions between builders/rewriters/listeners is _very_ confusing.

I've untangled the nest a bit in terms of how it actually works in MLIR, but its a non-starter in Rust, so a different design is required, but one that ideally achieves similar goals. The key thing listeners enable in MLIR, is the ability for say, the pattern rewrite driver, to be notified whenever a rewrite modifies the IR in some way, and as a result, modify its own state to take those modifications into account. There are other things as well, such as listeners which perform expensive debugging checks without those checks needing to be implemented by each builder/rewriter, but the main thing I care about is the rewrite driver.

There are a couple of approaches that I think should work, but need to test them out first. In any case, that's the main reason things have been a bit slow in the last week.

hir2/src/ir/traits.rs
bitwalker: Yeah I mostly didn't want any of the type-level magic (and boilerplate) to be hand-written as things were evolving - modifying the macro and updating all of the traits at once was far easier, and is really the only reason why the `derive!` macro still exists. However, as you've pointed out, it would be best to define an attribute macro for this as well, just like `#[operation]`, but I've punted on that since the `derive!` macro gets the job done for now, and not many new op traits will be getting defined in the near term, but its certainly on the TODO list.

hir2/src/lib.rs
bitwalker: Yeah I've really blown a hole in that plan 😂. Once you need some of the nigh perma-unstable features like `ptr_metadata`, `fn_traits`, or `specialization`/`min_specialization`, it's basically game over for using stable, unless you have alternative options that allow you to sidestep the use of those features.

That said, a fair number of these are pretty close to stabilization. The ones that are problematic are:

* `specialization` and `rustc_attrs` are needed because `min_specialization` isn't _quite_ sufficient for the type-level magic I'm doing for automatically deriving verifiers for operations. If improvements are made to `min_specialization`, we can at least switch over to that. Longer term, we'd have to come up with an alternative approach to verification if we wanted to switch to stable though.
* `ptr_metadata` has an unclear stabilization path, and while it is basically table stakes for implementing custom smart pointer types, stabilizing it will require Rust to commit to some implementation details they are still unsure they want to commit to.
* `allocator_api` and `alloc_layout_extra` should've been stabilized ages ago IMO, but they are still being fiddled with. However, we aren't really leaning on these very much, and may actually be able to remove them.
* `coerce_unsized` and `unsize` are unlikely to be stabilized any time soon, as they've been a source of unsoundness in the language (not due to the features, but due to the traits themselves, which in stable are automatically derived). Unfortunately, they are also table stakes in implementing smart pointer types with ergonomics like the ones in libstd/liballoc, so its hard to avoid them.
* `try_trait_v2` and `try_trait_v2_residual` are purely ergonomic, so we could remove them in exchange for more verbose code in a couple places, not a big deal
* `fn_traits`, `unboxed_closures`, and `tuple_trait` are all intertwined with the ability to implement implementations of `Fn`/`FnMut`/`FnOnce` by hand, which I'm currently using to make the ergonomics of op builders as pleasant as possible. We might be able to tackle ergonomics in another way though, and as a result, no longer require these three features.

So, long story short, most of these provide some really significant ergonomics benefits - but if we _really_ wanted to ditch nightly for some reason, we could sacrifice some of those benefits and likely escape the need for most of these features. Verification is the main one that is not so easily abandoned, as the alternative is a lot of verbose and fragile boilerplate for verifying each individual operation - that's not to say it can't be done, just that I'm not sure the tradeoff is worth it.

hir-macros/src/lib.rs
bitwalker: I don't know if you say the `derive!` macro _before_ I implemented the `#[operation]` attribute macro, but...it was madness 😅.

Now it is much cleaner (and more maintainable!), but it does a lot under the covers, so the main tradeoff is lack of visibility into all of that. That said, I tried to keep the macro focused on generating boilerplate for things that feel pretty intuitive, e.g. deriving traits, op construction and verification, correspondence between fields of the struct definition and the methods that get generated. So hopefully we never need to actually look at the generated code except in rare circumstances.

hir2/src/ir/operation/name.rs
bitwalker: I'm definitely happy with how that has turned out so far, it enables a number of really neat tricks! You might also be interested in a bit of trivia about pointer metadata in Rust:

* `DynMetadata<dyn Trait>` is the type of metadata associated with a trait object for the trait `Trait`, and is basically just a newtype for a pointer to the vtable of the trait object. While you could specify `DynMetadata<T>` for some `T` that _isn't_ a trait, that doesn't actually mean anything, and is technically invalid AFAIK.
* `usize` is  the type of metadata associated with "unsized" types (except trait objects), and corresponds to the unit size for the pointee type. For example, `&[T]` is an unsized reference, and its pointer metadata is the number of `T` in the slice, i.e. a typical "fat" pointer; but the semantics of the metadata value depend on the type, so you can have custom types that use `usize` metadata to indicate the allocation size of the pointee, or something else along those lines.
* `()` is the type of metadata associated with all "sized" types, i.e. pointers to those types don't actually have metadata

bitwalker:
Thanks for reviewing things, and for the kind comments! As mentioned, things are quite close to finished now, the last thing that needs some fiddling is finding a Rust-friendly design for handling the interaction between patterns, builders/rewriters, and the rewrite driver (i.e. so that the rewrite driver is made aware of changes to the IR made by a rewrite pattern via the builder/rewriter).

Once that's out of the way, there is a short list of tasks to switch over to this new IR:

* Port dominator tree and loop tree analyses on top of the new IR
* Port data flow analysis framework on top of the new IR
* Port liveness analysis and spills transformation using the data flow analysis framework
* Wire up lowering from the new IR to MASM
* Wire up lowering from Wasm to the new IR

hir2/src/lib.rs
greenhat: I can live with it. The verification is totally worth it!

hir2/src/ir/operation/name.rs
greenhat: Thank you! I'm definitely going to dig into it.

hir2/src/patterns/pattern.rs
```diff
@@ -0,0 +1,382 @@
...
+            let constant_builder = rewriter.create::<Constant, _>(span);
+            let constant: UnsafeIntrusiveEntityRef<Constant> =
+                constant_builder(Immediate::U32(2)).unwrap();
+            let shift = constant.borrow().result().as_value_ref();
+            let mul_builder = rewriter.create::<Mul, _>(span);
+            let mul = mul_builder(lhs, shift, Overflow::Wrapping).unwrap();
+            let mul = mul.borrow().as_operation().as_operation_ref();
+            log::trace!("replacing shl with mul");
+            rewriter.replace_op(op, mul);
+        }
```
greenhat, line 322: Ergonomics look great! The code gaves strong MLIR vibes! It'd take some time to getting use to the types, techniques, etc. but it worth it. I'm looking forward to start writing some rewrites.

greenhat (approved):
Looking great!

The component and module interfaces and instances feels generic and flexible enough to be able to represent Wasm CM.
Let's consider the following WIT:

```wit
package miden:basic-wallet@1.0.0;

use miden:base/core-types@1.0.0;

interface basic-wallet {
    use core-types.{core-asset, tag, recipient, note-type};
    receive-asset: func(core-asset: core-asset);
    send-asset: func(core-asset: core-asset, tag: tag, note-type: note-type, recipient: recipient);
}

interface aux {
    use core-types.{felt};
    process-list-felt: func(input: list<felt>) -> list<felt>;
}

interface foo {
    get-foo(a: list<felt>) -> list<felt>;
}

interface bar {
    get-bar(a: list<felt>) -> list<felt>;
}


world basic-wallet-world {
    include miden:core-import/all@1.0.0;
    
    import foo;
    import bar;

    export basic-wallet;
    export aux;
}
```

The component (high-level) imports (`foo` and `bar`) would be represented as module interfaces (one per component imported interface?) and the lowerings we will generate by creating a module that imports this module interface and exports the generated lowering functions with low-level types signatures which in turn be imports for the core Wasm module.
The component (high-level) exports (`basic-wallet` and `aux`) would be represented as module interfaces (one per component exported interface?) and the liftings we will generate by creating a module that imports the core Wasm module exports and exports the generated lifting functions with high-level types signatures.
We're going to need to give those generated liftings/lowerings modules access to the core Wasm module memory (Wasm memory

hir2/src/patterns/driver.rs
```diff
@@ -0,0 +1,1041 @@
...
+/// Returns `Ok(changed)` if the iterative process converged (i.e., fixpoint was reached) and no
+/// more patterns can be matched within the region. The `changed` flag is set to `true` if the IR
+/// was modified at all.
+///
+/// NOTE: This function does not apply patterns to the region's parent operation.
+pub fn apply_patterns_and_fold_region_greedily(
+    region: RegionRef,
+    patterns: Rc<FrozenRewritePatternSet>,
+    mut config: GreedyRewriteConfig,
+) -> Result<bool, bool> {
```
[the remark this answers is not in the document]
bitwalker: A dedicated type might be worth it, basically we want to signal whether or not convergence occurred, and whether it did or not, we also want to know whether the IR was changed.

hir2/src/dialects/hir/ops/component/interface.rs
```diff
@@ -0,0 +1,341 @@
...
+/// The fully-qualfied identifier of a component
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ComponentId {
+    /// The namespace in which the component is defined
+    pub namespace: SymbolName,
+    /// The name of this component
+    pub name: SymbolName,
+    /// The semantic version number of this component
+    pub version: Version,
+}
```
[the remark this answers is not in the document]
bitwalker: So I think `package` is `name` here, `namespace` is a given, as is `version`, while `interface` would be derived from the module name exported from a component. So on its own, a `ComponentId` doesn't represent all possible parts of a Wasm CM identifier, albeit still useful for identifying the component part.

hir2/src/patterns/driver.rs
```diff
@@ -0,0 +1,1041 @@
...
+        self
+    }
+
+    #[inline]
+    pub fn scope(&self) -> Option<RegionRef> {
+        self.scope.clone()
+    }
+
+    #[inline]
+    pub fn max_iterations(&self) -> Option<core::num::NonZeroU32> {
```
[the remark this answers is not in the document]
bitwalker: Definitely comes in handy! Another big benefit of using them are the niche optimizations you get when combined with `Option`, i.e. `Option<NonZeroU32>` has the same representation as a raw `u32`.

greenhat (approved):

Issue #346, opened 2024-10-21
Link the library from an account's Miden package when compiling the note script


greenhat:
I suggest passing the library from the Miden package as `LinkedLibrary`(via `--linked-lib`) to the midenc.

This would require to add `LinkedLibrary::Masp` variant that would parse the `Package` from the `masp` file and extract `CompiledLibrary` out of it.
Since `LinkLibrary` is in the `midenc-session` crate and `Package` is in `midenc-codegen-masm` crate, there is no way of parsing the `Package` in `LinkLibrary` impl. So we would need to re-arrange `Package` and/or `LinkLibrary`. `Package` seems to be destined to be in its own `midenc-package` crate. So I'd start there. Given that `PackageManifest::link_libraries` uses `LinkLibrary` they might end up in the same crate. 

EDIT: Following our discussion, the `miden-package` crate with `Package` type should be created in the vm repo.
```[tasklist]
### Tasks
- [ ] Extract `Package` into `miden-package` crate in the VM repo
```

greenhat:
This issue should be tackled after #347 is fixed.

PR #348, opened 2024-10-22
[3/x] Add note script compilation test

greenhat:
Ref #314 

**This PR is stacked on the #344 and should be merged after it**

This PR adds the P2ID note script compilation up to IR. The MASM compilation is blocked by #346.

This PR is also removing `ImportMetadata` (function digest) since we get digests from the linked libraries.


sdk/base-sys/src/bindings/types.rs
```diff
@@ -4,6 +4,12 @@ use miden_stdlib_sys::{Felt, Word};
 #[derive(Copy, Clone)]
 pub struct AccountId(Felt);
 
+impl AccountId {
+    pub fn as_felt(&self) -> Felt {
```
bitwalker, line 8: *Nit:*: Since `AccountId` is `Copy`, writing conversions like this as instead:

```rust
#[inline(always)]
pub const fn as_felt(self) -> Felt {
    self.0
```

will basically compile down to a no-op (possibly even without a copy in some cases, if the `AccountId` is not behind a reference/has only a single use).

Not something you need to fix, but could improve codegen in some cases, so mostly just worth keeping in the back of your mind.

Note - the use of `#[inline(always)]` here is because we know that doing so is _always_ strictly better than leaving it unlined. That's not always the case for small functions, but for accessors like this, it is pretty much always worth it IMO.

tests/rust-apps-wasm/rust-sdk/wit-sdk/miden.wit
```diff
@@ -138,6 +138,13 @@ interface core-types {
 
 }
 
+
+/// Note script interface that is expected to be implemented by the note script.
+interface note-script {
+    note-script: func();
```
bitwalker, line 144: Is this what we want to call the user-facing note script entrypoint? I would consider using `main`, so that writing note scripts feels mostly like writing an executable, or at least from a user perspective it is more apparent that `main` is being called by the system (Miden runtime), rather than from their own code. Not sure if `rustc` complains about that without `#![no_main]`, but  I think that would make note scripts feel a bit more intuitive.

Is there a convention for how these are named in Miden currently? We could just go with whatever that is too, but looking for ways to make things feel more intuitive/natural for users (to the extent we can).

tools/cargo-miden/src/lib.rs
```diff
@@ -240,6 +240,7 @@ where
                 )
                 .await
             })?;
+            dbg!(&wasm_outputs);
```
bitwalker, line 243: *Nit:* probably didn't mean to leave this in here

bitwalker (approved):

tests/rust-apps-wasm/rust-sdk/wit-sdk/miden.wit
greenhat: > Is this what we want to call the user-facing note script entrypoint? 

Yes. The names are drafts and are not fixed.

> I would consider using `main`, so that writing note scripts feels mostly like writing an executable, or at least from a user perspective it is more apparent that `main` is being called by the system (Miden runtime), rather than from their own code. Not sure if `rustc` complains about that without `#![no_main]`, but I think that would make note scripts feel a bit more intuitive.

I've tried to build the Wasm command component (`main` entrypoint) before, but the amount of `wasi:*` code it brought in was quite high. 
I revisited this approach, and it appears that all the `wasi:*` code could be removed by crafting a custom adapter. I made an [#350](https://github.com/0xPolygonMiden/compiler/issues/350) issue for this, but it would take some time. So I suggest we merge this PR as is and take care of it later.

> Is there a convention for how these are named in Miden currently? 

No. We can come up with one.

sdk/base-sys/src/bindings/types.rs
greenhat: Cool! `no-op`, ha! Thank you.

tools/cargo-miden/src/lib.rs
greenhat: Thanks for spotting this!

Issue #303, opened 2024-09-05
Cross-Context Procedure Invocation


bitwalker:
This is a tracking issue for the implementation of the `call` instruction, and more importantly, support for cross-context calls in Rust, e.g. invoking account methods from a note script.

## Background

Supporting this in Rust, properly, requires us to build on top of the [WebAssembly Component Model](https://component-model.bytecodealliance.org). This is because fundamentally, Rust does not have any notion of address space isolation, or more generally, a way to enforce shared-nothing boundaries. The Component Model requires cross-component communication to be done using a specific ABI, whose semantics do not permit sharing memory between caller and callee. This is not only dictated by the ABI, but is actually enforced by the WebAssembly runtime itself, so it is not possible to smuggle pointers across the boundary and access memory that way.

All of this is represented in Rust using three "tricks":

1. A Rust crate corresponds to a single component, thus all code in that crate (including code referenced from normal Cargo dependencies) is executed within the same "shared-everything" boundary.
2. Defining a component, requires expressing the interface(s) of that component using WIT (_WebAssembly Interface Types_). WIT only supports types which are valid in the Component Model ABI. Thus the functions that form the component boundary all, by construction, adhere to the Component Model ABI, enforcing the shared-nothing nature of the boundary.
3. In order to call into another component (from Rust), you need to add a new kind of dependency (component) to your project. Unlike normal Cargo dependencies, none of the code in the dependency is directly linked against. Instead, bindings are generated from the WIT description of the component, which not only handle the mechanical details of lifting/lowering the Rust ABI to the Component Model ABI, but also declare a link-time dependency on the actual implementation of those functions. At runtime, all components are instantiated and linked together. This prevents accidental inlining of code from another component written in Rust, into your own, and thus potentially breaking the assumptions of the caller and callee.

In addition to guaranteeing the shared-nothing semantics we want (corresponding to the shared-nothing semantics of Miden contexts), this also has the effect of enabling components to be implemented in any language, with consumers of those components not needing to be aware of what language was used to implement it. This is particularly desirable in the context of Miden, where we do not want to bake in Rust-specific details into accounts, note scripts, and more.

## Requirements

While we do not intend to support the full breadth of what is possible with the Component Model in the near term, the key requirements for us, in order to implement support for `call`, are:

- The ability to parse Wasm Component Model artifacts
- The ability to recognize that a function call crosses a component boundary, which we will use to distinguish normal procedure calls (`exec`) from cross-context procedure calls (`call`).
[...]

PR #353, opened 2024-11-14
[3/x] Link Miden packages without cross-context calls lifting/lowering

### LINKED ISSUE #346: Link the library from an account's Miden package when compiling the note script

### LINKED ISSUE #351: Wasm frontend makes all function in a module public (exported)
Example (basic_wallet) - https://github.com/0xPolygonMiden/compiler/blob/6cd29e17b34c5abef7f6328c33af06f8bf203344/tests/integration/expected/rust_sdk/basic_wallet.hir

Only the Wasm component exports (functions) should be exported.

greenhat:
Close #351 
Close #346 

Ref #303 

**This PR is stacked on the #356  and should be merged after it**

This PR links P2ID note script with the basic wallet Miden package, ignoring the lifting/lowering of the cross-context calls. The cross-context calls will be addressed in the subsequent PR(s).
Since this PR contains quite a few changes in various parts of the compiler, I suggest going by commits for the review process.

This PR makes the following decisions:
- Skip registering already registered modules with the assembler to avoid duplicate registration of the `intrinsics`, etc. modules;
- Recover Wasm CM interfaces as module names from the function names after assembling the library from (workaround for multiple interfaces in one Wasm core module and assembler using module name when building library exports); In the subsequent PRs `Component` will be used in compiler pipeline instead of `Module` and lifting/lowering will be done in separate modules so most likely this workaround will not be necessary
- In library exports, `#anon` namespace is used with Wasm CM interface (`miden:basic-wallet/basic-wallet@1.0.0`) resulting in `#anon::miden:basic-wallet/basic-wallet@1.0.0` string as the export module name;
- Populate `Linker::allow_missing` with linked library's exports (with removed `#anon` namespace);


bitwalker:
@greenhat Yes, high-level type signatures in Miden are something I'll be adding starting in a week or possibly at the start of the new year. Once added, they'll exist at the same level of abstraction as core Wasm signatures (albeit more useful, since not everything will be reduced to `i32`), and we'll rely on WIT + component types to tell us the high-level type signatures in the Canonical ABI for any component exports authored in Miden Assembly.

A couple notes about some of the decisions made here:

* Rather than using the `Anon` namespace, we should translate symbol names as described in the [Component Translation](https://github.com/0xPolygonMiden/compiler/discussions/360) discussion. The gist is that the WIT package name becomes the `LibraryNamespace` (specifically the `User` variant), sans version, items exported from the component use the export name as the first non-namespace path component, and so on (i.e. functions exported from an exported interface are additional components). Future changes to `LibraryPath` will make the syntax for fully-qualified paths a bit more like WIT identifiers, but for now the main thing we want to do is translate them in a way that retains useful semantics in Miden Assembly. We've largely already discussed this, but I'm just reiterating it here since it is relevant to decisions documented in this PR
* Likewise, the way in which module/function names are recovered I believe will need to be changed, but we can do that when merging the new IR into the frontend. This is also described in the Component Translation discussion, but essentially we can recover the original WIT interfaces (and their names/exports) using the component instances exported from the top-level component
* Once the new IR changes are merged, the `Linker` will no longer require us to explicitly allow "missing" symbols - so long as those symbols are accounted for via a dependency. The exception will be intrinsics of course, but those will be special-cased anyway. That doesn't affect anything here, just noting it for posterity.

I'm going to review/merge this under the assumption that we'll address any of the above in future PRs, but wanted to bring them up to ensure they are something we don't forget to deal with once the IR refactoring is merged (or as part of merging it).

codegen/masm/src/masm/program.rs
```diff
@@ -499,8 +516,54 @@ impl Library {
             .iter()
             .map(|rodata| (rodata.digest, rodata.to_elements()))
             .collect();
-        Ok(Arc::new(lib.with_advice_map(advice_map)))
+        let mut mast_forest = lib.mast_forest().as_ref().clone();
+        mast_forest.advice_map_mut().extend(advice_map);
+        let converted_exports = recover_wasm_cm_interfaces(&lib);
+        let lib = CompiledLibrary::new(Arc::new(mast_forest), converted_exports)?;
```
bitwalker, line 522: Note: When wiring up HIR2 to the frontend, the lowering from Wasm to HIR will need to create `Interface` operations in the top-level `Component` which contain `Function` ops corresponding to the exports of component instances in the original Wasm component. The `Component` operation will also contain component-level function exports and `Module`s corresponding to the core Wasm modules present in the original Wasm component. See [Component Translation](https://github.com/0xPolygonMiden/compiler/discussions/360) for further details.

The new codegen crate no longer has its own Miden Assembly IR, so the `Library` and `Program` structs no longer exist. To the extent that some of this stuff still exists (by which I mean things like assembling to MAST, late-stage linking and the like), it will either be done in the _new_ `Linker` (which is much simpler than the current one, and runs just prior to codegen, rather than as a initial part of lowering into HIR), or in the compiler driver (i.e. in an appropriate `Stage` of the compiler). I expect that a lot of this will no longer be necessary though, since the new representation is more faithful to the Component Model, so fewer contortions are required to lower into Miden Assembly than we currently are forced to do.

bitwalker (approved):

PR #356, opened 2024-11-26
[2/x] chore: remove pure `cargo-component`-based tests and artifacts, hand-written Wasm CM tests

greenhat:
**This PR is stacked on the #349 and should be merged after it**

We only care about `cargo-miden`s flavor of Wasm CM.


bitwalker (approved):
I think it's fine to remove these for now, but I think it is worth noting that we don't _only_ care about Wasm components as produced by `cargo-miden` - technically, other frontends can produce Wasm components that we can compile as well. `cargo-miden` outputs will necessarily be tailored to our first-party Rust frontend, so we should not make too many assumptions in the rest of the compiler based on idiosyncracies of Rust or `cargo-miden` specifically.

That said, we don't have other frontends at the moment, and it isn't clear what we would need the removed functionality for at this stage, so removing it makes sense to me.

PR #357, opened 2024-11-26
[5/x] Lifting/Lowering for cross-context calls

greenhat:
**This PR is stacked on the #358 and should be merged after it**

Ref #303 

This PR switches the compiler pipeline from `Module` to `Component` and adds two stages `LowerImportsCrossCtxStage` and `LiftExportsCrossCtxStage` for lifting and lowering components exports/imports for Miden cross-context `call`s. The implementation so far covers only scalar values as arguments and results, with heap-allocated values support coming later. The rodata is not handled in this PR.

Also, the following changes:
- Introduce `CallConv::CrossCtx` for Miden cross-context calls ABI;
- Add `rust_sdk_cross_ctx_account()` and `rust_sdk_cross_ctx_note()` tests with new Rust cargo-miden projects;

The example of the lifting/lowering MASM code generated can be observed in `cross_ctx_account.masm` and `cross_ctx_note.masm` files.


hir/src/function.rs
```diff
@@ -62,6 +62,13 @@ pub enum CallConv {
     ///
     /// In all other respects, this calling convention is the same as `SystemV`
     Kernel,
+    /// A function with this calling convention is expected to be called using the `call`
```
bitwalker, line 65: Note: Per the [Component Translation](https://github.com/0xPolygonMiden/compiler/discussions/360) discussion and our conversation today - it isn't necessary to represent the concept of context switching as an ABI unto itself, rather it is just one property of an ABI - arguably it is a property of a specific call site, not the callee itself, but for our purposes deriving it from ABI information works. I suggest that instead of `CrossCtx`, we specify three ABIs for use by our Wasm frontend: `Wasm`, `CanonLift`, and `CanonLower` (corresponding to core Wasm ABI and the respective primitives specified in the [Canonical ABI](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canon-lift)).

To elaborate on this:

* `Wasm` we already have of course, but it doesn't _really_ mean anything, and perhaps could be named something more generic. Essentially, this calling convention mostly indicates that: 
  a.) It came from a core Wasm module
  b.) We don't know anything about the "real" types of the arguments/results, i.e. it is mostly opaque from a type system perspective
  c.) The actual calling convention details of the function are encoded in its signature - by which I mean that if a specific convention was expected by the original source language, it has been compiled to Wasm such that those details are already baked in and represented by the core Wasm types that remain. Where it is unclear what the semantics are for a given type, we should consult the core Wasm spec for the `call` or `call_indirect` instructions.
* `CanonLift` and `CanonLower` correspond to their respective primitives as described in the [Canonical ABI](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#canon-lift) spec.
* Functions with the `CanonLift` ABI represent the actual component boundary, and calling/returning from them involves the actual context switch in the Miden VM. We'll revisit that part in a moment. This ABI is also the only one whose implementation details are _not_ fully specified by the Component Model/Canonical ABI. While the semantics of `canon lift` are described, the implementation is host-defined, and therefore is ours to dictate. This works out great for us, because this also happens to be where the idiosyncratic details of Miden's context switching must be dealt with, so this gives us flexibility.
* `CanonLower` is the only ABI other than `Wasm` that is callable from a `Wasm` ABI function. The actual details of this ABI are identical to `Wasm` from the perspective of a caller, however the use of a unique ABI for these functions allows us to identify them as synthetic, and to ensure that calls to a `CanonLift` function only occur via its `CanonLower` counterpart. A call to a `CanonLower` function can be thought of as a call to another `Wasm` ABI function, but with a layer of lifting/lowering inbetween, so what we end up with is actually a chain of calls composed of all three ABIs, i.e. `Wasm` (caller) -> `CanonLower` (wraps Canonical ABI function) -> `CanonLift` (wraps core function) -> `Wasm` (callee). A `CanonLower` function is _always_ paired with a `CanonLift` function, and is only ever called from a `Wasm` function. Conversely, `CanonLift` functions are never called directly from a `Wasm` function, only from a `CanonLower` function.
* Neither `Wasm` nor `CanonLower` require us to emit any special calling convention-related code, the details of the Canonical ABI itself are materialized in the implementation of the `CanonLower` and `CanonLift` functions.
* Both `Wasm` and `CanonLower` functions have type signatures consisting solely of core Wasm types.
* `CanonLift` functions have type signatures consisting solely of Component Model types. For the most part, we don't need to concern ourselves with these types, _except_ in regard to how to pass them across the context switch (i.e. as arguments in the call from `CanonLower` -> `CanonLift`, and as results when returning from `CanonLift`). The lifting/lowering of these types is described [here](https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#lifting-and-lowering-values), but the exact way in which we pass them across the `CanonLift` boundary is up to us.

With all that in mind, the only thing we need to determine if a function invocation should be lowered as `exec` or `call`, is to check the ABI of the callee. If it is `CanonLift`, then we use `call`, all other function invocations will use `exec`. As mentioned above, we also will validate all call sites to ensure that calls to `CanonLower` and `CanonLift` functions do not originate from invalid callers (e.g. direct from `Wasm` -> `CanonLift`). I should note that `CanonLower` and `CanonLift`, in terms of ABIs, _also_ encompass all of the other semantics as described in the Canonical ABI spec, so the fact that `CanonLift` corresponds to a context boundary is only one part of its contract.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
```diff
@@ -0,0 +1,170 @@
...
+/// For each component export ensure a module for each interface that has a lifting
+/// function, i.e. a function that takes parameters according to the cross-context Miden
+/// ABI, converts them to the Wasm CABI, calls the core Wasm module exported function,
+/// converts the results to the cross-context Miden ABI
+///
+/// After this stage all exported functons are expected to be called using the Miden ABI for
+/// cross-context calls, i.e. using the stack and the advice provider for arguments and results.
+pub struct LiftExportsCrossCtxStage;
+
+impl Stage for LiftExportsCrossCtxStage {
```
bitwalker, line 31: *Question:* I would expect us to emit IR from the Wasm frontend where the interfaces/component exports have already been materialized, i.e. when translating an exported component instance (as an interface), we would emit any synthetic exports as part of that translation. Is the current approach here just an artifact of the current IR structure, and can't really be changed until HIR2 is landed? If so, that's fine, but let's plan on moving this to the Wasm frontend as part of integrating HIR2, since it feels a bit out of place here.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        // dbg!(&component.exports());
+        // dbg!(&component.modules().keys());
+
+        let mut component_builder = ComponentBuilder::load(*component, &session.diagnostics);
+
+        let mut lifted_exports: BTreeMap<InterfaceFunctionIdent, ComponentExport> = BTreeMap::new();
+        let exports = component_builder.exports().clone();
+        for (id, export) in exports.into_iter() {
+            if let Canonical = export.function_ty.abi() {
```
bitwalker, line 55: This doesn't account for situations in which someone constructs IR that exports a function from the component with an ABI that is something other than Wasm-oriented, e.g. `SystemV`. To be clear, this isn't something our tools will produce, but someone else lowering to HIR may not do so through the Wasm frontend, and instead emit HIR components directly with whatever calling conventions they find appropriate.

Basically, I think we only ever synthesize these "lifting" functions when they correspond to a `canon lift` declaration, and only then. This is one of the reasons why synthesizing these functions in the frontend, where that information is present, is preferable.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
```diff
@@ -0,0 +1,170 @@
...
+            return Ok(input);
+        };
+
+        // dbg!(&component.exports());
+        // dbg!(&component.modules().keys());
+
+        let mut component_builder = ComponentBuilder::load(*component, &session.diagnostics);
+
+        let mut lifted_exports: BTreeMap<InterfaceFunctionIdent, ComponentExport> = BTreeMap::new();
+        let exports = component_builder.exports().clone();
```
bitwalker, line 53: *This note applies only to integrating HIR2 with the work in this PR:*

I would expect that the only exports from a component are either component functions (i.e. `CanonLift` ABI, from my other comment), or interfaces comprised of such functions. It looks like the way we're building components in the current IR is flattening the relationship between the component and the underlying module that provides the implementation of the component exports. Mostly just want to note that we'll need to make sure that when translating to HIR2 that we preserve more of the original component structure and visibilities.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
```diff
@@ -0,0 +1,170 @@
...
+                export.function.function, export.function.module
+            );
+            diagnostics.diagnostic(Severity::Error).with_message(message).into_report()
+        })?;
+    }
+    let span = export.function.function.span();
+    let call = builder.ins().exec(export.function, &params, span);
+    // dbg!(&sig);
+    let result = builder.first_result(call);
+    builder.ins().ret(Some(result), span);
```
bitwalker, line 158: Probably not relevant yet, because we haven't tested with any "interesting" types, but this is where we would need to handle returning results back across the context boundary, so we'll definitely have to flesh out the implementation here when we get into things that interact with memory, or involve representation changes and such.

midenc-compile/src/stages/cross_ctx/lower_imports.rs
```diff
@@ -0,0 +1,281 @@
...
+/// Generates lowering for imports for the cross-context calls according to the Miden ABI.
+///
+/// For each component import ensure a module for each interface that has a lowering function i.e. a
+/// function that takes the arguments according to the Wasm CABI, converts them to the cross-context
+/// Miden ABI, calls the imported function with the lowered arguments, takes the result
+/// (cross-context Miden ABI), and converts it to the Wasm CABI.
+/// The calls to the component import is switched to the generated lowering function.
+pub struct LowerImportsCrossCtxStage;
+
+impl Stage for LowerImportsCrossCtxStage {
```
bitwalker, line 28: Basically the same notes I had about the lifting stage apply here.

midenc-compile/src/stages/cross_ctx/lower_imports.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        // dbg!(&component.imports());
+        // dbg!(&component.modules().keys());
+
+        let mut component_builder = ComponentBuilder::load(*component, &session.diagnostics);
+
+        let mut lowered_imports: BTreeMap<FunctionIdent, ComponentImport> = BTreeMap::new();
+        let imports = component_builder.imports().clone();
+        for (core_import_func_id, import) in imports.into_iter() {
+            if let ComponentImport::MidenAbiImport(_) = import {
```
bitwalker, line 52: *Note:* All imports in a core Wasm module (whether they are imports of a `canon lower` declaration or of another core module's exports) have the same ABI (in terms of type signature and semantics) from a caller point of view. IMO we do not need to distinguish between them. It should be noted that an import of another component's exports, can only ever be called if it is synthesized via `canon lower`. In effect, the three ABIs I mentioned in my other comment are sufficient to encode all that information, and we should only be synthesizing lowering functions for `canon lower` declarations, so we don't need to consider ABI here at all - though that may have to wait until HIR2 is being integrated.

midenc-compile/src/stages/cross_ctx/lower_imports.rs
```diff
@@ -0,0 +1,281 @@
...
+            let message = format!(
+                "lowering function with name {} in module {} with signature \
+                 {import_lowered_sig:?} is already imported (function call) with a different \
+                 signature",
+                import_func_id.function, import_func_id.module
+            );
+            diagnostics.diagnostic(Severity::Error).with_message(message).into_report()
+        })?;
+    }
+    let call = builder.ins().call(import_func_id, &params, span);
```
bitwalker, line 166: *Note:* This is where we will need to handle any of the `CanonLift` ABI details I mentioned in my other comment, both when calling and when handling the return values. I know you haven't implemented any of that yet, since we're only dealing with scalars at the moment, but I'd leave a comment/TODO to that effect here (and in the place I mentioned in my corresponding comment in the lifting code).

midenc-compile/src/stages/cross_ctx/lower_imports.rs
```diff
@@ -0,0 +1,281 @@
...
+    });
+    // dbg!(&component_import);
+    // dbg!(&function_id);
+
+    Ok((component_import, function_id))
+}
+
+/// Rewrites the calls to the function `from` to the function `to` in the given function.
+/// Returns `true` if any call was rewritten.
+fn rewrite_calls(function: &mut Function, from: FunctionIdent, to: FunctionIdent) -> bool {
```
bitwalker, line 185: Is this needed anymore? We shouldn't need to rewrite any calls if we're synthesizing the `canon lift` and `canon lower` declarations, since calls will already be referencing those. Or am I misunderstanding what this is handling?

midenc-compile/src/stages/cross_ctx/lower_imports.rs
```diff
@@ -0,0 +1,281 @@
...
+                _op => (),
+            }
+
+            block.insts.push_back(UnsafeRef::from_box(inst));
+        }
+    }
+    dirty
+}
+
+/// Replaces calls to the imported functions with calls to the lowering functions.
```
bitwalker, line 221: Not sure about the current IR, but in HIR2 I expect we can avoid needing to touch the calls in the core module at all, since whether a function is a declaration (import) or a definition solely depends on whether it has a body or not. When we synthesize the body, none of the callers are affected.

I thought that was largely the same in the current IR, but my memory is a bit fuzzy on that detail.

midenc-debug/src/exec/executor.rs
```diff
@@ -85,10 +101,10 @@ impl Executor {
 
         log::debug!("executor created");
 
-        Ok(exec)
+        Ok(self)
     }
 
-    /// Set the contents of memory for the shadow stack frame of the entrypoint
+    /// Set the contents of memory for the shadow stack frame of the entrypint
```
bitwalker, line 107:
```suggestion
    /// Set the contents of memory for the shadow stack frame of the entrypoint
```

tests/integration/expected/rust_sdk_account_test/miden_sdk_account_test.hir
```diff
@@ -63,14 +63,14 @@
...
-                (let (v9 i64) (const.i64 1))
-                (let (v10 felt) (cast v9))
-                (let (v11 i64) (const.i64 2))
-                (let (v12 felt) (cast v11))
-                (let (v13 i64) (const.i64 3))
-                (let (v14 felt) (cast v13))
-                (let (v15 i64) (const.i64 4))
-                (let (v16 felt) (cast v15))
+                (let (v9 i32) (const.i32 1))
+                (let (v10 felt) (bitcast v9))
```
bitwalker, line 67: I'm not sure we should be allowing direct bitcasts between `i32` and `felt`, since that is obviously incorrect for signed values. Since in Wasm there isn't a distinction in terms of types between signed and unsigned, that's maybe less problematic, but in the IR type system there is a difference, which we'd generally want more explicit casts to represent.

I think the more semantically correct IR here would be a bitcast from `i32` to `u32`, and then an unsigned extension from `u32` to `felt`. Both of those are no-op casts once lowered, so there isn't a difference except in terms of the IR being slightly more verbose.

_Technically_, because the machine word size is a sort of weird superposition of `felt`/`u32`, the bitcast is safe, but definitely feels weird.

I don't know - what's your take?

tests/rust-apps-wasm/rust-sdk/cross-ctx-account/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+panic = "abort"
+
+[profile.dev]
+# Explicitly disable panic infrastructure on Wasm, as
+# there is no proper support for them anyway, and it
+# ensures that panics do not pull in a bunch of standard
+# library code unintentionally
+panic = "abort"
+opt-level = 1
+debug-assertions = false
```
bitwalker, line 34: Do we really want to disable debug assertions in debug builds? Those would generally be pretty useful in tests/troubleshooting. It will result in a bit more code, but the tradeoff is that we know pretty quick if something is obviously broken.

bitwalker (commented):
Overall, I think we could probably merge this as-is - _but_, once you start integrating HIR2, I think you'll find that much of this PR changes. Some of that I brought up in my comments, but obviously a fair bit is just inherently due to the differences in the IRs.

So I guess my question is - do we merge this, and then refactor it once HIR2 is merged, or merge HIR2, and then rework/recreate this PR on top of the new IR? On the one hand, the more code that has changed, the more work it is to integrate; but on the other, maybe you will find it easier to do the integration with all of this landed. What's your preference?

I think the only change I'd like to make to this PR, assuming it is practical to do so, is the calling convention changes I mentioned. _However_, if you think it is easier to do that change as part of the larger HIR2 integration, then I'm down to just merge this as-is, and go that route. Ultimately, whether they happen before or as part of integration doesn't make much of a difference, mostly just trying to avoid making things harder.

I'll get #332 wrapped up tomorrow (Tuesday), unfortunately ran out of time today with PR reviews, etc., so if you don't catch me with an update tonight before I sign off, I'll check in on this first thing in the morning.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
greenhat: My plan is to move it to the frontend during the HIR2 integration. 
Initially, I tried to figure out how to do it in the frontend, but IIRC due to the order of the translation in the frontend, I was missing some info on component import/export during the core module parsing. This, and the intention to build a prototype, pushed me to make it a separate stage(s). But since we'll revamp Wasm CM parsing (ditch inlining, etc.) and I now have a clearer picture of how it works (thanks to the prototype) I should be able to do it in the frontend.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
greenhat: I'm not sure I understand it. `Canonical` here means the Miden ABI. That's confusing. It's another type we have for describing the component imports/exports - https://github.com/0xPolygonMiden/compiler/blob/greenhat/i303-cross-ctx-lower/hir-type/src/lib.rs#L597-L608, and it'll be deleted once we migrate to the HIR2. We'll generate the synthetic functions in the frontend only for the `canon lift/lower` ones.

midenc-compile/src/stages/cross_ctx/lower_imports.rs
greenhat: Thanks! Done.

midenc-compile/src/stages/cross_ctx/lower_imports.rs
greenhat: Yes. I intend to ditch the component import/exports types in their current form during the HIR2 integration. 
This one is used to make the stage idempotent(re-entrable) in case of the added imports later, plus to skip the remnants of the SDK/intrinsics (`heap_base`) which should not be in the component imports to begin with.

midenc-compile/src/stages/cross_ctx/lower_imports.rs
greenhat: I intend to avoid replacing calls in the new HIR2 frontend. Here I'm rewriting them since the core Wasm import(function) in the note project for the `canon lower` has the same name (e.g. `miden:cross-ctx-account/foo@1.0.0#process-felt`) as the synthetic function generated for the `canon lift` in the account package, so I decided to rename the `canon lower` one (hardcoded prefix). In the HIR2 frontend, I plan to handle this in the core module import parsing, renaming them on the fly.

tests/integration/expected/rust_sdk_account_test/miden_sdk_account_test.hir
greenhat: This `bitcast` is produced by `Felt::from_u32` intrinsic at https://github.com/0xPolygonMiden/compiler/blob/greenhat/i303-cross-ctx-lower/frontend-wasm/src/intrinsics/felt.rs#L23-L27 so we expect to have only `u32` here. I refreshed my memory on the `bitcast` usage in `code_translator.rs` and we use it only when we are sure it's actually an unsigned value.
In general, I agree with you, we should not `bitcast` felts from `i32` values unless we sure it's an unsigned value.

tests/rust-apps-wasm/rust-sdk/cross-ctx-account/Cargo.toml
greenhat: Enabled. I probably got carried away trying to avoid any string formatting infra for panics, etc.

greenhat:
Thanks for the thorough review! I addressed all your notes and changed the `CallConv`. I'd like to merge it as is and take care of the rest in the HIR2 migration.

hir/src/function.rs
bitwalker: > The `canon lift` function description has CM type signatures, but the generated CanonLift function during the account code compilation has a "lowered" Miden cross-context ABI signature. 

It only has a single signature, the "high-level" type signature, there is no low-level signature for the `CanonLift` function at all. This is because it is called _only_ from a corresponding `CanonLower` function. The mechanics of the calling convention (when calling from `CanonLower` -> `CanonLift`) for those high-level types is up to us, i.e. how do we pass/return e.g. a `list`? However we see fit, because the `CanonLift` calling convention is host-defined. This calling convention is also where the context switch occurs, and the mechanics of that are also ours to define.

> For example, for the CM signature (list<felt>) -> list<felt> we will generate a synthetic function that uses advice provider, pass hashes and will have the "lowered" Miden CCABI (word) -> word signature.

Ah, I see where the confusion is coming from. You are assuming that we need to represent the calling convention implementation details in the function signature, resulting in needing two signatures (one to preserve the original information, one to encode the implementation details). This is not necessary though. The implementation details of the calling convention are _always_ derived from the original signature, so it is not necessary for us to encode them as a second signature, we just derive the latter as needed (essentially only during codegen). To be clear here, there is no "Miden ABI" or "Miden calling convention" per se - there _are_ constraints on how we can execute procedure calls in the Miden VM, and so any calling convention must be lowered to MASM in such a way that it works within those constraints. The "lowering" of calling conventions into Miden Assembly is responsible for emitting code that preserves the original calling convention semantics, but adheres to Miden's constraints at the VM level.

There are two places where we might want to do the actual work of preparing a function so that it is "Miden-compatible", by which I mean it can be lowered to Miden Assembly directly following the normal type layout rules, without violating any of the procedure invocation constraints:

1. During code generation, which is where the vast majority of these details will be handled, in particular all of the code related to laying out types in memory, interacting with the advice provider, operand stack management, etc.
2. As an IR transformation prior to codegen, so that we can ensure that code motion affecting function arguments is accounted for when we run the spills transformation. For example, if we have a `CanonLift` function whose signature is `list<felt> -> list<felt>`, we must rewrite the function (and any call sites) as if the signature was `digest -> digest`, as heap-allocated types cannot be passed across a context switch, so we instead store the lists in the advice provider, and pass the key in the advice map from which the actual list can be retrieved. The prologue of the function will then use that digest to extract the list into memory, and replace any uses of the original function argument with the list value once it has been read out of the advice provider. It should be noted that this transformation is implied/derived from the original signature, so we do not need to have the actual function definition to rewrite call sites, nor do we need to worry about call sites when rewriting the function.

The important thing to note about 2 is that while the actual IR is rewritten, the type signature is left unmodified! This is because the rewrite performed here doesn't change the semantics of the function at all, it is purely an artifact of representing those semantics in Miden Assembly.

I should also note that we do, in a sense, store both the original signature, and the "effective" signature here, because in HIR2 a `Function` has a `signature` attribute (a `Signature` value, corresponding to the original signature, and also providing things like ABI details), but we can also directly interrogate the types of its arguments and results as represented in the IR at any given point in time. However in almost all cases, the two will be the same. The only time we'd allow them to diverge is during codegen, in order to facilitate the rewrite I mentioned above.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
bitwalker: Yeah I'm saying we do not need to represent the notion of a "Miden ABI" at all, it is implied that lowering a function to Miden Assembly _by definition_ must adhere to the "Miden ABI". From a call site perspective, it is sufficient to know only the original callee signature (including its calling convention), because the low-level details of how to invoke the resulting Miden Assembly procedure can be completely derived from that signature.

Backing up a bit, and to help clarify conversations about this going forward, here's how I define the distinction between ABIs and calling conventions:

* An _ABI_ describes how a given set of types are laid out in memory, and how procedure calls _in general_ are executed by the machine, such that two applications can even begin to talk to one another. It is a very low-level, machine-dependent specification. For our purposes, ABI consists of:
  - The specification for how byte-addressable memory is laid out in Miden's word-addressable memory, and on the Miden operand stack
  - The bitwise representation of types in the IR type system, i.e. how they are laid out in byte-addressable memory
  - The specification of what constitutes valid _immediates_ in IR type system terms
  - The maximum addressable depth of the operand stack, and how overflow is handled by the VM
  - Operand stack effects of the various procedure invocation instructions: i.e. `exec` has none, `call` implicitly hides/restores the overflow portion of the operand stack and switches context, `dynexec` consumes the first element on the stack as a procedure reference, etc.
  - Program lifecycle, e.g. rodata initialization, global constructors, etc.
* A _calling convention_ encodes the information necessary to understand how to call a given function, i.e. how it expects to receive its arguments, and how you can expect to receive its results. A calling convention is generally specified in terms of a specific ABI. For example, the C calling convention can mean different things depending on the target ABI/architecture. In general, this includes:
  - The set of types specified by that convention
  - The way in which specific types are expected to be passed at the machine level (i.e. is a struct passed by value, or by reference, is there a threshold at which that happens). Is there a distinction between integers and floats, etc.
  - How to handle extension of integer values whose type is smaller than the machine word (i.e. are they sign-extended, or zero-extended). This isn't as important in a stack machine, but in a register machine it is essential, since different machine instructions can read the same register as different bit-widths, and failing to sign/zero-extend a value can result in silent corruption of computed results.
  - The range of possible function effects (e.g. are calls asynchronous, will they return, can they throw exceptions) and how those are handled
* In addition to those more general properties, any calling convention we support must also specify:
  - The procedure call protocol for passing arguments via the advice provider when the call involves a context switch, or when the callee arguments would overflow the addressable stack.
  - Whether the call involves a context switch (can also be considered a function effect)

So, with that in mind, what I'm trying to say is that the notion of a "Miden ABI" is implicit - we do not have any other target than the Miden VM, so by definition all of our calling conventions are/must be specified in terms of how they are represented at the Miden VM level.

So consider the `Wasm` calling convention, it specifies what types are allowed in function signatures using that convention (i.e. any type described in the core Wasm spec). _We_ have further defined how that convention is to be translated into Miden Assembly. When the IR is constructed, we will validate that there are no convention violations (i.e. that there aren't argument types with no corresponding Wasm representation). Only when we move to the code generation stage will we actually need to deal with calling convention details such as spilling of excess arguments, layout of types on the operand stack/in memory, etc.

Consider another example: the `CanonLift` convention. This one is maybe more interesting, because not only does the convention imply a context switch, it also permits a much broader range of types. Because of the context switch, we are required to handle certain types (i.e. heap-allocated types) a certain way, by passing them via the advice provider. We must also account for spilling of excess arguments, but again, due to the context switch we are required to handle spills a specific way as well (via the advice provider). Beyond those details, we have not yet defined specifically how the various allowed types permitted by the convention will be passed at the VM level, but that is entirely our perogative. For the most part, I suspect that we will treat most types the same we do in `Wasm`, as that already accounts for most (all?) scalar types allowed.

bitwalker (approved):
Looks good for now, I left a few more notes that I think will help clarify some of the confusing aspects of ABI vs calling convention, let me know if you still aren't sure what to make of that portion of my feedback.

Hopefully I can have all the HIR2 stuff ready for you by tomorrow, at the very least I there will be updates for you to review, and we can start to discuss any open questions of integrating it into the frontend.

hir/src/function.rs
greenhat: > > The `canon lift` function description has CM type signatures, but the generated CanonLift function during the account code compilation has a "lowered" Miden cross-context ABI signature.
> 
> It only has a single signature, the "high-level" type signature, there is no low-level signature for the `CanonLift` function at all. This is because it is called _only_ from a corresponding `CanonLower` function. The mechanics of the calling convention (when calling from `CanonLower` -> `CanonLift`) for those high-level types is up to us, i.e. how do we pass/return e.g. a `list`? However we see fit, because the `CanonLift` calling convention is host-defined. This calling convention is also where the context switch occurs, and the mechanics of that are also ours to define.
> 
> > For example, for the CM signature (list) -> list we will generate a synthetic function that uses advice provider, pass hashes and will have the "lowered" Miden CCABI (word) -> word signature.
> 
> Ah, I see where the confusion is coming from. You are assuming that we need to represent the calling convention implementation details in the function signature, resulting in needing two signatures (one to preserve the original information, one to encode the implementation details). This is not necessary though. The implementation details of the calling convention are _always_ derived from the original signature, so it is not necessary for us to encode them as a second signature, we just derive the latter as needed (essentially only during codegen). To be clear here, there is no "Miden ABI" or "Miden calling convention" per se - there _are_ constraints on how we can execute procedure calls in the Miden VM, and so any calling convention must be lowered to MASM in such a way that it works within those constraints. The "lowering" of calling conventions into Miden Assembly is responsible for emitting code that preserves the original calling convention semantics, but adheres to Miden's constraints at the VM level.

This helps and clarifies a lot. Thank you for the detailed explanation! It started to click for me. The high-level signature is surely enough to derive the low-level "implementation detail" one.

> 
> There are two places where we might want to do the actual work of preparing a function so that it is "Miden-compatible", by which I mean it can be lowered to Miden Assembly directly following the normal type layout rules, without violating any of the procedure invocation constraints:
> 
> 1. During code generation, which is where the vast majority of these details will be handled, in particular all of the code related to laying out types in memory, interacting with the advice provider, operand stack management, etc.
> 2. As an IR transformation prior to codegen, so that we can ensure that code motion affecting function arguments is accounted for when we run the spills transformation. For example, if we have a `CanonLift` function whose signature is `list<felt> -> list<felt>`, we must rewrite the function (and any call sites) as if the signature was `digest -> digest`, as heap-allocated types cannot be passed across a context switch, so we instead store the lists in the advice provider, and pass the key in the advice map from which the actual list can be retrieved. The prologue of the function will then use that digest to extract the list into memory, and replace any uses of the original function argument with the list value once it has been read out of the advice provider. It should be noted that this transformation is implied/derived from the original signature, so we do not need to have the actual function definition to rewrite call sites, nor do we need to worry about call sites when rewriting the function.
> 
> The important thing to note about 2 is that while the actual IR is rewritten, the type signature is left unmodified! This is because the rewrite performed here doesn't change the semantics of the function at all, it is purely an artifact of representing those semantics in Miden Assembly.

This part confuses me. From the previous comment, I understood that `CanonLift` and `CanonLower` synthetic functions are generated in the frontend, and now it appears that they are "implemented" in the IR transformation phase or the codegen. This leads me to think that the frontend generates empty synthetic functions which later "implemented" with the "Miden-compatible" lowering you described above either in the IR transformation phase or the codegen. Am I on the right track?

midenc-compile/src/stages/cross_ctx/lift_exports.rs
greenhat: Whoa! That's gold! I'm certainly guilty of conflating calling convention and ABI many times.

hir/src/function.rs
bitwalker: There are a couple different things here being accidentally conflated:

* A `CanonLower` function _declaration_ presents as just another `Wasm` function, to `Wasm` callers, from a semantics point of view. However, what it really doing is adapting a function with the `CanonLift` calling convention to the `Wasm` convention. An alternate way of thinking about it, is that we could allow `Wasm` functions to directly call `CanonLift` functions, but then we would need to emit the equivalent of the `CanonLower` implementation at every call site. This would generate a lot more code for no benefit.
* The _implementation_ of a `CanonLower` function emitted by the frontend must handle translating core Wasm types into their Canonical ABI representation. This could be represented with something like `cast`, but may need to be more explicit (depending on the type, I'm not sure). It then calls the corresponding `CanonLift` function with arguments in the Canonical ABI representation. Note that, at this point, no low-level calling convention details are being represented - only the type representation from core Wasm to Canonical ABI.
* The `CanonLift` _declaration_ is only callable from a `CanonLower` function, which can be thought of as a Canonical ABI caller (since internally it has converted types to the Canonical ABI). Much like how `CanonLower` presents as a `Wasm` function from the perspective of a `Wasm` caller `CanonLift` presents as a Canonical ABI function to a Canonical ABI caller, by internally adapting the underlying `Wasm` calling convention function to the Canonical ABI.
* The _implementation_ of a `CanonLift` function emitted by the frontend must handle translating Canonical ABI types into their core Wasm representation. Again, this could be represented with `cast`, but likely requires more explicit lowering code for certain types. Once complete, it then calls the underlying `Wasm` function. Again, no low-level calling convention details are being represented here - just the type representation change.

So in short:

* The frontend synthesizes the `CanonLift`/`CanonLower` functions with code that effects the type representation change
* The backend handles the low-level calling convention details (e.g. how Canonical ABI types are passed for the `CanonLift` convention, how to handle context switches, operand stack management, laying out types in memory, etc.)

bitwalker: I should also note that, we can't represent the type representation change purely using `cast` (except where the cast is obviously applicable, e.g. between integral types and such), because the IR type system doesn't distinguish between WebAssembly and Canonical ABI types, and shouldn't. Casts are essentially representation-agnostic, expressing only a conversion between IR types, and the default Miden ABI layout of those types may not match what is specified by the Canonical ABI. 

So by emitting (in the frontend) code to handle the representation change, we ensure that the code correctly lays things out according to the Canonical ABI rules, and then during codegen, we will know that the IR types corresponding to the function arguments/results are in Canonical ABI representation (as it is implied by the calling convention), and we will emit the low-level details of transferring arguments/results in that representation from caller to callee. In other words, the combination of IR type and calling convention imply a specific data layout when used as a function argument or result.

The Wasm aspect of all this is a bit weird. Basically, the Component Model is able to avoid explicitly representing the type representation change in the form of Wasm instructions because the host runtime speaks Wasm and can handle it automatically (and therefore more efficiently). Typically though, adapting types from one type system to another would require the frontend emit code to do that work somewhere, and then the low-level details are left up to the backend. We're in kind of an awkward position, as we must act as both frontend and host runtime: First, we materialize functions that perform the work that would normally have been emitted by a frontend on more traditional targets, to effect the type representation change. Second, we act as the host, by enforcing the shared-nothing boundary via context switching, and managing the movement of data across that boundary. Where it gets confusing is that, in the functions we synthesize to do this, we have more traditional calling convention details in the mix as well. So it gets a bit fuzzy here, in terms of being precise about what parts of the apparent calling convention are handled by the frontend, and what is handled by the backend. Hopefully my commentary here has made that distinction clearer, but I certainly can't blame you if it is still confusing.

midenc-compile/src/stages/cross_ctx/lift_exports.rs
bitwalker: Yeah me too, I've been playing fast and loose with the terminology in a lot of these conversations and now we're paying the price 😅

hir/src/function.rs
greenhat: Now, I feel that it finally clicked for me! Thank you for the detailed explanation!

Issue #363, opened 2024-12-17
Migrate the frontend to use the new HIR2


greenhat:
Compiled notes from PR comments:

### `anon` namespace

>Rather than using the `Anon` namespace, we should translate symbol names as described in the [Component Translation](https://github.com/0xPolygonMiden/compiler/discussions/360) discussion. The gist is that the WIT package name becomes the `LibraryNamespace` (specifically the `User` variant), sans version, items exported from the component use the export name as the first non-namespace path component, and so on (i.e. functions exported from an exported interface are additional components). Future changes to `LibraryPath` will make the syntax for fully-qualified paths a bit more like WIT identifiers, but for now the main thing we want to do is translate them in a way that retains useful semantics in Miden Assembly. We've largely already discussed this, but I'm just reiterating it here since it is relevant to decisions documented in this PR
>Likewise, the way in which module/function names are recovered I believe will need to be changed, but we can do that when merging the new IR into the frontend. This is also described in the Component Translation discussion, but essentially we can recover the original WIT interfaces (and their names/exports) using the component instances exported from the top-level component

from https://github.com/0xPolygonMiden/compiler/pull/353?new_mergebox=true#issuecomment-2547008283 :

### Generate synthetic functions for Wasm CM `canon lift/lower` in the frontend

See https://github.com/0xPolygonMiden/compiler/pull/357 comments for more details.

```[tasklist]
### Core Wasm module translation (for Miden VM program)
- [x] translate simple(integer addition) function
- [x] translate `fib` program
- [x] fix the rest of the Wasm ops translation and restore `code_translator::tests` (commented)
- [x] translate `Call` op
- [x] translate globals and data segments
- [x] translate `BrTable`
- [x] translate `If/Else` (not used by rustc)
```
```[tasklist]
### Wasm component translation (for accounts, note scripts, etc.)
- [x] translate `felt` intrinsics
- [x] translate `stdlib` calls (ABI transformations)
- [x] translate Wasm CM, `CanonLift` and `CanonLower` synthetic functions
- [x] translate miden tx kernel calls (ABI transformations)
```

bitwalker:
[Here's](https://github.com/0xPolygonMiden/compiler/pull/357#discussion_r1889134473) another comment that you may find useful, that more clearly defines ABI and calling convention terminology, and how I envision us approaching those concepts in the compiler.

[...]

PR #367, opened 2025-01-02
Migrate the frontend to use the new IR

greenhat:
Ref #363 

This PR establishes the `midenc-frontend-wasm2` crate (a copy of the original) and modifies the translation to generate the HIR2 Component. The core Wasm module code path (non-rollup Rust code) is partially implemented to handle the translation of the simple (addition) Rust code (see `pure_rust_hir2` test). The missing translation for more complex code will be implemented in the subsequent PRs.
Note, that the Wasm CM code path is using the sketched types and only deals with component imports and exports so far.


dialects/hir/src/builders/function.rs
```diff
@@ -1,36 +1,38 @@
...
-    ValueRef,
+    dialects::builtin::*, AsCallableSymbolRef, Block, BlockRef, Builder, Immediate, Listener,
+    OpBuilder, Overflow, Region, RegionRef, Report, SourceSpan, Type, UnsafeIntrusiveEntityRef,
+    Usable, ValueRef,
 };
 
 use crate::*;
 
-pub struct FunctionBuilder<'f> {
+pub struct FunctionBuilder<'f, L: Listener> {
```
bitwalker, line 9: I'm not sure this is actually needed anymore, no? It looks like the `FunctionBuilder` in `midenc_dialect_hir` already provides equivalent functionality (and more complete, i.e. most instructions are already supported there). If we need this custom one, maybe it would be better to adapt the `midenc_dialect_hir` version to be more generic, than to derive an almost identical one in this crate.

I haven't finished reviewing this PR yet though, so I may be missing the reason why we've kept this around for now.

bitwalker: Wait, I'm dumb, ignore this 😅

bitwalker (approved):

PR #373, opened 2025-01-14
chore: switch from `wasm32-unknown-unknown` to `wasm32-wasip1` 

greenhat:
As a workaround for our old rustc version (fails in fib tests). See https://github.com/rust-lang/libm/issues/214#issuecomment-2574228266 
Besides, that I think we should stick to `wasm32-wasi` everywhere.


greenhat:
@bitwalker Although I stacked this PR on the https://github.com/0xPolygonMiden/compiler/pull/372, the merging order is not important here.

bitwalker:
I didn't need to make this change to get our build working, I think only the old `wasm32-wasi` target was removed, so that was causing tests to fail, but removing that target via `rustup` was sufficient to get things working again (although I combined it with an update of the stable/nightly toolchains, so there may be some fixes I picked up that I am not aware of).

I think we should continue to use `wasm32-unknown-unknown` for tests which are not Component Model based, since it is technically the appropriate target for pure Rust projects that don't target the rollup, and it doesn't hurt to have a check on any unintentional assumptions we make due to the nature of `wasm32-wasip1` or `wasm32-wasip2` targets.

Given that, I'm going to close this for now I think, if you disagree, feel free to reopen and let's discuss further.

PR #377, opened 2025-01-29
Implement CondBr, Br, Loop translations; add `fib` test  

greenhat:
Ref #363 

**This PR is based on the #367**

This PR implements `CondBr`, `Br` and `Loop` ops along with the missing parts in the `SSABuilder` to successfully translate `fib` test code (see `fib_hir2` test).

### DIFF
--- a/dialects/hir/src/ops/binary.rs
+++ b/dialects/hir/src/ops/binary.rs
@@ -573,6 +573,22 @@ pub struct Neq {
 
 infer_return_ty_for_binary_op!(Neq as Type::I1);
 
+/// Inequality comparison with immediate
+#[operation(
+    dialect = HirDialect,
+    implements(InferTypeOpInterface)
+)]
+pub struct NeqImm {
+    #[operand]
+    lhs: AnyInteger,
+    #[attr]
+    rhs: i32,
+    #[result]
+    result: Bool,
+}
+
+infer_return_ty_for_binary_op!(NeqImm as Type::I1);
+
 /// Greater-than comparison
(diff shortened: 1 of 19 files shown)


dialects/hir/src/ops/binary.rs
```diff
@@ -573,6 +573,22 @@ pub struct Neq {
 
 infer_return_ty_for_binary_op!(Neq as Type::I1);
 
+/// Inequality comparison with immediate
+#[operation(
+    dialect = HirDialect,
+    implements(InferTypeOpInterface)
+)]
+pub struct NeqImm {
```
bitwalker, line 581: We shouldn't require these types of operations anymore (i.e. variants that accept an explicit immediate argument). The operand folding infrastructure can match on operations with constant operands and fold them where appropriate, and in all other circumstances will tell us whether a given operand is immediate or not.

The fewer op variants we have, the better, but this in particular is one of the ways the new IR is nicer than the old one. To be clear, you can still have builder methods like `neq_imm`, just have it emit both a constant op and a `neq`

bitwalker (approved):
Just one note about the new op you added, but I'll merge this anyway for now, we can tweak that in a subsequent PR.

dialects/hir/src/ops/binary.rs
greenhat: Got it. I'll remove them ASAP. 
I was suspected this when I saw `*Imm` ops missing, but then I found `ShlImm`. :)

PR #397, opened 2025-02-11
frontend: translate `mem::intrinsics::heap_base`

### LINKED ISSUE #396: Translate `heap_base` memory intrinsic needed for `BumpAlloc`
#363

greenhat:
Close #396 

This PR add translation for `mem::intrinsics::heap_base` by creating a module(`mem::intrinsics`) with an "empty" function (`heap_base`) since the real function implementation is linked at the codegen.

### DIFF
--- a/hir2/src/dialects/builtin/builders/component.rs
+++ b/hir2/src/dialects/builtin/builders/component.rs
@@ -1,8 +1,9 @@
+use super::ModuleBuilder;
 use crate::{
     dialects::builtin::{
         ComponentRef, InterfaceRef, ModuleRef, PrimInterfaceBuilder, PrimModuleBuilder,
     },
-    Builder, Ident, Op, OpBuilder, Report, Spanned,
+    Builder, FunctionIdent, Ident, Op, OpBuilder, Report, Signature, Spanned, SymbolTable,
 };
 
 pub struct ComponentBuilder {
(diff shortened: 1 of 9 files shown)


hir2/src/dialects/builtin/builders/component.rs
```diff
@@ -34,6 +35,25 @@ impl ComponentBuilder {
 
     pub fn define_module(&mut self, name: Ident) -> Result<ModuleRef, Report> {
         let builder = PrimModuleBuilder::new(&mut self.builder, name.span());
-        builder(name)
+        let module_ref = builder(name)?;
+        let is_new = self
+            .component
+            .borrow_mut()
+            .symbol_manager_mut()
+            .insert_new(module_ref, crate::ProgramPoint::Invalid);
```
greenhat, line 43: @bitwalker I'm wondering if the symbol registration is better to be done in the op builder (module in this case)? When I first encountered this (function creation in the module) I was quite surprised that new function is not automatically registered in module symbol table.

bitwalker: Good catch, I must've never added the logic to automatically register new symbol ops when inserted into a symbol table op. That should probably be done as some sort of hook that fires anytime an op is inserted into a region (or removed from one), but for now we'll have to just implement the logic in the op-specific builders manually.

bitwalker (approved):

PR #399, opened 2025-02-14
feat: initial draft of control flow lifting transform

bitwalker:
This PR implements a generic control flow lifting transformation framework, and an implementation of it for the HIR dialect.

The tranform converts unstructured control flow operations to structured control flow ops, e.g. `hir.cond_br` to `hir.if`. This includes complex control flow such as loops with loop-carried variables, reduce-style loops, etc.

Not all control flow is guaranteed to be lifted however. Currently, the primary limitation is that all return-like operations in a region must be equivalent (i.e. the same operation type), otherwise some control flow will be left unstructured.

For now, on the backend, this will result in a legalization error, as we do not have lowerings for unstructured control flow ops at this time. In the future we plan to lower directly to structured control flow, so this is more of a temporary solution for us anyway; but may also be useful for other compiler frontends.

---

@greenhat This is marked as a draft until I've finished testing and working out any kinks. I'll add the commit with the pipeline once that is ready. Fair warning, a lot of the code here is pretty rough, as it is pretty directly translated from C++ and I didn't have much time to spend re-architecting it to be as Rust-y as I'd like. Since long term we aim to lower directly to structured control flow anyway, I don't actually care all that much for now, but we may have to come back and clean this up at some point.


greenhat (approved):
LGTM!

Issue #403, opened 2025-02-17
Cannot create `GlobalSymbol` from `GlobalVariable` found via symbol resolution


greenhat:
The mutable borrow of the `GlobalVariable` at https://github.com/0xPolygonMiden/compiler/blob/greenhat/i363-more-tests-hir2/hir2/src/ir/operation.rs#L461-L461 fails due to the borrow of the same `GlobalVariable` from the resolved symbol.

EDIT: In `GlobalSymbolBuilder` we're trying to mutably borrow the GV reference that was passed to it as parameter.

To reproduce, uncomment the above linked code in #405 and run `cargo test code_translator::tests::globals`

greenhat:
Fixed in https://github.com/0xPolygonMiden/compiler/pull/405/commits/67273a79aeaf60c6b1a873e8752c06f84e585150

PR #405, opened 2025-02-17
fix(frontend): implement Wasm `global.get` and `set` translation

greenhat:
#363 

It turned out that I missed `global.set` and `get` Wasm ops last time. This PR implements them and adds a frontend test `code_translator::tests::globals`

### DIFF
--- a/hir2/src/ir/operation.rs
+++ b/hir2/src/ir/operation.rs
@@ -416,7 +416,7 @@ impl Operation {
         symbol: impl AsSymbolRef,
     ) {
         let attr_name = attr_name.into();
-        let mut symbol = symbol.as_symbol_ref();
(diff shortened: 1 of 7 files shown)


hir2/src/ir/operation.rs
```diff
@@ -458,7 +458,10 @@ impl Operation {
             self.set_attribute(attr_name, Some(attr));
         }
 
-        symbol.borrow_mut().insert_use(user);
+        // TODO: restore after https://github.com/0xPolygonMiden/compiler/issues/403
+        // is fixed
+        //
+        // symbol.borrow_mut().insert_use(user);
```
greenhat, line 464: @bitwalker I have not found a quick solution for #403 so I temporarily commented it out.

bitwalker (approved):

PR #407, opened 2025-02-18
[1/x] Order modules passed to the assembler and fix "intrinsics::i32 module not found" when compiling the Collatz example

### LINKED ISSUE #406: Collatz conjecture example compilation fails
```rust
fn entrypoint(mut n: u32) -> u32 {
    let mut steps = 0;
    while n != 1 {
        if n % 2 == 0 {
            n /= 2;
        } else {
            n = 3 * n + 1;
        }
        steps += 1;
    }
    steps
}
```

`cargo miden build` fails with `error: undefined module 'intrinsics::i32'`

greenhat:
Close #406 

EDIT: The reason for #406 is that the target compilation module is passed first to the assembler, before the intrinsics modules. The assembler sees the call to an intrinsic function, but the intrinsics module is not loaded, so it bails out.

This PR:
- Orders the program modules being passed to the assembler so that intrinsics module go first, before the target compilation module that depends on them. 
- Refactors the code that recognizes the intrinsics module names across the codebase to use the constants defined once instead of strings. 
- Adds the Collatz conjecture example.


bitwalker:
@greenhat This is a bug with the assembler - it should not require modules in any particular order, so to fix this properly, we'll need to do it there. I was like 99% sure I already did that when I refactored the assembler last year, but things have been changing, so a regression may have been introduced. We can merge this as a temporary workaround, but this issue will almost certainly surface again while the assembler has any kind of expectation on the order in which modules are passed to it.

/cc @plafer @paracetamolo In case something here rings a bell

bitwalker (approved):

plafer:
I believe this was introduced with https://github.com/0xPolygonMiden/miden-vm/pull/1643 - @paracetamolo I think we had discussed this; probably is worth fixing now

bobbinth:
> I believe this was introduced with [0xPolygonMiden/miden-vm#1643](https://github.com/0xPolygonMiden/miden-vm/pull/1643) - @paracetamolo I think we had discussed this; probably is worth fixing now

Do we already have an issue to fix this in `miden-vm` repo?

plafer:
Created https://github.com/0xPolygonMiden/miden-vm/issues/1669

PR #415, opened 2025-02-25
wip: switch compiler to hir2

bitwalker:
This PR makes the switch from the old HIR1 pipeline to the HIR2 pipeline. It does not remove the old HIR1 code yet, that will come later.

@greenhat This is still a work-in-progress, as a few things are stubbed out while I've gone through to update all the places in the code where old bits are referenced. As such, the tests haven't been run yet (and will largely fail until I finish tying up loose ends). The frontend-wasm2 crate should be isolated from breakages though, so I think this shouldn't affect you too much, until you get to the point of wanting to run the pipeline, and by then I hope to have the rest of this wrapped up.

**FINAL EDIT:** There is a regression in the optimality of the code being generated, as the HIR1 treegraph-based instruction scheduler hasn't been ported over yet. The result is an excess of stack management instructions as operands are located all over the operand stack, rather than being available just-in-time. Implementing the same kind of treegraph-based scheduling rewrite is actually much easier and more robust in the new IR, I just haven't done it yet. I plan to address this in the next release.


codegen/masm2/src/artifact.rs
```diff
@@ -94,3 +100,248 @@ impl Rodata {
         Ok(felts)
     }
 }
+
+inventory::submit! {
+    midenc_session::CompileFlag::new("test_harness")
```
greenhat, line 105: I suspect this causes an error `-> Command midenc: Argument names must be unique, but 'test_harness' is in use by more than one argument or group` when I run tests locally. Interestingly, it does not happen on CI (cached old version of `clap_builder`?).

hir2/src/dialects/builtin/builders/world.rs
```diff
@@ -38,4 +40,25 @@ impl WorldBuilder {
         let builder = PrimComponentBuilder::new(&mut self.builder, name.span());
         builder(ns, name, ver)
     }
+
+    /// Declare a new world-level module `name`
+    pub fn declare_module(&mut self, name: Ident) -> Result<ModuleRef, Report> {
```
greenhat, line 45: Ha! For some reason, I thought that `World` can contain only `Component`s. OTOH it makes sense to link stdlib and tx_kernel as modules.

bitwalker: Yeah my thinking is that "plain" modules, e.g. the Miden stdlib or whatever, can be declared/translated as world-level modules (since we can then just use their "real" symbol names when calling procedures exported from them). It makes the structure a bit more fluid, but feels like it is a better way to represent modules that aren't sourced from a Wasm component (e.g. intrinsics).

codegen/masm2/src/artifact.rs
bitwalker: Ah, that's happening because the test crate is linking both `midenc-hir` and `midenc-hir2` which both define the same flag.

dialects/hir/src/canonicalization/simplify_br_to_block_with_single_pred.rs
```diff
@@ -0,0 +1,74 @@
+use alloc::rc::Rc;
+
+use midenc_hir2::*;
+
+use crate::{ops::Br, HirDialect};
+
+/// Simplify a branch to a block that has a single predecessor. This effectively merges the two
+/// blocks.
+pub struct SimplifyBrToBlockWithSinglePred {
```
greenhat, line 9: Nice!

PR #417, opened 2025-02-28
Wasm component translation

### LINKED ISSUE #363: Migrate the frontend to use the new HIR2

greenhat:
Close #363 

This PR concludes the HIR2 migration and adds Wasm component translation:
- component exports translation (`CanonLift`), synth functions on the `Component` level;
- component imports translation (`CanonLower`), synth functions on the `Module` level + expected counterpart (export) in the external `Component`;
- integration tests using Wasm CM are enabled up to green stages, and all encountered errors are submitted as issues and  referenced in the test's code.


greenhat:
@bitwalker This PR is ready. I'm going to do some frontend refactoring and cleanup in the next PR. The old frontend can be removed, and I'll do it in the subsequent PR.

greenhat:
@bitwalker Now, this PR is properly ready. Nothing is left unimplemented, compared to the old frontend. 

I'm going to do some frontend refactoring and cleanup in the next PR. The old frontend can be removed, and I'll do it in the subsequent PR.

codegen/masm2/src/lib.rs
```diff
@@ -34,6 +36,10 @@ pub use self::{
 pub fn register_dialect_hooks(context: &midenc_hir2::Context) {
     use midenc_dialect_hir as hir;
 
+    context.register_dialect_hook::<builtin::BuiltinDialect, _>(|info, _context| {
```
bitwalker, line 39: You'll want to add these ops to the `eval` crate's `register_dialect_hooks` function as well, since the `Eval` trait is registered as a operation trait.

frontend-wasm2/src/code_translator/mod.rs
```diff
@@ -139,15 +131,15 @@ pub fn translate_operator(
...
+                    let imm = builder.ins().imm(Immediate::I64(0), span);
+                    let cond = builder.ins().neq(cond, imm, span)?;
+                    state.push1(builder.ins().select(cond, arg1, arg2, span)?);
                 }
                 ty => panic!("unsupported value type for 'select': {ty}"),
             };
         }
         Operator::Unreachable => {
-            builder.ins().unreachable(span);
+            builder.ins().unreachable(span)?;
```
bitwalker, line 142: *Note:* I think I already updated this, but just in case, the need for dedicated instruction builder, i.e. the reason why we have the `.ins()` builder method, is gone. Now any implementation of `OpBuilder` or `Rewriter` can be used to build ops, just by bringing the dialect-specific builder trait into scope, e.g. `UndefinedBehaviorOps` in the case of `unreachable`.

I believe I already updated the codebase to use the new style though, so probably this is just something that will go away after a rebase.

frontend-wasm2/src/code_translator/tests.rs
```diff
@@ -1,9 +1,9 @@
 use core::fmt::Write;
-use std::{any::Any, rc::Rc};
+use std::rc::Rc;
 
 use expect_test::expect_file;
 use midenc_hir2::{
-    dialects::builtin::{self, Module},
+    dialects::builtin::{self},
```
bitwalker, line 6:
```suggestion
    dialects::builtin,
```

frontend-wasm2/src/component/build_ir.rs
```diff
@@ -32,27 +29,24 @@ pub fn translate_component(
...
-    let (mut component_types_builder, parsed_root_component) =
+    let (mut component_types_builder, mut parsed_root_component) =
         parse(config, wasm, context.session())?;
-    let dialect = context.get_or_register_dialect::<BuiltinDialect>();
-    dialect.expect_registered_name::<midenc_hir2::dialects::builtin::Component>();
-    // context.get_or_register_dialect::<HirDialect>();
     // Extract component name from exported component instance
     let id = {
-        let instance = parsed_root_component
+        let instance: String = parsed_root_component
```
bitwalker, line 36: *Nit:* This type annotation can be elided no? The call to `.to_string()` disambiguates it, and the signature of `parse_component_id` requires a `String` anyway.

Doesn't really matter too much, but Rust already can be a bit noisy at times with types.

frontend-wasm2/src/component/lift_exports.rs
```diff
@@ -0,0 +1,105 @@
...
+        .copied()
+        .map(|ba| ba as ValueRef)
+        .collect();
+
+    // NOTE: handle CC lifting/lowering for non-scalar types
+    // see https://github.com/0xPolygonMiden/compiler/issues/369
+
+    let exec = fb
+        .ins()
+        .exec(core_export_func_ref, core_export_func_sig, args.to_vec(), span)
```
bitwalker, line 88:
```suggestion
        .exec(core_export_func_ref, core_export_func_sig, args, span)
```

frontend-wasm2/src/component/lift_exports.rs
```diff
@@ -0,0 +1,105 @@
...
+        .expect("failed to build an exec op");
+
+    let borrow = exec.borrow();
+    let results_storage = borrow.as_ref().results();
+    let results: Vec<ValueRef> =
+        results_storage.iter().map(|op_res| op_res.borrow().as_value_ref()).collect();
```
bitwalker, line 94:
```suggestion
    let results = ValueRange::<2>::from(borrow.results().all());
```

*Note:* Prefer to use `ValueRange` any time you need to operate on owned/borrowed ranges of operands, results, or block arguments. It will avoid allocating unless absolutely needed, and handles coercing all of those value types into `ValueRef`. The constant argument just dictates how many values the owned variant can hold before allocating on the heap, but if you never materialize the ownership, it just dictates how large the type is on the stack.

frontend-wasm2/src/component/lower_imports.rs
```diff
@@ -0,0 +1,119 @@
+//! lowering the imports into the Miden ABI for the cross-context calls
+
+use std::{cell::RefCell, rc::Rc, str::FromStr};
```
bitwalker, line 3:
```suggestion
use alloc::rc::Rc;
use core::{cell::RefCell, str::FromStr};
```

In preparation for running the compiler in the browser, we should make sure to only reference libstd behind a `#[cfg(feature = "std")]` gate. Generally speaking though, if we can reference a type in libcore we should, failing that, liballoc.

frontend-wasm2/src/module/function_builder_ext.rs
```diff
@@ -90,7 +89,7 @@ impl Listener for SSABuilderListener {
             debug_assert!(!is_filled, "you cannot add an instruction to a block already filled");
         }
 
-        if let Some(branch) = op.as_trait::<dyn BranchOpInterface>() {
+        if op.as_trait::<dyn BranchOpInterface>().is_some() {
```
bitwalker, line 92:
```suggestion
        if op.implements::<dyn BranchOpInterface>() {
```

frontend-wasm2/src/translation_utils.rs
```diff
@@ -147,3 +150,21 @@ pub fn sig_from_func_type(
         visibility,
     }
 }
+
+/// Parse `builtin::ComponentId` from the Wasm CM component name in the
+/// `namespace:package/interface@version` format
+pub fn parse_component_id(wasm_component_name: String) -> builtin::ComponentId {
```
bitwalker, line 156: This function can probably just be replaced with `wasm_component_name.parse::<builtin::ComponentId>().unwrap()`, since `ComponentId` implements `FromStr`.

hir2/src/ir/context.rs
```diff
@@ -276,6 +276,35 @@ impl Context {
         }
     }
 
+    /// Changes the destination of the `branch_inst` instruction to `new_dest` if the current
+    /// destination is `old_dest`.
+    ///
+    /// NOTE: Panics if `branch_inst` is not a branch instruction.
+    pub fn change_branch_destination(
```
bitwalker, line 283: I don't think we should attach this to `Context`. I would make this a new trait method of `BranchOpInterface` instead.

Additionally, we need to ensure that `old_dest` and `new_dest` have the same number of successor operands, otherwise the resulting IR will be invalid.

hir2/src/ir/context.rs
```diff
@@ -276,6 +276,35 @@ impl Context {
...
+        new_dest: BlockRef,
+    ) {
+        let mut borrow = branch_inst.borrow_mut();
+        let op = borrow.as_mut().as_operation_mut();
+        assert!(
+            op.as_trait::<dyn BranchOpInterface>().is_some(),
+            "expected branch instruction, got {branch_inst:?}"
+        );
+
+        let dest_operand_groups: Vec<usize> = op
```
bitwalker, line 296: *Nit:* In general I would like to avoid heap allocations as much as possible in the core of the compiler. In pretty much every situation where you'd reach for `Vec`, `SmallVec` can be substituted, with the choice of size picked based on what is likely to be sufficient for most scenarios (i.e. it isn't important to make the `SmallVec` large enough to handle all inputs on the stack, just avoid allocations on average). Similarly, `String` can be replaced with `CompactString` for the same sorts of uses (for short strings anyway, for anything longish, just use `String`).

hir2/src/ir/context.rs
```diff
@@ -276,6 +276,35 @@ impl Context {
...
+            .successors()
+            .iter()
+            .filter(|succ| succ.block.borrow().successor() == old_dest)
+            .map(|succ| succ.operand_group as usize)
+            .collect();
+        for dest_group in dest_operand_groups {
```
bitwalker, line 302: The implementation here is actually incorrect in general, operand group indices and successor indices are not guaranteed to be the same (and in fact, I don't think they ever are, as the first operand group is generally reserved for non-successor operands of an operation).

What you want instead is simply:

```rust
for successor_info in op.successors_mut().iter_mut() {
    if successor_info.successor() == old_dest {
        assert_eq!(old_dest.borrow().num_arguments(), new_dest.borrow().num_arguments());
        successor_info.block.borrow_mut().set(new_dest);
    }
}
```

I've been meaning to add a `fn successor_iter_mut(&mut self) -> impl DoubleEndedIterator<Item = OpSuccessorMut<'_>> + '_` helper to `Operation` to make this even nicer, but just haven't gotten around to it yet, but the above snippet avoids constructing the `EntityRange` for the successor operands, since those aren't needed here.

hir2/src/transforms/canonicalization.rs
```diff
@@ -100,7 +100,7 @@ impl Pass for Canonicalizer {
             return Ok(());
         };
         let op = {
-            log::debug!("applying canonicalization to {}", &*op);
+            log::debug!("applying canonicalization to {}", op.name());
```
bitwalker, line 103: *Nit:* If this is to avoid excess noise in the tracing output, I would add another `log::trace!` right below that prints the full op, as that is often quite useful in tracing output at certain stages.

Additionally, I've been trying to specify a log target for all of these `log` invocations, e.g. `log::trace!(target: "canonicalize", "...")`. Something granular enough to be able to narrow trace output down to a specific area of compilation, but not so narrow that you can't see everything related to, say, a specific pass or analysis.

To give you a sense of how I've been going at it so far, off the top of my head, some of the targets that exist now are:

* `driver`
* `codegen`
* `cfg-to-scf`, `spills`, and a few other pass/analysis-specific targets
* `dataflow-solver`
* `eval`

bitwalker (changes requested):
Looks good! Other than rebasing on `next`, one or two comments here are things that I'd like to fix up before merging, but overall great work!

dialects/hir/src/transforms/spill.rs
```diff
@@ -37,6 +37,11 @@ impl Pass for TransformSpills {
         // We need to drop our mutable reference while computing the analysis results
         let function = unsafe { FunctionRef::from_raw(&*op) };
         drop(op);
+        if function.borrow().body().is_empty() {
```
greenhat, line 40: @bitwalker Is this the correct method and place for skipping function declarations?

bitwalker: I'd probably use `is_declaration` when possible, which is a method on the `Symbol` trait, and since you have a `FunctionRef` there, all you need is to have the trait in scope. Internally though it just checks if the body region is empty, and more generally (i.e. with non-Symbol operations as well) that's what you'd have to do.

frontend-wasm2/src/code_translator/mod.rs
greenhat: Yep, it's gone after rebase. I removed `.ins()` in the new code.

codegen/masm2/src/lib.rs
greenhat: This file was deleted in your branch, so don't see after rebase. Since the tests are not failing I assume you also added them in `codegen/masm/src/lib.rs`, so we should be fine.

frontend-wasm2/src/component/build_ir.rs
greenhat: Yep. Thanks. Done.

frontend-wasm2/src/component/lift_exports.rs
greenhat: Done.

frontend-wasm2/src/component/lift_exports.rs
greenhat: Nice! Thank you! Fixed.

frontend-wasm2/src/module/function_builder_ext.rs
greenhat: Nice! Thanks! Done.

frontend-wasm2/src/translation_utils.rs
greenhat: Thanks! I removed the function.

hir2/src/transforms/canonicalization.rs
greenhat: > _Nit:_ If this is to avoid excess noise in the tracing output, I would add another `log::trace!` right below that prints the full op, as that is often quite useful in tracing output at certain stages.

IIRC, `&*op` caused an aliasing error. I tested it now and `&*op` works fine, so I restored it.

> Additionally, I've been trying to specify a log target for all of these `log` invocations, e.g. `log::trace!(target: "canonicalize", "...")`. Something granular enough to be able to narrow trace output down to a specific area of compilation, but not so narrow that you can't see everything related to, say, a specific pass or analysis.
> 
> To give you a sense of how I've been going at it so far, off the top of my head, some of the targets that exist now are:
> 
> * `driver`
> * `codegen`
> * `cfg-to-scf`, `spills`, and a few other pass/analysis-specific targets
> * `dataflow-solver`
> * `eval`

Sounds great! Will do as well. I enable tracing very often lately.

hir2/src/ir/context.rs
greenhat: Yep, it's much nicer. Thanks! Done.

hir2/src/ir/context.rs
greenhat: Got it. Will do.

hir2/src/ir/context.rs
greenhat: Thank you! Done.

dialects/hir/src/transforms/spill.rs
greenhat: Nice! Thanks! Done.

greenhat:
Thank you for a thorough review! I addressed all the comments. Please do another round.

codegen/masm2/src/lib.rs
bitwalker: Yeah the `masm2` crate was moved to `masm` (replacing the old `masm` contents)

bitwalker (approved):

PR #419, opened 2025-03-04
fix: collatz example readme

partylikeits1983:
Fixing the reference to "fibonacci" in the collatz example readme


greenhat (approved):
Thank you for spotting this!

PR #420, opened 2025-03-04
feat: add is-prime example

partylikeits1983:
Adding a basic example of computing wether or not an integer is prime or not. I tested the function with values from this website: https://www.math.utah.edu/~pa/MDS/primes.html

For comparison, I tried running SP1's example for proving if the number $$29$$ is prime or not. `top` said my CPU usage was 1000%. After a few minutes of the program running and the fans turning on my MBP M2, I decided to kill the program. 

Using the Miden compiler, proving that the number $$2147482583$$ is prime is near instant on the same computer.


examples/is-prime/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+```bash
+cargo miden build --release
+```
+
+## Run
+
+```bash
+midenc run target/miden/release/is_prime.masp --inputs inputs.toml
+```
```
bobbinth, line 13: This is just executes the program (rather than execute & prove), right?

greenhat: Yes, `midenc run` only executes the program and does not generate a proof. Since comparing our performance to other VMs would be a frequent case, should we add a command to `midenc` to execute and generate a proof?

examples/is-prime/src/lib.rs
```diff
@@ -0,0 +1,35 @@
+#![no_std]
+
+#[cfg(not(test))]
+#[panic_handler]
+fn my_panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
+}
+
+/// Returns 1 if integer is prime
+#[no_mangle]
```
greenhat, line 10:
```suggestion
```

`#[no_mangle]` is only needed for the `entrypoint` function because we are tracking it during the compilation.

examples/is-prime/README.md
bobbinth: We could do this, but I wonder if it may be better to just add support for working with `.masp` files in `miden-vm`. It should be simple enough and wouldn't require re-implementing a bunch of other functionality here (e.g., `analyze`, `debug`, etc. commands).

greenhat (changes requested):
Looking great! Thank you very much! Just a minor nit.

examples/is-prime/README.md
bitwalker: Yeah, my preference would be to support Miden packages in the VM proper, the need for `midenc run` is basically one of temporary necessity.

bitwalker (approved):

PR #432, opened 2025-03-10
feature(sdk): Low-level account storage API

### LINKED ISSUE #430: Low-level account storage API in Miden SDK
### Implementation details

The tx kernel account functions definitions in Miden SDK are in https://github.com/0xPolygonMiden/compiler/blob/greenhat/i363-hir2-wasm-cm/sdk/base-sys/src/bindings/account.rs

For the first iteration, use functions consistent with transaction kernel's:

- `set_storage_item` / `get_storage_item` for getting/setting value slots.
- `set_storage_map_item` / `get_storage_map_item` for reading, updating storage maps.

Their implementation should transform the parameters and result to/from Miden ABI. The transaction kernel implementation is at https://github.com/0xPolygonMiden/miden-base/blob/main/crates/miden-lib/asm/miden/account.masm#L140-L266 (ALL_CAPS means the value has type Word).

Also, if we could encapsulate them in a module, that may be slightly better ergonomics. For example, `storage::get_item`, `storage::set_item`, `storage::get_map_item`, `storage::set_map_item`.j

The Wasm CM names should then be added to the existing tx kernel account functions signature definitions in https://github.com/0xPolygonMiden/compiler/blob/greenhat/i363-hir2-wasm-cm/frontend-wasm2/src/miden_abi/tx_kernel/account.rs along with their WIT counterparts in https://github.com/0xPolygonMiden/compiler/blob/greenhat/i363-hir2-wasm-cm/tests/rust-apps-wasm/rust-sdk/wit-sdk/miden-core-import.wit and transformation strategy assigned in https://github.com/0xPolygonMiden/compiler/blob/greenhat/i363-hir2-wasm-cm/frontend-wasm2/src/miden_abi/transform.rs
For the `get*` functions, since the MASM code returns `Word`, we should use the `ReturnViaPointer` ABI transformation strategy.

### Testing

Assemble a project example in `examples` folder and add the integration test in https://github.com/0xPolygonMiden/compiler/blob/greenhat/i363-hir2-wasm-cm/tests/integration/src/rust_masm_tests/examples.rs
The test project should use the structure similar to the cross_ctx_account project used in https://github.com/0xPolygonMiden/compiler/blob/greenhat/i363-hir2-wasm-cm/tests/integration/src/rust_masm_tests/rust_sdk.rs (i.e. use Miden SDK, WIT files, and built via `rust_source_cargo_miden`).
The goal is to compile MASM and check the generated ABI transformation code.

greenhat:
Close #430 

**This PR is stacked on the #449 and should be merged after it**

This PR adds the low-level account storage API - thin wrapper around tx kernel functions.


bitwalker (approved):
LGTM!

greenhat:
@bitwalker I've rebased this branch on #449, and it's ready to merge.

tests/integration/src/rust_masm_tests/examples.rs
```diff
@@ -9,6 +9,21 @@ use proptest::prelude::*;
...
 use crate::{cargo_proj::project, CompilerTest, CompilerTestBuilder};
 
+#[test]
+fn storage_example() {
+    let config = WasmTranslationConfig::default();
+    let mut test =
+        CompilerTest::rust_source_cargo_miden("../../examples/storage-example", config, []);
+
+    test.expect_wasm(expect_file!["../../expected/examples/storage_example.wat"]);
+    test.expect_ir_unoptimized(expect_file![
```
bitwalker, line 19: Don't we generally want to be producing expects against only the optimized IR? My understanding is that we are only really interested in the unoptimized IR when there is an issue that might have its origin in the Wasm -> HIR translation, as otherwise tests could get pretty expensive compiling the same program twice.

I'd like to implement deep-cloning of IR entities, not only for use in transformations, but so that we could capture the IR at different stages of the pipeline, similar to how we capture the IR just prior to assembly for our "default" IR expect tests. Until that's implemented though, capturing the IR at different stages in a frozen state isn't possible, as the underlying data is being rewritten as it goes through the pipeline.

Anyway, wanted to double check whether or not this was here for testing, or if you planned to introduce expects for both unoptimized and optimized IR.

bitwalker (approved):
LGTM! Just one question and we can merge

PR #437, opened 2025-03-18
test: add unoptimized IR eval test, `is_prime_reduce` example

greenhat:
This PR:
- add unoptimized IR (straight from the frontend) eval check (green) for `is_prime` integration test;
- add `is_prime_reduce` example and integration test with the reduced code that still manifest the issue;


greenhat:
@bitwalker I also find it helpful to comment out https://github.com/0xPolygonMiden/compiler/blob/greenhat/switch-to-hir2-is-prime-reduce/midenc-compile/src/stages/rewrite.rs#L66-L69 to minimize HIR since the issue is in `LiftControlFlowToSCF`.

bitwalker:
I'm closing this now that we've identified and fixed the bugs that led to the is_prime test failing, and the main thing in this PR that was useful was the addition of the unoptimized compilation, but you've added that in #432 as well, so I think that can supercede this one. I'd rather not add the reducer to our `examples` though, but since it's just a reduced form of `is_prime`, it's redundant anyway.

PR #449, opened 2025-03-21
chore: update rust toolchain, dependencies

bitwalker:
* Updated to most recent nightly
* Removed unused dependencies in many places
* Removed uses of std features by default in many places
* Removed uses of dependencies which are re-exported by midenc-hir, in downstream crates that depend on midenc-hir already.
* Updated to latest wasmparser, which required some updates in midenc-frontend-wasm
* Removed unnecessary dependency on compiler crate in examples


Makefile.toml
```diff
@@ -191,6 +191,51 @@ description = "Clean build artifacts"
...
+    { name = [
+        "audit",
+    ] },
+]
+
+[tasks.unused]
+category = "Dependencies"
+description = "Check for unused dependencies"
+command = "cargo"
+args = ["machete", "--with-metadata"]
```
greenhat, line 225: Nice! I was wondering if we should run it as a CI job.

greenhat (approved):
Looking great!

Makefile.toml
bitwalker: Probably a good idea! I've held off for the time being as it can produce some false positives depending on build profile, but we could add it to CI and just see if it ends up being annoying or useful!

PR #452, opened 2025-03-24
fix: missing wasmparser feature causing cargo build failure

bitwalker:
This fixes the issue `cargo make build` failing, and a couple related fixes found in the process.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -165,7 +165,7 @@ args = [
     "build",
     "-p",
     "midenc",
-    "--out-dir",
+    "--artifact-dir",
     "${MIDENC_BIN_DIR}",
 ]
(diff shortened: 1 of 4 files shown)


partylikeits1983:
The command `cargo make build` now results in a successful build. However, the commands: 

```
cargo install --path midenc --locked
```

and 

```
cargo install --path tools/cargo-miden --locked
```

still fail to build. 


Does `cargo make build` install `cargo miden` and `midenc`?

greenhat (approved):

bitwalker:
@partylikeits1983 It currently only builds `midenc`, it doesn't install it - `cargo make install` does that. We could extend it to install both though!

Issue #433, opened 2025-03-12
`ApplyRewritesStage` removes too much code in `storage-example` example


greenhat:
Discovered while in #432 

Comparison highlights:

- `test_storage*` bodies stripped of the calls to the imported functions;
- synthetic functions for imports stripped of almost all the code;

Rust code at https://github.com/0xPolygonMiden/compiler/blob/greenhat/i430-acc-storage-low/examples/storage-example/src/lib.rs#L29-L43

HIR compiled **without** `ApplyRewritesStage` at https://github.com/0xPolygonMiden/compiler/blob/greenhat/i430-acc-storage-low/tests/integration/expected/examples/storage_example_unoptimized.hir

HIR compiled **with** `ApplyRewritesStage` at https://github.com/0xPolygonMiden/compiler/blob/greenhat/i430-acc-storage-low/tests/integration/expected/examples/storage_example.hir

bitwalker:
I'll take a look at this in the morning

bitwalker:
Fixed in #454

PR #454, opened 2025-03-25
fix: bug in would_be_trivially_dead_even_if_terminator

bitwalker:
This PR fixes the issue reported in #433.

In the process of troubleshooting this and fixing the bug, I also:

* Added some missing log targets
* Added a utility pass which can be added to the pipeline to print any operation, optionally with a filter applied. Comes in handy when trying to see the IR output for a specific function. Longer term, I need to implement the IR printing functionality of the `PassManager` itself, but for now this works in a pinch.

There is a missing lowering for `builtin.global_symbol` that causes the storage example to fail to compile all the way to MASM, but I'll fix that in a subsequent PR.

I had to disable the `cross-ctx-account` test due to an error `undefined module 'intrinsics::mem'`, which I haven't looked into yet, but since it was uncovered by this fix, my guess is that it is a pre-existing issue.


tests/integration/src/rust_masm_tests/rust_sdk.rs
```diff
@@ -127,6 +127,7 @@ fn rust_sdk_p2id_note_script() {
 }
 
 #[test]
+#[ignore = "until undefined module 'intrinsics::mem' issue is fixed"]
```
greenhat, line 130: Should we create an issue for this?

tests/integration/src/rust_masm_tests/abi_transform/stdlib.rs
```diff
@@ -17,6 +17,7 @@ use proptest::{
 use crate::CompilerTest;
 
 #[test]
+#[ignore = "until builtin.global_symbol is implemented"]
```
greenhat, line 20: Should we create an issue for this?

greenhat (approved):
Looking great! Just a couple of questions.

tests/integration/src/rust_masm_tests/rust_sdk.rs
bitwalker: I'm looking into both this and the other issue today, and once I know more about where the error comes from, I'll open an issue for it. Right now it's unclear if it's a bug, or something actually missing somewhere.

tests/integration/src/rust_masm_tests/abi_transform/stdlib.rs
bitwalker: Nah, I'm already working on this one, and will have a PR open shortly

PR #458, opened 2025-03-26
fix: missing lowering for global_symbol, 'undefined module' error during assembly

bitwalker:
@greenhat This fixes the two issues discovered in #454 

I ended up also doing the `FunctionIdent` to `SymbolPath` refactoring, since most of the uses were in the handling of intrinsics/imports anyway, and I that code anyway as part of fixing the "undefined module" issue.

See the individual commits to review the changes separately. The one that fixes the undefined module error has some extra descriptive text that describes the changes in more detail.

I was able to re-enable the blake3 and cross-account-ctx tests with these changes.


frontend/wasm/src/callable.rs
```diff
@@ -0,0 +1,75 @@
+use midenc_hir::{dialects::builtin::FunctionRef, interner::Symbol, Signature, SymbolPath};
+
+use crate::intrinsics::Intrinsic;
+
+/// Local core Wasm module functon or processed module import to be used for the translation of the
+/// Wasm `call` op.
+#[derive(Clone)]
+pub enum CallableFunction {
```
greenhat, line 8: This is so much nicer! Thank you! 
I remember trying the enum, but I did not like the result and ditched it half-way through. It appears that I was missing `Intrinsic`.

greenhat (approved):
Looking great!

Issue #352, opened 2024-10-30
`test_blake3_hash` test is failing on the VM from the `next` branch


greenhat:
It appears that `u32assert` at https://github.com/0xPolygonMiden/compiler/blob/6cd29e17b34c5abef7f6328c33af06f8bf203344/tests/integration/expected/abi_transform_stdlib_blake3_hash.masm#L208 is blowing up.

It is reproducible in #349 branch.

Trace:

```
[tests/integration/src/rust_masm_tests/abi_transform/stdlib.rs:51:9] &ibytes = [
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
]
[tests/integration/src/rust_masm_tests/abi_transform/stdlib.rs:51:9] &frame = [
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    2,
    86016,
]
[tests/integration/src/rust_masm_tests/abi_transform/stdlib.rs:51:9] rs_out = [
    42,
    218,
    131,
    193,
    129,
    154,
    83,
    114,
    218,
    225,
    35,
    143,
    193,
    222,
    209,
    35,
    200,
    16,
    79,
    218,
    161,
    88,
    98,
    170,
    238,
    105,
    66,
    138,
    24,
    32,
    252,
    218,
]

Stack Trace:
 |-> #exec::#main in <unavailable>
 `-> entrypoint in <unavailable>:

Last 5 Instructions (of current frame):
 |   dup0
 |   dup4
 |   push(28)
 |   add
 |   pad
 `-> <error occured here>


Last 5 Instructions (any frame):
 |   dup0
 |   dup4
 |   push(28)
 |   add
 |   pad
 `-> <error occured here>


Last Known State (at most recent instruction which succeeded):
 | Frame Pointer: 1073741824 (starts at 2^30)
 | Operand Stack: [0, 11671305615285950913, 1048544, 1048544, 1048576, 15253023094703789604, 11671305615285950885, 8635338869442206704, 14327958870441829769, 4492299921045254967, 17467091117607601070, 5096858464874356363, 86018, 1310720, 1376256, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 
[...]

PR #462, opened 2025-03-28
feat: implement support for element-addressable memory

### LINKED ISSUE #374: Migrate to element-addressable memory VM
Implemented in https://github.com/0xPolygonMiden/miden-vm/pull/1598

bitwalker:
This PR does the following:

1. Refactors the code generator to emit memory operations based on a element-oriented address space, rather than word-oriented.
2. Updates our dependency on the various Miden VM crates we depend on to use the latest stable release (i.e. 0.13.1)
3. Fixes the incomplete handling of global variables and data segments in the code generator and linker. This was the reason why #352 occurred, and that issue is fixed by these changes and the test re-enabled. It would have affected any program relying on global variables/data segments though, so it was a big oversight on my part. I hadn't caught it earlier because I thought the full blake3 test was still being run, but it was actually partially commented out, and our other tests apparently don't use globals/segments, so it never caused failures elsewhere (at least, not that are caught by our tests currently).

Closes #374

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,7 +56,7 @@ dependencies = [
  "getrandom 0.2.15",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
@@ -604,7 +604,7 @@ dependencies = [
  "p256",
  "parse_arg",
  "pretty_env_logger",
- "rand_core",
+ "rand_core 0.6.4",
  "rpassword",
  "semver 1.0.26",
  "serde",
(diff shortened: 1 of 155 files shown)


greenhat (approved):
Looking good! Thank you for the swift response! 

I've only glanced through the `mem.masm` changes, though. I cannot find tests I added in https://github.com/0xPolygonMiden/compiler/pull/290 and https://github.com/0xPolygonMiden/compiler/pull/269.
It appears that we lost them during the migration.

bitwalker:
I'll look at restoring those today, not sure what happened there!

PR #468, opened 2025-03-29
test(codegen): implement initial tests for memory intrinsics

bitwalker:
This adds back the level of coverage we previously had for `load_sw` and `load_dw` intrinsics, fixes various bugs, and makes some general improvements to our testing infrastructure in preparation for future tests of this type.


greenhat (approved):
Looking great!

PR #478, opened 2025-04-03
fix: note script compilation

### LINKED ISSUE #476: Fix note script compilation in the new pipeline
```
  |-> at frontend/wasm/src/module/module_translation_state.rs:128:36
  `-> unexpected import 'miden:cross-ctx-account/foo@1.0.0/process-felt'

```

in `rust_sdk_cross_ctx_note` test

greenhat:
Close #476

TODO:

- [x] fix: prefix module path with `SymbolNameComponent::Root` for core Wasm module
- [x] fix `hir.call` missing lowering
- [x] order modules by dependency when passing to the Assembler;
- [x] treat the entrypoint as an absolute path
- [x] parse version in `ComponentId` parsing
- [ ] assertion failed on execution at cycle 389 on `u32lt`

### DIFF
--- a/codegen/masm/src/lower/component.rs
+++ b/codegen/masm/src/lower/component.rs
@@ -45,7 +45,18 @@ impl ToMasmComponent for builtin::Component {
                 let name = masm::ProcedureName::from_raw_parts(masm::Ident::from_raw_parts(
                     Span::new(entry_id.function.span, entry_id.function.as_str().into()),
                 ));
-                let path = LibraryPath::new(entry_id.module).into_diagnostic()?;
+
+                let path = if component_path.to_string() == "root_ns:root@1.0.0" {
+                    // We're inside the synthethic "wrapping" component for pure Rust program
+                    // compilation. Since the user does not know about it their entrypoint does not
+                    // include the synthetic component id. Append the user-provided entrypoint
+                    // module
+                    component_path.clone().append_unchecked(entry_id.module)
+                } else {
+                    // We're compiling a Wasm component and the component id is included
+                    // in the entrypoint.
+                    LibraryPath::new(entry_id.module).into_diagnostic()?
+                };
                 Some(masm::InvocationTarget::AbsoluteProcedurePath { name, path })
             }
             None => None,
--- a/codegen/masm/src/lower/component.rs
+++ b/codegen/masm/src/lower/component.rs
@@ -1,7 +1,9 @@
 use alloc::{collections::BTreeSet, sync::Arc};
 
+use miden_assembly::LibraryPath;
 use midenc_hir::{
-    dialects::builtin, pass::AnalysisManager, FunctionIdent, Op, SourceSpan, Span, Symbol, ValueRef,
+    diagnostics::IntoDiagnostic, dialects::builtin, pass::AnalysisManager, FunctionIdent, Op,
+    SourceSpan, Span, Symbol, ValueRef,
 };
 use midenc_hir_analysis::analyses::LivenessAnalysis;
 use midenc_session::diagnostics::{Report, Spanned};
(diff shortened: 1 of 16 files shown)


frontend/wasm/src/component/translator.rs
```diff
@@ -397,9 +397,10 @@ impl<'a> ComponentTranslator<'a> {
                     }
 
                     let module_path = SymbolPath {
-                        path: smallvec![SymbolNameComponent::Component(Symbol::intern(
-                            *arg_module_name
-                        ))],
+                        path: smallvec![
+                            SymbolNameComponent::Root,
```
bitwalker, line 401: Just to clarify: this is bound as `module_path`, but the path would resolve at the top-level of the world, and AFAIK pretty much everything is nested within a component (with the exception of MASM modules which we're planning to represent without the outer `Component`). Is this actually referring to the component path, or the core Wasm module that goes inside it?

greenhat: It refers to the component path. The `module_path` is `miden:cross-ctx-account/foo@1.0.0`, which looks like an absolute path, and is later accessed at https://github.com/0xMiden/compiler/blob/greenhat/i476-note-script-compilation/frontend/wasm/src/module/module_translation_state.rs#L119-L128 where it's expected to be root. The Wasm CM uses this path along with the function name as the key to match module instantiation arguments (via `canon lower`) to module imports. The core Wasm module import is parsed in WAT - https://github.com/0xMiden/compiler/blob/greenhat/i476-note-script-compilation/tests/integration/expected/rust_sdk/cross_ctx_note.wat#L45-L45

hir/src/dialects/builtin/ops/component/interface.rs
```diff
@@ -69,18 +72,25 @@ impl TryFrom<&SymbolPath> for ComponentId {
...
             None => return Err(InvalidComponentIdError::MissingNamespace),
             Some(name) => match name.as_str().split_once(':') {
-                Some((ns, name)) => (SymbolName::intern(ns), SymbolName::intern(name)),
+                Some((ns, name)) => match name.split_once('@') {
+                    Some((name, version)) => (
+                        SymbolName::intern(ns),
+                        SymbolName::intern(name),
+                        Version::parse(version).map_err(InvalidComponentIdError::InvalidVersion)?,
+                    ),
+                    None => return Err(InvalidComponentIdError::MissingVersion),
```
bitwalker, line 84: Is this truly an error? My recollection of the Component Model spec is that the version is supposed to be optional AFAIK

codegen/masm/src/lower/component.rs
```diff
@@ -43,7 +45,18 @@ impl ToMasmComponent for builtin::Component {
                 let name = masm::ProcedureName::from_raw_parts(masm::Ident::from_raw_parts(
                     Span::new(entry_id.function.span, entry_id.function.as_str().into()),
                 ));
-                let path = component_path.clone().append_unchecked(entry_id.module);
+
+                let path = if component_path.to_string() == "root_ns:root@1.0.0" {
```
greenhat, line 49: @bitwalker I fixed the "undefined module" error with this quick hack, but I feel there should be a better way. Maybe mark the synthetic component with an attribute so we can recognize it without matching the ID?

hir/src/dialects/builtin/ops/component/interface.rs
greenhat: Good catch! I've been staring at versions for so long, I forgot that it's optional. 
Should we make `ComponentId::version` optional? I don't like the idea of putting a hardcoded default version if it's not specified.

bitwalker: I'll check and see what the spec indicates an elided version implies.

bitwalker (approved):

hir/src/dialects/builtin/ops/component/interface.rs
bitwalker: According to the Component Model spec, the version is definitely optional, but it does not specify what a missing version implies. My guess is that it just means any version will satisfy it. As for what we should do with it, I'm not sure. Do you know if we always observe versioned paths in the compiled Wasm binary? If so, maybe we just require a version as you've sketched out here. Not sure if there is a good reason to allow paths with an unspecified version. I'll punt on this until we have a specific issue come up.

greenhat: So far, we have __only__ versioned paths in the compiled binaries. AFAIR, I could not use unversioned paths in WIT. 
I'd require a version for now, but since we're going to use digests for our dependencies, versions seem redundant, so I'd explore if there is a possibility to ditch them.

PR #493, opened 2025-04-28
docs: update README link to correct repository path

Elena343-ai:
This PR updates the documentation reference link to point to the correct repository path.

Changes made:
- Updated the broken link from `https://github.com/0xPolygonMiden/compiler/frontend-wasm/README.md`
- To the correct path: `https://github.com/0xMiden/compiler/blob/next/frontend/wasm/README.md`

This ensures users can properly access the documentation about unsupported Wasm core types, instructions and features.

### DIFF
--- a/docs/src/design/frontends.md
+++ b/docs/src/design/frontends.md
@@ -5,4 +5,4 @@
 TODO
 
 For the list of the unsupported Wasm core types, instructions and features, see the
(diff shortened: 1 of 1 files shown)


bitwalker:
Thanks for the PR, but we do not currently accept trivial documentation changes like this due to spam. I'll make sure these issues are addressed though! In the future, feel free to mix in changes like this with more substantive changes, even if not necessarily related, we will still accept them.

PR #506, opened 2025-05-15
fix: install tasks for cargo-miden, midenc

bitwalker:
This fixes an issue with `cargo make install-midenc`, as well as implements `cargo make install-cargo-miden`.


greenhat (approved):
LGTM!

PR #507, opened 2025-05-15
fix(codegen): emit trace events to fix stack traces in debugger

bitwalker:
This reinstates call frame tracking for debugging purposes, using a pair of trace events that we listen for in the debug executor in order to construct stack traces.

### DIFF
--- a/sdk/stdlib-sys/src/intrinsics/debug.rs
+++ b/sdk/stdlib-sys/src/intrinsics/debug.rs
@@ -1,0 +1,14 @@
+#[link(wasm_import_module = "miden:core-import/intrinsics-debug@1.0.0")]
+extern "C" {
+    #[link_name = "break"]
+    fn extern_break();
+}
+
+/// Sets a breakpoint in the emitted Miden Assembly at the point this function is called.
+#[inline(always)]
+#[track_caller]
+pub fn breakpoint() {
+    unsafe {
+        extern_break();
+    }
+}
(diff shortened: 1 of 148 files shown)


sdk/stdlib-sys/src/intrinsics/mod.rs
```diff
@@ -1,5 +1,6 @@
 use core::ops::{Deref, DerefMut};
 
+pub mod debug;
```
greenhat, line 3: I'm not sure that I understand why it's public.

sdk/stdlib-sys/src/intrinsics/debug.rs
```diff
@@ -0,0 +1,14 @@
+#[link(wasm_import_module = "miden:core-import/intrinsics-debug@1.0.0")]
```
greenhat, line 1: This interface with the `break` function should be in our core-import WIT file at `tests/rust-apps-wasm/rust-sdk/wit-sdk/miden-core-import.wit`, otherwise wit-component errors on core module import. I'm in the middle of splitting our WIT files and moving them to SDK crates at #505, and I can take care of it there.

greenhat (approved):
Looking good, just minor notes.

sdk/stdlib-sys/src/intrinsics/debug.rs
bitwalker: Ah noted, I'll leave the fix to you then!

sdk/stdlib-sys/src/intrinsics/mod.rs
bitwalker: Unlike the other two modules, which we're wildcard re-exporting the contents of from the root, we only need to export the `debug` module and require fully-qualified paths instead to access its declarations.

PR #508, opened 2025-05-15
feat: switch to stable vm, link against real miden-lib

bitwalker:


Cargo.toml
```diff
@@ -134,19 +135,15 @@ midenc-debug = { version = "0.0.8", path = "midenc-debug" }
...
-    "parsing",
-    "derive",
-    "extra-traits",
-    "printing",
-] }
-proc-macro2 = "1.0"
-quote = "1.0"
 
-semver = { version = "1.0", default-features = false }
+[patch.crates-io]
```
greenhat, line 139: Nice!

midenc-session/src/libs.rs
```diff
@@ -23,8 +22,6 @@ use crate::{
 
 pub static STDLIB: LazyLock<Arc<CompiledLibrary>> =
```
greenhat, line 23: We should probably ditch `STDLIB` in favor of `StdLibrary::default().as_ref().clone()` as well for consistency.

greenhat (approved):
Looking good!

PR #511, opened 2025-05-20
feature(cargo-miden): add new project template for note script

### LINKED ISSUE #475: New project template for note script in `cargo-miden`
Add a note script Rust project template and add the `--note-script` option to `cargo miden new` command to create a new note script project

greenhat:
Close #475 

This PR switches to tag `v0.10.0`(set in https://github.com/0xMiden/rust-templates/pull/12) to fetch new project templates, adds note script test for new project template test.

### DIFF
--- a/docs/src/usage/cargo-miden.md
+++ b/docs/src/usage/cargo-miden.md
@@ -11,7 +11,7 @@ a template to spin up a new Miden project in Rust, and takes care of orchestrati
 > make sure you have it installed first:
 >
 > ```bash
-> rustup toolchain install nightly-2025-01-16
+> rustup toolchain install nightly-2025-03-20
 > ```
 >
 > NOTE: You can also use the latest nightly, but the specific nightly shown here is known to
--- a/docs/src/usage/midenc.md
+++ b/docs/src/usage/midenc.md
@@ -17,7 +17,7 @@ hands dirty.
 > make sure you have it installed first:
 >
 > ```bash
-> rustup toolchain install nightly-2025-01-16
+> rustup toolchain install nightly-2025-03-20
 > ```
 >
 > NOTE: You can also use the latest nightly, but the specific nightly shown here is known to


tools/cargo-miden/tests/build.rs
```diff
@@ -81,6 +85,20 @@ fn build_new_project_from_template(template: &str) -> Package {
     let restore_dir = env::current_dir().unwrap();
     let temp_dir = env::temp_dir();
     env::set_current_dir(&temp_dir).unwrap();
+
+    if template == "--note" {
+        // create the counter contract cargo project since the note depends on it
```
bitwalker, line 90: I wonder if we shouldn't expect that users will have already created the account contract, or indicate that in the documentation, or perhaps check for if it exists, and if not, raise an error/prompt for user confirmation. Alternatively/additionally, we could add a `--ignore-missing-account` flag that would skip the check and just create the note template non-interactively.

As an aside, I think the crates generated by `cargo miden new` should probably be more general than they currently are (i.e. they are currently generators for example projects, but that makes them less useful as a shortcut for creating a new empty crate with the boilerplate already generated for you, ala `cargo new`).

Thoughts?

bitwalker (approved):
Looks good, just one question/nit before I merge

tools/cargo-miden/tests/build.rs
bitwalker: Makes sense, though that approach is pretty much tailored towards first-time use of `cargo miden` only, rather than ongoing use once you already understand the relationship between the different crate types. Take `cargo new` for example, it generates a all-but-empty crate, but with enough boilerplate stubbed out so you can see how to expand on it, or alternatively, it is minimal enough that replacing it with your own code isn't a pain; and either way, it is useful both for first-time Rust devs, as well as seasoned devs; and it understands workspaces if that's the direction you want to go.

The downside of this implementation of `cargo miden new --note` is that it is ideally suited for first-timers, who don't yet have any understanding of how to lay out their project, and how the components interact/relate to one another. Once that understanding is gained though, suddenly `cargo miden new --note` is annoying to use, because they have to do a considerable amount of cleanup in order to use the generated crate for their own purposes.

I'm not sure what the best approach is here, but I think ultimately it would be better if `cargo miden new` defaulted to the more general form, while allowing first-timers an easy way to request that a more thorough example project be generated, maybe via a new command, e.g. `cargo miden example <example type/name>` (I'm not that opinionated about naming here). I think both forms are valuable for different reasons, but the former is going to be more common. Conflating the two types of generated crate is probably part of the issue (i.e. our "new" command generates effectively a full example, rather than just a skeleton for the given project type).

greenhat: > Makes sense, though that approach is pretty much tailored towards first-time use of `cargo miden` only, rather than ongoing use once you already understand the relationship between the different crate types. Take `cargo new` for example, it generates a all-but-empty crate, but with enough boilerplate stubbed out so you can see how to expand on it, or alternatively, it is minimal enough that replacing it with your own code isn't a pain; and either way, it is useful both for first-time Rust devs, as well as seasoned devs; and it understands workspaces if that's the direction you want to go.
> 
> The downside of this implementation of `cargo miden new --note` is that it is ideally suited for first-timers, who don't yet have any understanding of how to lay out their project, and how the components interact/relate to one another. Once that understanding is gained though, suddenly `cargo miden new --note` is annoying to use, because they have to do a considerable amount of cleanup in order to use the generated crate for their own purposes.

I agree, it's too much of cleanup if what you really want a clean slate project.

> I'm not sure what the best approach is here, but I think ultimately it would be better if `cargo miden new` defaulted to the more general form, while allowing first-timers an easy way to request that a more thorough example project be generated, maybe via a new command, e.g. `cargo miden example <example type/name>` (I'm not that opinionated about naming here). I think both forms are valuable for different reasons, but the former is going to be more common. Conflating the two types of generated crate is probably part of the issue (i.e. our "new" command generates effectively a full example, rather than just a skeleton for the given project type).

I like the `cargo miden example` approach! We can put it prominently in docs and in the `cargo miden --help` to steer the newcomers to it. I made https://github.com/0xMiden/compiler/issues/514 for gathering feedback and discussing implementation details.

PR #516, opened 2025-05-21
Remove `verify` from `OperationBuilder` and `SameOperandsAndResultType` verification implementation

### LINKED ISSUE #456: Fix/improve operation verification
Currently, the `OperationBuilder` runs verification when an operation is built. Unfortunately, because this happens before the op is inserted into a block, it means that various operation traits with associated validation rules, fail automatically. This has caused us to disable/comment out those validation rules temporarily.

After a review, we can fix this situation as follows:

* Remove the call to `op.verify()` in `OperationBuilder`
* Re-enable/uncomment all of the disabled validation rules
* Rely on the `PassManager` to run verification when so configured. This is a built-in feature of the `PassManager` implementation, so it is not necessary for us to _also_ run verification when constructing ops, especially because it presents the problem described above.

lima-limon-inc:
Closes #456

The issue states that currently `Operation::verify` is called inside the `OperationBuilder` which is called before operations are added to blocks, which could lead to failure in operation verification. Because of this, `SameTypeOperands`' verification function had been commented out.

This PR removes the call to `Operation::verify` from the `OperationBuilder` and moves it to the beggining of `PassManager::run`. `Operation::verify` was already called inside the `PassManager::run` (more precisely, inside the `OpToOpPassAdaptor::run_pipeline`). An additional `verify` call  was added at the beginning of the `PassManager` pipeline to "compensate" its removal in the OperationBuilder and ensure `Operation` integrity before applying passes. Now, the `PassManger` is responsible for `Operation` integrity verification, which is done before passes are applied and also after each pass applies its modifications. 

With the call to `verify` in the `OperationBuilder` now removed, `SameTypeOperands`' verification function was re-enabled and so was its test (`derived_op_verifier_test`).

Additionally, an implementation for `SameOperandsAndResultType`'s verification function was added (which was marked [here](https://github.com/0xMiden/compiler/blob/526167a3afc8218c3475073fe4701f928ad8bf5b/hir/src/ir/traits/types.rs#L73-L75) as a TODO item). Following the comment's requirements, `SameTypeOperands` was marked as a super-trait.

To support this, the `derive!` macro was expanded to support super-traits. The macro supports an arbitrary amount of traits; but it is important to note that the traits must be comma separated instead of the usual "+" separated format. As far as I am aware, it is not possible to use "+" as a separator inside a `macro_rules!` context.

With the super-trait now added, operations that implemented `SameOperandsAndResultType` but did not implement `SameTypeOperands` were updated to now explicitly implement both. This was done explicitly in order to replicate Rust's `#[derive]` macro's behavior of explicitly declaring all derived traits.

Finally, a unit test was added to test `SameOperandsAndResultType`'s verify function. For this, a test operation `InvalidOpsWithReturn` was added and registered in the test dialect.

Side note: I tried to keep the `SameOperandsAndResultType`'s doc comment relatively short in order for it to show up in its entirety when a compilation error shows up, it is currently displayed like so:

```
72  | / derive! {
73  | |     /// Op expects all operands and results to be of the same type.
74  | |     /// NOTE: Operations that implement this trait must also explicitly implement
75  | |     /// [`SameTypeOperands`]. This can be achieved by using the "traits" filed in the [#operation]
...   |
87  | |     pub trait SameOperandsAndResultType: SameTypeOperands {}
    | |               ------------------------- required by a bound in this trait

```

### DIFF
--- a/hir/src/dialects/test/ops/binary.rs
+++ b/hir/src/dialects/test/ops/binary.rs
@@ -74,25 +74,3 @@ impl InferTypeOpInterface for Shl {
         Ok(())
     }
 }
-
-/// Invalid operation that breaks the SameOperandsAndResultType trait (used for testing).
-#[operation(
-    dialect = TestDialect,
-    traits(BinaryOp, SameTypeOperands, SameOperandsAndResultType),
-    implements(InferTypeOpInterface)
-)]
-pub struct InvalidOpsWithReturn {
-    #[operand]
-    lhs: AnyInteger,
-    #[operand]
-    rhs: AnyInteger,
-    #[result]
-    result: AnyUnsignedInteger,
-}
-
-impl InferTypeOpInterface for InvalidOpsWithReturn {
-    fn infer_return_types(&mut self, _context: &Context) -> Result<(), Report> {
-        self.result_mut().set_type(Type::U64);
-        Ok(())
-    }
-}
--- a/hir/src/derive.rs
+++ b/hir/src/derive.rs
@@ -148,9 +148,9 @@ mod tests {
 
     use crate::{
         attributes::Overflow,
-        dialects::test::{self, Add, InvalidOpsWithReturn},
+        dialects::test::{self, Add},
         pass::{Nesting, PassManager},
-        Builder, BuilderExt, Context, Op, Operation, Report, Spanned,
+        Builder, BuilderExt, Context, Op, Operation, Report, Spanned, Value,
     };
 
     derive! {
@@ -255,12 +255,20 @@ mod tests {
         let mut builder = context.clone().builder();
         builder.set_insertion_point_to_end(block);
 
-        let op_builder = builder.create::<InvalidOpsWithReturn, _>(SourceSpan::default());
-        let op = op_builder(lhs, rhs);
-        let op = op.unwrap();
+        let op_builder = builder.create::<Add, _>(SourceSpan::default());
+        let op = op_builder(lhs, rhs, Overflow::Wrapping);
+        let mut op = op.unwrap();
+
+        // NOTE: We override the result's type in order to force the SameOperandsAndResultType
+        // verification function to trigger an error
+        {
+            let mut binding = op.borrow_mut();
+            let mut result = binding.result_mut();
+            result.set_type(Type::U64);
+        }
 
         // Construct a pass manager with the default pass pipeline
-        let mut pm = PassManager::on::<InvalidOpsWithReturn>(context.clone(), Nesting::Implicit);
+        let mut pm = PassManager::on::<Add>(context.clone(), Nesting::Implicit);
         // Run pass pipeline
         pm.run(op.as_operation_ref()).unwrap();
     }
--- a/hir/src/dialects/test.rs
+++ b/hir/src/dialects/test.rs
@@ -115,7 +115,6 @@ impl DialectRegistration for TestDialect {
 
     fn register_operations(info: &mut DialectInfo) {
         info.register_operation::<ops::Add>();
-        info.register_operation::<ops::InvalidOpsWithReturn>();
         info.register_operation::<ops::Mul>();
         info.register_operation::<ops::Shl>();
         info.register_operation::<ops::Ret>();


lima-limon-inc:
Update: The tests are now failing because `cfg_to_scf_lift_conditional_early_exit` builds a `Neg` operation with a `i1` and `u32` as operands. So `SameTypeOperands`' verify function fails (since it was enabled in the PR).

I'm working on a fix

dialects/scf/src/transforms/cfg_to_scf.rs
```diff
@@ -524,7 +524,8 @@ mod tests {
         builder.cond_br(v331, block35, [], block36, [], span)?;
 
         builder.switch_to_block(block34);
-        let v345 = builder.eq(v343, v330, span)?;
+        let mut v345 = builder.eq(v343, v330, span)?;
+        v345.borrow_mut().set_type(Type::U32);
```
bitwalker: Overriding the type like this manually is incorrect. To cast a type you'll want to do the following:

```rust
let v345 = builder.eq(v343, v330, span)?;
let v346 = builder.zext(v345, Type::U32, span)?;
```

This performs a semantically-correct cast from the `i1` type to `u32` by zero-extending the extra bits. For other types of casts, there are also `sext`, `trunc`, `bitcast`, `inttoptr`, and `ptrtoint` - each of which has unique semantics that get lowered differently, so in general you want to pay attention to things such as whether the cast is from a larger bitwidth to a smaller one, or vice versa (`trunc` vs `zext`/`sext`); signedness (`zext` vs `sext`); and whether the cast is between integer and pointer types (`inttoptr` vs `ptrtoint`). The `bitcast` operation is an optimization for casts between integral types of the same bitwidth where we just want to reinterpret the bits as the target type. Bitcasts are also used for casting pointer types between different pointee types. Bitcasts are effectively no-ops during code generation.

Hope that helps!

hir/src/dialects/test/ops/binary.rs
```diff
@@ -74,3 +74,25 @@ impl InferTypeOpInterface for Shl {
...
     }
 }
+
+/// Invalid operation that breaks the SameOperandsAndResultType trait (used for testing).
+#[operation(
+    dialect = TestDialect,
+    traits(BinaryOp, SameTypeOperands, SameOperandsAndResultType),
+    implements(InferTypeOpInterface)
+)]
+pub struct InvalidOpsWithReturn {
```
bitwalker: You shouldn't need to define a new operation type to test the verifier. I would instead simply construct an `Add` operation with mismatched operand types to trigger a verification error.

Since this operation implements `InferTypeOpInterface` (and `Add` does as well), by definition the result type will be inferred to match the operand type, so we're really just testing that verification fails if the operand types are mismatched, and that's fine for the purpose of the test anyway.

hir/src/ir/traits/types.rs
```diff
@@ -68,12 +68,67 @@ derive! {
...
+    /// )]
+    /// pub struct SomeOp {
+    ///       (...)
+    /// }
+    pub trait SameOperandsAndResultType: SameTypeOperands {}
+
+    verify {
+        fn operands_and_result_are_the_same_type(op: &Operation, context: &Context) -> Result<(), Report> {
+            let mut operands = op.operands().iter();
+            if let Some(first_operand) = operands.next() {
```
bitwalker: I think we should panic if the operation has no operands but derives `SameTypeOperands` (and by extension `SameOperandsAndResultType`), since the trait is meaningless for an operation with no operands. I would prefer to panic in that situation vs raise a compiler diagnostic, since it represents an internal compiler error that can't meaningfully be fixed without modifying the operation definition.

bitwalker (changes requested):
Nice work! I've noted a couple of small changes to make, but otherwise it looks good to me!

dialects/scf/src/transforms/cfg_to_scf.rs
lima-limon-inc: Thank you very much for the explanation bitwalker. I am taking note of those casting functions for future PRs.

Changed here: https://github.com/0xMiden/compiler/pull/516/commits/6069a1c379ceb43f91e2d100a814499c170fd3d6

hir/src/ir/traits/types.rs
lima-limon-inc: Great point! I changed this behavior in this commit: https://github.com/0xMiden/compiler/pull/516/commits/6069a1c379ceb43f91e2d100a814499c170fd3d6

On a related note, in this commit https://github.com/0xMiden/compiler/pull/516/commits/72d6df2ce20703b8a4cd179dd0296227c44976bf I added a similar check for operations marked with the `SameOperandsAndResultType` which are missing a `result`. 
What do you think?

hir/src/ir/traits/types.rs
```diff
@@ -68,12 +68,71 @@ derive! {
...
+            let first_operand = operands.next().unwrap_or_else(|| panic!("Operation: {} was marked as having SameOperandsAndResultType, however it has no operands.",
+                                                                         op.name()));
+            let (expected_ty, set_by) = {
+                let operand = first_operand.borrow();
+                let value = operand.value();
+                (value.ty().clone(), value.span())
+            };
+
+            let results = op.results();
+            if results.is_empty() {
```
bitwalker: This can be simplified to:

```rust
assert!(!results.is_empty(), "...", <format args..>);
```

hir/src/dialects/test/ops/binary.rs
lima-limon-inc: > You shouldn't need to define a new operation type to test the verifier. I would instead simply construct an `Add` operation with mismatched operand types to trigger a verification error.

Question, which verifier are you referring to?  Wouldn't constructing an `Add` operation with mismatched operand types trigger `SameTypeOperands`'s verification function to error instead of `SameOperandsAndResultType`'s? 

If I'm not mistaken, construction of an `Add` op with mismatched operand types is being tested in the [`derived_op_verifier_test`](https://github.com/0xMiden/compiler/pull/516/files#diff-5ada58d2a6bf0fefdaeffee3ea7d74478d2db0042279f810afd1b27d0b466ad5R208-R224) test,  note [these](https://github.com/0xMiden/compiler/pull/516/files#diff-5ada58d2a6bf0fefdaeffee3ea7d74478d2db0042279f810afd1b27d0b466ad5R214) lines.

The initial idea behind `InvalidOpsWithReturn` was for it to test `SameOperandsAndResultType` verification function, hence the new test added [here](https://github.com/0xMiden/compiler/pull/516/files#diff-5ada58d2a6bf0fefdaeffee3ea7d74478d2db0042279f810afd1b27d0b466ad5R239-R265).  

> Since this operation implements `InferTypeOpInterface` (and `Add` does as well), by definition the result type will be inferred to match the operand type, so we're really just testing that verification fails if the operand types are mismatched, and that's fine for the purpose of the test anyway.

I hardcoded `InferTypeOpInterface` implementation for `InvalidOpsWithReturn` so it always returned `U64` (see [here](https://github.com/0xMiden/compiler/pull/516/files#diff-3aa7b16f6e1632367508d000636be2f5eb1ec8189fd612a578cd09e39f27f7deR93-R98)). It was the only way I managed to make `InvalidOpsWithReturn` build correctly, but fail during trait validation. Is there a mechanism that I'm missing? As an aside if I remove the `InferTypeOpInterface` trait from the operation, then it fails to build the op, stating:  `invalid operation`

Do note that if this test is contrived, then it can be removed all together.   

Thanks in advance and apologies if I'm missing something

hir/src/ir/traits/types.rs
lima-limon-inc: Great catch, changed in this commit: https://github.com/0xMiden/compiler/pull/516/commits/9d4bc0e81383ddb188b6b81eb4cc5a0398bd2815

hir/src/ir/traits/types.rs
bitwalker: Good catch!

hir/src/dialects/test/ops/binary.rs
bitwalker: So in general, `SameOperandsAndResultType`, pretty much implies `InferTypeOpInterface` (not literally, but if you are defining an op whose result type _must_ be the same as the operands, then implementing `InferTypeOpInterface` is trivial, and would guarantee that the result always matches the operand type from which it is inferred.

With that in mind, testing `SameOperandsAndResultType` pretty much requires doing so with an op that doesn't implement `InferTypeOpInterface` (which would be weird, but is allowed), or overriding the result type _after_ inference. As you found, all of our current test ops that implement `SameOperandsAndResultType` also implement `InferTypeOpInterface` (and that's the case with all of our other dialects as well), so I think the best way to test the verifier here is to simply override the result type after inference (e.g. with `Add`), just to force invalid IR to trip the verifier.

lima-limon-inc: Great! Thanks for the clarification.

With this in mind, what do you think about this refactor: https://github.com/0xMiden/compiler/commit/10728a747373c8fbbad6f5fae43d71c03dbbcbba

I think I do have to use the `set_type` method in this case, right?  I tried to do a proper cast via `zext`/`sext`, but I was not able to use that method. As far as I can tell, that method is implemented for the `ArithOpBuilder` trait, which is implemented in the `midenc-dialect-arith` crate. So when I try to import it in `derive.rs`' `Cargo.toml` (`hir/Cargo.toml`), I get a `cyclic package dependency` error.

Regardless, maybe in this case does merit the use of `set_type` use, since it is more of an explicit override rather than a proper cast.

bitwalker: > I think I do have to use the set_type method in this case, right? 

Yes, correct, we're basically creating invalid IR on purpose, and the only real way to do it is to use low-level APIs like `set_type`.

> With this in mind, what do you think about this refactor: https://github.com/0xMiden/compiler/commit/10728a747373c8fbbad6f5fae43d71c03dbbcbba

LGTM!

Issue #435, opened 2025-03-12
Switch to the `wasm32-wasip2` target


greenhat:
This target was quite recently promoted to Tier 2, see https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support/wasm32-wasip2.md

I explored the feasibility of switching to it from the `wasm32-wasip1` we're currently using. 

The main difference between them is that in `wasm32-wasip1` rustc produces the core Wasm module and then `cargo-component` builds the Wasm component from it using `wit_component::ComponentEncoder`. In `wasm32-wasip2` rustc produces the Wasm component using a special linker https://github.com/bytecodealliance/wasm-component-ld which is a wrapper around `wasm-ld` and `wit_component::ComponentEncoder`. 

The only reason why we cannot switch to `wasm32-wasip2` is that we do type remapping (using our types in the generated Rust bindings) and it's not supported in `wasm32-wasip2`. In `wasm32-wasip1` workflow the remapped types are provided to the `wit_component::ComponentEncoder` at https://github.com/bytecodealliance/cargo-component/blob/pub-cargo-arguments-parse-from/src/lib.rs#L955-L955 and so far, there is no way to provide the remapped types to the rustc so they'd be passed to `wasm-component-ld` which in turn does not have an option to take them anyway.

TODO:
 - [x] Test the feasibility by forking the `cargo-component` and switching it to use `wasm32-wasip2` and skipping the component encoding; **YES**
 - [x] Check the type remapping works and study the changes for compiled WAT across the tests **YES**; 
 - [x] Test unsatisfied imports, if we can forgo the WIT interface for our low-level Rust bindings for MASM stdlib and tx kernel; **NOPE, wasm-component-ld calls wit-component encoder, which requires all imports to be satisfied**
 - [x] Re-use `cargo-component` parts and switch to `wasm32-wasip2` target;

bitwalker:
[...]

PR #517, opened 2025-05-21
Copy `cargo-component` code into `cargo-miden` as a module

greenhat:
Ref #435 

This is the first PR in the series of PR for switching to `wasm32-wasip2` target. The goal of this PR is to just copy the code as is and eliminate the `cargo-component` code from the review process. The switch to the `wasm32-wasip2` target will be done in the subsequent PRs.

This PR removes the `cargo-component` dependency and copies **only** the relevant code from the `cargo-component` crate and brings it as a module in the `cargo-miden` crate. The code is unaltered besides build fixes and removed git revision stamp in the generated component binary. We still use `wasm32-wasip1`.


bitwalker (approved):
LGTM, my only suggestion might be to rename the `cargo_component` directory/module to just `component`, since ultimately this code is going to be integrated tightly into `cargo-miden`, so references to `cargo-component` will largely go away over time anyway. Your call on that though.

Issue #510, opened 2025-05-20
Infinite loop in `LiftControlFlowToSCF` when new project template (counter-contract) compiled with `dev` profile (without `--release`)


greenhat:
### How to reproduce

1. Uncomment https://github.com/0xMiden/compiler/blob/next/tools/cargo-miden/tests/build.rs#L106-L126 in tests;
2. Run `MIDENC_TRACE=trace cargo test test_templates -- --nocapture`

Log (looping part):
```
[TRACE cfg-to-scf] visiting multiplexer block argument v128080: i32 at index 0
[TRACE cfg-to-scf] arg is undef on this edge
[TRACE cfg-to-scf] new successor operand is v696
[TRACE cfg-to-scf] visiting multiplexer block argument v128081: u32 at index 1
[TRACE cfg-to-scf] arg corresponds to discriminator index
[TRACE cfg-to-scf] new successor operand is v692
[TRACE cfg-to-scf] redirecting edge ^block32 -> ^block83 with 4 arguments starting at offset 0
[TRACE cfg-to-scf] multiplexer block ^block95621 has 2 arguments
[TRACE cfg-to-scf] extra arguments for edge will begin at 2
[TRACE cfg-to-scf] discriminator index, if present, will be Some(1)
[TRACE cfg-to-scf] visiting multiplexer block arguments for edge
[TRACE cfg-to-scf] visiting multiplexer block argument v128080: i32 at index 0
[TRACE cfg-to-scf] arg corresponds to original block argument at index 0
[TRACE cfg-to-scf] new successor operand is v128078
[TRACE cfg-to-scf] visiting multiplexer block argument v128081: u32 at index 1
[TRACE cfg-to-scf] arg corresponds to discriminator index
[TRACE cfg-to-scf] new successor operand is v697
[TRACE cfg-to-scf] creating switch, exclusions = [], multiplexer argc = 2
[TRACE cfg-to-scf] adding target ^block84 (at index 0) with 0 arguments from offset 0
[TRACE cfg-to-scf] adding target ^block83 (at index 1) with 1 arguments from offset 0
[TRACE cfg-to-scf]  empty branch region for edge ^block32 -> ^block95621
[TRACE cfg-to-scf]  empty branch region for edge ^block32 -> ^block95621
[TRACE cfg-to-scf]  empty branch region for edge ^block32 -> ^block83
[TRACE cfg-to-scf] creating edge multiplexer ^block95624 for [^block83, ^block84] with extra arguments []
[...]

PR #520, opened 2025-05-22
test: add `counter_contract_debug_build` test to reproduce #510

greenhat:
Add debug build artifacts support in integration test suite.


bitwalker (approved):

greenhat:
@bitwalker Initially, I tried to have `expect` tests for Wasm/HIR/MASM. But on linux(CI) the produced Wasm binary was different from the one I got on macos. After a dozen of tries, I gave up and removed the `expect` tests.

PR #522, opened 2025-05-22
fix(cfg-to-scf): incorrect folding of continuation region blocks

bitwalker:
This fixes the issue reported in #510, adds a new test case that covers the structure of the IR that triggered it, and adds some additional instrumentation I found helpful when diagnosing it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3734,6 +3734,7 @@ name = "midenc-dialect-scf"
 version = "0.0.8"
 dependencies = [
  "bitvec",
+ "env_logger 0.11.5",
  "expect-test",
  "log",
  "midenc-dialect-arith",
--- a/dialects/scf/Cargo.toml
+++ b/dialects/scf/Cargo.toml
@@ -26,3 +26,4 @@ bitvec.workspace = true
 
 [dev-dependencies]
 expect-test.workspace = true
+env_logger.workspace = true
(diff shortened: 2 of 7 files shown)


greenhat (approved):
Awesome!

bitwalker:
@greenhat Not sure why that one test is failing, I can't see any actual diff between the expected and actual output, and the test passes locally. I think I'll have to recreate the expect output or something, but maybe you've seen this happen before or will catch something I missed. I'll follow up in the morning though.

greenhat:
> @greenhat Not sure why that one test is failing, I can't see any actual diff between the expected and actual output, and the test passes locally. I think I'll have to recreate the expect output or something, but maybe you've seen this happen before or will catch something I missed. I'll follow up in the morning though.

I'll take care of it.

greenhat:
> @greenhat Not sure why that one test is failing, I can't see any actual diff between the expected and actual output, and the test passes locally. I think I'll have to recreate the expect output or something, but maybe you've seen this happen before or will catch something I missed. I'll follow up in the morning though.

The expected test fails in https://github.com/0xMiden/compiler/actions/runs/15197450891/job/42747931408?pr=522 because the Wasm binary on disk during `expect_wasm` check is the **core** Wasm module and not a Wasm component. I suspect that since we're running `rust_sdk_cross_ctx_account` and `rust_sdk_cross_ctx_note` in parallel and the latter builds account project (twice actually) we caught a bad timing when the note script overwrites account Wasm component binary with core Wasm binary (generated by rustc). We have a core Wasm binary from the rustc which then overwritten by `cargo-component` when it generates a Wasm component out of it. I'm looking into it and if there is no quick fix I'll make an issue.

PR #523, opened 2025-05-23
test: merge `rust_sdk_cross_ctx_account` and `rust_sdk_cross_ctx_note` tests

greenhat:
Following https://github.com/0xMiden/compiler/pull/522#issuecomment-2903478063

To avoid account Wasm binary overwrite when they run in parallel. The odds of this happening were increased by setting one cargo target folder in `Makefile.toml`.

This is a quick fix, and we need to figure out a way to use isolated target folders in each test.

### DIFF
--- a/tests/integration/src/rust_masm_tests/rust_sdk.rs
+++ b/tests/integration/src/rust_masm_tests/rust_sdk.rs
@@ -128,19 +128,19 @@ fn rust_sdk_p2id_note_script() {
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


bitwalker (approved):
I don't think we want to set a unique target directory for every test, or we'll end up doing a ton of re-fetching/compiling dependencies. Either we need to be surgical with it, or we need some kind of way to ensure that two tests with shared filesystem resources aren't stomping on each other. Not sure what the best approach is, so I think this is the best we can do for now.

PR #527, opened 2025-05-23
refactor: simplify function builder to enable better instrumentation

bitwalker:
The FunctionBuilder was requiring us to hold a mutable reference to the underlying Function, which meant that if we wished to print the current Function while tracing is enabled in the compiler, it would always fail when a FunctionBuilder is in scope.

This change refactors it to take a FunctionRef instead, and dynamically borrow as needed to avoid holding a borrow for the entire lifetime of the FunctionBuilder.


greenhat (approved):
Looking good!

PR #529, opened 2025-05-24
chore: update org references from 0xPolygonMiden to 0xMiden

sky-coderay:
Hi, replace all occurrences of "0xPolygonMiden" with "0xMiden" across changelogs and metadata to reflect the org rename


bitwalker:
Hi, thanks for the PR! Unfortunately, we don't accept these type of contributions at this time due to spam. Instead, we prefer to see these simple doc changes submitted alongside more meaningful changes such as bug fixes, etc.

PR #530, opened 2025-05-25
Add `midenc-expect-test` made from the `expect_test` but with unified diffs (short)

greenhat:
This PR adds `midenc-expect-test` crate built from the `expect-test` crate code but swapped diff to unified diff.

### Motivation
The diffs are far too long and insertion and deletions are not visible in CI log due to stripped colors.
Plus, we might be interested in ignoring some diffs (value numbering changes, comments in Rust bindings, etc.). Which makes our changes not feasible to merge into the upstream.


greenhat:
> @greenhat Let's move the new crate under `tools`, rather than `test`

Done.

PR #533, opened 2025-05-27
Arbitrary `struct` results support for cross-context calls

### LINKED ISSUE #539: Arbitrary `struct` shape values in cross-context call result
After #537

greenhat:
Ref #537 
Close #539

This PR:

- implements Wasm CM indirect lowering shim+fixup modules bypass with `canon lower` reconstruction #537;
- adds lifting/lowering for the arbitrary `struct` in the cross-context `call` result;
- adds `truncate_stack` Miden stdlib call at the end of the lifted (`call`-able) function at the MASM codegen;

I stumbled upon the small (less than 32 bits) struct field fail, but since this PR is already huge, I will fix it in the subsequent PR. I made #559 for tracking.

Note: I suggest reviewing this PR by commits.


greenhat:
@bitwalker 
TLDR; We need `call_indirect` and a significant chunk of shim+fixup parsing code in the frontend to support lowering of the return-via-pointer scheme in Wasm CM.

I created a test (with 2 cargo-miden projects) where the account project has a function that returns the `Word` struct which was compiled down to return-via-pointer scheme. When we call this function from the note project `wit-component` generates **two** extra core Wasm modules (shim and fixup) (at https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wit-component/src/encoding.rs#L1070-L1182) for indirect lowering.

My understanding (backed by Opus 4) of why shims are necessary:

In our case:

1. The main module imports `process-word` which takes a `Word` type (4 felt values)
2. This function requires passing parameters through memory (compound type)
3. The function signature is: `(param f32 f32 f32 f32 i32)` where:
  - The 4 f32 values are the felt elements
  - The i32 is a pointer to memory where the result will be written.

### Instantiation Order Problem

Looking at the component structure at https://github.com/0xMiden/compiler/blob/greenhat/i369-heap-allocated-values/tests/integration/expected/rust_sdk/cross_ctx_note_word.wat:
```wasm
(core instance (;3;) (instantiate 0    ; Main module instantiation
    (with "miden:cross-ctx-account-word/foo@1.0.0" (instance 2))
))
...
(alias core export 3 "memory" (core memory (;0;)))   ; Memory comes from instance 3
...
(core func (;3;) (canon lower (func 2) (memory 0)))  ; Lowering needs memory
```

The issue is:
- The main module needs its imports satisfied before it can be instantiated
- But the memory only exists after the module is instantiated
- This creates a circular dependency!

### Component Model Rules

The component model has strict rules about when things can be referenced:
- You can only reference exports from an instance after it's been instantiated
- You can't have circular dependencies in instantiation
- Core modules can only import from other core instances, not from component-level constructs

### The Shim Solution

The shims break this circular dependency:

```wasm
; Step 1: Create shim module with indirect calls
(core module (;1;)   ; Shim module
  (func $indirect-process-word (param ...)
    call_indirect ...)  ; Call through table
)

; Step 2: Instantiate main module with shim imports
(core instance (;3;) (instantiate 0
  (with "miden:cross-ctx-account-word/foo@1.0.0" (instance 2)) ; instance 2 has shim
))

; Step 3: Now we have memory, create the real lowering
(alias core export 3 "memory" (core memory (;0;)))
(core func (;3;) (canon lower (func 2) (memory 0)))

; Step 4: Wire up the table with fixup module
(core instance (;5;) (instantiate 2  ; Fixup module fills the table
  (with "" (instance 4))
))
```

### Why This Design?

This design allows:
- Deferred binding: The actual function implementation is bound after all modules are instantiated
- Memory access: The lowering can access the main module's memory once it exists
- Type safety: The component model can validate all the types at compile time
- Flexibility: Different modules can share the same imports with different memory layouts

The shim approach is essentially a "promise" pattern - the main module is promised that its imports will be satisfied, and the actual wiring happens after everything is instantiated. This is why you see the two-phase approach, with the shim module providing the interface and the fixup module doing the actual wiring.

frontend/wasm/src/component/canon_abi_utils.rs
```diff
@@ -0,0 +1,175 @@
...
+            .into());
+        }
+    }
+
+    Ok(())
+}
+
+/// Aligns ("rounds down") a value to the specified alignment (must be a power of 2)
+/// from https://github.com/WebAssembly/component-model/blob/main/design/mvp/CanonicalABI.md#alignment
+pub fn align_to(ptr: u32, alignment: u32) -> u32 {
```
bitwalker: Let's move this to the `Alignable` trait in `hir-type` as `align_down`, to mirror `align_up` which has the opposite rounding behavior.

frontend/wasm/src/component/canon_abi_utils.rs
```diff
@@ -0,0 +1,175 @@
...
+    fb: &mut FunctionBuilderExt<B>,
+    base_ptr: ValueRef,
+    ty: &Type,
+    offset: &mut u32,
+    values: &mut SmallVec<[ValueRef; 8]>,
+    span: SourceSpan,
+) -> WasmResult<()> {
+    // Align offset to the type's alignment requirement
+    let alignment = ty.min_alignment() as u32;
+    *offset = align_to(*offset, alignment);
```
bitwalker: *Issue*: This seems incorrect to me. Loads are required to be aligned, so I would expect `offset` to always be aligned for `ty`. If it isn't, then that probably should be an error that we catch and investigate.

We should put the onus on the caller to ensure that `base_ptr + offset` is minimally-aligned for `ty`. Especially since alignment rules can depend on context (e.g. the alignment of the first field of a struct is based on the minimal alignment of the _struct_, not the field).

frontend/wasm/src/component/canon_abi_utils.rs
```diff
@@ -0,0 +1,175 @@
...
+        | Type::U8
+        | Type::I16
+        | Type::U16
+        | Type::I32
+        | Type::U32
+        | Type::I64
+        | Type::U64
+        | Type::Felt => {
+            let ptr_type =
+                Type::from(PointerType::new_with_address_space(ty.clone(), AddressSpace::Byte));
```
bitwalker, line 30: Just FYI, `PointerType::new` defaults to `AddressSpace::Byte`

frontend/wasm/src/component/canon_abi_utils.rs
```diff
@@ -0,0 +1,175 @@
...
+
+            let src_offset = fb.i32(*offset as i32, span);
+            let src_addr = fb.add_unchecked(base_ptr, src_offset, span)?;
```
bitwalker: *Issue:* We're aligning `offset` to `ty.min_alignment`, not the actual _address_ of the load, which is going to be `base_ptr + offset`. So our computed alignment could be totally wrong, depending on the alignment of `base_ptr` itself.

The alignment requirements of a type are always based on the absolute address, not relative address. So if `base_ptr` is not minimally-aligned, then aligning `offset` is also not going to be minimally-aligned, even if it is minimally-aligned relative to `base_ptr`. If we needed to load an aligned value from an unaligned address, then we have to do the work to move the data to an aligned location first, and then load it.

frontend/wasm/src/component/canon_abi_utils.rs
```diff
@@ -0,0 +1,175 @@
...
+        // Struct types are loaded field by field
+        Type::Struct(struct_ty) => {
+            // For each field in the struct, use the pre-calculated field offset
+            let struct_base_offset = *offset;
+            for field in struct_ty.fields() {
+                // Use the field's offset within the struct, plus the struct's base offset
+                let mut field_offset = struct_base_offset + field.offset;
+
+                // Recursively load the field
+                load(fb, base_ptr, &field.ty, &mut field_offset, values, span)?;
```
bitwalker: Is there any reason why we're passing `base_ptr` + and offset here, rather than just aligning `base_ptr` for a given type, and then passing in the aligned pointer?

The reason I'm asking, is because otherwise we're going to compute incorrect addresses for the load, because alignments are going to be computed from the current offset, rather than the address of the load.

Ultimately, I don't see much value in passing an offset here; we also don't need to maintain one here AFAICT, because the field offsets for `struct_ty` are already computed to take into account the size of the fields + alignment.

frontend/wasm/src/component/canon_abi_utils.rs
```diff
@@ -0,0 +1,175 @@
...
+pub fn store<B: ?Sized + Builder>(
+    fb: &mut FunctionBuilderExt<B>,
+    base_ptr: ValueRef,
+    ty: &Type,
+    offset: &mut u32,
+    values: &mut impl Iterator<Item = ValueRef>,
+    span: SourceSpan,
+) -> WasmResult<()> {
+    // Align offset to the type's alignment requirement
+    let alignment = ty.min_alignment() as u32;
```
bitwalker: I think my notes about `load` apply to `store` as well, so I won't replicate them here, but just pointing out that it looks to exhibit the same issues.

frontend/wasm/src/component/flat.rs
```diff
@@ -174,6 +174,8 @@ pub fn assert_core_wasm_signature_equivalence(
     wasm_core_sig: &Signature,
     flattened_sig: &Signature,
 ) {
+    // dbg!(&wasm_core_sig);
```
bitwalker: I'd like to avoid committing these if we can, even if we're using them.

bitwalker (commented):
Just leaving my first set of notes, not quite finished reviewing yet. I'm about to start on the lifting code in the frontend, but the rest after that should go pretty quick.

frontend/wasm/src/component/canon_abi_utils.rs
greenhat: Yes, but I felt that specifying it explicitly helps my comprehension.

frontend/wasm/src/component/canon_abi_utils.rs
greenhat: I removed it and rely on ptr being aligned.

frontend/wasm/src/component/canon_abi_utils.rs
greenhat: Yes, you're right. I removed the alignment.

frontend/wasm/src/component/canon_abi_utils.rs
greenhat: Good point. I removed the offset, so it should not be an issue.

frontend/wasm/src/component/canon_abi_utils.rs
greenhat: Good question! I think it works because I ignore the offset parameter when processing a struct. But yeah, I removed the offset altogether. Thank you!

frontend/wasm/src/component/canon_abi_utils.rs
greenhat: Done.

frontend/wasm/src/component/flat.rs
greenhat: Sure. Thanks! Removed.

greenhat:
> Just leaving my first set of notes, not quite finished reviewing yet. I'm about to start on the lifting code in the frontend, but the rest after that should go pretty quick.

I rebased this branch and addressed all the notes. I don't want to force push and mess with your review progress. Let me know when I can force push.

bitwalker:
I'm out today and tomorrow, but I'll try to find time to finish it up. Sorry for the delay - too many plates in the air at the same time 😅

greenhat:
No rush! It does not block me and can wait. Justed wanted to make sure it's on your radar but I choose a bad timing. 😃

frontend/wasm/src/component/lift_exports.rs
```diff
@@ -60,7 +52,100 @@ pub fn generate_export_lifting_function(
...
+    // Extract flattened result types from the exported component-level function type
+    let flattened_results = flatten_types(&export_func_ty.results).map_err(|e| {
+        let message = format!(
+            "Failed to flatten result types for exported function {}: {}",
+            core_export_func_path, e
+        );
+        diagnostics.diagnostic(Severity::Error).with_message(message).into_report()
+    })?;
+
+    assert!(
```
bitwalker, line 141: *Suggestion:* If we hit this, we aren't going to know what function triggered it and what the signature was - I'd either add that information to the message, or emit an actual diagnostic and then bail.

frontend/wasm/src/component/lift_exports.rs
```diff
@@ -81,8 +166,106 @@ pub fn generate_export_lifting_function(
...
+    // 2. Return it as individual values (tuple)
+
+    let exec = fb.exec(core_export_func_ref, core_export_func_sig, args, span)?;
+
+    let borrow = exec.borrow();
+    let results = borrow.results().all();
+
+    // The core function should return a single pointer (as i32)
+    assert_eq!(results.len(), 1, "expected single result");
+    let result_ptr = results.into_iter().next().unwrap().borrow().as_value_ref();
```
bitwalker:
```suggestion
    let result_ptr = results[0].borrow().as_value_ref();
```

`EntityRange` implements `Index`, so you shouldn't have to convert it to an iterator just to get the first element.

frontend/wasm/src/component/lift_exports.rs
```diff
@@ -81,8 +166,106 @@ pub fn generate_export_lifting_function(
...
+    assert_eq!(
+        export_func_ty.results.len(),
+        1,
+        "expected a single result in the component-level export function"
+    );
+    let result_type = &export_func_ty.results[0];
+
+    load(&mut fb, result_ptr, result_type, &mut return_values, span)?;
+
+    assert!(
```
bitwalker, line 203: *Suggestion:* Same as my previous one regarding the information available to us if this assertion is raised

frontend/wasm/src/component/lower_imports.rs
```diff
@@ -83,12 +150,155 @@ pub fn generate_import_lowering_function(
...
+    );
+
+    // Import lowering: The lowered function takes a pointer as the last parameter
+    // where results should be stored. The import function returns a pointer to the result.
+    // We need to:
+    // 1. Call the import function (it returns a tuple to the flattened result)
+    // 2. Store the data from the tuple to the output pointer which expect to hold
+    //    flattened result
+
+    // Get the pointer argument (last argument) where we need to store results
```
bitwalker, line 201: It's unusual to pass an `sret` pointer as the last parameter, rather than inserting it at the front of the parameter list. Is there any particular reason why we use the last argument? I'm aware that our SDK bindings have tended to pass it as the last parameter, but that was an arbitrary choice, and it might make sense to try and align with normal ABI conventions for this type of thing, unless there are other constraints.

frontend/wasm/src/component/lower_imports.rs
```diff
@@ -83,12 +150,155 @@ pub fn generate_import_lowering_function(
...
+    let output_ptr = args.last().expect("expected pointer argument");
+    let args_without_ptr: Vec<_> = args[..args.len() - 1].to_vec();
+
+    // Call the import function - it will return a tuple to the flattened result
+    let call = fb.call(import_func_ref, new_import_func_sig, args_without_ptr, span)?;
+
+    let borrow = call.borrow();
+    let results_storage = borrow.as_ref().results();
+    let results: Vec<ValueRef> =
```
bitwalker: *Nit:* I think you can just use `ValueRange` here, e.g. `let results = borrow.as_ref().results().as_value_range().into_owned();`

frontend/wasm/src/component/lower_imports.rs
```diff
@@ -83,12 +150,155 @@ pub fn generate_import_lowering_function(
...
+    fb.br(exit_block, [], span)?;
+    fb.seal_block(exit_block);
+    fb.switch_to_block(exit_block);
+
+    // Return according to the core function signature
+    if core_func_sig.results().is_empty() {
+        fb.ret([], span)?;
+    } else {
+        // The core function should not have any results when using out-pointer
+        panic!("Core function should not have results when using out-pointer pattern");
```
bitwalker:
```suggestion
    // Return according to the core function signature
    // NOTE: The core function should not have any results when using out-pointer
    assert!(core_func_sig.results().is_empty(), "Core function should not have results when using out-pointer pattern");
    fb.ret([], span)?;
```

frontend/wasm/src/component/parser.rs
```diff
@@ -507,7 +507,9 @@ impl<'a, 'data> ComponentParser<'a, 'data> {
         self.validator.component_canonical_section(&s).into_diagnostic()?;
         for func in s {
             let types = self.validator.types(0).unwrap();
-            let init = match func.into_diagnostic()? {
+            let canonical_func = func.into_diagnostic()?;
+            log::debug!("Processing canonical function: {:?}", canonical_func);
```
bitwalker: *Nit:* Make sure to specify a `target` for `log` messages, e.g. `log::debug!(target: "component-parser", ...);`

Try to pick targets that are not overly granular, but not too coarse - it's a bit of an art. The idea is that if you are trying to see what is going on during some specific phase of compilation, you can use the target to drill  down to the part you are interested in, and filter out the noise. Too granular, and you can't see enough context; too coarse, and you get too much output.

frontend/wasm/src/component/parser.rs
```diff
@@ -588,7 +590,7 @@ impl<'a, 'data> ComponentParser<'a, 'data> {
                 | wasmparser::CanonicalFunction::StreamCloseWritable { .. }
                 | wasmparser::CanonicalFunction::StreamCloseReadable { .. } => unimplemented!(),
             };
-            // dbg!(&init);
+            log::debug!("Adding canonical initializer: {:?}", init);
```
bitwalker: *Nit:* Missing `log` target

frontend/wasm/src/component/shim_bypass.rs
```diff
@@ -0,0 +1,136 @@
...
+///
+/// Shim modules have these characteristics:
+/// 1. Export a table named "$imports"
+/// 2. Export functions (may have numeric names or other patterns)
+/// 3. Functions contain only call_indirect instructions
+/// 4. Usually have minimal imports
+pub fn is_shim_module(module: &ParsedModule) -> bool {
+    log::trace!(
+        "Checking if module is shim module. Exports: {:?}",
+        module.module.exports.keys().collect::<Vec<_>>()
```
bitwalker: You can use `midenc_hir::formatter::DisplayValues` here to avoid materializing the `Vec`

frontend/wasm/src/component/translator.rs
```diff
@@ -405,6 +573,7 @@ impl<'a> ComponentTranslator<'a> {
             ModuleDef::Static(static_module_idx) => {
                 let parsed_module = self.nested_modules.get_mut(*static_module_idx).unwrap();
                 for module_arg in args {
+                    // dbg!(&module_arg);
```
bitwalker: *Nit:* Looks like this got left in on accident

tests/integration/src/rust_masm_tests/rust_sdk.rs
```diff
@@ -45,10 +45,9 @@ fn rust_sdk_cross_ctx_account_and_note() {
         config.clone(),
         [],
     );
-    let artifact_name = test.artifact_name().to_string();
-    test.expect_wasm(expect_file![format!("../../expected/rust_sdk/{artifact_name}.wat")]);
-    test.expect_ir(expect_file![format!("../../expected/rust_sdk/{artifact_name}.hir")]);
-    test.expect_masm(expect_file![format!("../../expected/rust_sdk/{artifact_name}.masm")]);
+    test.expect_wasm(expect_file![format!("../../expected/rust_sdk/cross_ctx_account.wat")]);
```
bitwalker, line 48: *Question:* Any particular reason to make these hardcoded?

tests/rust-apps-wasm/rust-sdk/cross-ctx-account-word/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+// Global allocator to use heap memory in no-std environment
+#[global_allocator]
+static ALLOC: miden::BumpAlloc = miden::BumpAlloc::new();
+
+// Required for no-std crates
+#[cfg(not(test))]
+#[panic_handler]
+fn my_panic(_info: &core::panic::PanicInfo) -> ! {
+    loop {}
```
bitwalker, line 18: *Nit:* We should prefer `core::arch::wasm32::unreachable()` now - it's stable, and doesn't rely on LLVM lowering infinite loops to a trap

frontend/wasm/src/component/lift_exports.rs
greenhat: Thanks! Done in https://github.com/0xMiden/compiler/pull/533/commits/91310d5358ff12249f677c54c8234e9cc0a39dc3

frontend/wasm/src/component/lift_exports.rs
greenhat: Done in https://github.com/0xMiden/compiler/pull/533/commits/91310d5358ff12249f677c54c8234e9cc0a39dc3

frontend/wasm/src/component/lower_imports.rs
greenhat: It's a CanonABI (wit-bindgen) generated lowering for the component import, and it's not our decision. 

Since we don't commit generated `binding.rs`, here is the excerpt for the `process-mixed` function in the https://github.com/0xMiden/compiler/blob/91310d5358ff12249f677c54c8234e9cc0a39dc3/tests/rust-apps-wasm/rust-sdk/cross-ctx-note-word/src/lib.rs?plain=1 `cross-ctx-note-word` test that returns a large struct (see `wit_import4` external function):
```rust
pub fn process_mixed(input: MixedStruct) -> MixedStruct {
    unsafe {
        #[repr(align(8))]
        struct RetArea([::core::mem::MaybeUninit<u8>; 24]);
        let mut ret_area = RetArea([::core::mem::MaybeUninit::uninit(); 24]);
        let MixedStruct { f: f0, a: a0, b: b0, c: c0 } = input;
        let miden::Felt { inner: inner1 } = a0;
        let miden::Felt { inner: inner2 } = c0;
        let ptr3 = ret_area.0.as_mut_ptr().cast::<u8>();
        #[cfg(target_arch = "wasm32")]
        #[link(
            wasm_import_module = "miden:cross-ctx-account-word/foo@1.0.0"
        )]
        unsafe extern "C" {
            #[link_name = "process-mixed"]
            fn wit_import4(_: i64, _: f32, _: i32, _: f32, _: *mut u8);
        }
        #[cfg(not(target_arch = "wasm32"))]
        unsafe extern "C" fn wit_import4(
            _: i64,
            _: f32,
            _: i32,
            _: f32,
            _: *mut u8,
        ) {
            unreachable!()
        }
        unsafe {
            wit_import4(
                _rt::as_i64(f0),
                _rt::as_f32(inner1),
                _rt::as_i32(b0),
                _rt::as_f32(inner2),
                ptr3,
            )
        };
        let l5 = *ptr3.add(0).cast::<i64>();
        let l6 = *ptr3.add(8).cast::<f32>();
        let l7 = *ptr3.add(12).cast::<i32>();
        let l8 = *ptr3.add(16).cast::<f32>();
        let result9 = MixedStruct {
            f: l5 as u64,
            a: miden::Felt { inner: l6 },
            b: l7 as u32,
            c: miden::Felt { inner: l8 },
        };
        result9
    }
}
```

frontend/wasm/src/component/lower_imports.rs
greenhat: Thanks! Done in https://github.com/0xMiden/compiler/pull/533/commits/3029661eee9e24056eee26ecab9309c950c401e7

frontend/wasm/src/component/parser.rs
greenhat: Thanks! Specified throughout the frontend in https://github.com/0xMiden/compiler/pull/533/commits/ecf212961f3191c8f76ac9bcc608a09c8fd7ddf1

frontend/wasm/src/component/parser.rs
greenhat: Done in https://github.com/0xMiden/compiler/pull/533/commits/ecf212961f3191c8f76ac9bcc608a09c8fd7ddf1

frontend/wasm/src/component/shim_bypass.rs
greenhat: Done in https://github.com/0xMiden/compiler/pull/533/commits/e2ee88b7958599d0b38b5b03e3e84b2d76dd9973

frontend/wasm/src/component/translator.rs
greenhat: Thanks. Cleaned up in the whole frontend.

tests/integration/src/rust_masm_tests/rust_sdk.rs
greenhat: Yes. So the expected files are easier to find for an LLM agent.

tests/rust-apps-wasm/rust-sdk/cross-ctx-account-word/src/lib.rs
greenhat: Extracted into https://github.com/0xMiden/compiler/issues/570, since LSP was not happy with this switch.

greenhat:
> Overall, looks good! I'm marking this approved so I'm not holding things up, but would like to clean the following up before merging:
> 
> * Make sure all `log` tracing specifies the `target` field, and provide a useful target based on your gut feeling on what makes the most sense

Done in https://github.com/0xMiden/compiler/pull/533/commits/ecf212961f3191c8f76ac9bcc608a09c8fd7ddf1 

> * Related to logging, use `midenc_hir::formatter::DisplayValues` to render lists/vectors/sets of values without materializing a `SmallVec` or `Vec`. It'll reduce the impact on compilation performance in debug builds. It renders the values comma-separated without any surrounding `[`/`{`/`(`, so you provide those yourself in the format string based on what makes the most sense.

Done in https://github.com/0xMiden/compiler/pull/533/commits/024de0d95b899b179672039c0f80d6ab161e7164
 
> Optionally, it might be a good idea to have some unit tests around the type signature conversion code, so we can ensure there aren't any edge cases we've missed, but not sure how amenable the code is to being unit tested like that. Something to think about anyway.

Good point! Added in https://github.com/0xMiden/compiler/pull/533/commits/b293bb3bfdd93c77b7495d1c9afcc49df6639bee

Issue #404, opened 2025-02-17
Op builder should register the op in the symbol table


greenhat:
Following https://github.com/0xPolygonMiden/compiler/pull/397#discussion_r1953810426

>That should probably be done as some sort of hook that fires anytime an op is inserted into a region (or removed from one), but for now we'll have to just implement the logic in the op-specific builders manually.

PR #546, opened 2025-05-28
Add Op registration into symbol table inside OpBuilder 

lima-limon-inc:
Closes: #404

This PR implements logic to add an Operation which implements the `Symbol` trait to its parent `SymbolTable`, after the Operation is built.

This is done via the `EntityListItem::on_inserted` and `EntityListItem::on_removed` hooks, which are run after an `Operation` is added or removed removed from an `EntityList` respectively. This executed shortly after `Operation` creation, during the final stages of the `OperationBuilder::build` function.

Originally, `Operation`s were borrowed before entering this function, which caused errors when trying to add them to their parent's `SymbolTable`. To prevent this, some functions related to `Operation` insertion were moved to `OperationRef` and now avoid borrowing the underlying `Operation`.

### DIFF
--- a/hir/src/ir/operation.rs
+++ b/hir/src/ir/operation.rs
@@ -181,33 +181,32 @@ impl EntityWithParent for Operation {
 }
 impl EntityListItem for Operation {
     fn on_inserted(this: OperationRef, _cursor: &mut EntityCursorMut<'_, Self>) {
-        let parent = this.nearest_symbol_table();
-        if let Some(mut parent) = parent {
-            // NOTE: We use OperationName, instead of the Operation itself, to avoid borrowing.
-            if this.name().implements::<dyn Symbol>()
-                && parent.name().implements::<dyn SymbolTable>()
-            {
-                // NOTE: We call `unwrap()` here because we are confident that these function calls
-                // are valid thanks to the `implements` check above.
-                let mut symbol_table = parent.borrow_mut();
-                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().unwrap();
-                let mut sym_manager = sym_manager.symbol_manager_mut();
+        // NOTE: We use OperationName, instead of the Operation itself, to avoid borrowing.
+        if this.name().implements::<dyn Symbol>() {
+            let parent = this.nearest_symbol_table();
+            if let Some(mut parent) = parent {
+                if parent.name().implements::<dyn SymbolTable>() {
+                    // NOTE: We call `unwrap()` here because we are confident that these function calls
+                    // are valid thanks to the `implements` check above.
+                    let mut symbol_table = parent.borrow_mut();
+                    let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().unwrap();
+                    let mut sym_manager = sym_manager.symbol_manager_mut();
 
-                let symbol_ref = this.borrow().as_symbol_ref().unwrap();
+                    let symbol_ref = this.borrow().as_symbol_ref().unwrap();
 
-                let is_new = sym_manager.insert_new(symbol_ref, ProgramPoint::Invalid);
-                assert!(
-                    is_new,
-                    "Unable to insert {} in symbol table of {}: symbol {} is already registered \
-                     to another operation {}.",
-                    this.name(),
-                    parent.name(),
-                    symbol_ref.borrow().name(),
-                    sym_manager.lookup(symbol_ref.borrow().name()).unwrap().borrow().name()
-                );
-            };
+                    let is_new = sym_manager.insert_new(symbol_ref, ProgramPoint::Invalid);
+                    assert!(
+                        is_new,
+                        "Unable to insert {} in symbol table of {}: symbol {} is already \
+                         registered to another operation {}.",
+                        this.name(),
+                        parent.name(),
+                        symbol_ref.borrow().name(),
+                        sym_manager.lookup(symbol_ref.borrow().name()).unwrap().borrow().name()
+                    );
+                }
+            }
         }
-
         let order_offset = core::mem::offset_of!(Operation, order);
         unsafe {
             let ptr = UnsafeIntrusiveEntityRef::as_ptr(&this);
@@ -223,20 +222,21 @@ impl EntityListItem for Operation {
     }
 
     fn on_removed(this: OperationRef, _list: &mut EntityCursorMut<'_, Self>) {
-        let parent = this.nearest_symbol_table();
-        if let Some(mut parent) = parent {
-            // NOTE: We use OperationName, instead of the Operation itself, to avoid borrowing.
-            if this.name().implements::<dyn Symbol>()
-                && parent.name().implements::<dyn SymbolTable>()
-            {
-                let mut symbol_table = parent.borrow_mut();
-                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().unwrap();
-                let mut sym_manager = sym_manager.symbol_manager_mut();
+        // NOTE: We use OperationName, instead of the Operation itself, to avoid borrowing.
+        if this.name().implements::<dyn Symbol>() {
+            let parent = this.nearest_symbol_table();
+            if let Some(mut parent) = parent {
+                // NOTE: We use OperationName, instead of the Operation itself, to avoid borrowing.
+                if parent.name().implements::<dyn SymbolTable>() {
+                    let mut symbol_table = parent.borrow_mut();
+                    let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().unwrap();
+                    let mut sym_manager = sym_manager.symbol_manager_mut();
 
-                let symbol_ref = this.borrow().as_symbol_ref().unwrap();
+                    let symbol_ref = this.borrow().as_symbol_ref().unwrap();
 
-                sym_manager.remove(symbol_ref);
-            };
+                    sym_manager.remove(symbol_ref);
+                };
+            }
         }
     }
 }
--- a/hir/src/ir/operation.rs
+++ b/hir/src/ir/operation.rs
@@ -1125,6 +1125,14 @@ impl Operation {
 
 /// Insertion
 impl Operation {
+    pub fn insert_at_start(&mut self, block: BlockRef) {
+        self.as_operation_ref().insert_at_start(block);
+    }
+
+    pub fn insert_at_end(&mut self, block: BlockRef) {
+        self.as_operation_ref().insert_at_end(block);
+    }
+
     pub fn insert_before(&mut self, before: OperationRef) {
         assert!(
             self.parent().is_none(),
--- a/hir/src/ir/operation.rs
+++ b/hir/src/ir/operation.rs
@@ -301,12 +301,10 @@ impl OperationRef {
     pub fn nearest_symbol_table(&self) -> Option<OperationRef> {
         let mut parent = self.parent_op();
         while let Some(parent_op) = parent.take() {
-            let op = parent_op.borrow();
-            if op.implements::<dyn SymbolTable>() {
-                drop(op);
+            if parent_op.name().implements::<dyn SymbolTable>() {
                 return Some(parent_op);
             }
-            parent = op.parent_op();
+            parent = parent_op.parent_op();
         }
         None
     }
--- a/hir/src/ir/operation.rs
+++ b/hir/src/ir/operation.rs
@@ -189,10 +189,16 @@ impl EntityListItem for Operation {
                 && parent.name().implements::<dyn SymbolTable>()
             {
                 let mut symbol_table = parent.borrow_mut();
-                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().unwrap();
+                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().expect(
+                    "Could not cast parent operation {parent.name()} as SymbolTable, even though \
+                     it implements said trait",
+                );
                 let mut sym_manager = sym_manager.symbol_manager_mut();
 
-                let symbol_ref = this.borrow().as_symbol_ref().unwrap();
+                let symbol_ref = this.borrow().as_symbol_ref().expect(
+                    "Could not cast operation {this.name()} as Symbol, even though it implements \
+                     said trait",
+                );
 
                 sym_manager.insert_new(symbol_ref, ProgramPoint::Invalid);
             };
--- a/hir/src/ir/operation.rs
+++ b/hir/src/ir/operation.rs
@@ -218,7 +218,30 @@ impl EntityListItem for Operation {
         to.borrow_mut().invalidate_op_order();
     }
 
-    fn on_removed(_this: OperationRef, _list: &mut EntityCursorMut<'_, Self>) {}
+    fn on_removed(this: OperationRef, _list: &mut EntityCursorMut<'_, Self>) {
+        let parent = this.nearest_symbol_table();
+        if let Some(mut parent) = parent {
+            // NOTE: We are using OperationName here to check if the Operation implements symbol to
+            // avoid borrowing if possible
+            if this.name().implements::<dyn Symbol>()
+                && parent.name().implements::<dyn SymbolTable>()
+            {
+                let mut symbol_table = parent.borrow_mut();
+                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().expect(
+                    "Could not cast parent operation {parent.name()} as SymbolTable, even though \
+                     it implements said trait",
+                );
+                let mut sym_manager = sym_manager.symbol_manager_mut();
+
+                let symbol_ref = this.borrow().as_symbol_ref().expect(
+                    "Could not cast operation {this.name()} as Symbol, even though it implements \
+                     said trait",
+                );
+
+                sym_manager.remove(symbol_ref);
+            };
+        }
+    }
 }
 
 impl EntityParent<Region> for Operation {
--- a/hir/src/ir/operation.rs
+++ b/hir/src/ir/operation.rs
@@ -319,7 +319,6 @@ impl OperationRef {
         self.parent().and_then(|block| block.parent())
     }
 
-    /// TODO(fabrio): maybe remove later?
     /// Returns the nearest [SymbolTable] from this operation.
     ///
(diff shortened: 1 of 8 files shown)


bitwalker (changes requested):
I think unfortunately we're going to have to revisit the approach implemented in this PR.

Fundamentally, symbol registration is completely separate from operation creation. Similar to the issues driving #516, the creation of an operation that implements `Symbol` happens before it is ever added to the IR graph, and thus it isn't a given that we even know the `SymbolTable` (or have a reference to it handy) at the time the `Symbol` is created. Even if we were able to require one, doing so would create a situation where the `SymbolTable` says that some `Symbol` is registered with it, even though the `Symbol` op hasn't been added to the `Block` representing the body of the `SymbolTable` op. 

In short, (un)registering a `Symbol` with a parent `SymbolTable` must happen when a `Symbol` is inserted into/removed from the body of a `SymbolTable`, whether that happens via a `Builder` trait method (e.g. `Builder::insert`), or manually using one of the low-level `Operation` APIs, e.g. `Operation::insert_at_end`.

I think the best mechanism we have to achieve this is using the `EntityListItem::on_inserted` and `EntityListItem::on_removed` callbacks implemented for `Operation` impl. These callbacks are invoked any time an operation is added or removed from a `Block`. If `self.implements::<Symbol>()` in those callbacks, and if the parent operation whose `Block` we're inserting into/removing from implements `SymbolTable`, then we can take the appropriate action to modify the `SymbolTable` appropriately.

I'm not sure we have any other options. The alternative is to go one level higher, and modify all of the low-level IR APIs that insert/remove/transfer operations to handle symbol (de)registration as part of their implementation, but if we can handle it via the `EntityListItem` callbacks, that would be the best-case scenario, since all of the low-level IR APIs invoke those anyway, and it would be far more future-proof. The only potential issue might be with regard to borrowing of the parent `SymbolTable` op while we're inserting/removing from one of its blocks. In principle, we should be able to do that, since symbol table modifications don't require borrowing the block - but there might be situations where both the `SymbolTable` op _and_ the block are being borrowed already, in which case we will fail when we attempt to borrow the `SymbolTable` op again. I think the solution in that case would be to fix up the places where we're borrowing the op unnecessarily (by dropping the borrow until after we're done modifying the block), but we'd just have to see if we hit any awkward cases where that isn't an option, or something else I'm not thinking of gets in the way.

Apologies that this wasn't better elaborated on in the original issue, some of these constraints have also come up since that issue was originally created, so its possible the commentary there also laid out an incorrect solution, I'd have to go back and refresh my memory. In any case, I hope this provides a clear explanation of why the implementation here won't work as-is, and how to tackle it using some of the other facilities provided in the low-level internals of the IR. I think in the end the implementation using `EntityListItem` callbacks will be way simpler anyway, but definitely ping me if you have questions or run into any issues.

hir/src/ir/operation.rs
```diff
@@ -180,7 +180,30 @@ impl EntityWithParent for Operation {
...
+        if let Some(mut parent) = parent {
+            // NOTE: We use OperationName, instead of the Operation itself, to avoid borrowing.
+            if this.name().implements::<dyn Symbol>()
+                && parent.name().implements::<dyn SymbolTable>()
+            {
+                let mut symbol_table = parent.borrow_mut();
+                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().expect(
+                    "Could not cast parent operation {parent.name()} as SymbolTable, even though \
+                     it implements said trait",
+                );
```
bitwalker:
```suggestion
                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().unwrap();
```

This can be considered infallible if `implements` returns true - so `unwrap` is fine here. All kinds of things would be broken if that wasn't the case, so it just obscures what is going on this function to have an expect message IMO.

hir/src/ir/operation.rs
```diff
@@ -180,7 +180,30 @@ impl EntityWithParent for Operation {
...
+                let mut sym_manager = sym_manager.symbol_manager_mut();
+
+                let symbol_ref = this.borrow().as_symbol_ref().expect(
+                    "Could not cast operation {this.name()} as Symbol, even though it implements \
+                     said trait",
+                );
```
bitwalker:
```suggestion
                let symbol_ref = this.borrow().as_symbol_ref().unwrap();
```

Same as above

hir/src/ir/operation.rs
```diff
@@ -180,7 +180,30 @@ impl EntityWithParent for Operation {
...
+
+                let is_new = sym_manager.insert_new(symbol_ref, ProgramPoint::Invalid);
+                assert!(is_new, "{} already exists in {}", this.name(), parent.name());
```
bitwalker: *Nit:* This error won't be very helpful, since it is only mentioning the operation names, e.g. `builtin.function`. What we actually want to say here is something like:

> unable to insert '{this.name()}' in symbol table of '{parent.name()}': symbol '{symbol_ref.borrow().name()}' is already registered to another operation

hir/src/ir/operation.rs
```diff
@@ -189,15 +212,35 @@ impl EntityListItem for Operation {
...
+        if let Some(mut parent) = parent {
+            // NOTE: We use OperationName, instead of the Operation itself, to avoid borrowing.
+            if this.name().implements::<dyn Symbol>()
+                && parent.name().implements::<dyn SymbolTable>()
+            {
+                let mut symbol_table = parent.borrow_mut();
+                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().expect(
+                    "Could not cast parent operation {parent.name()} as SymbolTable, even though \
+                     it implements said trait",
+                );
```
bitwalker:
```suggestion
                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().unwrap();
```

hir/src/ir/operation.rs
```diff
@@ -180,7 +180,30 @@ impl EntityWithParent for Operation {
     type Parent = Block;
 }
 impl EntityListItem for Operation {
-    fn on_inserted(this: UnsafeIntrusiveEntityRef<Self>, _cursor: &mut EntityCursorMut<'_, Self>) {
+    fn on_inserted(this: OperationRef, _cursor: &mut EntityCursorMut<'_, Self>) {
+        let parent = this.nearest_symbol_table();
```
bitwalker: We should gate the symbol registration on whether or not `this` implements `Symbol`, rather than looking up the nearest symbol table first (an expensive operation). For operations which are not symbols (the vast majority), this is going to do a lot of work that is then ignored because `this` isn't even a `Symbol`

hir/src/ir/operation.rs
```diff
@@ -189,15 +212,35 @@ impl EntityListItem for Operation {
...
-        to: &mut EntityList<Self>,
-    ) {
+    fn on_transfer(_this: OperationRef, _from: &mut EntityList<Self>, to: &mut EntityList<Self>) {
         // Invalidate the ordering of the new parent block
         let mut to = to.parent();
         to.borrow_mut().invalidate_op_order();
     }
+
+    fn on_removed(this: OperationRef, _list: &mut EntityCursorMut<'_, Self>) {
+        let parent = this.nearest_symbol_table();
```
bitwalker: Same issue here as `on_inserted` - we should only do this work if we know that `this` implements `Symbol` first

hir/src/ir/operation.rs
```diff
@@ -189,15 +212,35 @@ impl EntityListItem for Operation {
...
+                let sym_manager = symbol_table.as_trait_mut::<dyn SymbolTable>().expect(
+                    "Could not cast parent operation {parent.name()} as SymbolTable, even though \
+                     it implements said trait",
+                );
+                let mut sym_manager = sym_manager.symbol_manager_mut();
+
+                let symbol_ref = this.borrow().as_symbol_ref().expect(
+                    "Could not cast operation {this.name()} as Symbol, even though it implements \
+                     said trait",
+                );
```
bitwalker:
```suggestion
                let symbol_ref = this.borrow().as_symbol_ref().unwrap();
```

hir/src/ir/operation.rs
```diff
@@ -227,6 +270,59 @@ impl Operation {
     }
 }
 
+/// Insertion
+impl OperationRef {
+    pub fn insert_at_start(&self, mut block: BlockRef) {
```
bitwalker:
```suggestion
    pub fn insert_at_start(self, mut block: BlockRef) {
```

hir/src/ir/operation.rs
```diff
@@ -227,6 +270,59 @@ impl Operation {
...
+            self.parent().is_none(),
+            "cannot insert operation that is already attached to another block"
+        );
+        {
+            let mut block = block.borrow_mut();
+            block.body_mut().push_front(*self);
+        }
+    }
+
+    pub fn insert_at_end(&self, mut block: BlockRef) {
```
bitwalker:
```suggestion
    pub fn insert_at_end(self, mut block: BlockRef) {
```

hir/src/ir/operation.rs
```diff
@@ -227,6 +270,59 @@ impl Operation {
...
+            self.parent().is_none(),
+            "cannot insert operation that is already attached to another block"
+        );
+        {
+            let mut block = block.borrow_mut();
+            block.body_mut().push_back(*self);
+        }
+    }
+
+    pub fn insert_before(&mut self, before: OperationRef) {
```
bitwalker:
```suggestion
    pub fn insert_before(self, before: OperationRef) {
```

hir/src/ir/operation.rs
```diff
@@ -227,6 +270,59 @@ impl Operation {
...
+        let mut block = before.parent().expect("'before' block is not attached to a block");
+        {
+            let mut block = block.borrow_mut();
+            let block_body = block.body_mut();
+            let mut cursor = unsafe { block_body.cursor_mut_from_ptr(before) };
+            cursor.insert_before(*self);
+        }
+    }
+
+    pub fn insert_after(&mut self, after: OperationRef) {
```
bitwalker:
```suggestion
    pub fn insert_after(self, after: OperationRef) {
```

hir/src/ir/operation.rs
```diff
@@ -241,6 +337,31 @@ impl OperationRef {
...
+
+    /// Returns a handle to the nearest containing [Operation] of this operation, if it is attached
+    /// to one
+    pub fn parent_op(&self) -> Option<OperationRef> {
+        self.parent_region().and_then(|region| region.parent())
+    }
+
+    /// Returns a handle to the containing [Region] of this operation, if it is attached to one
+    pub fn parent_region(&self) -> Option<RegionRef> {
+        self.parent().and_then(|block| block.parent())
```
bitwalker:
```suggestion
        self.grandparent()
```

hir/src/ir/operation.rs
```diff
@@ -1033,54 +1154,20 @@ impl Operation {
...
-    pub fn insert_at_start(&mut self, mut block: BlockRef) {
-        assert!(
-            self.parent().is_none(),
-            "cannot insert operation that is already attached to another block"
-        );
-        {
-            let mut block = block.borrow_mut();
-            block.body_mut().push_front(self.as_operation_ref());
-        }
+    pub fn insert_at_start(&mut self, block: BlockRef) {
```
bitwalker: This will always fail, because by definition we're holding a mutable borrow of the operation, so the call to `insert_at_start` will fail with an aliasing error when it tries to borrow the operation again.

We should either require that these methods take an immutable reference, or remove them entirely.

I'm in favor of the latter personally - using these methods off of `&Operation`/`&mut Operation` is very likely to cause aliasing conflicts, and it would be better to encourage using the methods of `OperationRef` instead, to better scope the borrows.

hir/src/ir/operation.rs
```diff
@@ -1033,54 +1154,20 @@ impl Operation {
...
-    pub fn insert_at_end(&mut self, mut block: BlockRef) {
-        assert!(
-            self.parent().is_none(),
-            "cannot insert operation that is already attached to another block"
-        );
-        {
-            let mut block = block.borrow_mut();
-            block.body_mut().push_back(self.as_operation_ref());
-        }
+    pub fn insert_at_end(&mut self, block: BlockRef) {
```
bitwalker: Same issue here re: aliasing

hir/src/ir/operation.rs
```diff
@@ -1033,54 +1154,20 @@ impl Operation {
...
+        self.as_operation_ref().insert_at_end(block);
     }
 
     pub fn insert_before(&mut self, before: OperationRef) {
```
bitwalker: Same issue here re: aliasing

hir/src/ir/operation.rs
```diff
@@ -1033,54 +1154,20 @@ impl Operation {
...
-        {
-            let mut block = block.borrow_mut();
-            let block_body = block.body_mut();
-            let mut cursor = unsafe { block_body.cursor_mut_from_ptr(before) };
-            cursor.insert_before(self.as_operation_ref());
-        }
+        self.as_operation_ref().insert_before(before);
     }
 
     pub fn insert_after(&mut self, after: OperationRef) {
```
bitwalker: Same issue here re: aliasing

bitwalker (changes requested):
Looks good! There are a couple of issues/nits that we need to address before merging, but they are small things that you were largely already aware of I think, based on your last comment.

If you could also rebase on `next` and squash your commits, I'll get this merged ASAP

hir/src/ir/operation.rs
lima-limon-inc: > This can be considered infallible if `implements` returns true - so `unwrap` is fine here. All kinds of things would be broken if that wasn't the case, so it just obscures what is going on this function to have an expect message IMO.

Noted! I'll keep it in mind

lima-limon-inc: I updated them both in this commit: https://github.com/0xMiden/compiler/commit/daafae4ad14cd9d8878481a5417dab5c0a0e85aa

I also added a small comment to explain the reasoning behind it.

hir/src/ir/operation.rs
lima-limon-inc: Answered here: https://github.com/0xMiden/compiler/pull/546#discussion_r2200879817

hir/src/ir/operation.rs
lima-limon-inc: > unable to insert '{this.name()}' in symbol table of '{parent.name()}': symbol '{symbol_ref.borrow().name()}' is already registered to another operation

Good point, I added it here: https://github.com/0xMiden/compiler/pull/546/commits/6c3f9eb56f79b35caa36d475a265aa04ebfb761a. 
Additionally, I also added the name of the operation that is already in the symbol table, like so: https://github.com/0xMiden/compiler/pull/546/commits/1fb680211d1a6693c7928f3ab547d8a2693d0f30

What do you think?

hir/src/ir/operation.rs
lima-limon-inc: Changed here: https://github.com/0xMiden/compiler/pull/546/commits/18518f50d171cc9409f871aeec6fed7df854502f

hir/src/ir/operation.rs
lima-limon-inc: Noted, great point. I changed it here: https://github.com/0xMiden/compiler/pull/546/commits/27f333fd1e69713b5cb5764315fbc8d593048d0c

What do you think? Is the logic too deeply nested or does it look fine?

hir/src/ir/operation.rs
lima-limon-inc: Also changed here: https://github.com/0xMiden/compiler/pull/546#discussion_r2201430227

hir/src/ir/operation.rs
lima-limon-inc: Also changed here: https://github.com/0xMiden/compiler/commit/18518f50d171cc9409f871aeec6fed7df854502f

hir/src/ir/operation.rs
lima-limon-inc: I changed all of these in this commit: https://github.com/0xMiden/compiler/pull/546/commits/a80fc2f611bcd83b7141fcbaa0739c58920bd036

hir/src/ir/operation.rs
lima-limon-inc: Changed here: https://github.com/0xMiden/compiler/pull/546#discussion_r2201439375

hir/src/ir/operation.rs
lima-limon-inc: Changed here: https://github.com/0xMiden/compiler/pull/546#discussion_r2201439375

hir/src/ir/operation.rs
lima-limon-inc: Changed here: https://github.com/0xMiden/compiler/pull/546#discussion_r2201439375

hir/src/ir/operation.rs
lima-limon-inc: Changed here: https://github.com/0xMiden/compiler/pull/546/commits/1ecfbb58b9acb0286e00fe323af2257588a4b41c

hir/src/ir/operation.rs
lima-limon-inc: Noted, completely agree. I removed them here: https://github.com/0xMiden/compiler/commit/21a90c54819abbde86da5f2243bff30ec1121fb6

hir/src/ir/operation.rs
lima-limon-inc: Also addressed here: https://github.com/0xMiden/compiler/pull/546#discussion_r2201487016

hir/src/ir/operation.rs
lima-limon-inc: Also addressed here: https://github.com/0xMiden/compiler/pull/546#discussion_r2201487016

hir/src/ir/operation.rs
lima-limon-inc: Also addressed here: https://github.com/0xMiden/compiler/pull/546#discussion_r2201487016

bitwalker (approved):
Looks good! Sorry it took awhile for me to get back and do a final pass on this!

lima-limon-inc:
> Looks good! Sorry it took awhile for me to get back and do a final pass on this!

No worries! Thank you for your helpful feedback ^_^

Issue #543, opened 2025-05-28
Compilation test for basic-wallet and p2id note-script examples


greenhat:
After #538

1. Move both basic-wallet and p2id projects from `tests/rust-apps-wasm/rust-sdk/` to `examples`;
2. Add a test compiling them.

greenhat:
Done in https://github.com/0xMiden/compiler/pull/549

PR #549, opened 2025-05-30
Compile basic-wallet and p2id note scripts

### LINKED ISSUE #551: Ensure the data segments are word(16 byte) aligned
Following https://github.com/0xMiden/compiler/pull/549#discussion_r2116503896

greenhat:
Ref #543
Close #551

This PR:
- Moves basic-wallet and p2id note script projects to the `example` folder along with an integration test;
- Add data segment aligning and merging in the frontend;
- Fixes import module names to be in sync with the WIT files (Miden SDK).


codegen/masm/src/lower/component.rs
```diff
@@ -316,7 +316,7 @@ impl MasmComponentBuilder<'_> {
             self.init_body
                 .push(Op::Inst(Span::new(span, Inst::SysEvent(masm::SystemEventNode::PushMapVal))));
             // write_ptr
-            assert!(rodata.start.addr.is_multiple_of(4), "rodata segments must be word-aligned");
+            assert!(rodata.start.is_word_aligned(), "rodata segments must be word-aligned");
```
greenhat, line 319: @bitwalker I'm hitting this assert in p2id note script compilation due to multiple data segments in core Wasm module (see https://github.com/0xMiden/compiler/blob/greenhat/i543-compile-basic-wallet/tests/integration/expected/examples/p2id.wat#L611-L612)
The second `.data` is not word (16-byte) aligned. 
I'm trying to figure out what is the reason for having this assert. So far, my conclusion it that it prevents adjacent data segment overwrite when loading them from the advice provider, which loads words. Is this correct?
I've tried the `--merge-data-segments` option passed to `lld` with no success. I'm thinking of merging them in the frontend. How does it sound?

bitwalker: The assert was to catch if we ever observed rodata segments that weren't at least word-aligned, since handling under-aligned segments requires additional special handling that we do not currently have. At the time, my assumption is that generally LLVM was going to place data segments on a page-aligned boundary, which would by definition guarantee they are word-aligned, but it appears that isn't always the case - but that's why the assert was added, so we would be immediately aware of any program which violated that assumption.

The correct solution here, IMO, is as follows:

1. If the data segment is >= word-aligned, no special handling is required
2. If the data segment is < word-aligned, pad the start of the segment to the nearest word-aligned address. This requires zero-padding the data as well, so that the hash of the segment is correct, and the unhashing of the data works as expected.
  a. If padding the segment would cause the segment to overlap with a previous segment, then the segments should be merged, filling the space between the end of the lower-addressed segment and the start of the subsequent segment with zeros.
  b. If padding the segment does not result in overlap, then we're done.

I think doing this check/transformation in the frontend makes the most sense, as you've guessed, since we only really care about keeping segments compact (i.e. merging all segments could result in a lot of excess data being hashed/unhashed which is just zeros, if the intervening space between segments is large). It isn't important to preserve the unmerged segment information if we determine that merging is required for alignment reasons.

I should note that the requirement for word-aligned segments is driven off the requirements of the instructions/procedures which are used to hash/unhash the data into memory.

greenhat: Thank you! Done in https://github.com/0xMiden/compiler/pull/549/commits/1d7f2b21a66a7b20b84a2c747caca5086cf5f074

frontend/wasm/src/module/build_ir.rs
```diff
@@ -203,19 +206,39 @@ fn build_data_segments(
     module_builder: &mut ModuleBuilder,
     diagnostics: &DiagnosticsHandler,
 ) -> WasmResult<()> {
+    // First, collect all data segments into ResolvedDataSegment structures
+    let mut resolved_segments = Vec::new();
```
bitwalker: *Nit:* I'd use `SmallVec` for things like this, especially since most modules we'll compile will have < 3 segments. The less temporary heap allocation we do during compilation, the better.

frontend/wasm/src/module/data_segments.rs
```diff
@@ -0,0 +1,266 @@
...
+            16 - (self.offset % 16)
+        }
+    }
+
+    /// Align this segment to word boundary by adjusting offset and prepending zeros
+    pub fn align_to_word_boundary(&mut self) {
+        let padding = self.padding_needed();
+        if padding > 0 {
+            // Adjust offset down to nearest word boundary
+            self.offset = self.offset - (self.offset % 16);
```
bitwalker: *Question:* Is there any reason why we can't just do `self.offset -= padding`? I would expect that the padding reflects the adjustment to the offset required to achieve word-alignment, and then we avoid effectively recomputing the value of `padding` here.

greenhat: Great question! I found an error in `padding_needed()`. It was calculated for offset adjustment up and not down. I fixed it and use it here in https://github.com/0xMiden/compiler/pull/549/commits/c4a04cd647640baeccdf50f530372a71c5430f51. Thank you!

frontend/wasm/src/module/build_ir.rs
greenhat: Thanks! Done.

greenhat:
@bitwalker All fixed! Please do another round.

frontend/wasm/src/module/data_segments.rs
```diff
@@ -24,20 +26,17 @@ impl ResolvedDataSegment {
...
     /// Align this segment to word boundary by adjusting offset and prepending zeros
     pub fn align_to_word_boundary(&mut self) {
         let padding = self.padding_needed();
         if padding > 0 {
             // Adjust offset down to nearest word boundary
-            self.offset = self.offset - (self.offset % 16);
+            self.offset -= padding;
             // Prepend zeros to maintain data at correct offset
             let mut new_data = vec![0u8; padding as usize];
             new_data.extend_from_slice(&self.data);
```
bitwalker:
```suggestion
```suggestion
            let mut new_data = Vec::with_capacity(padding as usize + self.data.len());
            new_data.resize(padding as usize, 0);
            new_data.extend_from_slice(&self.data);
```

This avoids allocating a `Vec` on the heap with `padding` bytes, and then immediately re-allocating a new `Vec` during `extend_from_slice`. Instead, it allocates once, fills the initial `padding` bytes with zeros, and then fills the remaining capacity with `self.data`.

bitwalker (approved):
LGTM. I've proposed one small suggested change, but am marking this approved so it can be merged by you when you're ready.

frontend/wasm/src/module/data_segments.rs
greenhat: Thank you! Committed.

Issue #550, opened 2025-05-30
Update `BumpAlloc` + SDK Rust bindings to support element-addressable memory


greenhat:
Besides the `BumpAlloc` itself (in `sdk/alloc`) also update `sdk/base-sys/src/bindings/note.rs` SDK bindings.

bitwalker:
@greenhat What do you think about having @otrho tackle this? Would give him some familiarity with parts of the SDK in case we want to have him pick off tasks around that in the near future

PR #553, opened 2025-06-04
Enable `get_inputs` ABI transformation test + fixes

### LINKED ISSUE #552: Enable `test_get_inputs` test for ABI transformation for returning a list via pointer
Before tackling #544, let's make sure we test the `ListReturn` ABI transformation strategy.

greenhat:
Close #552 
Ref #550 

**This PR is stacked on the #549 and should be merged after it**

Note: I suggest reviewing this PR by commits.

This PR:
- fixes "empty" `ret` in the ABI transformation to return intended results
- fix `note` Rust bindings to use element-addressable memory
- updates the `get_inputs` test to use Miden asserts 
- removes pure `cargo` build process support from `CompilerTest` in favor of `cargo-miden`


bitwalker:
> removes pure cargo build process support from CompilerTest in favor of cargo-miden

@greenhat Isn't there still value in being able to orchestrate pure Rust compilation without `cargo-miden`? Invoking `cargo-miden` does simplify some things, but:

* Removes our ability to experiment with Rust compilation without `cargo-miden` in the loop (i.e. we must modify `cargo-miden` to support whatever arbitrary scenario we want to test first).
* Will result in three subprocesses being created for each test rather than just two, since instead of `test -> cargo -> rustc`, we end up with `test -> cargo-miden -> cargo -> rustc`, where `->` here represents creation of a sub-process to perform some work. The overhead for pure Rust tests could add up quickly.

The tradeoff is that maintaining the pure `cargo` build support requires us to be aware of the behaviors that diverge between pure `cargo` builds and `cargo-miden` builds.

Anyway, wanted to get your take on whether you'd still prefer to remove pure `cargo` build support with those considerations in mind (and perhaps you already have considered them).

frontend/wasm/src/miden_abi/mod.rs
```diff
@@ -187,7 +187,11 @@ pub fn define_func_for_miden_abi_transformation(
     func_builder.br(exit_block, results, span).expect("failed br");
     func_builder.seal_block(exit_block);
     func_builder.switch_to_block(exit_block);
-    func_builder.ret(None, span).expect("failed ret");
+    let arg_vals: Vec<ValueRef> = {
```
bitwalker: *Nit:* Another case where `SmallVec` is preferred.

bitwalker (approved):
LGTM, a couple comments, but I've marked this as approved so you can merge when you are ready.

frontend/wasm/src/miden_abi/mod.rs
greenhat: Thanks! Done.

bitwalker:
Sounds good! I agree that keeping it doesn't seem to be worth it then.

PR #555, opened 2025-06-10
Add the test that executes counter contract and note script on the testnet

### LINKED ISSUE #554: Run counter-contract and note example on the local devnet
Before tackling #544 let's fix the simpler example.

greenhat:
Close #554 

This PR:
- adds a new counter contract test executed against the Miden testnet using the client directly;

TODO:
- [ ] fix the `TransactionProvingError(ProvenTransactionBuildFailed(AccountUpdateSizeLimitExceeded { account_id: V0(AccountIdV0 { prefix: 13040896844101492496, suffix: 497889488833280 }), update_size: 72217 }))` error


greenhat:
@bitwalker I've tried to use `Scenario` type for the `counter_contract` test on the testnet, and I stumbled on one lacking feature. When I want to get the note ID when I'm creating the output note so that I can use it later when consuming the note. I haven't figured out an ergonomic way to implement it with its builder-like API. So I temporarily wrote a new test using the Miden client directly.

Besides that, I've been thinking about the target audience for the testing harness, and I think it might be too opinionated for the end users. Meaning that they would have to familiarize themselves with Miden client API anyway when they will be writing their off-chain code. And now we are forcing them to also learn an opinionated framework to write their tests instead of just letting them use the already familiar Miden client API that they've used in their off-chain code.

bitwalker:
@greenhat To be clear, the testing harness in the integration test suite is for _internal_ use only, the public test harness would almost certainly look different, and likely incorporate the efforts by the node and client teams in terms of describing test setup/scenarios.

If the `Scenario` type isn't achieving the goal of allowing us to declaratively express test scenarios at a high level, while hiding the boilerplate/details of the underlying client orchestration, then it isn't suited to its purpose, and I think it should be rewritten/refactored so that it does. I still think writing tests directly against the `miden-client` APIs in our integration test suite is far too low-level at this point in time, and is going to cause a lot of churn as that API evolves. Ideally we could keep that churn contained in a single place, more or less, but it really depends on if there are repeated patterns in the tests.

greenhat:
@bobbinth @bitwalker @partylikeits1983 
I dug into the order of the word elements on the stack in the stdlib and `miden-base` API. It turned out that the order of the felts on the stack when passing a word is consistent across the limited bindings we have for the standard library API and `miden-base` API. The word is expected with the least significant felt on top of the stack, i.e., `[f0, f1, f2, f3]` word is expected to be as `[f3, f2, f1, f0]` on the stack. For some reason I assumed the opposite.

However, I found that `hash_1to1` expects the 8 4-byte parts(felts) in the order with the most significant bytes on top of the stack. So it seems, when it comes to multiple felts that are not words the order expected on the stack is different.

In the upcoming type signatures and automated Rust bindings generation we need to differentiate words from multiple felts and encode the words in reverse order in the Rust bindings.

tests/integration-node/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+midenc-frontend-wasm.workspace = true
+midenc-session.workspace = true
+proptest.workspace = true
+rand = "0.9"
+temp-dir = "0.1"
+tokio.workspace = true
+uuid = { version = "1.10", features = ["v4"] }
+
+# For accessing the test builder from the main integration tests
+miden-integration-tests = { path = "../integration" }
```
bitwalker, line 35: Maybe we should move shared test code like that into a `midenc-testing` crate, and then share it between `miden-integration-tests` (which we should probably rename to `midenc-integration-tests) and this crate?

Having this crate depend on `miden-integration-tests` sort of defeats the point of splitting it out into its own crate IMO.

tests/integration-node/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "miden-integration-node-tests"
```
bitwalker, line 2: Let's name this `midenc-integration-network-tests`, since the tests aren't really testing the node (which the name kind of implies), instead we're testing programs against a local version of the network (i.e. testnet)

We should use the `midenc-` prefix for any compiler crates, even if they aren't published, since the `miden-` prefix is heavily used by other parts of the project, and it helps us to avoid naming conflicts, and provides a bit more context for where the crate fits. Ideally we'd namespace all Miden crates better than we do now, but that ship has mostly already sailed, the compiler is really the only one that has been consistent about using its own prefix.

tests/integration-node/Cargo.toml
greenhat: Good point! I agree.

tests/integration-node/Cargo.toml
greenhat: > Let's name this `midenc-integration-network-tests`, since the tests aren't really testing the node (which the name kind of implies), instead we're testing programs against a local version of the network (i.e. testnet)

I agree the name is a bit misleading. When I see the `network` in the name I'm thinking of the tests that are relying on the external services. Also, testnet means for me the "official" instance of the chain running by the creators for the testing purposes. So when I bootstrap the local node in my test I'm creating the new instance of the chain on every test run. For some reason "local node" describes it for me the best. How about `midenc-integration-local-chain-tests`?

greenhat:
I'd like to merge it as is. I made https://github.com/0xMiden/compiler/issues/629 to address the comments in the next PR.

PR #556, opened 2025-06-11
chore: remove lto = "thin" from release profile

lima-limon-inc:
Issue raised in this conversation: https://github.com/0xMiden/midenup/pull/5#discussion_r2141001835


lima-limon-inc:
It doesn't, the default lto for release is false.
Source: https://doc.rust-lang.org/cargo/reference/profiles.html#release

PR #564, opened 2025-06-17
fix: delayed registration of scf dialect causes canonicalizations to be skipped

### LINKED ISSUE #563: SCF canonicalisation rewriters may never be invoked.
As a blunt test I put a false assertion into each of the constructors (i.e., `new()` methods) for every rewrite implementer under `dialects/scf/src/canonicalization` and ran the full test suite with `cargo make test`.

None of the assertions triggered, implying they are probably not registered properly, nor invoked by the canonicaliser.

At the very least it seems to imply they are never tested, even by the integration tests.

bitwalker:
When the `Canonicalizer` pass is initialized by the `PassManager`, it visits each of the dialects registered with the `Context` at that time, and populates the set of canonicalization patterns for all operations registered with that dialect. Once constructed, any dialect registrations after that point will not have the canonicalization patterns of their operations included in the set applied by the `Canonicalizer`.

The registration of the SCF dialect was occuring implicitly, when the `CFGToSCFTransform` pass executed and encountered a `cf` dialect operation to transform. This is well after the point where the compiler's pass pipeline is instantiated and initialized. As a result, the canonicalization patterns of the SCF dialect were never registered with the `Canonicalizer`, because the SCF dialect itself was not registered at the point where the `Canonicalizer` performs its own initialization.

This change modifies the two existing transform passes such that they force registration of the dialects they operate on. Really only the `CFGToSCFTransform` pass is relevant here, as it is the only current dialect that has its registration delayed so long, which is a short-term consequence of the frontend not emitting any SCF ops.

Closes #563


otrho (approved):
Great!  I wasn't crazy when I couldn't get the rewriters to run. :)

Looking at the test diffs it would seem it'd be useful to have a final renamer pass on the IR just to avoid all the noisy tiny value name diffs.

PR #572, opened 2025-06-23
Invoke account's `init` in the lifted function prologue

### LINKED ISSUE #560: Call `init` to load data segments in the prologue of the generated lifted component export fuction
Discovered while working on #555

greenhat:
Close #560


greenhat:
@bitwalker I'm not sure if I understand what test scenario should be constructed in the IR for multiple data segments you described on our call yesterday. My fix at https://github.com/0xMiden/compiler/pull/572/commits/64bb1aba63f74e0ed0102ed9ba31e5f7622510c9 fixes the global table offset to be after the last data segment instead of putting the globals on the default reserved memory offset. The default reserved memory offset used when inserting globals coincided with the data segment offset, leading to globals overwriting the data segment part. 
I assume that the last data segment has the largest offset, thus, no matter how many data segments we have the globals will end up after all the data segments.

bitwalker:
> @bitwalker I'm not sure if I understand what test scenario should be constructed in the IR for multiple data segments you described on our call yesterday. 

My suggestion was that we create a test that does the following:

* Creates a basic component skeleton in HIR
* In randomized order (e.g. using `[]::shuffle`), insert N non-overlapping data segments (where N is `1..=3`) and 2 global variables. Where one of the global variables is a pointer to data in one of the inserted data segments, one of them is just a simple integer value (like a counter basically) - the differing types here are just to represent global variables that have a dependency on a data segment vs those that don't.
* Emit a program entrypoint in the component module that reads the value of both global variables and asserts that their values match what is expected. NOTE: The global variable referencing a data segment requires that the data segment contain some non-zero data at the address referenced by the global variable so that this test is useful.
* Compute the data/global layout for the component, and verify that none of the computed layouts overlap, and that the global variables are always laid out following the last data segment, and that in the case of the global variable that is a pointer; that the computed address of that pointer refers to a location in the corresponding data segment.
* Convert the component to MASM and execute it to verify that all of the test assertions in the emitted code pass.

The idea here is to capture a few different details:

1. That the order in which data segments and global variables are encountered during lowering does not affect the correctness of the data layout computed by the backend. If the order of the global variables is random, we'd expect that the order they appear in the final layout could differ, but it should never be the case that they overlap with any of the data segments that were defined.
2. That the number of data segments does not affect the correctness of the data layout
3. That whether a global variable references data in a data segment or not does not affect the correctness of the data layout (we currently don't do anything that would cause a difference in behavior, but in the future we may, and this will ensure we catch if any changes in behavior break the data layout).
4. That the computed layout information is correctly reflected in the compiled code, and verified by executing that code.
5. Catch future regressions

> My fix at [64bb1ab](https://github.com/0xMiden/compiler/commit/64bb1aba63f74e0ed0102ed9ba31e5f7622510c9) fixes the global table offset to be after the last data segment instead of putting the globals on the default reserved memory offset. 

So this is what it _should_ have been doing, but as you found, the bug was that after we have visited the component and found all the data segments and global variables, even though we correctly compute the new global table offset,  we never actually adjust the offsets of the inserted global variables when the new global table offset is greater than the initially computed global table offset (based on the reserved pages). Just doing that alone is sufficient to ensure that all of the globals never overlap with any of the data segments.

> I assume that the last data segment has the largest offset, thus, no matter how many data segments we have the globals will end up after all the data segments.

That isn't actually the case - the data segments can appear in any order. The fix here is to simply determine if the result of computing a possibly-modified global table offset, as performed by this bit:

```rust
        // 3. Layout global variables in the next page following the last data segment
        let next_available_offset = self.segment_layout.next_available_offset();
        self.globals_layout.global_table_offset = core::cmp::max(
            (self.reserved_memory_pages * self.page_size).next_multiple_of(4),
            next_available_offset,
        );
```

Is different than the initially-computed value of `self.globals_layout.global_table_offset`, and if so, we just need to _also_ revisit all of the entries in `self.globals_layout.offsets` and adjust the offsets by the difference. So if `next_available_offset` happens to be 64 bytes larger than `self.globals_layout.global_table_offset`, we would adjust all of the entries in `self.globals_layout.offsets` up by 64.

codegen/masm/src/artifact.rs
```diff
@@ -330,6 +330,7 @@ impl MasmComponent {
             let mut block = masm::Block::new(span, Vec::with_capacity(64));
             // Invoke component initializer, if present
             if let Some(init) = self.init.as_ref() {
+                dbg!("invoke");
```
bitwalker: *Nit:* Looks like you left this in unintentionally

codegen/masm/src/linker.rs
```diff
@@ -63,93 +63,92 @@ impl LinkInfo {
...
-    pub fn link(mut self, component: &builtin::Component) -> Result<LinkInfo, LinkerError> {
-        // Gather information needed to compute component data layout
-
-        // 1. Verify that the component is non-empty
-        let body = component.body();
-        if body.is_empty() {
-            // This component has no definition
-            return Err(LinkerError::Undefined);
-        }
+/// Links a component and computes its data layout.
```
bitwalker: I think this change can be reduced to just iterating over all of the entries in the `GlobalVariableLayout` and adjusting the offsets when `next_available_offset` is greater than the initially-computed value of `global_table_offset`.

The linker is currently quite simple, but it's purpose is to encapsulate all of the work done to link a component, so converting this into a more specialized function makes sense in isolation, but not in the broader context of how I expect the linker stage to evolve over time.

greenhat:
> > @bitwalker I'm not sure if I understand what test scenario should be constructed in the IR for multiple data segments you described on our call yesterday.
> 
> My suggestion was that we create a test that does the following:

Got it. Thank you for the detailed explanation! I extracted it into the https://github.com/0xMiden/compiler/issues/574 

> > My fix at [64bb1ab](https://github.com/0xMiden/compiler/commit/64bb1aba63f74e0ed0102ed9ba31e5f7622510c9) fixes the global table offset to be after the last data segment instead of putting the globals on the default reserved memory offset.
> 
> So this is what it _should_ have been doing, but as you found, the bug was that after we have visited the component and found all the data segments and global variables, even though we correctly compute the new global table offset, we never actually adjust the offsets of the inserted global variables when the new global table offset is greater than the initially computed global table offset (based on the reserved pages). Just doing that alone is sufficient to ensure that all of the globals never overlap with any of the data segments.
> 
> > I assume that the last data segment has the largest offset, thus, no matter how many data segments we have the globals will end up after all the data segments.
> 
> That isn't actually the case - the data segments can appear in any order. The fix here is to simply determine if the result of computing a possibly-modified global table offset, as performed by this bit:
> 
> ```rust
>         // 3. Layout global variables in the next page following the last data segment
>         let next_available_offset = self.segment_layout.next_available_offset();
>         self.globals_layout.global_table_offset = core::cmp::max(
>             (self.reserved_memory_pages * self.page_size).next_multiple_of(4),
>             next_available_offset,
>         );
> ```
> 
> Is different than the initially-computed value of `self.globals_layout.global_table_offset`, and if so, we just need to _also_ revisit all of the entries in `self.globals_layout.offsets` and adjust the offsets by the difference. So if `next_available_offset` happens to be 64 bytes larger than `self.globals_layout.global_table_offset`, we would adjust all of the entries in `self.globals_layout.offsets` up by 64.

I'm confused. 

1. `self.segment_layout.next_available_offset()` only analyses the _last_ segment, which might not have the largest offset. I mean, if we have two segments and the last one has an offset smaller than the first one, then we might end up overlapping globals with the first data segment.

2. The `GlobalVariableLayout` is created _after_ we determined the new global table offset, and we insert globals with the correct offset. Why do we need to revisit them and adjust their offsets?

codegen/masm/src/linker.rs
greenhat: > I think this change can be reduced to just iterating over all of the entries in the `GlobalVariableLayout` and adjusting the offsets when `next_available_offset` is greater than the initially-computed value of `global_table_offset`.

Yes, but I don't see it as "reduced". I found creating them after we are done with data segments with correct offsets much simpler and more readable. 

> The linker is currently quite simple, but it's purpose is to encapsulate all of the work done to link a component, so converting this into a more specialized function makes sense in isolation, but not in the broader context of how I expect the linker stage to evolve over time.

I thought about that and decided to refactor it into a function in the separate reversible commit. My reason for the refactor was to avoid adjusting the offsets in the `Linker` state in `link()` and ditch the state altogether. Do you want me to revert the refactor back `Linker`?

codegen/masm/src/artifact.rs
greenhat: Thank you! Removed.

codegen/masm/src/linker.rs
bitwalker: I'd like to keep `Linker`. That said, if you think there is a cleaner way to implement the internals of `Linker` that handle the data segment/global variable layout work, that's fine.

> Yes, but I don't see it as "reduced".

My point was that the actual fix required is much smaller than the change made here, even accounting for the switch from `Linker` to a standalone function. That doesn't mean that the resulting code would be the cleanest solution by any means, but there is a difference between what fixes the bug in question, and the other clean up/refactoring bits. In this case, I'd prefer a refactoring that preserves `Linker` but improves the code as much as possible otherwise.

bitwalker:
> I'm confused.
> 
> 1. self.segment_layout.next_available_offset() only analyses the last segment, which might not have the largest offset. I mean, if we have two segments and the last one has an offset smaller than the first one, then we might end up overlapping globals with the first data segment.

I did a quick check of `DataSegmentLayout`, and it turns out that it does ensure that the internal vector of the layout is ordered by the segment offset when calling `insert`. Thus `next_available_offset` is always looking at the segment with the greatest offset.

This invalidates my previous concern about the segments in the layout possibly not being ordered.

> 2. The GlobalVariableLayout is created after we determined the new global table offset, and we insert globals with the correct offset. Why do we need to revisit them and adjust their offsets?

I was talking about how to modify the implementation _before_ your changes, i.e. the only thing required to fix the original bug is to adjust the offsets in the layout after computing the new global table offset. The original implementation default-initialized the `GlobalVariableLayout` with the reserved pages accounted for, but if the segments added in the layout would extend beyond the reserved space, that would force the global table to be relocated, so we needed to shift the offsets of the globals in the table, but weren't.

Your refactoring just defers constructing the `GlobalVariableLayout` until the end, but the result is the same either way.

Like I mentioned in my other comment, I'm fine with cleaning up the current implementation if it simplifies things, so long as we're also preserving the `Linker` for now.

codegen/masm/src/linker.rs
greenhat: Makes sense. I restored `Linker` and implemented the global offsets update in a separate method.

greenhat:
> > 2. The GlobalVariableLayout is created after we determined the new global table offset, and we insert globals with the correct offset. Why do we need to revisit them and adjust their offsets?
> 
> I was talking about how to modify the implementation _before_ your changes, i.e. the only thing required to fix the original bug is to adjust the offsets in the layout after computing the new global table offset. The original implementation default-initialized the `GlobalVariableLayout` with the reserved pages accounted for, but if the segments added in the layout would extend beyond the reserved space, that would force the global table to be relocated, so we needed to shift the offsets of the globals in the table, but weren't.
> 
> Your refactoring just defers constructing the `GlobalVariableLayout` until the end, but the result is the same either way.
> 
> Like I mentioned in my other comment, I'm fine with cleaning up the current implementation if it simplifies things, so long as we're also preserving the `Linker` for now.

It makes sense to update the global offsets if we keep the `Linker` intact. I implemented it as a `GlobalVariableLayout::update_global_table_offset()`. Please do another round.

PR #575, opened 2025-06-27
Add a rewriter pass for folding redundant yields from SCF control flow.

### LINKED ISSUE #440: Implement SCF canonicalization to elide redundant yields from regions
This issue aims to solve something I've observed in the IR resulting from the CFG-to-SCF lifting transformation. Essentially, the structured control flow operations produce results which correspond to values which escape one/all of the regions of that operation. In some cases, the same value is yielded/returned from all regions, but the definition of the value actually dominates the parent operation of the region(s), i.e. the yield and corresponding result value is unnecessary/redundant, and can be elided completely, simply using the original value.

## Example

This is an example of something I've actually observed, and while a bit pathological, exhibits the problem perfectly:

```
v0 = ... : i32
v1 = ... : i32
v2 = scf.index_switch v0 {
    case 1: {
        scf.yield v1
    }
    case 2: {
        scf.yield v1
    }
    default: {
        scf.yield v1
    }
}
...
```

In this case, all uses of `v2` could be replaced with `v1`, and the entire `scf.index_switch` could be elided, not just the redundant `scf.yield` argument and `scf.index_switch` result.

## Approach

This should be implemented as a canonicalization over any operation that implements `RegionBranchOpInterface`, as opposed to a canonicalization of a specific concrete operation, similar to how the `SplitCriticalEdges` canonicalization is defined to apply to any `BranchOpInterface` op (see the pass in `dialects/cf/src/canonicalization/split_critical_edges.rs` for more details). We will want to instantiate the canonicalization on a concrete operation though (i.e. see how `SplitCriticalEdges` is used as an example). In other words, the canonicalization is generic, but the instantiation is not.

**NOTE:** Define the canonicalization in the `dialects/scf` crate.

Roughly, we want to do the following:

* Construct the set of operand/result pairs which are returned from every child region which branches back to the parent operation. The set of such regions can be obtained via the `RegionBranchOpInterface` trait's `get_successor_regions` method, by passing `RegionBranchPoint::Child(current_region)`, and filtering out any regions for which the resulting successor region iterator does not consist solely of a successor for which `is_parent` returns true. The results produced by the parent operation as a result of that branch can be obtained from the successor object via its `successor_inputs` method. The successor operands must be obtained by querying for the terminator op of the region entry block (our SCF ops only allow single-block regions), and casting it to a reference of `RegionBranchTerminatorOpInterface`, and calling `get_successor_operands(RegionBranchPoint::Parent)` on it. You can then zip the two together, construct a tuple of the values, and construct a `SmallSet` from them which can be unioned across all matching regions. In addition to the yield/result set, you'll want to keep track of the terminator operation ref, as it will need to be rewritten if the canonicalization can be applied. It may be necessary to have an auxiliary data structure to track the actual `OpOperand` that was yielded, as we must ensure that when we later are removing an operand, that we remove the specific instance of a value, not every use of a value. For computing the sets though, we want the `ValueRef` corresponding to the operand.

* Compute the set intersection across all the returned sets. If the resulting intersection is empty, there is nothing to do.

* Otherwise, we need to prune the set of any pairs where the yielded value does not dominate the parent operation. You can do this by calling `parent_region` on the yielded value (which must be set), and then checking if that region is in the set of regions belonging to the parent operation.

* At this point, every result/yield pair in the set represents a redundancy. We first need to replace all uses of the result with the yielded value; then remove the result from the parent op; then visit each of the terminator ops and remove the corresponding yield. Ensure that you use `rewriter.replace_all_uses_of_value_with` when replacing uses of the redundant results.

* If _all_ of the operation results are unused, _and_ `op.is_memory_effect_free() == true`, then we can remove the entire operation, rather than just the unused yields/results. This can be done instead of rewriting the results individually, by using `rewriter.replace_op_with_values`, which will both erase the operation and rewrite all uses of the results with the provided values.

**NOTE:** When rewriting the terminator ops, ensure that you precede the modification with a call to `let _guard = rewriter.modify_op_in_place(terminator)` so that the patten rewrite driver knows to revisit the modified operations. You don't need to do this for the parent operation itself though.

## Testing

For now, we're largely just looking to verify the following:

* The canonicalization applies when we expect it to:
  1. When there is a mix of redundant and not-redundant results/yields across every region
  2. When all results/yields are redundant, the op has no memory effects, and thus the entire op is redundant
* The canonicalization does not apply when only a subset of yields return the same value to the parent
* The canonicalization is applied, but does not erase the op, if the op has memory effects, even if all results/yields are redundant. It should simply erase the redundant results.

We don't have a ton of testing infrastructure set up for the new IR just yet, but expect tests are a good option here. You can use the tests in `dialects/scf/transforms/cfg_to_scf.rs` as a reference. If you see something that would be useful to add on this front though, feel free to prototype it here, and we can incorporate those changes into other future tests.

## Notes

The following are just a few notes I think are important to keep in mind while working on this:

* All of our SCF operations consist of single-block regions, and those regions must be terminated by a `RegionBranchTerminatorOpInterface` operation. It is fine to raise an assertion if this assumption is violated, since it should've been caught elsewhere.
* If a value yielded from a child region is not defined in any region of the parent operation, by definition the value definition must dominate the parent operation, so we don't need a dominator tree for this rewrite.
* Make use of `ValueRange` to coerce ranges of different value-like types to a `ValueRef`. For example, block arguments and op results both are implementations of `Value`. Operands can also be coerced to a `ValueRef` (as the operand represents a use of a block argument/op result). You can always downcast a `ValueRef` to its concrete type if needed.
* When removing a result and the corresponding yield, you cannot remove the yielded operand by `ValueRef` identity, you must use the `OpOperand` identity, as multiple uses of the same value can be present, and you only want to remove the specific operand in question. * Related to the above, you cannot use the index of the result to infer the index of the corresponding yield, as some region terminators have two sets of operands, e.g. `scf.condition` which takes a boolean flag and then the set of values to yield to the successor.
* Many common APIs are exposed via the `Op` trait, which is a supertrait of `RegionBranchOpInterface`, but you may also want to access functions only exported by the underlying `Operation` type, which you can get a reference to via the `as_operation` trait method.
* There are a few confusing aspects to `RegionSuccessor`/`RegionBranchPoint` and some of their exported methods. `RegionBranchPoint` can refer both to the source of a branch _and_ the target of a branch, depending on the context. `RegionSuccessor::successor_inputs`, when called in the context of returning to a parent operation, currently returns the set of results of the parent operation which are defined as the result of the return, _not_ the set of successor operands at the region terminator. I'm planning on improving the terminology around this in the near future, but for now it is just something to keep in mind.
* If you notice any missing helpers/etc which feel like they should exist (I found a few just writing this up), feel free to implement them. For example, `ValueRange` should implement `Clone`, but for some reason I forgot to add the implementation of it. Ideally, `RegionSuccessor` would also provide a convenience for obtaining the predecessor operation when representing a branch from a child region to the parent, since it is quite common to want that.

otrho:
At this stage this is only `scf.if` and `scf.index_switch`.  It will try to replace the entire operation if all yields within are of the same values, or it will cull those yields down to only those which differ per region.

Something similar could be applied to `scf.while` but while I tried to get it going, it was both more complicated and felt like it was addressing a corner case.  Finding redundant values between the 'before' and 'after' blocks might be worthwhile, though it would be unlikely the compiler would generate such values.  And then lifting out redundant yields (of values which dominate the `while`) could also be done.

To test the output I've created a pass which runs a single rewriter.  It's handy to use a pass manager to run it because a) the `LiftControlFlowToSCF` transform can be run on `cf` input before calling the rewriter, and then b) it handles the greedy iteration nicely.

This pass and the helper functions to run it could be lifted into the parent `scf/lib.rs` or somewhere deeper if it were to be used for testing elsewhere.

Closes #440.

### DIFF
--- a/dialects/scf/Cargo.toml
+++ b/dialects/scf/Cargo.toml
@@ -25,6 +25,5 @@ midenc-hir-transform.workspace = true
 bitvec.workspace = true
 
 [dev-dependencies]
-midenc-dialect-hir.workspace = true
 midenc-expect-test.workspace = true
 env_logger.workspace = true
--- a/dialects/scf/Cargo.toml
+++ b/dialects/scf/Cargo.toml
@@ -25,5 +25,6 @@ midenc-hir-transform.workspace = true
 bitvec.workspace = true
 
 [dev-dependencies]
+midenc-dialect-hir.workspace = true
 midenc-expect-test.workspace = true
 env_logger.workspace = true
--- a/dialects/scf/src/canonicalization/fold_redundant_yields.rs
+++ b/dialects/scf/src/canonicalization/fold_redundant_yields.rs
@@ -140,7 +140,7 @@ impl RewritePattern for FoldRedundantYields {
                 let mut new_opands = SmallVec::<[ValueRef; 4]>::default();
 
                 // Make a copy of the old operands, except for the redundant value, except in the
-                // case where they're all redudant where we need no operands.
+                // case where they're all redundant where we need no operands.
                 if !all_results_are_redundant {
                     for old_opand in term_op.borrow().operands().iter() {
                         if !redundant_result_positions.contains(&old_opand.index()) {
--- a/dialects/scf/src/canonicalization/fold_redundant_yields.rs
+++ b/dialects/scf/src/canonicalization/fold_redundant_yields.rs
@@ -68,6 +68,13 @@ impl RewritePattern for FoldRedundantYields {
                 "All region block terminators must impl RegionBranchTerminatorOpInterface.",
             );
 
+            // For now we only support regions with a simple `yield` terminator.  This may change
+            // in the future if/when we support other RegionBranchOps (e.g., `while`).
+            let term_op_name = term_op.as_operation().name();
+            if term_op_name.dialect() != "scf" || term_op_name.name() != "yield" {
+                return Ok(false);
+            }
+
             // Save the terminator and each of its opands paired with their indices.
             term_ops.push(term_op_ref);
             region_yields.push(
--- a/dialects/scf/src/canonicalization/fold_redundant_yields.rs
+++ b/dialects/scf/src/canonicalization/fold_redundant_yields.rs
@@ -170,14 +170,13 @@ impl RewritePattern for FoldRedundantYields {
 mod tests {
     use alloc::{boxed::Box, format, rc::Rc, sync::Arc, vec::Vec};
 
-    use builtin::{BuiltinOpBuilder, FunctionBuilder};
-    use midenc_dialect_arith::ArithOpBuilder;
     use midenc_dialect_cf::{ControlFlowOpBuilder, SwitchCase};
-    use midenc_dialect_hir::HirOpBuilder;
     use midenc_expect_test::expect_file;
     use midenc_hir::{
-        dialects::builtin,
-        pass,
+        dialects::{
+            builtin::{self, BuiltinOpBuilder, FunctionBuilder},
+            test::TestOpBuilder,
+        },
         pass::{Pass, PassExecutionState},
         patterns::{
             FrozenRewritePatternSet, GreedyRewriteConfig, RewritePattern, RewritePatternSet,
@@ -316,14 +315,14 @@ mod tests {
         let if_sum_rhs = builder.append_block_param(if_final, Type::U32, span);
 
         let input = builder.current_block().borrow().arguments()[0].upcast();
-        let redundant_val = builder.u32(11, span);
+        let redundant_val = builder.u32(11, span)?;
 
-        let zero = builder.u32(0, span);
+        let zero = builder.u32(0, span)?;
         let is_zero = builder.eq(input, zero, span)?;
         builder.cond_br(is_zero, if_then, [], if_else, [], span)?;
 
         builder.switch_to_block(if_then);
-        let then_non_redundant_val = builder.u32(22, span);
+        let then_non_redundant_val = builder.u32(22, span)?;
         builder.br(if_final, [redundant_val, then_non_redundant_val], span)?;
 
         let switch_on_one_case = SwitchCase {
@@ -348,15 +347,15 @@ mod tests {
         )?;
 
         builder.switch_to_block(switch_on_one_block);
-        let switch_on_one_non_redundant_val = builder.u32(33, span);
+        let switch_on_one_non_redundant_val = builder.u32(33, span)?;
         builder.br(if_final, [redundant_val, switch_on_one_non_redundant_val], span)?;
 
         builder.switch_to_block(switch_on_two_block);
-        let switch_on_two_non_redundant_val = builder.u32(44, span);
+        let switch_on_two_non_redundant_val = builder.u32(44, span)?;
         builder.br(if_final, [redundant_val, switch_on_two_non_redundant_val], span)?;
 
         builder.switch_to_block(switch_default_block);
-        let switch_default_non_redundant_val = builder.u32(55, span);
+        let switch_default_non_redundant_val = builder.u32(55, span)?;
         builder.br(if_final, [redundant_val, switch_default_non_redundant_val], span)?;
 
         // Add all the results together along with the redundant value to give them all users.
@@ -399,8 +398,8 @@ mod tests {
         let sum_rhs = builder.append_block_param(switch_final_block, Type::U32, span);
 
         let input = builder.current_block().borrow().arguments()[0].upcast();
-        let redundant_val0 = builder.u32(11, span);
-        let redundant_val1 = builder.u32(22, span);
+        let redundant_val0 = builder.u32(11, span)?;
+        let redundant_val1 = builder.u32(22, span)?;
 
         let switch_on_one_case = SwitchCase {
             value: 1,
@@ -426,7 +425,7 @@ mod tests {
         builder.br(switch_final_block, [redundant_val0, redundant_val1], span)?;
 
         builder.switch_to_block(switch_on_two_block);
-        let zero = builder.u32(0, span);
+        let zero = builder.u32(0, span)?;
         let is_zero = builder.eq(input, zero, span)?;
         builder.cond_br(is_zero, if_then_block, [], if_else_block, [], span)?;
 
@@ -478,10 +477,10 @@ mod tests {
         let sum_rhs = builder.append_block_param(if_final_block, Type::U32, span);
 
         let input = builder.current_block().borrow().arguments()[0].upcast();
-        let redundant_val0 = builder.u32(11, span);
-        let redundant_val1 = builder.u32(22, span);
+        let redundant_val0 = builder.u32(11, span)?;
+        let redundant_val1 = builder.u32(22, span)?;
 
-        let zero = builder.u32(0, span);
+        let zero = builder.u32(0, span)?;
         let is_not_zero = builder.neq(input, zero, span)?;
         builder.cond_br(is_not_zero, if_then_block, [], if_else_block, [], span)?;
 
@@ -562,9 +561,9 @@ mod tests {
 
         let input = builder.current_block().borrow().arguments()[0].upcast();
 
-        let redundant_val11 = builder.u32(11, span);
-        let redundant_val22 = builder.u32(22, span);
-        let redundant_val33 = builder.u32(33, span);
+        let redundant_val11 = builder.u32(11, span)?;
+        let redundant_val22 = builder.u32(22, span)?;
+        let redundant_val33 = builder.u32(33, span)?;
 
         let switch_on_one_case = SwitchCase {
             value: 1,
@@ -593,8 +592,8 @@ mod tests {
         )?;
 
         builder.switch_to_block(switch_on_one_block);
-        let switch_on_one_non_redundant_val0 = builder.u32(100, span);
-        let switch_on_one_non_redundant_val1 = builder.u32(101, span);
+        let switch_on_one_non_redundant_val0 = builder.u32(100, span)?;
+        let switch_on_one_non_redundant_val1 = builder.u32(101, span)?;
         builder.br(
             switch_final_block,
             [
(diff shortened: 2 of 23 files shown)


dialects/scf/src/ops.rs
```diff
@@ -688,3 +694,21 @@ impl RegionBranchTerminatorOpInterface for Yield {
...
+    fn has_no_effect(&self) -> bool {
+        true
+    }
+
+    fn effects(&self) -> EffectIterator<::midenc_hir::effects::MemoryEffect> {
+        EffectIterator::from_smallvec(::midenc_hir::smallvec![])
+    }
+}
+
+impl AlwaysSpeculatable for Yield {}
```
bitwalker: No need to implement this one manually - it's done by adding it to the list of `traits` in the `#[operation]` macro

dialects/scf/Cargo.toml
```diff
@@ -25,5 +25,6 @@ midenc-hir-transform.workspace = true
 bitvec.workspace = true
 
 [dev-dependencies]
+midenc-dialect-hir.workspace = true
```
bitwalker: It looks to me like we don't need this dependency anywhere, so probably should be removed

dialects/scf/src/canonicalization/fold_redundant_yields.rs
```diff
@@ -0,0 +1,959 @@
...
+            for idx in &redundant_result_positions {
+                op.results_mut().group_mut(0).erase(*idx);
+            }
+
+            // And remove the redundant terminator operands.
+            for mut term_op in term_ops {
+                let mut new_opands = SmallVec::<[ValueRef; 4]>::default();
+
+                // Make a copy of the old operands, except for the redundant value, except in the
+                // case where they're all redudant where we need no operands.
```
bitwalker:
```suggestion
                // case where they're all redundant where we need no operands.
```

dialects/scf/src/canonicalization/fold_redundant_yields.rs
```diff
@@ -0,0 +1,959 @@
...
+                        if !redundant_result_positions.contains(&old_opand.index()) {
+                            new_opands.push(old_opand.borrow().as_value_ref());
+                        }
+                    }
+                }
+
+                // Update the terminator with the new operands.
+                let _guard = rewriter.modify_op_in_place(term_op);
+                let mut term_op_mut = term_op.borrow_mut();
+                term_op_mut.set_operands(new_opands);
```
bitwalker, line 161: There is a bug hiding here. We're assuming that the terminator here is an `scf.yield`, or more precisely, that the terminator has no other non-forwarded operands, and that all of the operands are in a single operand group; which is the case for `scf.yield`, but not for all of the SCF terminator ops.

While we're not actually supporting `scf.while` with this currently (and maybe never), it makes for a good example of why the assumption here isn't safe. The `before` region of `scf.while` always terminates with an `scf.condition`, rather than `scf.yield`; and it has two operand groups - the first holding the condition, and the second holding all of the operands forwarded to the successor region. If we were to set the operands of `scf.condition` using `Operation::set_operands` as  we do here, we'd end up breaking the op's structure, because `set_operands` erases all existing operands and operand groups, and populates a single operand group with all of the given operands. Further, the condition operand of `scf.condition` is not actually forwarded, i.e. it is consumed by the `scf.condition` itself, so I'm assuming that we'd actually be unintentionally dropping the condition operand above because it would not be contained in `region_yields`, and so not be present in the set of operands provided to `set_operands` here.

Because this op is implemented for `RegionBranchOpInterface`, we shouldn't make any assumptions about which concrete ops it will be applied to (as those could change over time). We _could_ reject any op whose regions do not all terminate with `scf.yield` so that the assumptions we're making here hold, and that would be acceptable because it would still be applicable to any `RegionBranchOpInterface`, it just would fail to match things like `scf.while` (again, totally fine).

If we don't do that, then I think we'd need to be more precise about how we interact with the operands of the terminator. Namely, ensure that we also track the operand group for each operand, and use `OpOperandRangeMut::set_operands` on each operand group that is represented in `region_yields` instead of using `Operation::set_operands`, so that we are preserving the operand structure of the underlying terminator operation.

To be clear, I think the implementation here works fine on `scf.if` and `scf.index_switch`, my issue is more that if we add more structured control flow ops in the future, or add other valid SCF terminators that have a different structure/semantics than `scf.yield`, that this rewrite could end up breaking them unexpectedly if we assume it is safe to apply to them. Fixing the issue I pointed out is more about preserving the notion that we can safely apply this canonicalization to any `RegionBranchOpInterface` op now or in the future without introducing a bug.

bitwalker (changes requested):
Looks great! Nice work!

I did find one issue, that I think we should address before merging. I think it's a simple fix of just rejecting any input op that doesn't terminate all of its regions with `scf.yield`. If we wanted to make it more general, there is another approach that would also work, but I think the more narrow solution is fine for now. See the comment I left for more details on the actual bug.

Otherwise, everything else looks solid to me!

dialects/scf/Cargo.toml
otrho: It's needed to add a memory effect using `store` in one of the tests.  That's why I put it in `dev-dependencies` but if there's a simpler way to add effectful ops to the test without importing `dialect/hir` then sure.  I had a quick look around and only found `ub` stuff and thought `store` is the obvious choice.

https://github.com/0xMiden/compiler/blob/otrho/redundant-yield-elision/dialects/scf/src/canonicalization/fold_redundant_yields.rs#L883-L885

dialects/scf/src/canonicalization/fold_redundant_yields.rs
otrho: Yeah, I deliberately assumed they were yields when dropping `while` support, but you're right, they need to be more general.

But I've implemented the explicit check for `yield` instead for now, although I wasn't sure what the best approach would be.  Instead of checking the terminator implements `RegionBranchTerminatorOpInterface` I could've replaced it with something more concrete?  But instead I left that intact and just abort if the op name isn't `scf.yield`, with a comment.

dialects/scf/Cargo.toml
bitwalker: Ah, I missed that. We have a `test` dialect in the `midenc-hir` crate that we use for this purpose, by adding any ops we need purely for testing purposes; that way we can write better tests in some of the foundational crates.

Here it isn't as essential to use `test` vs one of the other crates - but since the effect infra is defined in `midenc-hir`, it probably makes sense to implement `load` and `store` ops in the `test` dialect so that we can implement tests with effects in `midenc-hir` itself.

dialects/scf/src/canonicalization/fold_redundant_yields.rs
```diff
@@ -0,0 +1,966 @@
...
+            let term_op = term_op.as_trait::<dyn RegionBranchTerminatorOpInterface>().expect(
+                "All region block terminators must impl RegionBranchTerminatorOpInterface.",
+            );
+
+            // For now we only support regions with a simple `yield` terminator.  This may change
+            // in the future if/when we support other RegionBranchOps (e.g., `while`).
+            let term_op_name = term_op.as_operation().name();
+            if term_op_name.dialect() != "scf" || term_op_name.name() != "yield" {
+                return Ok(false);
+            }
```
bitwalker:
```suggestion
            if !term_op.as_operation().is::<crate::YieldOp>() {
                return Ok(false);
            }
```

Much easier to use `Operation::is::<T>()` for this sort of thing 🙂

otrho: Yeah, great, as I was writing that clumsy string compare I was thinking there *must* be a better way to do this...

bitwalker: Shit, sorry, forgot the op name is `Yield`, not `YieldOp` 😬

dialects/scf/Cargo.toml
otrho: OK.  `midenc-hir::dialects::test` already implement `load` and `store` AFAICT but they can't be imported across crates due to the `cfg(test)` gating.

Since every dialect is in its own crate, it would need to implement its own `test` dialect for ops it needs only for testing.  The solution here might be to create a test dialect and `mem_effect` dummy op purely for `scf`..?

otrho:
> Also, sorry about not giving you a heads up about this earlier, but our repos require signed commits - have you set that up before? If not, and you aren't sure how to do it, send me a message on Slack and we'll get it sorted. Once it is set up, you'll have to force git to sign the commits that GitHub shows as not verified (I think partial verification is OK, not 100% on that though).

Looks like it's working now.  Of the 4 commits the two which were taken from your 'suggested change' comments are partially verified, but still green, so it looks like it can merge. :+1:

dialects/scf/Cargo.toml
otrho: @bitwalker I've added a `TestOpBuilder` for `dialects/scf` to replace the use of `midenc-hir` for `store` in the tests.  If you're happy with that final commit then I think it can be merged.  Thanks.

dialects/scf/src/test_ops.rs
bitwalker, line 1: How about we just remove the `#[cfg(test)]` gating of the "real" `test` dialect? There isn't any real reason to have it gated, other than removing things that weren't needed downstream, but we're obviously past that. Then this module can go away entirely.

bitwalker: Once we do that, we can merge this 👍

otrho: I've given this a go and it's failing with a classic OO multiple inheritence problem AFAICT.  My tests are using the `arith` op builder (usually for building conditionals or adding uses to values) and the `midenc-hir` test op builder is also implementing a couple of those ops.

* Weirdly the `.u32()` test op builder returns a `Result` but the `arith` op builder doesn't.  So that should be fixed?
* `.eq()` and `.neq()` are missing from the test op builder, so I can add those.

But this seems to be more work than necessary.  I'm honestly inclined to revert to the original solution I had which just included the `dialects/hir` crate as a `dev-dependencies` in `dialects/scf` to get the `store()` op.  We're including the other 'peer' dialects as actual dependencies anyway, and it's by far the simplest solution.  No need to faff around with `midenc-hir`, no need to implement a `test_ops` crate and everything is consistent.

PR #576, opened 2025-06-30
Expose `hmerge` VM op in Rust

### LINKED ISSUE #573: Expose `hmerge` instruction on the Rust side
### Implementation details

`hmerge` VM operation docs - https://0xmiden.github.io/miden-docs/imported/miden-vm/src/user_docs/assembly/cryptographic_operations.html

We want to expose in our Miden SDK as `fn merge(values: &[Digest; 2]) -> Digest`, where digest is `pub struct Digest(Word)`.

See how stdlib's blake3 hash function is exposed in sdk/stdlib-sys/src/stdlib/crypto/hashes.rs (`hash_1to1`, note return-by-pointer pattern) and make similar functions for `merge` at sdk/stdlib-sys/src/intrinsics/crypto.
Add the low-level(extern) function signature at sdk/stdlib-sys/wit/miden-core-intrinsics.wit (see blake3 function at sdk/stdlib-sys/wit/miden-core-stdlib.wit)
Catch the low-level function import at the Wasm frontend similar to blake3 (`hash-one-to-one` and `hash_1to1`) and translate it into the `HMerge` HIR operation (create it).
In codegen convert the HIR op to MASM `Instruction::HMerge`.

greenhat:
Close #573 

**This PR is stacked on the #587 and should be merged after  #it**

This PR exposes `hmerge` VM op on the Rust side as:
```rust
/// Computes the hash of two digests using the Rescue Prime Optimized (RPO)
/// permutation in 2-to-1 mode.
///
/// This directly maps to the `hmerge` VM instruction.
///
/// # Arguments
/// * `digests` - An array of two digests to be merged. The function internally
///   reorders them as required by the VM instruction (from [A, B] to [B, A] on the stack).
#[inline]
pub fn merge(digests: [Digest; 2]) -> Digest {
```


greenhat:
@bitwalker ping

sdk/stdlib-sys/src/intrinsics/crypto.rs
```diff
@@ -0,0 +1,119 @@
...
+pub fn merge(digests: [Digest; 2]) -> Digest {
+    unsafe {
+        let mut ret_area = ::core::mem::MaybeUninit::<Word>::uninit();
+        let ptr = ret_area.as_mut_ptr() as *mut Felt;
+        // The VM hmerge instruction expects the second digest first, then the first digest
+        // (i.e., [B, A] order when merging digests A and B).
+        // See: https://0xmiden.github.io/miden-docs/imported/miden-vm/src/user_docs/assembly/cryptographic_operations.html
+        let first_digest = &digests[0];
+        let second_digest = &digests[1];
+        extern_hmerge(
```
bitwalker: I wonder if we might not generate significantly better code by having the underlying intrinsic do the load of the two words from the pointer itself, using `mem_loadw`. This approach will result in 8 individual loads of the field elements, so a 4x penalty

tests/integration/src/rust_masm_tests/instructions.rs
```diff
@@ -323,3 +330,84 @@ test_unary_op_total!(bnot, !, u32);
 test_unary_op_total!(bnot, !, u16);
 test_unary_op_total!(bnot, !, u8);
 test_unary_op_total!(bnot, !, bool);
+
+#[test]
+fn test_hmerge() {
```
bitwalker, line 335: Related to my previous comment, this test doesn't really demonstrate the problem I was getting at, because the function is so simple, the optimizer is able to elide the construction of the arrays so that all of the felt arguments to the function are passed through directly to `hmerge`. This is only possible because the arguments are already in the expected order, and the call to `hmerge` is happening without anything else happening in the function other than the logical grouping of the arguments (again, in the same exact order as expected by the `hmerge` intrinsic), so all of the loads/stores that would happen are effectively redundant and get stripped away.

If we instead considered a function that takes two digests and called `hmerge`, I think we'd then see that it must do a lot more work to get things into position for the call to `hmerge`, whereas just passing a pointer to an array to the intrinsic and letting it handle loading things on to the operand stack is going to be a lot more efficient.

To be clear, I'm not saying you have to change this test, I'm just making the observation that we can't use the MASM generated by this test as a useful proxy for the quality of the codegen in general.

bitwalker (approved):
Looks good!

I did have one nit that I think is worth considering carefully before you merge this, in regards to of performance of the generated code. Since hashing is likely to be quite important, and this is a fairly low-level primitive, ensuring that we make it as efficient as possible is worth the time investment IMO. With word-oriented primitives, we're almost always going to be better off using MASM intrinsics to do the heavy lifting of moving things into place on the operand stack, compared to using Rust to do so, since MASM can make use of the word-oriented ops like `mem_loadw` more precisely than we can (especially after things like SROA - scalar replacement of aggregates - has been applied by LLVM, and so we can't easily optimize loads/stores of groups of felts).

sdk/stdlib-sys/src/intrinsics/crypto.rs
greenhat: Good catch! I tried to pass the pointer to `[Digest; 2]` but since the `Digest` is defined using a tuple the in-memory representation is not what can be loaded with `mem_loadw`. When https://github.com/0xMiden/compiler/issues/597 is implemented, we will be able to do it. I made https://github.com/0xMiden/compiler/issues/598 to keep it on the radar.

bitwalker: > but since the Digest is defined using a tuple the in-memory representation is not what can be loaded with mem_loadw

Could you elaborate on this a bit? I would expect a tuple of four felts to have the same layout as an array of four felts, so it is surprising that they actually differ here.

greenhat: You are right, tuple is fine. My brain was so fried after chasing #595, so I misdiagnosed #600 as #597. I repurposed #598 and left it open to use the pointer to `[Digest; 2]` after #600 is fixed.

greenhat:
@bitwalker I made the `hmerge` intrinsic to accept the pointer to the digest and load them via `mem_loadw`. Because of #600, I put in a workaround of constructing an array. I'd like to merge it and come back to it after #600 is fixed. I made #598 to keep on the radar.

PR #587, opened 2025-07-04
fix: `Rodata::bytes_to_elements()` check if `into_remainder()` returns an empty iterator

### LINKED ISSUE #586: `Rodata::bytes_to_elements()` for 32 byte input produces 10 felts (expected 8)
Discovered while working on #576 (data segment hit 32 bytes).

greenhat:
Close #586

There is an issue to address this `into_remainder()` weird behavior - https://github.com/rust-lang/rust/issues/116000

### DIFF
--- a/codegen/masm/src/artifact.rs
+++ b/codegen/masm/src/artifact.rs
@@ -97,18 +97,20 @@ impl Rodata {
         let mut iter = bytes.iter().copied().array_chunks::<4>();
         felts.extend(iter.by_ref().map(|chunk| Felt::new(u32::from_le_bytes(chunk) as u64)));
         if let Some(remainder) = iter.into_remainder() {
-            let mut chunk = [0u8; 4];
-            for (i, byte) in remainder.into_iter().enumerate() {
-                chunk[i] = byte;
+            if remainder.len() > 0 {
+                let mut chunk = [0u8; 4];
+                for (i, byte) in remainder.into_iter().enumerate() {
+                    chunk[i] = byte;
+                }
+                felts.push(Felt::new(u32::from_le_bytes(chunk) as u64));
             }
-            felts.push(Felt::new(u32::from_le_bytes(chunk) as u64));
         }
 
         let size_in_felts = bytes.len().next_multiple_of(4) / 4;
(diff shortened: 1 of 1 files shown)


codegen/masm/src/artifact.rs
```diff
@@ -89,27 +89,29 @@ impl Rodata {
...
         use miden_processor::Felt;
 
         let mut felts = Vec::with_capacity(bytes.len() / 4);
         let mut iter = bytes.iter().copied().array_chunks::<4>();
         felts.extend(iter.by_ref().map(|chunk| Felt::new(u32::from_le_bytes(chunk) as u64)));
         if let Some(remainder) = iter.into_remainder() {
-            let mut chunk = [0u8; 4];
-            for (i, byte) in remainder.into_iter().enumerate() {
-                chunk[i] = byte;
+            if remainder.len() > 0 {
```
bitwalker, line 100: Would this actually fix the bug we saw? I thought we ended up with 10 elements, but it seems to me that neither the code here before or after would've resulted in anything other than a number of elements divisible by 4

bitwalker (approved):
LGTM!

codegen/masm/src/artifact.rs
greenhat: Yes, it fixes the bug. See the 32 byte case test in `test_bytes_to_elements_edge_cases` test below and added `debug_assert_eq!` in the `bytes_to_elements` body. 
As I mentioned and linked the issue in the PR description, `into_remainder()` returns `Some(empty_iterator)` when the remainder is zero. This leads to `chunk` being appended and then `abs_diff` decides that padding is 1 (8 - 9), resulting in 10 elements.

bitwalker: I think the confusion for me is why we're computing the extra padding here - the total size of the data should always be in word-aligned units, so even if we were appending an extra 4-element chunk, I wouldn't expect us to then add padding that makes the resulting number of elements not divisible by 4. The part I don't get, is why adding an empty chunk or not throws that part of the calculation off (and more precisely, my assumption is that how we're computing padding is the problem).

bitwalker (approved):

PR #588, opened 2025-07-04
fix: add empty `struct` support in `#[component]` attribute macro

greenhat:


sdk/base-macros/src/lib.rs
```diff
@@ -378,14 +365,36 @@ pub fn component(
...
+        syn::Fields::Named(fields) => {
+            // Process fields: extract storage info, generate Default parts, update builder
+            let field_inits = match process_fields(fields, &mut acc_builder) {
+                Ok(inits) => inits,
+                Err(e) => return e.to_compile_error().into(),
+            };
+            generate_default_impl(struct_name, &field_inits)
+        }
+        syn::Fields::Unit => {
+            // For unit structs, generate simple Default impl
```
bitwalker, line 379: *Question:* Why not leave this up to the user? I'd be surprised if any macro generated a `Default` impl for me, and it'll cause a confusing error if they attempt to define one.

That's assuming the struct here isn't one we generated of course, but then it'd be odd not to use `derive` ourselves.

examples/basic-wallet/src/lib.rs
```diff
@@ -23,8 +23,9 @@ use bindings::exports::miden::basic_wallet::*;
 
 bindings::export!(MyAccount with_types_in bindings);
 
-use miden::{Asset, NoteType, Recipient, Tag};
+use miden::{component, Asset, NoteType, Recipient, Tag};
 
+#[component]
```
bitwalker, line 28: Per my other comment, I think we should just expect users to add `#[derive(Default)]`, and document that we expect these types to implement `Default`. It'll be a lot clearer to readers of the code then that the trait is implemented for this type - not that I can think of a scenario where they would be manually instantiating it, but just more as a general Rust practice I guess.

sdk/base-macros/src/lib.rs
greenhat: Since we need to implement `Default` when there is a storage field (like in https://github.com/0xMiden/compiler/blob/c24703176cb235f6f231dc34b1d02a39799c6028/examples/storage-example/src/lib.rs?plain=1#L51-L64) we're signaling that we are taking care of the `Default` implementation for accounts. Plus, I suppose this is our answer to "How do I get the account instance?" types of user questions. It'd be consistent to provide the `Default` implementation in all the cases and not only when storage fields are present. 

I agree that it might be confusing for many. Maybe, we should not use `Default` implementation for account instantiation since it's too generic and come up with our own `get_instance()` method/trait?

examples/basic-wallet/src/lib.rs
greenhat: See my comment above.

sdk/base-macros/src/lib.rs
bitwalker: We can discuss on our call today, but I'll merge this for now

PR #589, opened 2025-07-07
expose Miden stdlib `hash_elements` procedure in the Miden SDK

### LINKED ISSUE #583: Expose `hash_memory` stdlib procedure on the Rust side
>And ideally, we'd also have another function exposed that looks like [`Rpo256::hash_elements()`](https://github.com/0xMiden/crypto/blob/next/miden-crypto/src/hash/rescue/rpo/mod.rs#L195) - this can be implemented via a procedure in the standard library (`std::crypto::rpo::hash_memory`).

>And a cherry on top could be adding [`Rpo256::merge_many()`](https://github.com/0xMiden/crypto/blob/next/miden-crypto/src/hash/rescue/rpo/mod.rs#L165), this could also be done with the standard library procedure (`std::crypto::rpo::hash_memory_words`).

The `std::crypto::rpo::hash_memory` docs:
```
#! Inputs:  [ptr, num_elements]
#! Outputs: [HASH]
#!
#! Where:
#! - ptr is the  memory address of the first element to be hashed. This address must be
#!   word-aligned - i.e., divisible by 4.
#! - num_elements is the number of elements to be hashed.
```

greenhat:
Close #583 

This PR:
- Exposes `hash_memory` Miden stdlib procedure as `hash_elements` in the SDK;
- Adds tests to minimally reproduce #595;


sdk/stdlib-sys/src/stdlib/crypto/hashes.rs
```diff
@@ -174,3 +192,41 @@ pub fn sha256_hash_1to1(input: [u8; 32]) -> [u8; 32] {
...
+
+/// Computes the hash of a sequence of field elements using the Rescue Prime Optimized (RPO)
+/// hash function.
+///
+/// This maps to the `std::crypto::rpo::hash_memory` procedure in the Miden stdlib.
+///
+/// # Arguments
+/// * `elements` - A slice of field elements to be hashed
+#[inline]
+pub fn hash_elements(elements: &[Felt]) -> Digest {
```
greenhat: @bitwalker Since Miden stdlib `hash_memory` which we are calling underneath expects a word-aligned pointer to the felts, we need to make sure that `elements` slice is aligned. See my implementation below. It's compiled down to Wasm `memory.copy` which is a byte copy and besides being hugely expensive copying byte by byte it does not work with felts - see https://github.com/0xMiden/compiler/issues/594 

I'm thinking of switching to the `pub fn hash_elements(elements: Vec<Felt>) -> Digest` signature. Since our BumpAlloc is word-aligned, we can assume that `elements` is word-aligned and get away with an assert.

greenhat:
@bitwalker `hash_elements` is working and ready for review. For the optimizations (word- and double word-sized inputs) I made #601 and will implement it after I fix #595.

greenhat:
@bitwalker ping

bitwalker (approved):

PR #590, opened 2025-07-08
refactor: `ExampleCommand` implementation, bug fixes, add comments

greenhat:


examples/basic-wallet/rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "nightly-2025-03-20"
+components = ["rustfmt", "rust-src", "clippy"]
+targets = ["wasm32-unknown-unknown", "wasm32-wasip1", "wasm32-wasip2"]
```
bitwalker: We probably only need `wasm32-wasip2` mentioned here, no? The other targets would never be used, but would take time to install.

examples/p2id-note/rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "nightly-2025-03-20"
+components = ["rustfmt", "rust-src", "clippy"]
+targets = ["wasm32-unknown-unknown", "wasm32-wasip1", "wasm32-wasip2"]
```
bitwalker: Same here

bitwalker (approved):
LGTM. One suggested change, but not important

examples/basic-wallet/rust-toolchain.toml
greenhat: Good catch! I've gone through the examples and removed not-used targets in https://github.com/0xMiden/compiler/pull/590/commits/cd03b72977fb83367c06ef0df51cc6b7968cd90a

PR #591, opened 2025-07-09
refactor: operation verification

bitwalker:
This supercedes #516 as I needed to rebase the PR branch and add a commit on top, but don't have permissions on the lambdaclass repo for that.


docs/src/design/ir.md
```diff
@@ -771,7 +771,8 @@ The primary difference between `Rewriter` and `Builder`, is that the `Rewriter`
 
 A key aspect of the IR design, is to enable as much boilerplate as possible to be generated from the description of an operation, especially when it comes to verifying the validity of an operation using the type constraints of it's operands and results, and the traits/interfaces it implements.
 
-Operations are currently validated when constructed by the `OperationBuilder` that underpins the `BuildableOp` implementation generated by the `#[operation]` proc-macro attribute. There are some known issues with doing it here, so we are likely to revisit this at some point, but for now this is the first place where verification occurs. It may also be triggered manually at any point by calling the `Operation::verify` method.
+Operations are currently validated by the `PassManager`. Before executing the pass pipeline, the `PassManager` does a recursive validation of all the operations to make sure that operations have been built correctly according to their traits. Moreover, after executing each pass, the `PassManager` runs validation a second to make sure that the transformation hasn't broken any operation invariant.
```
greenhat, line 774:
```suggestion
Operations are currently validated by the `PassManager`. Before executing the pass pipeline, the `PassManager` does a recursive validation of all the operations to make sure that operations have been built correctly according to their traits. Moreover, after executing each pass, the `PassManager` runs validation a second time to make sure that the transformation hasn't broken any operation invariant.
```

greenhat: I'm struggling to find where the validation is ran the second time (after each pass). Should it be covered with a test?

lima-limon-inc: > I'm struggling to find where the validation is ran the second time (after each pass).

If I'm not mistaken, this validation is already performed by the compiler, here: https://github.com/0xMiden/compiler/blob/e51c05ab851e128fbb760db09f13bf8627aeb747/hir/src/pass/manager.rs#L976-L977. 
As far as I can tell, `verify` is always set to true, since it set to true in the `PassManager::new()` constructor ([here](https://github.com/0xMiden/compiler/blob/e51c05ab851e128fbb760db09f13bf8627aeb747/hir/src/pass/manager.rs#L103)

> Should it be covered with a test?

I think this is "technically" tested, since the tests call `PassManager::run`. Afterwards, the call stack goes like this:


```
PassManager::run - 1)
└── run_passes - 2)
    └── OpToOpPassAdaptor::run_pipeline - 3)
        ├── initial_verify - 4) (this is the bit that got added in this PR)
        └── OpToOpPassAdaptor::run - 5)
            └── second_verification - 6)
```


1): [Link](https://github.com/0xMiden/compiler/blob/e51c05ab851e128fbb760db09f13bf8627aeb747/hir/src/pass/manager.rs#L117)
2): [Link](https://github.com/0xMiden/compiler/blob/e51c05ab851e128fbb760db09f13bf8627aeb747/hir/src/pass/manager.rs#L163)
3): [Link](https://github.com/0xMiden/compiler/blob/e51c05ab851e128fbb760db09f13bf8627aeb747/hir/src/pass/manager.rs#L181)
4): [Link](https://github.com/0xMiden/compiler/pull/591/files#diff-526b0a945631700bb9ff00bdc261a7c0370d8180e51c958c975b7afa09a55e1bR848-R853)
5): [Link](https://github.com/0xMiden/compiler/pull/591/files#diff-526b0a945631700bb9ff00bdc261a7c0370d8180e51c958c975b7afa09a55e1bR874)
6): [Link](https://github.com/0xMiden/compiler/blob/e51c05ab851e128fbb760db09f13bf8627aeb747/hir/src/pass/manager.rs#L976-L993)

However, do let me know if you'll like me to add a more explicit test.

bitwalker: I think it'd probably be a good idea to add a simple test that runs a pass that modifies the IR by making it invalid. We'd then expect the `PassManager` to raise a verification error if everything is working as expected.

This is being technically tested, but I think it's useful to test our implicit assumptions about some of these things, so that in the future, we get a test failure if we violate those assumptions.

I'm going to merge this PR as-is, but if you want to open a new PR with the test, I'll keep an eye out for it and merge it ASAP.

greenhat: @lima-limon-inc Thank you for the detailed reply! I totally blanked that we already have the second verification. The explicit test for it would be nice to have.

lima-limon-inc: > @lima-limon-inc Thank you for the detailed reply! I totally blanked that we already have the second verification. The explicit test for it would be nice to have.

Noted! I'll get to work on it. I created a small issue to track this, here: https://github.com/0xMiden/compiler/issues/605

PR #593, opened 2025-07-09
wip: update to latest miden-vm release

bitwalker:
This PR transitions us from Miden 0.14 to 0.16, except for the client/node crates which we are still waiting on. Until those crates are released, this remains a draft PR.


codegen/masm/src/artifact.rs
```diff
@@ -45,7 +44,7 @@ pub struct Rodata {
     /// The component to which this read-only data segment belongs
     pub component: builtin::ComponentId,
     /// The content digest computed for `data`
-    pub digest: Digest,
+    pub digest: Word,
```
greenhat, line 47: What is the reason to ditch the Digest?

bobbinth: I think it was somewhat confusing when to use `Word` and when to use `Digest` as they were frequently used interchangeably and so we decided to consolidate them. Here is the discussion that led up to it: https://github.com/0xMiden/crypto/issues/59#issuecomment-2849086705.

greenhat: Got it. Thanks!

bitwalker: `Word` supercedes it upstream in `miden_core`/`miden_crypto`, and it provides all of the same features, plus some. There's no reason to have a `Digest` type now IMO.

greenhat:
@bitwalker Let's get this merged. I need this in #555 to update the client/node. Although the client/node crates are published, I suggest disabling/removing the testnet test since I'm working on it in #555 and I'll deal with client/node there.

PR #599, opened 2025-07-11
fix: prevent panic in filestem() method with invalid paths

strmfos:
Replace double unwrap() calls with a safe chain of and_then() and unwrap_or() to handle 
invalid file paths gracefully. This prevents potential panics when:
1. Path has no file name component (e.g., "/")
2. File name contains invalid UTF-8 characters
3. Other edge cases where file_stem() or to_str() might return None


bitwalker:
Closing as spam

PR #602, opened 2025-07-16
fix: swapped and "frozen" function arguments case in `HirLowering::schedule_operands()`

### LINKED ISSUE #600: Swapped and "frozen" function arguments
Discovered while revamping the `hmerge` intrinsic to accept digests as a pointer at #576 

I include the test from https://github.com/0xMiden/compiler/blob/db446df63fd22974a048ee84b0ddb3e70d74700d/tests/integration/src/rust_masm_tests/instructions.rs?plain=1#L411 since it is the best way to explain the bug. See the first comment.

```rust
fn test_arg_order() {
    // This test reproduces the "swapped and frozen" function arguments issue
    // discovered while working on hmerge VM op
    // The issue manifests in "intrinsic" function parameters being in the wrong order
    // (see assert_eq before the call and inside the function)
    // on the stack AND the their order is not changing when the parameters are
    // swapped at the call site (see expect_masm with the same file name, i.e. the MASM
    // do not change when she parameters are swapped).
    fn main_fn_template(digest_ptr_name: &str, result_ptr_name: &str) -> String {
        format!(
            r#"
    (f0: miden_stdlib_sys::Felt, f1: miden_stdlib_sys::Felt, f2: miden_stdlib_sys::Felt, f3: miden_stdlib_sys::Felt, f4: miden_stdlib_sys::Felt, f5: miden_stdlib_sys::Felt, f6: miden_stdlib_sys::Felt, f7: miden_stdlib_sys::Felt) -> miden_stdlib_sys::Felt {{
        let digest1 = miden_stdlib_sys::Digest::new([f0, f1, f2, f3]);
        let digest2 = miden_stdlib_sys::Digest::new([f4, f5, f6, f7]);
        let digests = [digest1, digest2];
        let res = merge(digests);
        res.inner.inner.0
    }}

    #[inline]
    pub fn merge(digests: [Digest; 2]) -> Digest {{
        unsafe {{
            let digests_ptr = digests.as_ptr().addr() as u32;
            // ATTENTION: the digests_ptr is correct (__stack_pointer - 96)
            // see wat/hir
            assert_eq(Felt::from_u32(digests_ptr as u32), Felt::from_u32(1048480));

            let mut ret_area = ::core::mem::MaybeUninit::<Word>::uninit();
            let result_ptr = ret_area.as_mut_ptr().addr() as u32;
            // ATTENTION: the result_ptr is expected to be 1048544 (__stack_pointer - 96 + 64)
            // see wat/hir
            assert_eq(Felt::from_u32(result_ptr as u32), Felt::from_u32(1048544));

            intrinsic({} as *const Felt, {} as *mut Felt);

            Digest::from_word(ret_area.assume_init())
        }}
    }}

    #[no_mangle]
    fn intrinsic(digests_ptr: *const Felt, result_ptr: *mut Felt) {{
        // ATTENTION: the digests_ptr is expected to be 1048480 (__stack_pointer - 96)
        // see assert_eq above, before the call
        assert_eq(Felt::from_u32(digests_ptr as u32), Felt::from_u32(1048544));
        // ATTENTION: the result_ptr is expected to be 1048544 (__stack_pointer - 96 + 64)
        // see assert_eq above, before the call
        assert_eq(Felt::from_u32(result_ptr as u32), Felt::from_u32(1048480));
    }}
        "#,
            digest_ptr_name, result_ptr_name
        )
    }

    let config = WasmTranslationConfig::default();
    let main_fn_correct = main_fn_template("digests_ptr", "result_ptr");
    let mut test = CompilerTest::rust_fn_body_with_stdlib_sys(
        "arg_order",
        &main_fn_correct,
        config.clone(),
        [],
    );

    test.expect_wasm(expect_file![format!("../../expected/arg_order_correct.wat")]);
    test.expect_ir(expect_file![format!("../../expected/arg_order_correct.hir")]);
    test.expect_masm(expect_file![format!("../../expected/arg_order.masm")]);

    let args = [
        Felt::ZERO,
        Felt::ZERO,
        Felt::ZERO,
        Felt::ZERO,
        Felt::ZERO,
        Felt::ZERO,
        Felt::ZERO,
        Felt::ZERO,
    ];
    eval_package::<Felt, _, _>(&test.compiled_package(), [], &args, &test.session, |trace| Ok(()))
        .unwrap();

    // Lets swap the "digests_ptr" and "result_ptr" and test it
    let main_fn_swapped = main_fn_template("result_ptr", "digests_ptr");
    let mut test_swapped =
        CompilerTest::rust_fn_body_with_stdlib_sys("arg_order", &main_fn_swapped, config, []);

    test_swapped.expect_wasm(expect_file![format!("../../expected/arg_order_swapped.wat")]);
    test_swapped.expect_ir(expect_file![format!("../../expected/arg_order_swapped.hir")]);

    // NOTE: the MASM did not change (see the same file name in the expect_masm call above)
    // compared to the "digests_ptr, result_ptr" in the previous expect_masm call
    test_swapped.expect_masm(expect_file![format!("../../expected/arg_order.masm")]);

    eval_package::<Felt, _, _>(
        &test_swapped.compiled_package(),
        [],
        &args,
        &test_swapped.session,
        |trace| Ok(()),
    )
    .unwrap();
}
```

To reproduce it, run `cargo make test test_arg_order` in the #576 branch.

greenhat:
Close #600

This PR:
- Fixes two-operand op optimized stack scheduling at `HirLowering::schedule_operands()`;
- simplifies the `test_arg_order` test and moves it to the `misc` module;


bitwalker:
So, I was going to attempt to reproduce the issue to investigate this further, and found something interesting.

If you take HEAD of `next` (6b8dd63 as of this writing), and _just_ add the `test_func_arg_order` test to the integration test suite, and add the following on line 26:

```rust
    crate::testing::setup::enable_compiler_instrumentation();
```

And then run the test with:

```
MIDENC_TRACE=codegen=trace cargo nextest run -E 'package(miden-integration-tests)' -- test_func_arg_order
```

The test fails with the following output:

```
[TRACE codegen] lowering public builtin.function @entrypoint(v0: i32, v1: i32) -> i32 {
    ^block4(v0: i32, v1: i32):
        v3 = hir.exec @root_ns:root@1.0.0/arg_order/intrinsic(v0, v0) : i32
        builtin.ret v3;
    };
[TRACE codegen] dropping unused operands at: v3 = hir.exec @root_ns:root@1.0.0/arg_order/intrinsic(v0, v0) : i32
[TRACE codegen] found unused operands [] with constraints []
[TRACE codegen] scheduling operands for binary op: v3 = hir.exec @root_ns:root@1.0.0/arg_order/intrinsic(v0, v0) : i32
[TRACE codegen]   b = v0 = Move
[TRACE codegen]   a = v0 = Copy
  |-> at codegen/masm/src/lower/lowering.rs:88:44
  `-> invalid operand stack index (1): only 1 operands are available
  help: set the `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Aside from the actual reason this fails, notice that the call to `intrinsic` passes `v0` _twice_, not `v0` and `v1`, as you'd expect based on the original Rust code.

There are a few things here:

* The error here occurs because the manual handling for `args.len() == 2` is not accounting for when the two arguments are the same SSA value, and only one copy of that SSA value is on the stack (i.e. we require a copy to be made). This may also explain the "frozen" swap issue, but might be a separate bug.
* It looks to me like either we're translating the Wasm for this function incorrectly, or some kind of miscompilation is happening from Rust -> Wasm, or maybe in a canonicalization/rewrite in our pipeline. The HIR reflects something that is clearly incorrect (i.e. that we're aliasing what should be an unaliasable mutable reference). I can't tell why this is happening, but haven't dug into it yet. We _definitely_ need to understand why this is happening.
* The test comment here explains that the test is to reproduce #600, but the test does not reproduce that issue (or at least, it can't get to the point of reproducing it because of this other issue). I'm guessing the latter.

Obviously you wouldn't hit this when adding the `implements::<BinaryOp>()` check to that block - but I'm surprised you didn't hit it prior to that (unless the test was modified _after_ that change, so the reproducer is actually better this time around, in that it uncovered the issue more precisely). Anyway, wanted to bring this to your attention. I'll try to poke at it more tonight, but if you don't hear any updates from me, I'd take a look on your end and see what you find.

codegen/masm/src/lower/lowering.rs
```diff
@@ -276,6 +271,8 @@ pub trait HirLowering: Op {
                 }
             }
 
+            dbg!("exit");
```
bitwalker: Looks like this was left in accidentally

codegen/masm/src/lower/lowering.rs
```diff
@@ -121,9 +123,6 @@ pub trait HirLowering: Op {
                     }
                     (Constraint::Move, Constraint::Copy) => {
                         emitter.dup(1, span);
-                        if !is_duplicate {
```
bitwalker, line 124: 🤦

bitwalker (approved):
LGTM!

codegen/masm/src/lower/lowering.rs
greenhat: Thanks. Removed.

PR #604, opened 2025-07-17
fix: start `Linker::reserved_memory_pages` on the next page after rustc reservation for the Rust stack

### LINKED ISSUE #595: Unexpected values in a `Vec` after passing it to an inlined function
Found while working on #589 

To reproduce, run the (ignored) test `cargo make test test_hash_element_smoke`. 

See the three `assert_eq` calls at https://github.com/0xMiden/compiler/blob/747fe6769f382d14a0cd6e8048a7c57cdda27ae3/tests/integration/src/rust_masm_tests/abi_transform/stdlib.rs?plain=1#L121-L124 . 
Instead of the expected `1,2,3` values, the vector holds `8,address,8` values. The Wasm and HIR look fine to me. I suspect the bug is in the MASM codegen, triggered by the what seems like a remnants of the `hash_elements` function call before inlining (storing the capacity, pointer, and the length of the `elements` vector on the Rust stack). These remnants do not have to be there since the function call was inlined, but the `elements` vector is allocated at an entirely different address, and they should not overlap.

greenhat:
Close #595

This PR fixes the bug that manifests as if a `Vec` was allocated at the same address as the Rust stack. The root cause was that we put the first global variable (`__stack_pointer`) at the same address as the Rust stack pointer. Since the value of the `__stack_pointer` is the Rust stack address and the `BumpAlloc` instance is allocated on the stack, the `BumpAlloc::top` pointer got the value of the Rust stack address (`__stack_pointer`) and `BumpAlloc::maybe_init()` constructor skipped the `BumpAlloc::top` initialization.

This PR:
- adds an extra memory page to `DEFAULT_RESERVATION` to put the Wasm globals after the Rust `static` (needed when no data segments present);
- adds an extra memory page after the data segments before putting Wasm globals to accommodate Rust `static` (Rust puts them after data segments).

### DIFF
--- a/codegen/masm/src/linker.rs
+++ b/codegen/masm/src/linker.rs
@@ -5,6 +5,7 @@ use midenc_hir::{
 
 const DEFAULT_PAGE_SIZE: u32 = 2u32.pow(16);
 /// Currently, Wasm modules produced by rustc reserve 16 pages for the Rust stack
+/// (see __stack_pointer global variable value in Wasm).
 const DEFAULT_RESERVATION: u32 = 16;
 
 pub struct LinkInfo {
@@ -74,7 +75,10 @@ pub struct Linker {
 
 impl Default for Linker {
     fn default() -> Self {
-        Self::new(DEFAULT_RESERVATION, DEFAULT_PAGE_SIZE)
+        // We start our reserved memory from the next page after the rustc reserved for the
+        // Rust stack to avoid overlapping.
+        let reserved_memory_pages = DEFAULT_RESERVATION + 1;
+        Self::new(reserved_memory_pages, DEFAULT_PAGE_SIZE)
     }
 }
 
--- a/codegen/masm/src/linker.rs
+++ b/codegen/masm/src/linker.rs
@@ -147,15 +147,20 @@ impl Linker {
         // 3. Layout global variables in the next page following the last data segment
         let next_available_offset = self.segment_layout.next_available_offset();
         let reserved_offset = (self.reserved_memory_pages * self.page_size).next_multiple_of(4);
+        // We add a page after the data segments to avoid overlapping with Rust `static` vars which
+        // are placed after data segments (if present).
+        let next_available_offset_after_rust_statics = next_available_offset + DEFAULT_PAGE_SIZE;
         log::debug!(target: "linker",
-            "next_available_offset from segments: {:#x}, reserved_offset: {:#x}, \
+            "next_available_offset (after Rust statics) from segments: {:#x}, reserved_offset: {:#x}, \
              segment_count: {}",
-            next_available_offset,
+            next_available_offset_after_rust_statics,
             reserved_offset,
             self.segment_layout.len()
         );
-        self.globals_layout
-            .update_global_table_offset(core::cmp::max(reserved_offset, next_available_offset));
+        self.globals_layout.update_global_table_offset(core::cmp::max(
+            reserved_offset,
+            next_available_offset_after_rust_statics,
+        ));
         log::debug!(target: "linker",
             "global_table_offset set to: {:#x}",
             self.globals_layout.global_table_offset()
--- a/codegen/masm/src/linker.rs
+++ b/codegen/masm/src/linker.rs
@@ -6,7 +6,9 @@ use midenc_hir::{
 const DEFAULT_PAGE_SIZE: u32 = 2u32.pow(16);
 /// Currently, Wasm modules produced by rustc reserve 16 pages for the Rust stack
 /// (see __stack_pointer global variable value in Wasm).
-const DEFAULT_RESERVATION: u32 = 16;
+// We start our reserved memory from the next page after the rustc reserved for the
+// Rust stack to avoid overlapping with Rust `static` vars.
+const DEFAULT_RESERVATION: u32 = 17;
 
 pub struct LinkInfo {
     component: builtin::ComponentId,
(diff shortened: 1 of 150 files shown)


bitwalker (approved):

codegen/masm/src/linker.rs
```diff
@@ -74,7 +75,10 @@ pub struct Linker {
 
 impl Default for Linker {
     fn default() -> Self {
-        Self::new(DEFAULT_RESERVATION, DEFAULT_PAGE_SIZE)
+        // We start our reserved memory from the next page after the rustc reserved for the
```
bitwalker: Let's move this to the `DEFAULT_RESERVATION` definition itself.

codegen/masm/src/linker.rs
```diff
@@ -5,6 +5,7 @@ use midenc_hir::{
 
 const DEFAULT_PAGE_SIZE: u32 = 2u32.pow(16);
 /// Currently, Wasm modules produced by rustc reserve 16 pages for the Rust stack
+/// (see __stack_pointer global variable value in Wasm).
 const DEFAULT_RESERVATION: u32 = 16;
```
bitwalker: Let's define this as `17`, and document that this reserves 16 pages for the shadow stack, and 1 page for the Rust global allocator, which is not allocated as a normal global variable.

greenhat:
> > The 1048576 is also the address where we put the first global variable (self.reserved_memory_pages * self.page_size == 1048576)
>
> That should be totally fine - the stack grows downwards from that address, and that address is excluded from the stack itself (i.e. nothing stack-allocated can be allocated there). So the fact that the `__stack_pointer` global is allocated at the top of the stack isn't in and of itself a bug.
>
> > The`i32.const 1048576` is `&self`
>
> This appears to me to be the actual root cause here. It looks to me like Rust is assuming a specific memory address for the start of the heap (i.e. immediately after the 16 pages it reserves for the stack). That address, `1048576` is _not_ the stack, it is the start of the global heap (i.e. Rust here has assigned the global variable for the `BumpAlloc` global allocator at the start of the global heap, which makes sense).

Yes, you're right. I mistakenly assumed since `_stack_pointer == 1048576` that `1048576` the start of the stack, but `1048576-1` is the start of stack.

> This presents us with an issue - Rust (or LLVM) is not using a global variable for the global allocator, but is instead placing it on the heap _without_ doing so via the global allocator (which obviously it can't do, because the global allocator can't allocate itself). I'm fairly sure this issue is specific to the global allocator itself, and not a general problem with globals in Rust.

I tested with another `static` (struct) and looked at our `test_static_mut` test, and they don't end up as Wasm global vars. I suppose the Rust (or LLVM) is not using Wasm global variables for `static`s. It ends up as an address on the global heap. I think `BumpAlloc::new()` does not allocate anything (allocations are not allowed in statics anyway), and for that reason it does not need an allocator. I noticed that the `#[global_allocator]` instance ends up last (higher address) on the global heap if more `static` vars are present.

> So I think the change in this PR is a reasonable solution to the problem (we're essentially giving Rust an entire page for the global allocator, which albeit wasteful, ensures that we aren't trying to cut things too close by doing something like assuming the specific layout of `BumpAlloc` in our linker).

> The one thing we should maybe try to confirm though, is whether this issue is specific to the global allocator global, or something that Rust does more generally (i.e. some globals translate to Wasm as actual globals, while some are heap allocated statically by Rust/LLVM). If this is not specific to just the global allocator, then we may need to rethink how we're laying out memory in the linker to avoid this issue resurfacing in other contexts. I think the odds of that are pretty low if we're giving Rust an entire page to work with, but it's to our benefit to understand exactly what Rust assumes when it comes to memory.

I agree. Since "large" statics end up in the data segment (see the `test_static_mut` test) we only need to reserve space for "small" statics and one page seems enough.

greenhat:
Wait, I just tested both "large" and "small" statics together, and the "large" goes into the data segment at the start of the global heap, with the "small" static going **after** it. But in the `Linker::link()` https://github.com/0xMiden/compiler/blob/239912ce4b52d75a49b1fba562a70c4248d856f1/codegen/masm/src/linker.rs?plain=1#L149 after the data segment we're putting Wasm globals **overlapping** the Rust statics which should end up there. 
I suggest we put another memory page after the data segment and put Wasm globals after it. What do you think?

bitwalker:
Well that's a bit unfortunate!

I think we need to figure out how Rust expects global allocators to manage the heap (i.e. how are they expected to obtain the start of the heap _not_ used by Rust itself). My guess here is that the global allocator (on Wasm), is expected to use `memory_size` to get the size of the heap in pages, assume that is all used, and allocate fresh pages from there. If that's the case, then we can do the same thing (not literally, but since we know the size of the Wasm memory statically, we can just round that up the nearest number of pages, and allocate our globals starting in the next page after that.

That's similar to what you are suggesting, but the difference is that this approach guarantees that we never overlap with space reserved by Rust, since the size of the Wasm memory will always be sufficiently large enough to hold all of the Rust globals. So we don't need to care about where the segments are, or how many, we can just work off the size of the Wasm heap (though we probably should validate that there are no segments starting on/after the start of the heap we plan to allocate stuff in).

I think the main hiccup here is that we currently don't propagate that information through to the backend, so all we have at the moment is the segments.

Let's start with your solution initially, just so we have a fix in place, but we should create an issue to allow the frontend to hint to the backend where to place the global heap. That hint would then take precedence over our default approach (your proposed solution). I'm not yet sure what that should look like though, so its not something we need to tackle right now.

greenhat:
@bitwalker I added an extra memory page after the data segments and addressed your comments. Please do another round.

PR #609, opened 2025-07-19
fix(codegen): missing result renaming in ptrtoint lowering

bitwalker:
While looking at our lowering for `hir.ptr_to_int`, I realized that we weren't correctly renaming the result value on the operand stack, which is pushed as a type, to an SSA value, because we were implementing the lowering manually, rather than going through `InstOpEmitter`. The effect of this is that subsequent operand stack scheduling that tried to look up that value by name, would fail to find it.

We weren't yet using `hir.ptr_to_int` anywhere, which is why this hasn't come up yet.


bitwalker:
It appears we aren't using `ptr_to_int` anywhere yet (none of our expect tests have it in the HIR), so that's why we haven't observed the error I mentioned.

PR #613, opened 2025-07-21
Add extensible benchmarking system with `is_prime` benchmark

partylikeits1983:
This PR adds a `benches/` directory with an extensible framework for benchmarking Miden compiler programs. The system automatically compiles Rust programs with `cargo miden build`, executes them with `midenc run`, and measures VM cycles and performance metrics.

**Usage:**
```bash
# Run is_prime benchmark
cargo run -p miden-compiler-benches --bin is_prime

# Test different inputs and iterations
cargo run -p miden-compiler-benches --bin is_prime -- --input 97 --iterations 3
```

The framework is designed to be easily extended for benchmarking other Miden programs by adding new binaries to `benches/src/`.

### DIFF
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "miden-compiler-benches"
+name = "midenc-benchmark-runner"
 version.workspace = true
 edition.workspace = true
 authors.workspace = true
--- a/benches/Cargo.toml
+++ b/benches/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "miden-compiler-benches"
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+documentation.workspace = true
+description = "Benchmarks for the Miden compiler"
+publish = false
+
+[[bin]]
+name = "is_prime"
+path = "src/is_prime.rs"
+
+[dependencies]
+anyhow.workspace = true
+clap.workspace = true
+serde = { workspace = true, features = ["std"] }
+toml.workspace = true
+
+[dev-dependencies]
+criterion = "0.5"
+
+[[bench]]
+name = "is_prime_bench"
+harness = false
--- a/benches/README.md
+++ b/benches/README.md
@@ -4,15 +4,33 @@ Benchmarks for measuring VM cycles and performance of Miden programs.
 
 ## Usage
 
+### From project root (recommended)
+
 ```bash
 # Run is_prime benchmark
-cargo run --bin is_prime
+cargo make bench --bin is_prime
 
 # Custom input
-cargo run --bin is_prime -- --input 97
+cargo make bench --bin is_prime -- --input 97
 
 # Multiple iterations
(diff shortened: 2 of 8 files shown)


benches/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "miden-compiler-benches"
```
bitwalker: Let's name this crate `midenc-benchmark-runner` or `midenc-benchmarks`. We try to use the `midenc-` prefix in this repo to avoid conflicts with other Miden repos/crates, and to keep things namespaced a bit cleaner.

benches/README.md
```diff
@@ -0,0 +1,36 @@
+# Miden Compiler Benchmarks
+
+Benchmarks for measuring VM cycles and performance of Miden programs.
+
+## Usage
+
+```bash
+# Run is_prime benchmark
```
bitwalker, line 26: We should add a task to the top-level `Makefile.toml` to allow for running these from the project root like so:

```
cargo make bench is_prime -- --input 97 --iterations 5
```

Which would be expanded to:

```
cargo run -p midenc-benchmark-runner --bin is_prime -- --input 97 --iterations 5
```

You should be able to do that with this task definition:

```toml
[tasks.bench]
category = "Benchmarks"
description = "Runs benchmarks written in Rust"
command = "cargo"
args = [
    "run",
    "@@remove-empty(CARGO_MAKE_CARGO_VERBOSE_FLAGS)",
    "-p",
    "midenc-benchmark-runner",
    "${@}",
]
```

benches/src/is_prime.rs
```diff
@@ -0,0 +1,101 @@
...
+//! and measures its execution performance in the Miden VM.
+
+use std::path::PathBuf;
+
+use anyhow::Result;
+use clap::{Arg, Command};
+use miden_compiler_benches::BenchmarkRunner;
+
+fn main() -> Result<()> {
+    let matches = Command::new("is_prime_benchmark")
```
bitwalker: I'd recommend defining an argument structure, like so:

```rust
use clap::Parser;

#[derive(Parser)]
struct Config {
    /// The number to test for primality
    #[arg(short = 'i', long, value_name = "NUMBER", default_value = "29")]
    input: usize,
    /// Path to the is_prime source file
    #[arg(short = 's', long, value_name = "PATH", default_value = "examples/is_prime/src/lib.rs")]
    source: PathBuf,
    /// Number of iterations to run
    #[arg(short = 'n', long, value_name = "COUNT", default_value = "1")]
    iterations: usize,
}
```

Then you can parse the arguments with just:

```rust
fn main() -> anyhow::Result<()> {
    let config = Config::parse()?;
    ...
}
```

bitwalker (changes requested):
Looks good! Thanks for putting this together!

I have a couple of tweaks I'd like to make before merging, but they are small things.

partylikeits1983:
I addressed feedback by renaming the crate to "midenc-benchmark-runner," adding a bench task to `Makefile.toml`, and updating the argument parsing in `benches/src/is_prime.rs`.

bitwalker (approved):

PR #625, opened 2025-07-31
Transaction script compilation support

### LINKED ISSUE #618: Update Miden SDK `tx:create_note` to the new signature
We're expecting `(Asset, Tag, NoteType, Recipient) -> NoteId` but in `miden-base` is has:
```
#! Inputs:  [tag, aux, note_type, execution_hint, RECIPIENT]
#! Outputs: [note_idx]
```

See https://github.com/0xMiden/miden-base/blob/fb286d3457e5b85cf1326ef7b2648d4ee0e35ed0/crates/miden-lib/asm/miden/tx.masm?plain=1#L149-L178

### LINKED ISSUE #621: Remove `send_asset` in `basic-wallet` example in favor of the new `move_asset_to_note` procedure
See basic wallet in the `miden-base` at https://github.com/0xMiden/miden-base/blob/fb286d3457e5b85cf1326ef7b2648d4ee0e35ed0/crates/miden-lib/asm/miden/contracts/wallets/basic.masm?plain=1#L31-L58

### LINKED ISSUE #622: Transaction script compilation

greenhat:
Close #618
Close #621
Close #622

Corresponding templates repo PR - https://github.com/0xMiden/rust-templates/pull/18

This PR introduces transaction script compilation support.

### Key Changes

- Added the `cargo miden new --tx-script` option to generate a tx script project;
- Added `project-kind` under `[package.metadata.miden]` in Cargo.toml to identify project kind (account, note script, tx script);
- Renamed both note script and tx script entrypoint function to be `run`;
- Added the `arg: Word` parameter to both note and tx scripts entrypoint functions;
- Added missing tx kernel bindings and changed existing ones where MASM signature changed;
- Finished the `test_basic_wallet_p2id_local` test against the local node instance with the scenario Faucet -> Alice -> Bob asset movement.
- Added `basic-wallet-tx-script` to `cargo miden example` command;

See `basic-walled-tx-script` at https://github.com/0xMiden/compiler/blob/5872ba29cc0807bfd25cf529b8880f7907984776/examples/basic-wallet-tx-script/src/lib.rs?plain=1#L29-L64

And the offchain code that calls it at https://github.com/0xMiden/compiler/blob/5872ba29cc0807bfd25cf529b8880f7907984776/tests/integration-node/src/node_tests/helpers.rs?plain=1#L342-L390

The updated `basic-wallet` Rust code is at https://github.com/0xMiden/compiler/blob/5872ba29cc0807bfd25cf529b8880f7907984776/examples/basic-wallet/src/lib.rs?plain=1#L29-L49


examples/basic-wallet-tx-script/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+mod bindings;
+
+use bindings::{exports::miden::base::transaction_script::Guest, miden::basic_wallet::basic_wallet};
+use miden::{intrinsics::advice::adv_push_mapvaln, *};
+
+struct BasicWalletTxScript;
+
+impl Guest for BasicWalletTxScript {
+    fn tx_script(arg: Word) {
```
bitwalker: I've been looking at the name of this method, and that of the note script type, and I really think we should make this more intuitive.

In particular, it would probably be a lot more intuitive to just name them `main`, `run`, `execute`, or `start` - something to that effect.

However, if we _need_ to distinguish them, then something like the following would be an improvement IMO:

* Use `execute_note` for a note script entrypoint. We don't need to mention "script" anywhere here, because "executing" a note is by definition executing a note script.
* Use an event handler styled name for the transaction script entrypoint, where the event refers to its place/role in the protocol. IIRC, something like `handle_notes_executed` would be most precise; but maybe something more general like `finalize_transaction` or whatever, would be more intuitive.

Basically, I think `note_script` and `tx_script` are probably not good names for a method/function that represents the entrypoint of those scripts. Maybe @bobbinth has some thoughts here.

As an aside, we should probably consider implementing the abstractions over the `Guest` trait that I described in one of our discussions earlier this year so that we can hide some of the boilerplate here, and use something more semantically meaningful without needing to hack together anything crazy. I can't recall off the top of my head where that conversation was though, so I'll have to dig around for it.

bitwalker (commented):
I know this is still in draft status, but figured I'd take a pass through early. No notes, aside from one about how we name the entrypoint for note and transaction scripts, but it isn't necessarily a blocker, just something that sticks out to me now.

examples/basic-wallet-tx-script/src/lib.rs
greenhat: Thank you! I was looking at them for so long that I didn't notice how ugly they are. I renamed them both to `run`. The `cargo-miden` has them hard-coded to set the entrypoint option, but the interface name and the target (note vs. tx script) are enough to distinguish them.

greenhat:
@bitwalker It's ready for review.

bitwalker (approved):

PR #627, opened 2025-08-01
Advice map loading in Miden SDK

### LINKED ISSUE #626: Advice map loading API in Miden SDK
### API

We want to expose the loading values from the advice map in our Miden SDK with the following functions:

```rust
/// Pushes a list of field elements onto the advice stack. The list is looked up in the advice map using `key` as the key.
/// Returns the number of elements pushed on the advice stack.
fn adv_push_mapvaln(key: Word) -> Felt;
/// Pops an arbitrary number of words from the advice stack and asserts it matches the commitment.
fn adv_load_preimage(num_words: Felt, commitment: Word) -> Vec<Word>;
```

`adv_push_mapvaln` uses `adv.push_mapvaln` VM instruction. It should be implemented as a compiler intrinsic at `codegen/masm/intrinsics/io.masm` after calling `adv.push_mapvaln` drop the word from the stack with the `dropw` op and pop (move to the stack) the first element (number of elements pushed on the advice stack) from the advice stack with `adv_push.1` op. This will be the intrinsic result.

`adv_load_preimage` uses `pipe_preimage_to_memory` (use `extern` it will be linked) Miden stdlib function (see https://0xmiden.github.io/miden-docs/imported/miden-vm/src/user_docs/stdlib/mem.html). Before calling it it allocates a `Vec` with `num_words` and passes its buffer pointer to the stdlib function along with commitment.
For returning a vec see `get_assets` Miden SDK function for the example.

### Implementation details

For an example of exposing the Miden VM instruction or stdlib function see how stdlib's blake3 hash function is exposed in sdk/stdlib-sys/src/stdlib/crypto/hashes.rs (`hash_1to1`, note return-by-pointer pattern).
Add the low-level(extern) function signature at sdk/stdlib-sys/wit/miden-core-intrinsics.wit (see blake3 function at sdk/stdlib-sys/wit/miden-core-stdlib.wit)
Catch the low-level function import at the Wasm frontend similar to blake3 (`hash-one-to-one` and `hash_1to1`) and translate it into the intrinsic (create it if needed).

### Test

The test should work similar to the `test_blake3_hash` test.
The code under test has the signature `(key: Word) -> Vec<Felt>` and should return loaded `Vec<Felt>`. The code should use `adv_push_mapvaln` and `pipe_preimage_to_memory` using the key as commitment. Before calling `eval_package` the `Vec<Felt>` should be generated (len % 4 == 0) and added to the advice map using its  RpoDigest as the key.

greenhat:
Close #626 

**This PR is stacked on the #555 and should be merged after it**

This PR adds the following functions to the Miden SDK:
```rust
/// Pushes a list of field elements onto the advice stack. The list is looked up in the advice map using `key` as the key.
/// Returns the number of elements pushed on the advice stack.
fn adv_push_mapvaln(key: Word) -> Felt;
/// Pops an arbitrary number of words from the advice stack and asserts it matches the commitment.
fn adv_load_preimage(num_words: Felt, commitment: Word) -> Vec<Felt>;
```


codegen/masm/intrinsics/io.masm
bitwalker, line 1: Maybe we should call this `advice.masm` instead? I don't think I'd know to look for these intrinsics in a module named `io.masm`

sdk/stdlib-sys/src/intrinsics/mod.rs
```diff
@@ -1,12 +1,15 @@
...
+mod debug;
 mod felt;
+mod io;
 mod word;
 
-pub use crypto::Digest;
+pub use crypto::*;
+pub use debug::*;
 pub use felt::*;
+pub use io::*;
```
bitwalker: I don't think it's a good idea to hide the namespacing and re-export everything from t he top level like this - I'd rather try to match the structure of the Miden standard library. Even if we do our own organization, I still think it aids discovery when things are namespaced.

For example, `miden_stdlib_sys::merge` (which I saw in the diff somewhere here) does not tell you what that function is at all, while `miden_stdlib_sys::crypto::merge` at least hints that it is a crypto-related operation.

bitwalker (changes requested):
Looks good! Just a couple of small things I'd like to change before merging.

codegen/masm/intrinsics/io.masm
greenhat: Yes. Advice ops are in IO ops section of the book, but IO might be far too generic for the intrinsic file name.

sdk/stdlib-sys/src/intrinsics/mod.rs
greenhat: Good point. What about the types? I suppose we export types from the root, or do we want to export them from specific crates as well?

codegen/masm/intrinsics/io.masm
greenhat: Done in https://github.com/0xMiden/compiler/pull/627/commits/d236627213dc210e2ca620fe2f3d9b6659cf7eb4

sdk/stdlib-sys/src/intrinsics/mod.rs
greenhat: Done in https://github.com/0xMiden/compiler/pull/627/commits/d236627213dc210e2ca620fe2f3d9b6659cf7eb4

greenhat:
Thank you! I addressed your notes. Please do another round.

PR #630, opened 2025-08-07
fix: HEAP_END to represent the Miden limit in the Rust address space

greenhat:
The linear memory heap must not spill over into the region reserved for procedure locals, which begins at `2^30` in Miden's address space. In Rust address space it should be `2^30 * 4` but since it overflows the usize which is 32-bit on wasm32 we use `u32::MAX`. "-1" is to avoid `xor` op in the compiled Wasm.

### DIFF
--- a/sdk/alloc/src/lib.rs
+++ b/sdk/alloc/src/lib.rs
@@ -16,9 +16,11 @@ const PAGE_SIZE: usize = 2usize.pow(16);
 /// We require all allocations to be minimally word-aligned, i.e. 16 byte alignment
 const MIN_ALIGN: usize = 16;
 
-/// The linear memory heap must not spill over into the region reserved for procedure
-/// locals, which begins at 2^30 in Miden's address space.
-const HEAP_END: *mut u8 = (2usize.pow(30) / 4) as *mut u8;
+/// The linear memory heap must not spill over into the region reserved for procedure locals, which
+/// begins at 2^30 in Miden's address space. In Rust address space it should be 2^30 * 4 but since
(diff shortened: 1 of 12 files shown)


sdk/alloc/src/lib.rs
```diff
@@ -16,9 +16,11 @@ const PAGE_SIZE: usize = 2usize.pow(16);
 /// We require all allocations to be minimally word-aligned, i.e. 16 byte alignment
 const MIN_ALIGN: usize = 16;
 
-/// The linear memory heap must not spill over into the region reserved for procedure
-/// locals, which begins at 2^30 in Miden's address space.
-const HEAP_END: *mut u8 = (2usize.pow(30) / 4) as *mut u8;
+/// The linear memory heap must not spill over into the region reserved for procedure locals, which
+/// begins at 2^30 in Miden's address space. In Rust address space it should be 2^30 * 4 but since
+/// it overflows the usize which is 32-bit on wasm32 we use u32::MAX. "-1" is to avoid `xor` op in
```
bitwalker, line 21: Could you remind me why we want to avoid `xor`? I'd like to fix that issue ASAP if that's something we actually need to avoid in general right now.

bitwalker (approved):
LGTM - but I would like to make sure we fix the `xor` thing ASAP, doesn't have to be here though.

sdk/alloc/src/lib.rs
greenhat: Ha! Great question! The only reason I avoided it was that I thought it was expensive. But I checked, and it's 1 cycle. :face_palm: I'm making a new PR to remove `-1`.

PR #632, opened 2025-08-07
fix: Aligned data segments overlap the `BumpAlloc` instance

### LINKED ISSUE #631: Data segments overlap the `BumpAlloc` instance
Discovered in https://github.com/0xMiden/compiler/pull/625

The second data segment was misplaced after the first was padded to the word boundary and its tail ended up at the address of the `BumpAlloc` instance.

The Rust/LLVM places global variable instances that are initialized to zeroes right after the last data segment in memory. This means we can merge all the data segments and pad the result to the word boundary safely since Rust expects zeroes after the last data segment anyway.

### Note
When compiling the Wasm component, I noticed that Rust/LLVM uses GOT (Global Offset Table). I presume it's for the Wasm component code to be position-independent and relocateble. It manifests with an extra global var in the core module. 

```wasm
(global $GOT.data.internal.__memory_base (;1;) i32 i32.const 0)
```
(see [example](https://github.com/0xMiden/compiler/blob/d236627213dc210e2ca620fe2f3d9b6659cf7eb4/tests/integration/expected/examples/basic_wallet.wat?plain=1#L49)) 

Which is initialized to zero and is used in the Wasm core module code as an offset for every hardcoded memory address (Rust globals, data segment access, etc.). I reckon, we can use it if we would want to relocate the data segments and Rust globals.

greenhat:
Close #631 

 **This PR is stacked on the #627 and should be merged after it**

This PR merges all the data segments and pads the result to the word boundary. 

The Rust/LLVM places global variable instances that are initialized to zeroes right after the last data segment in memory. This means we can merge all the data segments and pad the result to the word boundary safely since Rust expects zeroes after the last data segment anyway.


frontend/wasm/src/module/build_ir.rs
```diff
@@ -223,21 +223,18 @@ fn build_data_segments(
...
-        let init = ConstantData::from(segment.data);
+    if let Some(merged_segment) = merge_data_segments(resolved_segments)? {
+        let init = ConstantData::from(merged_segment.data);
         let size = init.len() as u32;
         if let Err(e) = module_builder.define_data_segment(
-            segment.offset,
+            merged_segment.offset,
             init,
-            segment.readonly,
+            merged_segment.readonly,
```
bitwalker, line 232: We'll need to be really careful with this - we shouldn't merge data segments with mixed `readonly` behavior (i.e. if one segment is `readonly` and the next isn't, we must keep them unmerged).

frontend/wasm/src/module/data_segments.rs
```diff
@@ -1,272 +1,380 @@
...
+    let mut all_readonly = true;
 
-        let self_end = self.offset + self.data.len() as u32;
-        let gap = other.offset.saturating_sub(self_end);
+    for (i, segment) in segments.iter().enumerate() {
+        if i > 0 {
+            merged_name.push('_');
+        }
+        merged_name.push_str(&segment.name);
+        all_readonly = all_readonly && segment.readonly;
```
bitwalker, line 91: We can't merge adjacent segments that don't have the same `readonly` status - we can only merge runs of segments with the same status.

I'm not sure how often we have multiple segments where some are readonly and others aren't, but we need to preserve the semantic distinction for sure. In the future, we'll need to enforce memory protection in some fashion, and we'll be reliant on the `readonly` flag to tell us whether that region of memory can be written to or not. We don't enforce it now, but if we don't preserve the information, it is useless.

bitwalker (changes requested):
Looks good overall, but there is one issue that we'll need to address before this can be merged - namely the merging of segments with differing `readonly` status. Handling it will complicate the implementation further, but is necessary to ensure we preserve that important semantic information.

PR #636, opened 2025-08-15
fix: intra block spills rewrite to process all ops in the block

### LINKED ISSUE #548: Stack access out of bounds error when lowering to MASM
Discovered while working on https://github.com/0xMiden/compiler/pull/547 and testing the maximum flattened number of function arguments(16) so probably not many users will hit it.

Error:
```
  |-> at codegen/masm/src/lower/lowering.rs:191:41
  `-> invalid stack index: only the first 16 elements on the stack are directly accessible, got 16
```

### How to reproduce

Either in https://github.com/0xMiden/compiler/pull/547 or after it is merged:

At https://github.com/0xMiden/compiler/blob/greenhat/i538-struct-arg/tests/rust-apps-wasm/rust-sdk/cross-ctx-account-word-arg/src/lib.rs#L60-L60 change:
```rust
    let felt4_sum = felt4;
```
to:
```rust
    let felt4_sum = felt4 * felt!(1);
```

Run `UPDATE_EXPECT=1 cargo test rust_sdk_cross_ctx_word_arg_account_and_note`

greenhat:
Close #548

Note: I suggest reviewing this PR on a per-commit basis.

This PR:
- Fixes: Intra-block spills rewrite to process all ops in the block instead of processing only the first one;
- Fixes: infinite loop in `dominated_by_slow_tree_walk`. During the `spills_loop_nest` test at some iteration `b_idom.level()` ends up being lower than `a_level`looping the `while` forever;
- Fixes: only keep spills that feed a live reload dominated by this spill;
- Fixes: handle empty dominance frontier at `DominanceFrontier::iterate_all`;
- Adds tests for the `TransformSpills` pass;
- Rewrites spill analysis tests.

### DIFF
--- a/dialects/hir/src/transforms/spill/tests.rs
+++ b/dialects/hir/src/transforms/spill/tests.rs
@@ -7,8 +7,8 @@ use midenc_expect_test::expect_file;
 use midenc_hir::{
     dialects::builtin::{BuiltinOpBuilder, Function, FunctionBuilder},
     pass::{Nesting, PassManager},
-    AbiParam, AddressSpace, Builder, Context, Ident, Op, OpBuilder, PointerType, ProgramPoint,
-    Report, Signature, SourceSpan, SymbolTable, Type, ValueRef,
+    AbiParam, AddressSpace, Builder, Context, Ident, Op, OpBuilder, PointerType, Report, Signature,
+    SourceSpan, Type, ValueRef,
 };
 
 use crate::{transforms::TransformSpills, HirOpBuilder};
@@ -29,7 +29,7 @@ fn materializes_spills_intra_block() -> TestResult<()> {
     let context = Rc::new(Context::default());
     let mut ob = OpBuilder::new(context.clone());
 
-    let mut module = ob.create_module(Ident::with_empty_span("test".into()))?;
+    let module = ob.create_module(Ident::with_empty_span("test".into()))?;
     let module_body = module.borrow().body().as_region_ref();
     ob.create_block(module_body, None, &[]);
     let func = ob.create_function(
@@ -42,7 +42,6 @@ fn materializes_spills_intra_block() -> TestResult<()> {
             [AbiParam::new(Type::U32)],
         ),
     )?;
-    module.borrow_mut().symbol_manager_mut().insert_new(func, ProgramPoint::Invalid);
     let callee_sig = Signature::new(
         [
             AbiParam::new(Type::Ptr(Arc::new(PointerType::new_with_address_space(
@@ -58,10 +57,6 @@ fn materializes_spills_intra_block() -> TestResult<()> {
     );
     let callee =
         ob.create_function(Ident::with_empty_span("example".into()), callee_sig.clone())?;
-    module
-        .borrow_mut()
-        .symbol_manager_mut()
-        .insert_new(callee, ProgramPoint::Invalid);
 
     {
         let mut b = FunctionBuilder::new(func, &mut ob);
@@ -155,7 +150,7 @@ fn materializes_spills_branching_cfg() -> TestResult<()> {
     let context = Rc::new(Context::default());
     let mut ob = OpBuilder::new(context.clone());
 
-    let mut module = ob.create_module(Ident::with_empty_span("test".into()))?;
+    let module = ob.create_module(Ident::with_empty_span("test".into()))?;
     let module_body = module.borrow().body().as_region_ref();
     ob.create_block(module_body, None, &[]);
     let func = ob.create_function(
@@ -168,7 +163,6 @@ fn materializes_spills_branching_cfg() -> TestResult<()> {
             [AbiParam::new(Type::U32)],
         ),
     )?;
-    module.borrow_mut().symbol_manager_mut().insert_new(func, ProgramPoint::Invalid);
 
     let callee_sig = Signature::new(
         [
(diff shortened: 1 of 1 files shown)


hir-analysis/src/analyses/spills.rs
```diff
@@ -2715,116 +2886,168 @@ mod tests {
...
+            let accn = b.add_unchecked(acc, load, span)?;
+            let one2 = b.u32(1, span);
+            let coln = b.add_unchecked(col, one2, span)?;
+            // Backedge: continue inner loop by jumping to its header
+            b.br(blk4, [coln, accn], span)?;
+
+            (blk1, blk3, blk4, blk5, v1)
+        };
 
-        // We expect a spill from block3 to block4, as due to operand stack pressure in the loop,
```
greenhat, line 2809: @bitwalker I'm puzzled by the old `spills_loop_nest` test. It expected a spill (see removed asserts and comments) but in the new test I see max stack pressure at 9 and there are no spills. I recreated the IR from the old test (see doc comment) - check the `expect!` below.
Any ideas why the old test expected a spill?

bitwalker: I'm not sure the old tests are going to be useful in the new IR - there are a lot of differences that are going to lead to differing results.

I would suggest instead writing new tests from scratch. The main things we're looking to verify are:

* That when there are enough values live simultaneously whose representation would overflow the Miden operand stack, that the spills analysis identifies this, and selects appropriate candidates to spill to procedure locals such that the remaining live values will fit. **This is the most important property**.
* That all inserted spills and reloads are replaced with appropriate stores/loads after the spills transform is applied. We don't necessarily need to test this explicitly, since we'd get an error during codegen if we ever failed to rewrite all spills/reloads, but any tests we write around spills should probably make sure this is the case explicitly/implicitly.
* That all uses of a spilled value refer to the nearest definition of that value (either a reload, or original definition). This is critical for correctness of the IR/maintaining the SSA representation.
* That a spill of a value only ever occurs once in any blocks strictly dominated by that spill, i.e. we don't redundantly re-spill a value that was already spilled. If we're re-spilling values that are already spilled along a given control flow path, we have a bug in our tracking of what values are in the register set, and which are in the spill set, since observing this implies that we think we haven't spilled the value yet. It wouldn't necessarily cause issues in and of itself, other than being redundant, but it implies we have a bug which might lead to actual correctness issues.
* That spills are inserted at control flow join points where a value was spilled along one path, but not along all of them, IFF the value dominates the join point (i.e. the value is defined along all paths, but only spilled on a subset of them). This is critical for correctness of the IR/maintaining the SSA representation. Any reload that occurs after the join point will lead to undefined behavior if the value is not spilled along all control flow paths leading to the reload.

greenhat:
@bitwalker Thanks! The tests seem to be covering all your points. It's ready for review.

dialects/hir/src/transforms/spill/tests.rs
```diff
@@ -0,0 +1,407 @@
...
+    // We expect spills and reloads to be required around the call and later uses:
+    // - At the `exec` call, operand stack pressure forces spilling of two live values.
+    //   The analysis selects the farthest next-use candidates, which here correspond to
+    //   the results used later.
+    // - Before the `store` and `ret` we must reload them back.
+    // After running TransformSpills, those spill/reload pseudos will be materialized as:
+    //   store_local <spilled>
+    //   load_local  : <type> #[local = lvN]
+    let before = func.as_operation_ref().borrow().to_string();
+    expect![[r#"
```
bitwalker: We probably should use `expect_file`, because of the issue that arises when multiple `expect` assertions need to be updated concurrently via `UPDATE_EXPECT=1` - it very easily corrupts the file if lines are added/removed as a result of the update, but it's racy, so not something that always happens.

dialects/hir/src/transforms/spill/tests.rs
```diff
@@ -0,0 +1,407 @@
...
+        Ident::with_empty_span("test::spill".into()),
+        Signature::new(
+            [AbiParam::new(Type::Ptr(Arc::new(PointerType::new_with_address_space(
+                Type::U8,
+                AddressSpace::Element,
+            ))))],
+            [AbiParam::new(Type::U32)],
+        ),
+    )?;
+    module.borrow_mut().symbol_manager_mut().insert_new(func, ProgramPoint::Invalid);
```
bitwalker: This isn't required anymore with the symbol management changes made by @lima-limon-inc. Adding/removing an `Operation` that implements `Symbol` from a `Block` in a region of an `Operation` that implements `SymbolTable`, automatically adds/removes the symbol to/from the table.

If that doesn't work, we have a bug and need to figure out why that isn't happening.

bitwalker (approved):
Looks good, though there are two small tweaks I left comments for. I've marked this approved though so once those are made you can go ahead and merge this.

dialects/hir/src/transforms/spill/tests.rs
greenhat: Thanks! Done in https://github.com/0xMiden/compiler/pull/636/commits/3464ae706e57e95bbb6b29c056db6bf56be1b3c1

dialects/hir/src/transforms/spill/tests.rs
greenhat: Nice! Done in https://github.com/0xMiden/compiler/pull/636/commits/9acba821b13efe2a417652092992f66dcc8c7130

PR #645, opened 2025-08-25
Add support for 128-bit wide arithmetic.

### LINKED ISSUE #592: Support for u128/i128 integers
We have some users who would like to make use of 128-bit integers in their programs, but we only have very minimal support implemented in the compiler currently:

* They are represented in the type system
* We can load/store them in memory
* We can perform equality checks/assertions on them

## Background: Rust, Wasm, and 128-bit integer support

WebAssembly does not have a native 128-bit integer type that supports 128-bit arithmetic, it does have `v128`, but this is nominally a vector type (e.g. SIMD), though it may also be treated as a single 128-bit type per the [spec:](https://webassembly.github.io/spec/core/syntax/types.html#vector-types):

> The type  corresponds to a 128 bit vector of packed integer or floating-point data. The packed data can be interpreted as signed or unsigned integers, single or double precision floating-point values, or a single 128 bit type. The interpretation is determined by individual operations.

The problem, however, is that `v128` does not support arithmetic operations, only bitwise operations. This means using `v128` isn't really useful to us.

However, Wasm does have experimental "wide arithmetic" operations, which with Rust can be enabled via the `wide-arithmetic` target feature, e.g. `RUSTFLAGS="-Ctarget-feature=+wide-arithmetic" cargo build ..`. Rust lowers 128-bit integers to pairs of `u64`, and with the wide arithmetic operations, e.g. `u64.add128`, we basically get the equivalent of native 128-bit integer support, for a limited set of operations.

## Supporting 128-bit integers in `midenc`

So that leaves us with the following tasks needed to effectively enable 128-bit integers in Rust programs compiled to Miden:

* Enable the `wide-arithmetic` feature in the set of supported Wasm features in `midenc-frontend-wasm`.
* Ensure we build Rust projects with the `wide-arithmetic` feature enabled when invoking `cargo` from `cargo-miden`
* Update our test suite infrastructure to enable the `wide-arithmetic` feature
* Implement two new specialized cast instructions for converting pairs of `u64` to a single `u128` value, and vice versa
* Implement translations of the wide arithmetic operations (e.g. `u64.add128`) by casting the `u64` limbs to a `u128` value, and performing the given operation on the `u128` value(s). For signed operations, we'd also issue a bitcast from `u128` to `i128`. For now, we must also then perform a cast to destruct the `u128` back into the `u64` limbs. These casts are effectively no-ops during code generation, but will create a bit of noise in the IR. However, the general idea here is to implement 128-bit integer operations on the 128-bit integer types, and map the Wasm representation to those. In the future, if/when 128-bit integers are natively supported, we will be able to just update our translation in the frontend, without having to modify anything in the rest of the compiler.
* Implement support for the 128-bit versions of the various wide arithmetic operations, e.g. `add`, in the backend using MASM intrinsics, similar to how we handle other integer types. This is the final step.

/cc @otrho

otrho:
This change adds support for the WebAssembly extensions for `wide-arithmetic`.

It makes changes to just about every part of the chain:
- Adds flags for `cargo miden` to pass to `rustc`.
- Adds support to the front-end for new ops `I64Add128`, `I64Sub128`, `I64MulWideS` and `I64MulWideU`.
- Adds new HIR ops `arith.join` and `arith.split` for constructing and deconstructing 128-bit values from 64-bit limbs.
- Adds an intrinsics module for the 128-bit Masm.
- Adds emitters for the new ops to call the new intrinsics.
- And adds property based tests for each new op.

## Issues

Remaining unaddressed issues are translation from Wasm limb ordering, and the memory representation of 128-bit values.

### Limb ordering

The former is a problem with how Wasm and Miden orders the limbs to 64-bit and 128-bit values.  Wasm expects 128-bit values to be represented by two 64-bit limbs in little endian order.

When they are passed to functions `(a: i128, b: i128)` is called as `(a_lo: i64, a_hi: i64, b_lo: i64, b_hi: i64)`.  Miden in turn splits each 64-bit value into 32-bit limbs in big endian order, so the above would become `(a_mid_lo: i32, a_lo_lo: i32, a_hi_hi: i32, a_mid_hi: i32, b_mid_lo: i32, ...)`.

This is weird and unfortunate, though if consistent may not pose a significant problem.  More testing needs to be done.  To handle the problem this PR changes the `ToMidenRepr` for `u128` in `midenc-debug/src/felt.rs`.

### Memory ordering

The latter problem, regarding memory representation of 128-bit values is similar.  When writing the 128-bit result of an operation to memory the Wasm writes the lower 64-bit limb to `ptr` and the upper 64-bit limb to `ptr+8`.

When writing 64-bit values to memory with `instrinsics::mem::store_dw` we write the upper 32-bits to `ptr` and the lower 32-bits to `ptr+1` (felt addressing).

So the four limbs `(hi_hi, mid_hi, mid_lo, lo_lo)` are written (`mid_lo`, `lo_lo`, `hi_hi`, `mid_hi`) to `ptr`.

To handle this problem the tests currently re-order the result values read from memory before comparing to the fully little endian expected result.

Closes #592.


midenc-debug/src/felt.rs
```diff
@@ -465,7 +465,10 @@ impl ToMidenRepr for u128 {
             RawFelt::new(u32::from_be_bytes([bytes[8], bytes[9], bytes[10], bytes[11]]) as u64);
         let lo_l =
             RawFelt::new(u32::from_be_bytes([bytes[12], bytes[13], bytes[14], bytes[15]]) as u64);
-        smallvec![hi_h, hi_l, lo_h, lo_l]
+
+        // The 64-bit limbs are little endian, (lo, hi), but the 32-bit limbs of those 64-bit
+        // values are big endian, (lo_h, lo_l) and (hi_h, hi_l).
+        smallvec![lo_h, lo_l, hi_h, hi_l]
```
greenhat, line 471: Should the `FromMidenRepr::from_felts` be changed to use the same order as well?

frontend/wasm/src/code_translator/mod.rs
```diff
@@ -336,6 +336,18 @@ pub fn translate_operator<B: ?Sized + Builder>(
...
+            let (rhs_hi, rhs_lo) = state.pop2();
+            let (lhs_hi, lhs_lo) = state.pop2();
+
+            let lhs = builder.join(lhs_hi, lhs_lo, span)?;
+            let rhs = builder.join(rhs_hi, rhs_lo, span)?;
+
+            let res = builder.add_wrapping(lhs, rhs, span)?;
+
+            let (res_hi, res_lo) = builder.split(res, span)?;
+            state.pushn(&[res_lo, res_hi]);
```
greenhat, line 350: Consider a comment confirming/explaining the order difference for split -> pushn since it is non-obvious, and future dev might want to "fix" it. :)
Also a doc comment for `split` about the result order `(hi, lo)` would be nice.

dialects/arith/src/ops/coercions.rs
```diff
@@ -229,3 +229,52 @@ impl Foldable for Sext {
...
+    fn infer_return_types(&mut self, _context: &Context) -> Result<(), Report> {
+        self.result_mut().set_type(Type::I128);
+        Ok(())
+    }
+}
+
+/// Split a 128bit integer into two 64 bit integers.
+#[operation(
+    dialect = ArithDialect,
+    traits(UnaryOp, SameTypeOperands),
```
greenhat, line 265: Do we need `SameTypeOperands` for the unary op?

dialects/arith/src/ops/coercions.rs
```diff
@@ -229,3 +229,52 @@ impl Foldable for Sext {
...
+
+/// Join two 64bit integers into one 128 bit integer.
+#[operation(
+    dialect = ArithDialect,
+    traits(BinaryOp, SameTypeOperands),
+    implements(InferTypeOpInterface, MemoryEffectOpInterface)
+)]
+pub struct Join {
+    #[operand]
+    lhs: Int64,
```
greenhat: Consider naming operands `hi` and `lo` to convey semantics.

midenc-debug/src/felt.rs
otrho: I would guess so, but the only place I see it used is in the [round trip test](https://github.com/0xMiden/compiler/blob/88ef016f8fcd956157243d48d7dc5b2d470ea80e/midenc-debug/src/felt.rs#L997-L1014) and it's currently passing.

It seems assymetrical but at the same time it's an artefact of the combinations of Wasm splitting-to-limbs combined with our splitting-to-limbs which causes the weird order.  We never go back to 128bits via a Wasm op, so it's all convention at that point.

dialects/arith/src/ops/coercions.rs
otrho: I did this weeks ago so I've forgotten the specifics, but I think there was a reason.  I can't really see why now, though all of the unary operators in [unary.rs](https://github.com/0xMiden/compiler/blob/next/dialects/arith/src/ops/unary.rs) have it too, so it might as well stay?

dialects/arith/src/ops/coercions.rs
```diff
@@ -229,3 +229,57 @@ impl Foldable for Sext {
...
+    traits(BinaryOp, SameTypeOperands),
+    implements(InferTypeOpInterface, MemoryEffectOpInterface)
+)]
+pub struct Join {
+    #[operand]
+    high_limb: Int64,
+    #[operand]
+    low_limb: Int64,
+    #[result]
+    result: SizedInt<128>,
```
bitwalker: Probably makes sense to introduce a type alias like `Int64` for 128-bit integers now that we actually have ops that produce them.

PR #647, opened 2025-08-26
fix: reverse only non-intrinsic modules when assembling a Program

greenhat:
In this PR:
- Pass intrinsics to the assembler for `Program` in the direct order (in sync with `Library`);
- Reverse non-intrinsic modules when assembling the `Program` since `Assembler::add_module` requires that all the dependencies have to be already passed.


greenhat:
@bitwalker Although https://github.com/0xMiden/miden-vm/issues/1669 is marked as fixed in v0.15 the order of `Assembler::add_module` is still important. I suppose the fix is in v0.16+.

Issue #648, opened 2025-08-26
cargo-miden error when building a crate in a workspace


greenhat:
How to reproduce:

- put basic-wallet, tx script and p2id note from examples in a workspace;
- try to build any crate using `--manifest-path` option while cwd is the workspace root;

greenhat:
#650 PR adds `cargo-miden` support to build the crate, which is a part of the workspace but it is disabled for now. 

Unfortunately, `cargo` ignores profiles defined in the Cargo.toml in the individual crate and expects them to be defined in the workspace Cargo.toml file. We end up with the default release/dev profile with all the panic machinery, etc. which cannot compile yet.
We should move our profile definitions to `.cargo/config.toml` and rewrite the `cargo-miden` and our test harness to not use the `--manifest-path` but change the current directory to the crate we're compiling and run `cargo` there.

For now we throw an error if the user tries to build a crate that is included in the workspace with the tips (move the crate out of the workspace or delete the workspace Cargo.toml).

otrho:
Makes sense to me. 👍

PR #650, opened 2025-08-27
fix: draft cargo-miden support (disabled for now) for building the crate which is included in the workspace

greenhat:
Ref #648 

This PR adds `cargo-miden` support to build the crate, which is a part of the workspace but it is disabled for now. 

Unfortunately,`cargo` ignores profiles defined in the Cargo.toml in the individual crate and expects them to be defined in the workspace Cargo.toml file. We end up with the default release/dev profile with all the panic machinery, etc. which cannot compile yet.
We should move our profile definitions to `.cargo/config.toml` and rewrite the `cargo-miden` and our test harness to not use the `--manifest-path` but change the current directory to the crate we're compiling and run `cargo` there.

For now we throw an error if the user tries to build a crate that is included in the workspace with the tips (move the crate out of the workspace or delete the workspace Cargo.toml).

### DIFF
--- a/tools/cargo-miden/src/lib.rs
+++ b/tools/cargo-miden/src/lib.rs
@@ -5,11 +5,12 @@
 
 use std::path::Path;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{bail, Result};
 use cargo_component::{
     config::{CargoArguments, Config},
-    load_component_metadata, load_metadata, run_cargo_command,
+    load_component_metadata, load_metadata, run_cargo_command, PackageComponentMetadata,
 };
+use cargo_metadata::{camino, Metadata};
 use clap::{CommandFactory, Parser};
 pub use commands::example_project::WIT_DEPS_PATH;
 use commands::{ExampleCommand, NewCommand};
@@ -17,6 +18,7 @@ use compile_masm::wasm_to_masm;
 use dependencies::process_miden_dependencies;
 use midenc_session::{ProjectType, RollupTarget, TargetEnv};
 use non_component::run_cargo_command_for_non_component;
+use path_absolutize::Absolutize;
 pub use target::{
     detect_project_type, detect_target_environment, target_environment_to_project_type,
 };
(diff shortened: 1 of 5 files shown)


bitwalker:
I suppose the thing to do here is to document the fact that the profiles need to be defined at the workspace level, but it is a bit annoying that `--manifest-path` doesn't work the way I thought it did (i.e. by making the provided path the "root" manifest).

examples/basic-wallet/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 // However, we could still use some standard library types while
 // remaining no-std compatible, if we uncommented the following lines:
 //
-extern crate alloc;
```
bitwalker, line 7: Why is this commented out? Seems to me that if you don't have the alloc crate linked, then we don't need to define a `#[global_allocator]` either. But also, we may as well remove the commented-out line as well if we don't actually need to link liballoc.

examples/storage-example/src/lib.rs
```diff
@@ -4,7 +4,7 @@
 // However, we could still use some standard library types while
 // remaining no-std compatible, if we uncommented the following lines:
 //
-extern crate alloc;
+// extern crate alloc;
```
bitwalker, line 7: Same here

bitwalker (approved):
LGTM, though I had one question. Marking this approved anyway

examples/basic-wallet/src/lib.rs
greenhat: I see examples as a starting point for our users to build their own code. This is why I want to leave it along with the comment. I commented out `extern crate alloc;` since it's not needed for this example code.

`#[global_allocator]` is required for no_std. Otherwise the following error is thrown:
```
error: no global memory allocator found but one is required; link to std or add `#[global_allocator]` to a static item that implements the GlobalAlloc trait
```

greenhat:
> I suppose the thing to do here is to document the fact that the profiles need to be defined at the workspace level, but it is a bit annoying that `--manifest-path` doesn't work the way I thought it did (i.e. by making the provided path the "root" manifest).

I think I found a better way - https://github.com/0xMiden/compiler/issues/648#issuecomment-3232892940

PR #656, opened 2025-09-03
Refactor operand scheduling for binary ops.

### LINKED ISSUE #614: Refactor and test two-operand stack scheduling in `HirLowering::schedule_operands`
Extract and split the `if is_binary {}` code into manageable pieces and cover them with unit tests. 
Consider encoding stack manipulations (`dup`, `swap`, etc.) separately and not using `emitter` to ease the testing.

otrho:
This PR has two commits and should probably be reviewed per commit.

The first commit changes the solver to use stacks of `ValueOrAlias` rather than `Operand`.  The old code tries very hard to keep the `.pos` member consistent with the place it sits in the stack, and by removing it we can instead just use the literal position of the value in the stack.

The second commit takes the binary op special case handling out of the top of `lowering.rs` and puts it into a proper tactic for the solver to use.  It still creates a solution for every possible binary operation though, and will take priority for binary ops.

To handle commutative ops, which is where most of the benefits for special casing come from, a `may_be_unordered` flag is sent deep into the solver.  When checking if a solution is valid it now checks for exact matching, or a relaxed match if the output may be unordered.

Testing for the new tactic uses full permutations of every possible input and ensures the output is valid, has at most 2 actions and has a check to ensure no regressions.

One part I'm unsure of, and is commented out in this PR, is handling cases where there are multiple copies of a single value already in the input stack and especially the case where they should both be moved.

When building the context for the solver [any dupes are found and converted into copies](https://github.com/0xMiden/compiler/blob/369bc7925775a5756cb8b6c8864cd6fe053c3537/codegen/masm/src/opt/operands/context.rs#L45-L49), regardless of the actual passed constraints.  I'm not sure why this is a thing, but when I attempted to test this scenario it foiled my efforts. :)  @bitwalker maybe you know why it's done and if it's still necessary.  Is it perhaps impossible for the input to have multiple copies of the same value?  I would guess not.

Closes #614.

### DIFF
--- a/codegen/masm/src/opt/operands/context.rs
+++ b/codegen/masm/src/opt/operands/context.rs
@@ -1,8 +1,9 @@
 use core::num::NonZeroU8;
 
 use midenc_hir::{self as hir, hashbrown, FxHashMap};
+use smallvec::SmallVec;
 
-use super::{SolverError, Stack, ValueOrAlias};
+use super::{SolverError, SolverOptions, Stack, ValueOrAlias};
 use crate::{Constraint, OperandStack};
 
 /// The context associated with an instance of [OperandMovementConstraintSolver].
@@ -12,17 +13,17 @@ use crate::{Constraint, OperandStack};
 /// copied operands.
 #[derive(Debug)]
 pub struct SolverContext {
+    options: SolverOptions,
     stack: Stack,
     expected: Stack,
-    allow_unordered: bool,
     copies: CopyInfo,
 }
 impl SolverContext {
     pub fn new(
         expected: &[hir::ValueRef],
-        allow_unordered: bool,
         constraints: &[Constraint],
         stack: &OperandStack,
+        options: SolverOptions,
     ) -> Result<Self, SolverError> {
         // Compute the expected output on the stack, as well as alias/copy information
         let stack = Stack::from(stack);
@@ -31,14 +32,18 @@ impl SolverContext {
         for (value, constraint) in expected.iter().rev().zip(constraints.iter().rev()) {
             let value = ValueOrAlias::from(*value);
             match constraint {
-                // If we observe a value with move semantics, then it is
-                // always referencing the original value
+                // If we observe a value with move semantics, then it is always referencing the
+                // original value
                 Constraint::Move => {
                     expected_output.push(value);
                 }
-                // If we observe a value with copy semantics, then the expected
-                // output is always an alias, because the original would need to
-                // be preserved
+                // If we observe a value with copy semantics, then the expected output is always an
+                // alias, because the original needs to be preserved.
+                //
+                // NOTE: Just because an expected operand has a copy constraint applied, does not
+                // mean that there aren't multiple copies on the input stack already - it indicates
+                // that all of those copies must be preserved, and a _new_ copy must be materialized
+                // by the solution we produce.
                 Constraint::Copy => {
                     expected_output.push(copies.push(value));
                 }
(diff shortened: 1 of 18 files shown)


codegen/masm/src/opt/operands/tactics/two_args.rs
```diff
@@ -0,0 +1,382 @@
...
+                             {:?}\ninput: {:?}\nexpected: {:?}\noutput: {:?}",
+                            constraint_pair[1],
+                            constraint_pair[0],
+                            &pending,
+                            &context.expected(),
+                            &builder.stack()
+                        );
+
+                        let num_actions = builder.take().len();
+                        assert!(builder.take().len() <= max_actions);
```
greenhat: `builder.take()` is called twice.

codegen/masm/src/opt/operands/tactics/two_args.rs
```diff
@@ -0,0 +1,382 @@
...
+                }
+            }
+        }
+
+        total_actions
+    }
+
+    #[test]
+    fn every_ordered_stack() {
+        // Take every permutation of a 5 element stack and each permutation of two operand contraints
```
greenhat:
```suggestion
        // Take every permutation of a 5 element stack and each permutation of two operand constraints
```

codegen/masm/src/opt/operands/tactics/two_args.rs
```diff
@@ -0,0 +1,382 @@
...
+        // and confirm that at most 1 action is required for an unordered solution.
+        let val_refs = generate_valrefs(3);
+        let total_actions = permute_stacks(&val_refs, 1, true);
+
+        // This number should only ever go down as we add optimisations.
+        midenc_expect_test::expect!["36"].assert_eq(&total_actions.to_string());
+    }
+
+    fn every_duplicated_stack_single_util(may_be_unordered: bool) -> usize {
+        // Take very permutation of a 4 element stack etc. where the two operands are the very same
```
greenhat:
```suggestion
        // Take every permutation of a 4 element stack etc. where the two operands are the very same
```

codegen/masm/src/opt/operands/context.rs
```diff
@@ -44,21 +47,25 @@ impl SolverContext {
...
+                .iter()
+                .eq(stack.iter().skip(stack.len() - expected_output.len()));
         if is_solved {
             return Err(SolverError::AlreadySolved);
         }
 
         Ok(Self {
             stack,
             expected: expected_output,
+            unordered: may_be_unordered,
```
greenhat: Consider naming both (field and parameter) the same name. As a side note, would `allow_unordered` fit better?

codegen/masm/src/opt/operands/solver.rs
```diff
@@ -147,6 +157,7 @@ impl OperandMovementConstraintSolver {
                 self.tactics.push(Box::new(Linear));
                 self.tactics.push(Box::new(CopyAll));
             }
+            self.tactics.push(Box::new(TwoArgs));
```
greenhat: Consider adding a debug_assert before running the solver to ensure `TwoArgs` is running first. Some red light if we mess with the order in the future.

greenhat (commented):
Looking good! Just one question and a few nits.

codegen/masm/src/opt/operands/tactics/two_args.rs
otrho: Ooh, nice catch.  That has unsurprisingly revealed some wrong assumptions I had about the scheduling.  I had it in my head that for a 3 element stack it would only need 1 action to solve, if unordered was allowed.  And so I made the test and it passed.  Wow.  But why was the total actions so high?  Dunno, whatever, it passes.

But of course this makes it pass.  The assumption I had was true, but only for Move/Move, so I've updated the test.  Thanks.

codegen/masm/src/opt/operands/tactics/two_args.rs
otrho: I can't tell you how often I write 'contraint' instead of 'constraint'.  Some sort of glitch.

codegen/masm/src/opt/operands/context.rs
otrho: Thanks.  This really bugged me.  I had `can_be_unordered` and thought yuck, but then couldn't think of anything better and used `may_be_unordered` in the end, but still yuck.  `allow_unordered` is much better. :facepalm: I've changed it everywhere.  The predicate to query it is `.unordered_allowed()` though since it's a little better there.

codegen/masm/src/opt/operands/solver.rs
otrho: I looked at this but whatever I could think of to put into an assert either couldn't work or was intrusively gnarly.

I think if the `TwoArgs` isn't last (or popped first) then the integration tests will probably fail, since none of the other tactics will account for commutivity.  But your point is if/when that happens it might not be obvious why, where an assert would clarify that.

greenhat: If the test only fails because of the expected file changes, it's easy to miss. I'd even put the assert right after this line so it'd also serve as documentation of why it should be first (to pop).

bitwalker (approved):

Issue #281, opened 2024-08-27
Hitting the "expected switch instructions to have been rewritten before stackification" panic in `BlockEmitter::emit_inst`when compiling rust enum


greenhat:
Hitting panic in the `BlockEmitter::emit_inst` https://github.com/0xPolygonMiden/compiler/blob/c2053e6b6f568486df7a05498a89f8c51d586a64/codegen/masm/src/codegen/emitter.rs#L246-L248

How to reproduce:
- uncomment `expect_masm` in `types::test_enum` integration test added in #280
 
https://github.com/0xPolygonMiden/compiler/blob/2ccdf23fa606ef0ba29f734394ad7dae6ff566a6/tests/integration/src/rust_masm_tests/types.rs#L11

bitwalker:
Hmm, looks like we never actually lower these to a chain of if/then/else in our pipeline. I'll add this to my TODO for the IR refactoring

bitwalker:
Fixed by the switch to HIR2

PR #665, opened 2025-09-09
Re-enabled the commented out `rust_masm_tests::types::test_enum` test.

otrho:
Missed when #281 was closed.


greenhat (approved):
Good find!

PR #666, opened 2025-09-09
[1/2] Authentication component rollup target

### LINKED ISSUE #660: New project template and `RollupTarget` for the authentication `AccountComponent`
The compiled Miden package is expected to export exactly one `auth__*` procedure.
Test by compiling the `auth_no_auth` MASM example from https://github.com/0xMiden/miden-base/blob/e4912663276ab8eebb24b84d318417cb4ea0bba3/crates/miden-lib/asm/account_components/no_auth.masm?plain=1 and using it with `miden-client` in some local network integration test.

### LINKED ISSUE #667: Don't export core Wasm function that was lifted as component export
Discovered in #666 

We're exporting both core Wasm and lifted component export functions: 
```
        "miden::base::authentication-component@1.0.0::auth-procedure",
        "miden:base/authentication-component@1.0.0::auth-procedure",
```

See the HIR - https://github.com/0xMiden/compiler/blob/4e46ea60c0f0453ea2704a2c8cf8d5f61916b34b/tests/integration/expected/rust_sdk/auth_component_no_auth.hir?plain=1
And the MASM - https://github.com/0xMiden/compiler/blob/4e46ea60c0f0453ea2704a2c8cf8d5f61916b34b/tests/integration/expected/rust_sdk/auth_component_no_auth.masm?plain=1

greenhat:
Close #660 
Close #667 

Corresponding template repo PR - https://github.com/0xMiden/rust-templates/pull/21

I suggest reviewing this PR on a per-commit basis.

This PR:
- Adds new project template for the authentication component (`cargo miden new --auth-component [NAME]`);
- Adds `example/auth-component-no-auth` example project for no-auth version;
- Adds a local network test using the `example/auth-component-no-auth` as an authentication component using `miden-client` in the counter contract scenario;
- Fixes a few issues discovered along the way.

### DIFF
--- a/examples/auth-component-no-auth/Cargo.lock
+++ b/examples/auth-component-no-auth/Cargo.lock
@@ -491,7 +491,7 @@ checksum = "32a282da65faaf38286cf3be983213fcf1d2e2a58700e808f83f4ea9a4804bc0"
 
 [[package]]
 name = "miden"
-version = "0.4.1"
+version = "0.5.0"
 dependencies = [
  "miden-base",
  "miden-base-sys",
@@ -531,7 +531,7 @@ dependencies = [
 
 [[package]]
 name = "miden-base"
-version = "0.4.1"
+version = "0.5.0"
 dependencies = [
  "miden-base-macros",
  "miden-base-sys",
@@ -540,7 +540,7 @@ dependencies = [
 
 [[package]]
 name = "miden-base-macros"
-version = "0.4.1"
+version = "0.5.0"
 dependencies = [
  "miden-objects",
  "proc-macro2",
@@ -552,7 +552,7 @@ dependencies = [
 
 [[package]]
 name = "miden-base-sys"
-version = "0.4.1"
+version = "0.5.0"
 dependencies = [
  "miden-stdlib-sys",
 ]
@@ -673,11 +673,11 @@ dependencies = [
 
 [[package]]
 name = "miden-sdk-alloc"
-version = "0.4.1"
+version = "0.5.0"
 
 [[package]]
 name = "miden-stdlib-sys"
-version = "0.4.1"
+version = "0.5.0"
 
 [[package]]
 name = "miden-verifier"


codegen/masm/src/lower/component.rs
```diff
@@ -217,7 +217,7 @@ impl MasmComponentBuilder<'_> {
             let init_body = core::mem::take(&mut self.init_body);
             let init = masm::Procedure::new(
                 Default::default(),
-                masm::Visibility::Public,
```
bitwalker, line 220: The init procedure needs to be public, so that it can be called from any `call`able procedure in the component that needs to initialize a fresh context in its prologue.

greenhat: I'm not sure that I understand why `init` needs to be public. We only invoke the `init` in our lifted exports functions (see an example at https://github.com/0xMiden/compiler/blob/4efe0635d73c2b7f94cbdd59557ec803c2426c3a/tests/integration/expected/examples/basic_wallet.masm?plain=1#L13-L21) which we place in the same module. The lifted export function is then `call`ed without any prologue and epilogue at the call site. So the lifted export functions are the only public procedures in the component.

See also the discussion that led to making `init` private at https://github.com/0xMiden/miden-base/issues/1877

bitwalker: It should be noted that the fact we're emitting a single module for a given Wasm binary is incidental here, it isn't an inherent structural thing, just an artifact of how Rust compiles to Wasm.

In any case, if we can guarantee that all exports from a component (including a program entrypoint, if present), are emitted to the same underlying MASM module, sure, the `init` procedure can be private in that case. I didn't think that was something we guaranteed, so much as it was a happy accident of how things get compiled from Rust.

I'm fine with making this private until there is an actual issue in practice, but wanted to raise the chance of it being an issue during review to make sure the change is actually well-motivated.

PR #676, opened 2025-09-14
updated `StorageMapAccess::get` return `Option<V>`

### LINKED ISSUE #673: SDK: make `StorageMapAccess::get` return `Option<V>`
Defined at https://github.com/0xMiden/compiler/blob/9f7ec43d21f81c17946258a8b4dc3168690f3548/sdk/base/src/types/storage.rs?plain=1#L33.

### Why
The underlying `storage::get_map_item` returns zero word ([0, 0, 0, 0]) in case there is no value associated with the given key. Let's make `get` more in sync with idiomatic Rust.

PoulavBhowmick03:
Fixes #673

### DIFF
--- a/sdk/base/src/types/storage.rs
+++ b/sdk/base/src/types/storage.rs
@@ -30,7 +30,7 @@ impl<V: Into<Word> + From<Word>> ValueAccess<V> for Value {
 
 pub trait StorageMapAccess<K, V> {
     /// Returns a map item value for `key` from the account storage.
-    fn get(&self, key: &K) -> V;
+    fn get(&self, key: &K) -> Option<V>;
     /// Sets a map item `value` for `key` in the account storage and returns (old_root, old_value)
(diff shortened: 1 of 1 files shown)


sdk/base/src/types/storage.rs
```diff
@@ -44,8 +44,8 @@ impl<K: Into<Word> + AsRef<Word>, V: From<Word> + Into<Word>> StorageMapAccess<K
 {
     /// Returns a map item value from the account storage.
     #[inline(always)]
-    fn get(&self, key: &K) -> V {
-        storage::get_map_item(self.slot, key.as_ref()).into()
+    fn get(&self, key: &K) -> Option<V> {
+        Some(storage::get_map_item(self.slot, key.as_ref()).into())
```
greenhat, line 48: The idea was to test if get_map_item returns zero word and if so - return None. See the Why section in #673.

greenhat (changes requested):
Thank you! Please check my comment.

PR #681, opened 2025-09-17
wip: update compiler to use miden-vm v0.19

bitwalker:
This PR implements the necessary changes to migrate to the v0.18 release of the various VM crates we rely on.

**NOTE:** This cannot be merged until v0.18 is released and `miden-base`, `miden-client`, and `miden-node` have migrated (and ideally also been released, but we could also pin to a git ref like `next` temporarily if need be). That should be some time next week I believe. Until then, any of our integration tests that depend on those crates will fail to compile.

The main things that change here are:

* The addition of procedure type signature information in library/package metadata
* The removal of the `account_component_metadata` field of `miden_mast_package::Package`, and the addition of first-class custom section support (where the account component metadata is now stored).
* The removal of many `miden_assembly_syntax::ast::Instruction` variants related to MASM's `push` instruction, replacing them with `Push(IntValue)` and `PushWord(WordValue)` variants. This required changing a bunch of uses of the old variants in the codegen backend, which has been refactored a bit to avoid hardcoding what `Instruction` variant is used in all those places, instead centralizing it in the emitter.
* Some minor changes to the `SyncHost`/`AsyncHost` traits of the processor


bitwalker:
@greenhat This PR is basically done, but I'm not familiar enough with the client changes to fix some of the broken test code due to changes in the client API. Feel free to check out my branch and push any fixes you have, and review/merge when you're ready. Let me know if you need me to look into anything

tools/cargo-miden/src/template.rs
```diff
@@ -0,0 +1,580 @@
...
+    clone_repository(repo, template_path, temp_dir.path())?;
+
+    Ok(TemplateSource {
+        root: temp_dir.path().to_path_buf(),
+        _keepalive: Some(temp_dir),
+    })
+}
+
+fn clone_repository(repo: &str, template_path: &TemplatePath, destination: &Path) -> Result<()> {
+    let mut command = Command::new("git");
```
bitwalker, line 131: I'd pass `--single-branch` too, so that other branches aren't fetched (which still happens even with `--branch`)

bitwalker: Though I think we should only do this when `--rev` is not used, since that option necessarily requires a full clone

greenhat: Thanks! Added.

PR #689, opened 2025-09-25
[1/x] feature: add `hash_words` in Miden SDK

### LINKED ISSUE #644: Expose `hash_memory_words` stdlib procedure on the Rust side

greenhat:
Close #644 

This PR adds `hash_words` in Miden SDK implemented with stdlib's `hash_memory`


sdk/stdlib-sys/src/stdlib/crypto/hashes.rs
```diff
@@ -199,7 +199,7 @@ pub fn sha256_hash_2to1(input: [u8; 64]) -> [u8; 32] {
 /// This maps to the `std::crypto::rpo::hash_memory` procedure in the Miden stdlib.
 ///
 /// # Arguments
-/// * `elements` - A slice of field elements to be hashed
+/// * `elements` - A Vec of field elements to be hashed
 #[inline]
 pub fn hash_elements(elements: Vec<Felt>) -> Digest {
```
bitwalker, line 204: Question: Is there any reason not to have this function take a slice rather than a `Vec`? Seems odd to pass a Vec by value here.

bitwalker (approved):
LGTM, just one question, but not anything that holds up this PR

sdk/stdlib-sys/src/stdlib/crypto/hashes.rs
greenhat: Yes. I did it to ensure the pointer is word-aligned because the underlying `hash_memory` expects a word-aligned address. Since our `BumpAlloc` produces word-aligned allocations I assume the pointer should always be word-aligned.

bitwalker: We should make an issue to add something akin to `Cow<'a, T>` that can be created from either an owned vector or a borrowed slice, and passes by reference when the reference is >= word-aligned. That way we can avoid cloning the data when we already know the data is word-aligned.

greenhat: Nice! Done - https://github.com/0xMiden/compiler/issues/690

PR #691, opened 2025-09-26
Migrate to `wit-bindgen::generate!` macro

### LINKED ISSUE #671: Consider generating Rust bindings using `wit_bindgen::generate!` macro instead of during the `cargo miden build` run
### Motivation

1. After the `cargo miden new` command opening the project in IDE results in LSP errors due to the missing `bindings.rs` file. It will be generated during the `cargo miden build` command run.

2. After editing the WIT interface the `bindings.rs` will be updated only after `cargo miden build` runs.

### How

Ditch the `bindings.rs` and its generation on `cargo miden build` and call the [`wit_bindgen::generate!`](https://docs.rs/wit-bindgen/latest/wit_bindgen/macro.generate.html#options-to-generate) macro directly in the user `lib.rs`. We can later wrap it in pursuing better ergonomics in #336.

Alternatively, we can introduce `cargo miden check` and put it in the vscode local setting file instead of `cargo check`. It'll generate the `bindings.rs` and call `cargo check` afterward. I don't like this approach since it's quite fragile and per-IDE specific.

greenhat:
Close #671 

TODO:
- [ ] migrate the rest of the `examples` and `rust_sdk` tests;
- [ ] migrate new project templates;


examples/auth-component-no-auth/src/lib.rs
```diff
@@ -19,13 +19,13 @@ fn my_panic(_info: &core::panic::PanicInfo) -> ! {
     loop {}
 }
 
-bindings::export!(AuthComponent with_types_in bindings);
+miden::miden_generate!();
```
bitwalker: *Nit:* Maybe we should rename the macro to just `generate!`, so that this can read as `miden::generate!()` - it's a minor aesthetic improvement, but a lot easier to remember and say out loud.

bitwalker (approved):
LGTM!

examples/auth-component-no-auth/src/lib.rs
greenhat: Good idea! I wanted the name to be different from the `wit_bindgen`'s  macros, but I have no solid reason for this.

greenhat: Done.

greenhat:
@otrho Heads up. `generate!` macro is merged.

PR #692, opened 2025-09-26
Migrate docs from mdbook to docusaurus

Keinberger:
# Summary

This PR migrates the existing documentation from **mdbook** to **Docusaurus**. This includes syntax changes and structural updates.

❗**ATTENTION**: Do not merge this PR yet as merging has to be done in a specific order to ensure a smooth transition from mdbook to Docusaurus for our docs web page.

## Note

1. This PR changes the markdown syntax from mdbook to Docusaurus. Please familiarize yourself with the new Docusaurus syntax.
2. The new markdown files have been moved to `docs/src`.
3. I separated the exported md files from non-exported md files that are intended as "internal docs". The exported files can be found inside of `docs/src/external`, internal docs inside of `docs/src/internal`.

## Questions & Clarifications

I kept the md files inside of `docs/src/internal` to keep supporting mdbook, so that the internal docs can be hosted as an mdbook on [0xmiden.github.io/miden-node/](https://0xmiden.github.io/miden-node/). These files are files that are currently not exported to the Miden book.

That said, I believe it would be better to migrate these files to Docusaurus syntax as well. This would allow us to integrate the internal docs into the external documentation in the future, while also ensuring that developers only need to work with a single, consistent syntax (Docusaurus Markdown) instead of switching between mdBook and Docusaurus.

@bitwalker What do you think?

### Syntax changes

As already mentioned before, this PR changes the markdown syntax of all docs markdown files.
Going forward all documentation will also be written using Docusaurus sytnax.

**NOTE:** Please review the updated Markdown files to understand the new format.

### Local testing

To test docs locally before pushing changes:

1. Cd into docs:
```bash
cd docs/external
```

2. Install packages:
```bash
npm install
```

4. Start the local dev server:
```bash
npm run start:dev
```

**Note:** The `index.md` file might not appear in the correct sidebar order during local dev. In production deployments it will. To guarantee proper ordering, ensure `index.md` includes:

```yaml
---
sidebar_position: 1
---
```

### New Docs Architecture

The documentation system no longer relies on cron jobs or local scripts. Instead, changes are propagated through a CI-driven process. Here are the key changes:

1. Removal of cron task for initiating a bash script

Instead of relying on a cron task to update the docs, every repository now integrates a "trigger-docs-deploy.yml" workflow, which automatically triggers a rebuild and redeployment of the docs upon changes to `next`.

2. Versioning support

Upon Testnet deployments, a dedicated workflow inside of the miden-docs repository can be triggered to create a snapshot of the docs for that specific version. This workflow has to be triggered manually.


.github/workflows/trigger-deploy-docs.yml
```diff
@@ -0,0 +1,21 @@
+name: Trigger Aggregator Docs Rebuild
+
+on:
+  push:
+    branches: [next]
```
bitwalker, line 5: We probably want to only trigger this workflow if files in the `docs` directory change, no?

.github/workflows/trigger-deploy-docs.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Send repository_dispatch to aggregator
+        uses: peter-evans/repository-dispatch@v3
```
bitwalker: Let's pick a specific SHA revision of this action - depending on tags (which are mutable) is extremely risky when secrets are involved. We need to do this with our other `uses`, but I don't want to add any new GHA dependencies without selecting a revision that we've reviewed ourselves and vetted.

bitwalker:
@Keinberger You'll also have to make sure commit signing is turned on and sign your commits, let me know if you need any help getting that set up - I think all of our repos currently require it in order to merge, and while I can bypass it for this PR, you'll need to get that set up anyway, so we may as well do it now 🙂

.github/workflows/trigger-deploy-docs.yml
Keinberger: Absolutely agreed, will fix!

.github/workflows/trigger-deploy-docs.yml
Keinberger: Absolutely agree with you on this one. I will pick a specific SHA ref and we can review it for safety.

Keinberger: How about we plan on taking this SHA ref?: [a628c95fd17070f003ea24579a56e6bc89b25766](https://github.com/peter-evans/repository-dispatch/commit/a628c95fd17070f003ea24579a56e6bc89b25766#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519)

It's the latest version that was released yesterday.

Keinberger:
Hey @bitwalker ,

I implemented the changes you asked for. Also implemented signed commits, my last commit is signed as you can see.

Furthermore, based on discussions I had with Philipp and Bobbin on the Docusaurus PR on the `miden-base` repository, I also included some more changes on this branch:
- `_category_.json` files are now in YAML format. That way we can add a comment in each file to explain the position parameter
- Added a small section in the README that explains the docs architecture (NPM dependencies and absorption into main docs page)
- Added a new `build-docs.yml` workflow to check md syntax for the exported docs

bitwalker (approved):

bitwalker:
Shit, I just remembered I wasn't supposed to merge this 🤦 - @Keinberger will there be any issues once the other repos are migrated?

PR #703, opened 2025-10-06
feat: enable `==` operator for `AccountId` type

### LINKED ISSUE #702: Enable `==` Comparison for `AccountId` Types
### Feature description

**Description:**
Currently, `AccountId` types from `miden` cannot be compared using the `==` operator because `PartialEq` is not implemented. This prevents simple equality checks between accounts in user code, for example:

```rust
fn reclaim_assets(consuming_account: AccountId) {
    let creator_account = note::get_sender();

    if consuming_account == creator_account {
        consume_assets();
    } else {
        panic!();
    }
}
```

Attempting this produces:

```
error[E0369]: binary operation `==` cannot be applied to type `miden::AccountId`
```

**Proposed Solution:**
Implement `PartialEq` (and optionally `Eq`) for `AccountId` to allow direct equality comparisons. 

### Why is this feature needed?

This would improve ergonomics and make account-based logic more intuitive.

partylikeits1983:
Resolves #702 

This PR adds on two new features to the `AccountId` type:

1) Enables the `==` operator to be used with the `AccountId` type.
2) Adds a `AccountId::from` method to convert two `Felt` types into an `AccountId`.

The P2ID script is updated to use these two new features.

### DIFF
--- a/examples/p2id-note/src/lib.rs
+++ b/examples/p2id-note/src/lib.rs
@@ -27,9 +27,11 @@ fn run(_arg: Word) {
     let inputs = note::get_inputs();
     let target_account_id_prefix = inputs[0];
     let target_account_id_suffix = inputs[1];
-    let account_id = account::get_id();
-    assert_eq(account_id.prefix, target_account_id_prefix);
-    assert_eq(account_id.suffix, target_account_id_suffix);
+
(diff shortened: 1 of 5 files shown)


greenhat (changes requested):
Looking good! Thank you! Just update the expected wat/hir/masm files.

bitwalker (approved):

PR #704, opened 2025-10-06
Add P2IDE note example

partylikeits1983:
This PR adds the P2IDE note as an example note.


tests/integration-node/src/node_tests/basic_wallet.rs
```diff
@@ -199,3 +199,202 @@ pub fn test_basic_wallet_p2id_local() {
...
+            &mut client,
+            note_package.clone(),
+            faucet_account.id(),
+            NoteCreationConfig {
+                assets: NoteAssets::new(vec![fungible_asset.into()]).unwrap(),
+                inputs: vec![
+                    alice_account.id().prefix().as_felt(),
+                    alice_account.id().suffix(),
+                    timelock_height.into(),
+                    reclaim_height.into(),
```
partylikeits1983: 4 inputs to the note

examples/p2ide-note/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+#[note_script]
+fn run(_arg: Word) {
+    let inputs = note::get_inputs();
+    let target_account_id_prefix = inputs[0];
+    let target_account_id_suffix = inputs[1];
+    let timelock_height = inputs[2];
+    let reclaim_height = inputs[3];
```
partylikeits1983, line 42: Note consumption fails when more than 2 inputs. If `timelock_height` & `reclaim_height` are commented out, note consumption works.

examples/p2ide-note/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+    // assert_eq(inputs.len().into(), Felt::from(4u32));
+
+    // sanity check
+    // assert_eq(timelock_height, Felt::from(1000u32));
+    // assert_eq(reclaim_height, Felt::from(1010u32));
+
+    // get block number
+    let block_number = tx::get_block_number();
+
+    assert!(block_number >= timelock_height);
```
greenhat, line 46: The reason for step 2 failure is this `assert!`. The `timelock_height` is 1000, but the test suite is bootstrapping the chain so the `block_number` should be low digits. If you set `timelock_height` to `1` this `assert!` is passing. After that the test fails at step 5.

partylikeits1983: I wish this was the issue. I updated the code to demonstrate that this is not the issue. The issue is that if you try to initialize more than 2 variables from note inputs:

```rs
let inputs = note::get_inputs();
let target_account_id_prefix = inputs[0];
let target_account_id_suffix = inputs[1];

let timelock_height = inputs[2]; // fails when uncommented
```

examples/p2ide-note/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+#[note_script]
+fn run(_arg: Word) {
+    let inputs = note::get_inputs();
+    let target_account_id_prefix = inputs[0];
+    let target_account_id_suffix = inputs[1];
+    
+    let timelock_height = inputs[2]; // fails when uncommented
```
partylikeits1983: fails when trying to initialize more than 2 inputs.

examples/p2ide-note/src/lib.rs
greenhat: Oh, this fails on step 5. In this case `send_asset_to_account` makes a note providing only 2 elements (account id) as note inputs. So there are only 2 elements in `inputs`.

examples/p2ide-note/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+fn run(_arg: Word) {
+    let inputs = note::get_inputs();
+    let target_account_id_prefix = inputs[0];
+    let target_account_id_suffix = inputs[1];
+    
+    let timelock_height = inputs[2];
+    let reclaim_height = inputs[3];
+
+    // make sure the number of inputs is 4
+    assert_eq(inputs.len().into(), Felt::from(4u32));
```
greenhat: The code before this `assert_eq` indexes inputs[0..=3] before checking how many items were provided, so a malformed note input will panic before this assert. Consider moving it to the start of the function.

tests/integration-node/src/node_tests/basic_wallet.rs
```diff
@@ -199,3 +199,212 @@ pub fn test_basic_wallet_p2id_local() {
...
+        eprintln!("\n=== Step 4: Alice creates p2ide note for Bob ===");
+
+        let transfer_amount = 10_000u64; // 10,000 tokens
+        let transfer_asset = FungibleAsset::new(faucet_account.id(), transfer_amount).unwrap();
+
+        let timelock_height = Felt::new(0);
+        let reclaim_height = Felt::new(0);
+
+        // Create the p2ide note with only 2 inputs (account ID)
+        // This is a simplified P2IDE note without timelock/reclaim functionality
```
greenhat: It seems like it's a full-fledged P2IDE note with a reclaim feature.

examples/p2ide-note/src/lib.rs
greenhat: There is a `felt!` macro in the SDK. So `Felt::from(4u32)` can be substituted with `felt!(4)`.

greenhat (approved):
Looking good! Thank you!

PR #706, opened 2025-10-08
Add a bitcast back to signed integer after a translating a `load` then 

otrho:
…n `zext`.


greenhat (approved):
Looking good! Thank you!

PR #708, opened 2025-10-08
Remove explicit `miden.wit` dependency and use the one bundled with Miden SDK

### LINKED ISSUE #694: Remove`miden.wit` in the Cargo.toml dependencies
Discovered while working on https://github.com/0xMiden/compiler/pull/691

### Why

Since `miden.wit` defines core Miden SDK types (Felt, Word, Asset, etc.) and interfaces we can consider it to be a prelude that should be implicitly imported in the WIT interface of every project. So there is no need to require it explicitly in the `[package.metadata.component.target.dependencies]` in Cargo.toml in every project.

### How

Since we [embed](https://github.com/0xMiden/compiler/blob/1b3f1f0a4b0d529909b9bfd652e37182b412e331/sdk/base/src/base_wit.rs?plain=1#L2) it in the `miden-base` crate we can write it on disc at `wit/deps` in our `miden_generate!` macro.

greenhat:
Close #694 

TODO:
- [ ] update new project templates


examples/auth-component-no-auth/.gitignore
```diff
@@ -1 +1,2 @@
-/target
\ No newline at end of file
+/target
+/wit-auto-generated
```
bitwalker: Why not place the auto-generated files somewhere in the `target` directory, so it gets cleaned up with other transient build artifacts? Doesn't seem like it would be necessary to keep it around or source control it.

bitwalker (changes requested):
Looks good, just one question/change regarding the choice of output directory for the generated files

examples/auth-component-no-auth/.gitignore
greenhat: I thought about `target` but decided against it since the user would most likely want to open the `miden.wit` file in their editor to look at the SDK types they can use in their account API. I think a non-target folder would convey it better.

greenhat: Essentially my line of thinking was that the user will not look inside the `target` folder for something to use as a reference. And moreover, the user's editor conceals the `target` directory completely. And since we are pushing`miden.wit` path to `wit-bindgen::generate!` behind the scenes, the user would have no idea that they need to look for the `miden.wit` file in the `target` folder.

bitwalker: Conventionally, generated files are placed in `OUT_DIR`, which is in the target directory - our generated LALRPOP grammars are handled this way, for example.

This is particularly the case for generated files of dependencies, where the contents are dependent on what version of the dependency they are generated from, rather than something the user controls.

Regardless, since this directory is to be gitignore'd, editors are going to treat them as things to be ignored anyway, so we may as well avoid adding yet more clutter to users' projects if we can help it. Picking a predictable place under `target` still provides an easy way to view this file if desired IMO.

bitwalker: As an aside, Zed doesn't hide the `target` directory (or other gitignore'd directories), instead they are just greyed out in the project panel (where files/directories are listed hierarchically), and their contents are not indexed for project-wide search. It is still easy to navigate to files you know are there using the project panel though. The same was true of Emacs when I was still actively using it. I would be surprised if other editors completely hid those things from view, but obviously there might be some configurations that do.

Doesn't really change the fact that generated stuff is often more awkward to access directly - but that's typically because one isn't supposed to look at them directly anyway, except infrequently. I think some of the ergonomics here are going to depend on the editor support for WIT though, and how WIT imports are resolved by a WIT LSP - it may be the case that we have to be _more_ explicit about where these files are, just so that basic LSP features work. That said, I'm not sure if there even is any LSP implementations for WIT yet.

greenhat: Yeah, you're right. Putting it in the `.gitignore` will probably hide it anyway in editors, so there is not much sense in keeping it outside the `OUT_DIR`. I moved it under the `OUT_DIR` in https://github.com/0xMiden/compiler/pull/708/commits/4757d18a2ddfc7b1490b5c13962d650c67900e84

PR #713, opened 2025-10-10
refactor(codegen-masm): remove petgraph from Linear tactic, add no_std SCC

### LINKED ISSUE #447: Remove use of `petgraph` in `midenc-codegen-masm`
In order to support no-std builds of the compiler, we must remove any usage of crates/dependencies which are not no-std compatible (or gate them behind #[cfg(feature = "std")].

In the midenc-codegen-masm crate, we make use of petgraph to compute the strongly connected components of a simple graph, representing the edges connecting operands that should be swapped to schedule the operand stack. This is overkill, and we already have code for computing SCCs elsewhere in the compiler. We should evaluate whether the existing SCC code can be used, or if it is not suitable, simply implement it in the operand stack solver code for that specific use case, since it is a simple algorithm.

radik878:
- Replace DiGraphMap + kosaraju_scc with a minimal no_std-friendly SimpleDiGraph and local Kosaraju SCC.
- Remove petgraph from codegen/masm/Cargo.toml.
- Use FxHashMap and SmallVec; avoid std-only types.

fix #447

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -347,7 +347,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "234113d19d0d7d613b40e86fb654acf958910802bcceab913a4f9e7cda03b1a4"
 dependencies = [
  "memchr",
- "regex-automata 0.4.9",
+ "regex-automata",
  "serde",
 ]
 
@@ -1678,8 +1678,8 @@ dependencies = [
  "aho-corasick",
  "bstr",
  "log",
- "regex-automata 0.4.9",
- "regex-syntax 0.8.5",
+ "regex-automata",
+ "regex-syntax",
 ]
 
 [[package]]
@@ -2057,7 +2057,7 @@ dependencies = [
  "globset",
  "log",
  "memchr",
- "regex-automata 0.4.9",
+ "regex-automata",
  "same-file",
  "walkdir",
  "winapi-util",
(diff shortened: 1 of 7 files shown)


codegen/masm/src/opt/operands/tactics/linear.rs
```diff
@@ -238,6 +234,133 @@ impl Tactic for Linear {
     }
 }
 
+/// Minimal, no_std-friendly directed graph for `Operand` nodes with SCC computation
+#[derive(Default)]
+struct SimpleDiGraph {
```
bitwalker: I think it'd be worthwhile to extract this to `hir/src/adt/digraph.rs`, and make it generic over the node type, but only provide an implementation of the graph for node types which implement `Hash`.

We may have other uses for a basic digraph data structure, so it makes sense to provide a common implementation that we can extend as needed.

Other than that, providing some tests for the provided APIs (namely `neighbors_outgoing`, `reverse`, and `sccs`) and we're good to go!

bitwalker (changes requested):
Looks great! I'd like to extract the basic digraph data structure so we can use it elsewhere when needed, but we don't need to expand on the APIs you've provided here now. See my other comment for details.

Aside from that, and a few tests to cover the new data structure, it'll be nice to finally close out this issue!

radik878:
> Looks great! I'd like to extract the basic digraph data structure so we can use it elsewhere when needed, but we don't need to expand on the APIs you've provided here now. See my other comment for details.
> 
> Aside from that, and a few tests to cover the new data structure, it'll be nice to finally close out this issue!

Thanks for review, ser, i've made corrections

bitwalker (approved):
Turns out `petgraph` now supports no-std builds, so I've modified this PR to use that instead, as well as clean up a few other dep-related things I spotted

PR #718, opened 2025-10-15
Fix a bug in `mem::store_sw` intrinsic which used incorrect masks in unaligned writes.

### LINKED ISSUE #719: Add tests for unaligned writes using the `mem::store_dw` intrinsic.
See the tests in #718 and replicate for `mem::store_dw`.

otrho:
`mem::store_dw` is probably similarly broken, I haven't checked.

Closes #719.

### DIFF
--- a/tests/integration/src/codegen/intrinsics/mem.rs
+++ b/tests/integration/src/codegen/intrinsics/mem.rs
@@ -765,14 +765,7 @@ fn store_unaligned_u32() {
         builder.ret(Some(result), SourceSpan::default()).unwrap();
     });
 
-    fn run_test(
-        package: &miden_mast_package::Package,
-        context: &std::rc::Rc<midenc_hir::Context>,
-        write_to: u32,
-        idx: u32,
-        expected0: u32,
-        expected1: u32,
-    ) {
+    let run_test = |offs: u32, expected0: u32, expected1: u32| {
         // Initialise memory with some known bytes.
         let initializers = [Initializer::MemoryBytes {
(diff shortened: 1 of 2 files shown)


otrho:
Ugh, typo in the commit message. :facepalm:

tests/integration/src/codegen/intrinsics/mem.rs
```diff
@@ -731,3 +731,99 @@ fn store_u8() {
...
+        // Return a constant to satisfy test infrastructure
+        let result = builder.u32(1, SourceSpan::default());
+        builder.ret(Some(result), SourceSpan::default()).unwrap();
+    });
+
+    fn run_test(
+        package: &miden_mast_package::Package,
+        context: &std::rc::Rc<midenc_hir::Context>,
+        write_to: u32,
+        idx: u32,
```
greenhat: `idx` -> "offset"? Also, the first three arguments can be removed since we're passing the same values on every call.

greenhat:
> `mem::store_dw` is probably similarly broken, I haven't checked.

Yeah, that might be the case. Worth checking out.

tests/integration/src/codegen/intrinsics/mem.rs
otrho: I made `run_test` a function (hence requiring those args to be passed) because I assumed there'd be some borrowing conflicts if it was a closure.  But a quick check says that `run_test` can be a closure afterall.  OK.

PR #721, opened 2025-10-16
refactor(cargo-miden): unify core Wasm and component build process and clean-up `cargo-component` remnants

greenhat:
Everything that I missed in #709.

### DIFF
--- a/tools/cargo-miden/src/config.rs
+++ b/tools/cargo-miden/src/config.rs
@@ -22,6 +22,7 @@ use std::{collections::BTreeMap, fmt, fmt::Display, path::PathBuf, str::FromStr}
 
 use anyhow::{anyhow, bail, Context, Result};
 use cargo_metadata::Metadata;
+use midenc_session::ColorChoice;
 use parse_arg::{iter_short, match_arg};
 use semver::Version;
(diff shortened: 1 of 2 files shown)


tools/cargo-miden/src/config.rs
```diff
@@ -357,6 +355,42 @@ impl Args {
     }
 }
 
+/// The supported color options of `cargo`.
+#[derive(Default, Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub enum Color {
```
bitwalker: We can probably use the implementation in `midenc_session`, no?

tools/cargo-miden/src/non_component.rs
bitwalker, line 1: Should we just remove this whole file? Or are you planning to bring it back at some point.

tools/cargo-miden/src/config.rs
greenhat: Thanks! Done.

tools/cargo-miden/src/non_component.rs
greenhat: Oops. Thanks! Removed.

PR #722, opened 2025-10-17
[1/x] Account WIT interface generation

### LINKED ISSUE #720: WIT generation
Relevant issues and discussions:
#336 
#698 

Experiment if we could generate a WIT interface from the Rust code using a macro.

greenhat:
Close #720

Corresponding template repo PR that switches new project templates to the WIT generation: https://github.com/0xMiden/rust-templates/pull/30

The notes below are put as doc and module comments as well.

### How to use it

1. Add `#[component]` on you `impl MyAccountType {`.
2. Add `#[export_type]` on every defined type that is used in the public(exported) method signature.

Example:
```rust

#[export_type]
pub struct StructA {
    pub foo: Word,
    pub asset: Asset,
}

#[export_type]
pub struct StructB {
    pub bar: Felt,
    pub baz: Felt,
}

#[component]
struct MyAccount;

#[component]
impl MyAccount {
    pub fn (&self, a: StructA) -> StructB {
        ...
    }
}
```

Foreseeing the case where a user forgets to put `#[export_type]` on their type the friendly error message with a hint is displayed.

### How it works

1. `#[export_type]` collects the type descriptions.
2. `#[component]` on `impl` generates a WIT interface with all public methods, calls `wit_bindgen::generate!` and implements the `Guest` trait.
3. The user-defined types are mapped to the WIT types so in the generated bindings the user-defined types are used instead of the WIT-derived.

The WIT file is written in a subfolder under the `target` folder.

### Escape hatch

In a small fraction of the cases where the WIT generation is not possible (think a type defined only in an external WIT file) or not desirable, the WIT generation can be disabled:

To disable WIT interface generation:
- Don't use the `#[component]` attribute macro in the `impl MyAccountType` section;

To use the manually crafted WIT interface:
- Put the WIT file in the `wit` folder;
- call `miden::generate!();` and `bindings::export!(MyAccountType);`
- implement `impl Guest for MyAccountType`;

These steps are described in the doc comments to the `#[component]` macro.
See the [storage-example](https://github.com/0xMiden/compiler/blob/4c69e27fb72e6188ab9f2b27b557bd45c0cb9f96/examples/storage-example/src/lib.rs?plain=1#L24-L38) using this escape hatch and opting out of the WIT generation.

### Next steps

1. Implement proper `enum` and no `Vec` and `Option` support since there is no support for them in the frontend yet. In this PR I implemented `enum` without the data constructors, and it works fine. I'm pretty sure at this point that `Vec` and `Option` are also doable.

### Tests

The new test module for macros-related tests is created at [`rust_sdk::macros`](https://github.com/0xMiden/compiler/blob/568fb8bf2be4fd62de54cf4f9cd3fc99c6b99bda/tests/integration/src/rust_masm_tests/rust_sdk/macros.rs?plain=1#L6) with a test for various shapes of the user-defined types and an account <> note script execution test.


greenhat:
> Regarding the `self` argument: the example doesn't really showcase how to use it, but I'm assuming we'll be able to use it to access storage, right? (e.g., `let x = self.foo.read();` or something like that?)

Yes, exactly! Like in the storage example - https://github.com/0xMiden/compiler/blob/681d52ab02f52c1536cc4d30ff89e9d5dfe8c344/examples/storage-example/src/lib.rs?plain=1#L47-L60 but via `self` instead of `my_account` instance.

> 
> Also, I wonder if at some point in the future, we could also use `self` for common account methods so that we could do something like:
> 
> ```rust
> pub fn receive_asset(&self, asset: Asset) {
>     self.add_asset(asset);
> }
> ```
> 
> The assumption here is that `add_asset()` method could be auto-implemented on `MyAccount` struct by the `#[component]` macro.

We can generate it in the macro if we can define what tx kernel functions we want to wrap this way. I don't think asking the user is a suitable option. The DX problem I see here is how to convey to the user what functions the macros add.

Alternatively, we can define these wrappers in the `BaseWallet` trait in the SDK:
```rust
trait BaseWallet {
    fn add_asset(&mut self, a: Asset) {
        account::add_asset(a);
    }
}
```

And then a user would simply add `impl BaseWallet for MyAccount {}` to access them. This way the user has a clearer picture, what functions are added and why. I'd prefer this option vs. macro magic.

greenhat:
> > We can generate it in the macro if we can define what tx kernel functions we want to wrap this way. I don't think asking the user is a suitable option. The DX problem I see here is how to convey to the user what functions the macros add.
> 
> I think these would be all functions that are exposed for the "active account" - like getting the account ID, reading account vault, adding assets to the vault etc. (we won't need to function for storage because storage would be accessed through the fields).
> 
> We could have something like `Account` trait - e.g., something like:
> 
> ```rust
> pub trait Account {
>     fn id(&self) -> AccountId {
>         miden::account::get_id()
>     }
> 
>     ...
> }
> ```
> 
> And then have the user implement it manually - e.g.:
> 
> ```rust
> #[component]
> struct MyAccount;
> 
> impl Account for MyAccount { }
> ```
> 
> But I think it may be better if the `#[component]` macro implemented it automatically.

Works for me. The `#[component]` macro can add `impl Account for MyAccount { }`. With the `Account` trait defined in the SDK.

> > Alternatively, we can define these wrappers in the `BaseWallet` trait in the SDK:
> 
> I think this could work too - but I was trying to address a different issue that is not related to the basic wallet: every account component "inherits" functions provided by the transaction kernel, and so, when we are inside the functions of the account component, it would be more natural to say something like `self.id()` (or maybe `self.account_id()`) rather than `miden::account::get_id()`.

Oh, I meant exactly what you described above with the `Account` trait. I just picked the wrong trait name.

bitwalker (approved):
LGTM, I didn't go crazy deep into the macro code, but this stuff is still in flux and I'm comfortable with what I've seen so far. Nice work!

PR #723, opened 2025-10-20
feature(cargo-miden): passthrough unknown options in `cargo miden build` to `midenc`

### LINKED ISSUE #320: Expose `midenc` CLI features (print, run, debug) through `cargo-miden`
### Motivation

`midenc` CLI provides features that are needed for a developer (print MASM, run, debug, etc., see the [guide](https://github.com/0xPolygonMiden/compiler/blob/eac2f71fcd567886b354952bdc368d1c68cba135/docs/usage/midenc.md#L1)). These features need to be exposed through `cargo-miden` so that the developer have access to the `midenc` CLI features without installing it.

```[tasklist]
### Features to expose
- [ ] Print Wasm, HIR, MASM
- [ ] Run the compiled package 
- [ ] Load the debugger with the compiled package
```

greenhat:
Close #320 

This PR refactors `cargo miden build` to parse all command-line arguments through the `midenc` compiler's `Compiler` struct instead of using a separate custom argument parser. This adds hidden cargo-specific options (`--release`, `--manifest-path`, `--workspace`, `--package`) to the `Compiler` struct in `midenc-compile`, which are recognized by `cargo miden build` and forwarded to the underlying `cargo build` invocation. The midenc options (currently only `--emit`) are passed to the compilation stage. 

**Note:** When the `--emit` option is passed it is ignored since `cargo-miden` uses the `-o` option for the Miden package file and it overrides the `--emit`. See created #724 for details. I'd like to tackle it separately since this PR is about options passthrough.

### DIFF
--- a/tools/cargo-miden/src/commands/build.rs
+++ b/tools/cargo-miden/src/commands/build.rs
@@ -40,7 +40,7 @@ impl BuildCommand {
         // This gives us a structured representation of all options.
         let compiler_opts = Compiler::try_parse_from(&self.args).map_err(|e| {
             // Render the clap error with full formatting (colors, suggestions, etc.)
-            anyhow::anyhow!("{}", e.render())
+            anyhow::anyhow!("failed to parse 'cargo miden build' arguments: {}", e.render())
         })?;
 
         // Extract cargo-specific options from parsed Compiler struct
@@ -310,7 +310,7 @@ fn run_cargo(wasi: &str, spawn_args: &[String]) -> Result<Vec<PathBuf>> {
     Ok(outputs)
 }
 
-pub fn spawn_cargo(mut cmd: Command, cargo: &Path) -> Result<Vec<Artifact>> {
+fn spawn_cargo(mut cmd: Command, cargo: &Path) -> Result<Vec<Artifact>> {
     log::debug!("spawning command {cmd:?}");
 
     let mut child = cmd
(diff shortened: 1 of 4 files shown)


greenhat:
@bitwalker In our discussion elsewhere I mentioned the `--manifest-path`, `--workspace`, etc. options might not be a good fit in `midenc` `Compiler` struct options and you suggested I look into codegen and unstable options as an example of compartmentalization. AFAIU, they are accessible only via a prefix (`-C` for codegen and `-Z` for unstable). But the premise of going with familiar `cargo` options is that they will be used without the prefix as one uses them in `cargo build --workspace` . If let's say we choose the `-B` prefix then `--workspace` can only be invoked as `cargo miden build -B --workspace`. 
There is a value in keeping the `cargo miden build` compatible with `cargo build` options. I'd prefer to avoid the compartmentalization and add these options in the [`Compiler`](https://github.com/0xMiden/compiler/blob/49fb859935185b70973aab370e1f53d7883d53ae/midenc-compile/src/compiler.rs?plain=1#L21) struct so they can be invoked without any prefix. What do you think?

bitwalker:
@greenhat I think that's fine, let's make them first-class options then, perhaps just decorated with `#[clap(hidden(true))]` (or whatever the option is, maybe it was `hide(true)`) to keep them out of the default help of `midenc`

PR #725, opened 2025-10-20
fix: panic in emit_template pre-scan and 128-bit bnot template

sashass1315:
Remove unsafe template(0) probe in template emission and register invokes per-instruction as we emit. This eliminates potential panics when templates are not valid for n=0.

Rewrite unary::bnot multi-limb logic to:
- Invert the top two limbs with safe swaps.
- For remaining limbs, bring limb to top with movup_from_offset(i), invert, then restore with movdn_from_offset(i); this avoids invalid offsets for i < 2 and preserves operand order.

Rationale: Previous code called template(0), which could construct invalid instructions such as movup_from_offset(0) leading to a panic. Also, the bnot template produced movup_from_offset(0/1) during iteration for 128-bit values. These changes make the flow safe across all limb counts while keeping semantics unchanged.

### DIFF
--- a/codegen/masm/src/emit/mod.rs
+++ b/codegen/masm/src/emit/mod.rs
@@ -1814,6 +1814,72 @@ mod tests {
     }
 
     #[test]
+    fn op_emitter_u128_bnot_shadowing_fix_test() {
+        let mut block = Vec::default();
+        let mut stack = OperandStack::default();
+        let mut invoked = BTreeSet::default();
+        let mut emitter = OpEmitter::new(&mut invoked, &mut block, &mut stack);
+
+        // Push a U128-typed operand; value is irrelevant for emission semantics here
+        emitter.push(Type::U128);
+        assert_eq!(emitter.stack_len(), 1);
+
+        let span = SourceSpan::default();
+        emitter.bnot(span);
+
+        // Type preserved, no panic
+        assert_eq!(emitter.stack_len(), 1);
+        assert_eq!(emitter.stack()[0], Type::U128);
+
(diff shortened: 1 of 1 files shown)


bitwalker:
Thanks for the submission, but we do not merge PRs that are not associated with a GitHub issue that has been created or vetted by one of the compiler team members, due to the burden on the team with all the AI-generated spam we get.

Setting that aside, there are other issues with this PR:

1. The issue stated in the description is not accurate. In particular, calling the template function is always safe, because template functions are idempotent, and may be called with any number of iterations. It is the template function itself which is responsible for ensuring that if they use the current iteration count parameter, that their use of that value is valid. It is perfectly fine if a panic occurs due to invalid use of that parameter, and we actually _want_ a panic if that should occur, as it represents a bug elsewhere in the compiler.
2. The solution for the stated `template(0)` case doesn't actually prevent the template function being called with a value of `0`, because any non-zero number of iterations is going to result in calling it with `template(0)` on the first iteration.
3. The bug in the lowering of multi-limbed `bnot` isn't due to the use of `emit_template`, but rather that the iteration count parameter of the template function we're passing to `emit_template` is shadowing the actual `n` we're trying to reference (i.e. the one bound by the match arm corresponding to the number of limbs), so we should just rename the iteration count parameter to `_` here and ignore it. The point of the current implementation there is to move up each limb from back to front, inverting the bits while at the top. After doing `n` iterations of that, the limbs will be back in the desired order, but inverted.

If you would like to modify this PR to remove the changes to `emit_template`, and address point 3 above as I described, with a test to cover it, we'd gladly accept that though.

PR #728, opened 2025-10-22
Fix codegen for `cf.select`of 64bit values.

otrho:


tests/integration/src/codegen/operations.rs
```diff
@@ -0,0 +1,191 @@
...
+        let cond_val = block.borrow().arguments()[0] as ValueRef;
+
+        let a_imm = builder.imm(a, span);
+        let b_imm = builder.imm(b, span);
+
+        let result_val = builder.select(cond_val, a_imm, b_imm, span).unwrap();
+
+        builder.ret(Some(result_val), span).unwrap();
+    });
+
```
otrho, line 27: Took me all morning to work out why the executor wouldn't work.  See the `XXX` comment below.

I still don't know exactly why, but if there are no initialisers passed to `eval_package()` -- at least when the package was built with `compile_test_module()` -- then it'll fail with an error before even attempting to run the main block.

Adding this dummy initialiser 'fixed' the problem, digging deeper to find out why should be another issue.

greenhat: Yes, feel free to make an issue. I also hit another initializer bug - multiple initializers did not work. I left a TODO at https://github.com/0xMiden/compiler/blob/c81da7edfbe965a07595b295b4c87ce62199c5cf/tests/integration/src/rust_masm_tests/abi_transform/stdlib.rs?plain=1#L140-L144 but forgot to make an issue.

PR #732, opened 2025-10-27
feat: add `--language` flag to `cargo miden new`

lima-limon-inc:
Closes https://github.com/0xMiden/midenup/issues/109

Adds a `--language` flag to `cargo miden new` which determines the language the generated project directory will use.

Currently two variants exists:
- `rust` which persists `cargo miden`'s current behavior 
- `assembly` which clones this repository https://github.com/0xMiden/project-template, which was linked in this `midenup` issue (https://github.com/0xMiden/midenup/issues/109).

### DIFF
--- a/tools/cargo-miden/src/commands/new_project.rs
+++ b/tools/cargo-miden/src/commands/new_project.rs
@@ -7,11 +7,20 @@ use anyhow::Context;
 use cargo_generate::{GenerateArgs, TemplatePath};
 use clap::{Args, Parser};
 
+// ======================== Rust Template Repository ===========================
+/// Repository containing all the available rust templates.
+const RUST_TEMPLATE_REPO: &str = "https://github.com/0xMiden/rust-templates";
+
 /// The tag used in checkout of the new project template.
 ///
-/// Before changing it make sure the new tag exists in the rust-templates repo and points to the
-/// desired commit.
-const TEMPLATES_REPO_TAG: &str = "v0.19.0";
+/// Before changing it make sure the new tag exists in the
+/// [[RUST_TEMPLATE_REPO]] repo and points to the desired commit.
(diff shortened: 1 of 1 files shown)


tools/cargo-miden/src/commands/new_project.rs
```diff
@@ -183,18 +214,32 @@ impl NewCommand {
...
+                        ..Default::default()
+                    }
                 }
-            }
+                Language::MidenAssembly => {
+                    if let Some(template) = self.template {
+                        anyhow::bail!(
+                            "'--{template}' flag is not valid with an 'assembly' project."
+                        );
+                    }
```
lima-limon-inc, line 235: With this, if a user passes any flag from `RustProjectTemplate`, `cargo miden` will fail displaying the following message:

```
cargo miden new --account  --language assembly project 
'--account' flag is not valid with an 'assembly' project.
```

This check could be removed if [the assembly project repository](https://github.com/0xMiden/project-template) supported those templates as well.

tools/cargo-miden/src/commands/new_project.rs
```diff
@@ -183,18 +214,32 @@ impl NewCommand {
...
+                    TemplatePath {
+                        git: Some(ASSEMBLY_TEMPLATE_REPO.into()),
+                        branch: Some("ajl-add-counter-contract".into()),
```
greenhat, line 238: When https://github.com/0xMiden/project-template/pull/5 is merged the `ajl-add-counter-contract` branch will most likely be deleted and `cargo miden new --assembly` will stop working. We don't want this to happen.

Ideally, I'd like to see a stable branch name here (`main`, etc.). The last commit in the branch is from two month ago. @partylikeits1983 what do you think? When will https://github.com/0xMiden/project-template/pull/5 be merged?

PR #741, opened 2025-10-31
[2/x] feature: `cargo miden new` without options creates a Miden project

### LINKED ISSUE #737: Revamp the `cargo miden new` command to create a Miden project (workspace)
As discussed in https://github.com/0xMiden/compiler/discussions/646#discussioncomment-14814243 we want the `cargo miden new` to create a workspace structure consisting of multiple types of projects (contracts, Rust crates) mixed with miden client settings.

I suggest we make the `cargo miden new [NAME]` to create this workspace structure (Miden project) instead of creating an account Rust crate as we have now. And leave the individual contract project types (account, note script, etc.) under the `--[TYPE]` switch as we have today. For an account run `cargo miden new --account my-account`, for a note script run `cargo miden new --note p2id`, etc. 

/cc @Keinberger

greenhat:
Close #737 

**This PR is stacked on #739 and should be merged after it.**

This PR changes the behavior of the `cargo miden new` so without options it creates a Miden project (workspace) by taking `https://github.com/Keinberger/miden-project-environment`. The ability of create new contracts is preserved via options (`--account`, `--note`, etc.).

### DIFF
--- a/tools/cargo-miden/src/cli.rs
+++ b/tools/cargo-miden/src/cli.rs
@@ -19,7 +19,7 @@ pub struct CargoMidenCli {
 /// Subcommands supported by `cargo-miden`.
 #[derive(Debug, Subcommand)]
 pub enum CargoMidenCommand {
-    /// Create a new Miden project from a template.
+    /// Create a new Miden project (default) or a contract from a template (see Options).
     New(NewCommand),
     /// Compile the current crate to Miden package.
     Build(BuildCommand),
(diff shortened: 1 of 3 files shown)


bitwalker:
This feels backwards to me? Definitely very surprising behavior. IMO we should require an explicit project type argument and default to a Rust project. Just because we're going to point people to create MASM projects in the short term doesn't mean we should be setting up the tooling that way, it just does not make any sense to me for our Rust tooling to ever default to a non-Rust project type.

bitwalker (approved):

PR #765, opened 2025-11-11
Change the memory representation of 64-bit, dual-limbed values to be little-endian.

otrho:
In memory the low (least significant) 32-bit limb is now loaded or stored from/to `addr` and the high 32-bit limb from/to `addr+1`.  On the stack the limb order is still big-endian.

~For testing the `ToMidenRepr` array of felts are now little-endian.  The `FromMidenRepr` is still a big-endian array of felts, as per the stack.~

For testing now the values are placed in memory explicitly as little-endian and are limb-swapped upon return.  Changes should be made to the `miden-debug` crate to also support little-endian memory.

### DIFF
--- a/codegen/masm/intrinsics/mem.masm
+++ b/codegen/masm/intrinsics/mem.masm
@@ -229,7 +229,9 @@ end
 # to perform a sequence of shifts and masks to get the bits where they belong. This function
 # performs those steps, with the assumption that the caller has three values on the operand stack
 # representing any unaligned double-word value
-export.realign_dw # [chunk_hi, chunk_mid, chunk_lo, offset]
+#
+# Stack transition is [chunk_hi, chunk_mid, chunk_lo, offset] -> [shifted_lo, shifted_hi]
+export.realign_dw
     # We will refer to the parts of our desired double-word value
     # as two parts, `x_hi` and `x_lo`.
     # Re-align the high bits by shifting out the offset
@@ -283,9 +285,6 @@ export.realign_dw # [chunk_hi, chunk_mid, chunk_lo, offset]
 
     # OR the two halves together, giving us our second word, `x_lo`
     u32or # [x_lo, x_hi]
-
-    # Swap the words so they are in the correct order
-    swap.1 # [x_hi, x_lo]
 end
 
 # Shift two 32-bit words by the given offset.
@@ -330,13 +329,10 @@ export.load_dw # [addr, offset]
         swap.1 push.8 mul push.32 swap.1 sub swap.1        # [addr, bit_offset]
 
         # load the three elements containing the double-word on the stack and re-align
-        # NOTE: realign_dw expects the hi, mid, lo word inputs and returns hi, lo (i.e., everything
-        # is 64-bit big-endian limb oriented), so we must swap the output to be in the correct order
         dup.0 mem_load                                     # [e0, addr, bit_offset]
         dup.1 push.1 u32overflowing_add assertz mem_load   # [e1, e0, addr, bit_offset]
         movup.2 push.2 u32overflowing_add assertz mem_load # [e2, e1, e0, bit_offset]
         exec.realign_dw
-        swap.1
     end
 end
 


otrho:
OK, now after #681 the changes I've added to `ToMidenRepr` and `FromMidenRepr` need to go in the `miden-debug` crate.  I have no idea at this stage whether they would break things there, and/or if we should not use them for tests at all.

codegen/masm/intrinsics/mem.masm
```diff
@@ -317,22 +317,21 @@ export.load_dw # [addr, offset]
     # if the pointer is naturally aligned..
     if.true
         # drop offset
-        swap.1 drop            # [addr]
-        # load second element
-        dup.0 push.1 u32overflowing_add assertz # [addr + 1, addr]
-        mem_load               # [e1, addr]
-        # load first element
-        swap.1 mem_load        # [e0, e1]
+        swap.1 drop                              # [addr]
```
bitwalker: This appears to change the order of the bytes for _all_ double-word loads. I think we should constrain this behavior to only loads of `u64`/`i64`, since only that type gets special treatment in terms of its operand stack layout vs heap layout. We assume pretty pervasively that the layout of any type on the operand stack is in little-endian order, and that any special-case types are handled when we emit loads/stores of that type.

AIUI, the current behavior of `std::math::u64` is: big-endian limb order on the operand stack, little-endian order on the heap; but our backend assumes/assumes that the big-endian order holds for both storage locations, when that isn't true. I think the right fix here is to specifically address that bug, rather than change the underlying 'load_dw`/`store_dw` intrinsics.

Let me know if there is something I'm missing here, but I believe this PR would actually introduce a new bug, just one that we don't likely hit due to limited usage of these intrinsics currently.

bitwalker:
By the way, feel free to open PRs against `miden-debug` with any changes there, and I can get those released ASAP

codegen/masm/intrinsics/mem.masm
otrho: So codegen for  `hir.load : i64` should still emit an unchanged/big-endian `intrinsics::mem::load_dw` but then swap the limbs afterwards?

otrho: Hrmm, except at the moment the only place `load_dw` is used in the compiler is to lower 64-bit `hir.load`.

bitwalker: > So codegen for  hir.load : i64 should still emit an unchanged/big-endian intrinsics::mem::load_dw but then swap the limbs afterwards?

Yes, basically. Or we could have an intrinsic for loading `u64` values specifically, that loads them on to the operand stack in big-endian limb order - the main thing is that I would prefer `load_dw` and friends to maintain the little-endian byte ordering semantics that they've had up until now, as that is consistent with the rest of our memory model.

The idea behind `load_dw` (and `load_qw`, etc.) is to facilitate loading 4-byte words (so double-word here is an 8 byte load, and so on).

At the moment, we don't have loads of structs implemented in the backend, because our Rust compilation pipeline doesn't contain struct types (they've been converted to loads of some integral type). However, if one was to lower to HIR directly, we'd need to implement `load_struct`/`store_struct` in the backend using the `load_*` intrinsics as primitives. It's possible there are cases where we will need this even for Rust-compiled programs in the near future, but we've mostly been punting on this for now as it hasn't been needed.

In any case, the "correct" way for us to lay out `std::math::u64` would be to just use `load_dw` to get the limbs on the operand stack in little-endian order. The problem with just using `load_dw` though, as you know, is that Wasm wants the limbs on the operand stack in big-endian order, so we either need to reverse the limbs after `load_dw`, or implement a `u64`-specific pair of load/store intrinsics.

otrho: I still don't quite get what you mean here.

AFAICT all reads from memory have been little-endian _byte-wise_, and big-endian _limb-wise_, and any multi-limbed values have always been big-endian on the working stack, for Wasm and for MidenVM.

I've added a `load_unaligned_u64()` test to this PR now, which loads 64-bit words as little-endian byte-wise _and_ limb-wise, with the result big-endian on the stack (MSB on the top).

Is this not the behaviour we want?

The reason I even need this change is because Wasm is loading a struct with two 32bit members using a single 64bit read, alluding the the `load_struct` stuff you mention.

otrho:
Ping @bitwalker.

codegen/masm/intrinsics/mem.masm
```diff
@@ -326,16 +326,17 @@ export.load_dw # [addr, offset]
...
+        swap.1 push.8 mul push.32 swap.1 sub swap.1        # [addr, bit_offset]
+
+        # load the three elements containing the double-word on the stack and re-align
+        # NOTE: realign_dw expects the hi, mid, lo word inputs and returns hi, lo (i.e., everything
+        # is 64-bit big-endian limb oriented), so we must swap the output to be in the correct order
+        dup.0 mem_load                                     # [e0, addr, bit_offset]
+        dup.1 push.1 u32overflowing_add assertz mem_load   # [e1, e0, addr, bit_offset]
+        movup.2 push.2 u32overflowing_add assertz mem_load # [e2, e1, e0, bit_offset]
         exec.realign_dw
+        swap.1
```
bitwalker: We should modify `realign_dw` to remove the `swap.1` at the end of it (i.e. it is written to return the two 32-bit elements in big-endian order, when we don't need it to do that).

We could then remove the `swap.1` here, saving us some cycles.

bitwalker (approved):

PR #785, opened 2025-11-16
Fix source locations when trim-paths is being used

djolertrk:
Besides fixing the problem, it also covers:
  - Add lit test for it:
    `$ litcheck lit run --path bin tests/lit/source-location/`
  - Add HIR serialization in form of text


examples/basic-wallet/Cargo.toml
```diff
@@ -23,7 +23,9 @@ project-kind = "account"
 supported-types = ["RegularAccountUpdatableCode"]
 
 [profile.release]
+debug = true
 trim-paths = ["diagnostics", "object"]
 
 [profile.dev]
+debug = true
```
bitwalker: `debug = true` is the default for the `dev` profile, so no need to specify it here

examples/source-location-test/source_location_test.masp
bitwalker, line 1: We shouldn't commit these artifacts, as the format is still unstable

hir/src/ir/print.rs
```diff
@@ -280,6 +301,9 @@ impl PrettyPrint for OperationPrinter<'_> {
             doc + const_text(" ") + attrs
         };
 
+        // Add source location if available
```
bitwalker: I'd prefer to make this a configurable option, since it produces a lot of noise in the output otherwise

midenc-compile/src/stages/link.rs
```diff
@@ -1,16 +1,40 @@
...
+    Emit, Writer,
+    InputType, OutputMode, OutputType, ProjectType,
 };
 
 use super::*;
 
+/// A wrapper for emitting HIR as text
+struct HirText(String);
+
+impl Emit for HirText {
```
bitwalker: We should just implement `Emit` for `String` as one of the built-in impls of the trait. That way we don't need a wrapper like this just for outputting a string value.

midenc-compile/src/stages/link.rs
```diff
@@ -131,8 +155,16 @@ impl Stage for LinkStage {
             #[cfg(not(feature = "std"))]
             InputType::Real(_path) => unimplemented!(),
             InputType::Stdin { name, input } => {
+                // Populate trim_path_prefixes with the current directory
+                let trim_path_prefixes = {
```
bitwalker: I think this should be added as a compiler option, just like `current_dir`, so that `cargo-miden` can specify the full set of prefixes it knows about when invoking the compiler (including the location of the Cargo registry for dependency sources).

midenc-compile/src/stages/link.rs
```diff
@@ -181,8 +222,30 @@ fn parse_hir_from_wasm_file(
     let mut bytes = Vec::with_capacity(1024);
     file.read_to_end(&mut bytes).into_diagnostic()?;
     let file_name = path.file_stem().unwrap().to_str().unwrap().to_owned();
+
+    // Populate trim_path_prefixes with directories to try for resolving relative DWARF paths
+    // This allows resolving relative paths from trimmed DWARF debug info like "./src/lib.rs"
+    let trim_path_prefixes = {
```
bitwalker: Related to my above comment, this should be an explicit compiler option/handled by `cargo-miden`, rather than computing these paths here.

tests/lit/source-location/lit.suite.toml
```diff
@@ -0,0 +1,5 @@
+name = "source-location"
+patterns = ["*.wat", "*.wasm"]
```
bitwalker: We probably don't want the `.wasm` file type here, since it is a binary format, so it isn't suitable for lit tests.

bitwalker (changes requested):
Nice work @djolertrk! 

I have a few changes I'd like to make, that I've left comments for, but this looks good overall.

I think the `source-location-test` _probably_ shouldn't be an example, and instead defined as a regular test (see the integration test crate for examples of how we generate Cargo projects on the fly), but it's not a big deal for now - in general though, we try to reserve `examples` for specifically things that we want to use as documentation on how to do certain things with the compiler/SDK. Not a big deal though.

One thing to bear in mind too, is that we have generic IR printing infrastructure in place, i.e. the `Print` instrumentation in `hir/src/pass.rs`, that is enabled with a compiler option. This infra uses the `OperationPrinter` impls to actually do the printing, but what triggers the printing itself is the instrumentation pass. I'd prefer to only print IR using that - however the addition of the code to emit HIR if the HIR artifact type is requested is totally fine, just wanted to clarify the difference between the two types of output.

tests/lit/source-location/lit.suite.toml
djolertrk: I agree. :)

midenc-compile/src/stages/link.rs
djolertrk: Hmm, yes, works even better.

midenc-compile/src/stages/link.rs
djolertrk: Addressed, yes, thanks for the suggestion!

hir/src/ir/print.rs
djolertrk: Yep, added.

examples/source-location-test/source_location_test.masp
djolertrk: Ouch, yes....

djolertrk:
> Nice work @djolertrk!
> 
> I have a few changes I'd like to make, that I've left comments for, but this looks good overall.
> 
> I think the `source-location-test` _probably_ shouldn't be an example, and instead defined as a regular test (see the integration test crate for examples of how we generate Cargo projects on the fly), but it's not a big deal for now - in general though, we try to reserve `examples` for specifically things that we want to use as documentation on how to do certain things with the compiler/SDK. Not a big deal though.
> 

Hmm I see. I moved it to lit test directory, it is added as `tests/lit/source-location/test-project`, this may be a good enough for now.

> One thing to bear in mind too, is that we have generic IR printing infrastructure in place, i.e. the `Print` instrumentation in `hir/src/pass.rs`, that is enabled with a compiler option. This infra uses the `OperationPrinter` impls to actually do the printing, but what triggers the printing itself is the instrumentation pass. I'd prefer to only print IR using that - however the addition of the code to emit HIR if the HIR artifact type is requested is totally fine, just wanted to clarify the difference between the two types of output.

Got it, thanks :)

midenc-compile/src/compiler.rs
```diff
@@ -201,6 +201,29 @@ pub struct Compiler {
...
+    /// line, and column for each operation.
+    #[cfg_attr(
+        feature = "std",
+        arg(
+            long = "print-hir-source-locations",
+            default_value_t = false,
+            help_heading = "Output"
+        )
+    )]
+    pub print_hir_source_locations: bool,
```
bitwalker: Let's move this to the `UnstableOptions` struct, which contains other printer-specific compiler options that are sensitive to implementation details of the compiler

midenc-compile/src/compiler.rs
```diff
@@ -201,6 +201,29 @@ pub struct Compiler {
...
+    /// Specify path prefixes to try when resolving relative paths from DWARF debug info
+    #[cfg_attr(
+        feature = "std",
+        arg(
+            long = "trim-path-prefix",
+            value_name = "PATH",
+            help_heading = "Compiler"
+        )
+    )]
+    pub trim_path_prefixes: Vec<PathBuf>,
```
bitwalker: Let's move this to `UnstableOptions` as well, since this is also an internal implementation detail that is pretty specific to the Rust compilation pipeline and may change at any time depending on how Rust/LLVM decides to emit DWARF/debug info.

tests/lit/source-location/test.wat
```diff
@@ -0,0 +1,23 @@
+;; RUN: cargo build --release --target wasm32-unknown-unknown --manifest-path %S/test-project/Cargo.toml 2>&1
+;; RUN: env MIDENC_TRACE=debug bin/midenc compile %S/test-project/target/wasm32-unknown-unknown/release/source_location_test.wasm --entrypoint=source_location_test::test_assertion --trim-path-prefix=%S/test-project --emit=masm=- 2>&1 | filecheck %s
```
bitwalker: With the compiler options moved under `UnstableOptions`, just remember you'll need to pass them using the `-Z` prefix, e.g. `-Zprint-hir-source-locations`, etc.

PR #798, opened 2025-11-26
[1/x] Pass an account as a parameter to note and tx script

### LINKED ISSUE #787: Pass an account components as a parameter to the note script
As discussed in https://github.com/0xMiden/compiler/issues/698#issuecomment-3387066393 we need to find a way to "reconstruct" the account components interfaces (traits?) as they were defined in their crates, represent them as one entity (struct?) and pass it into the note script as a parameter.

greenhat:
Close #787 

This PR refactors the `generate!` macro to call `wit-bindgen` directly instead of using its `generate!` macro, and introduces an `Account` wrapper struct that aggregates methods from imported interfaces.

### `generate!` macro refactoring
- Replaced the `wit_bindgen::generate!` macro invocation with direct calls to `wit-bindgen-core` and `wit-bindgen-rust` libraries
- This enables post-processing of generated bindings before emitting them

### `Account` wrapper struct
- The `generate!` macro now automatically generates an `Account` struct in the `bindings` module
- This struct aggregates all public functions from imported WIT interfaces as methods
- Methods delegate to the original free functions in leaf modules

### Script macros (`#[note_script]`, `#[tx_script]`)
- Now accept an optional second parameter of type `Account`
- The parameter is instantiated via `Default::default()` and passed to the user's `run` function
- Example: `fn run(arg: Word, account: Account) { ... }`

### Misc
- Added collision detection for method names across different interfaces with clear error messages
- Extracted manifest path resolution into a new `manifest_paths` module

## Example

Before:
```rust
use crate::bindings::miden::basic_wallet::basic_wallet::receive_asset;

#[note_script]
fn run(_arg: Word) {
    receive_asset(asset);
}
```

After:
```rust
use crate::bindings::Account;

#[note_script]
fn run(_arg: Word, account: Account) {
    account.receive_asset(asset);
}
```

### DIFF
--- a/sdk/base-macros/src/generate.rs
+++ b/sdk/base-macros/src/generate.rs
@@ -1,4 +1,4 @@
-use std::{env, fs, path::PathBuf};
+use std::{collections::HashMap, env, fs, path::PathBuf};
 
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{quote, ToTokens};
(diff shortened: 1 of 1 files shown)


bobbinth:
> I decided to forgo asking users about the account components. My line of thinking was that the note script can have only account components as dependencies. So, whatever account components generated bindings (free functions in modules) are, we will gather them all as methods in the `Account` struct. So for the developer the `Account` here is whatever they list in the note script dependencies fused under one struct.

Makes sense! I think this approach works - though, we should document this somewhere very clearly (otherwise it might feel like there is too much magic happening).

examples/basic-wallet-tx-script/src/lib.rs
```diff
@@ -23,7 +23,7 @@ const ASSET_START: usize = 8;
 const ASSET_END: usize = 12;
 
 #[tx_script]
-fn run(arg: Word) {
+fn run(arg: Word, account: Account) {
```
bitwalker, line 26: Is there a good reason to pass `account` by value, rather than by reference (either immutable or mutable, depending on the needs of the script)?

One reason why I think that might be important, is that it would provide us enough information to be able to determine how a given script/note interacts with an account, as well as allow Rust to enforce at compile-time that you cannot call certain methods on the account because you don't have an appropriate reference type.

The bigger question I guess, is whether there is any value in non-mutable references as argument to a transaction script/note script. In theory, that would enforce read-only access - but I'm not clear on whether read-only scripts would be useful; and even if we had them, how other parts of our tooling would leverage that (i.e. depending on what type of reference your script entrypoint uses (e.g. `&` vs `&mut`), we'd encode in the resulting package metadata that the script is read-only or can mutate an account. That could make it easier to notify users that they are about to perform an action with mutable access to an account - but I don't know the exact mechanism by which that would work.

Anyhow, by passing the `account` parameter by-value, we don't even have the option of using reference mutability to encode that information, so it might be worth getting ahead of it a bit, by always passing a mutable reference for now (or if it's possible, we could look at the reference type of the method, and then have the proc macro machinery emit code to pass a reference of the appropriate type).

What do you think?

bitwalker (approved):
I'm marking this approved for now, but I do think we might want to revisit how the account is passed to script entrypoints to use a reference instead. See my other comment for more details on my thought process there.

examples/basic-wallet-tx-script/src/lib.rs
greenhat: Sorry, I addressed this in https://github.com/0xMiden/compiler/pull/801 but forgot to mention it in this PR.
Also, I'm linking here for account methods `self` mutability issue: https://github.com/0xMiden/compiler/issues/802.

PR #801, opened 2025-12-01
[2/x] `ActiveAccount` and `NativeAccount` traits to call tx kernel functions via `self.*` on an account

### LINKED ISSUE #733: Gather `account` tx kernel function under the `Account` trait
Extracted from https://github.com/0xMiden/compiler/pull/722#issuecomment-3435278566

Define the `Account` trait in the SDK:

```rust
trait Account {
    fn add_asset(&mut self, a: Asset) {
        account::add_asset(a);
    }
}
```

Make `#[component]` macro to add `impl Account for USER_TYPE {}` to be able to call `self.add_asset(asset);` in any `USER_TYPE` method.

### LINKED ISSUE #788: Require `&mut self` for the methods that mutate the account state (storage, assets)
In the account interface (see https://github.com/0xMiden/compiler/blob/bc0e1487e02a139ae279c1d6198084b7345fcb25/examples/basic-wallet/src/lib.rs?plain=1 for an example) the `&mut self` should be required for the methods that mutate the account state.

greenhat:
Close #733 
Close #788 

**This PR is stacked on #798 and should be merged after it.**

This PR:
- Adds `NativeAccount` and `ActiveAccount` traits wrapping functions from the corresponding tx kernel library modules;
- `#[component]` macro implements these traits for the user account type;
- `#[note_script]` macro implements `ActiveAccount` for the generated `Account` struct introduced in #798

### DIFF
--- a/sdk/base-macros/src/generate.rs
+++ b/sdk/base-macros/src/generate.rs
@@ -254,8 +254,10 @@ fn augment_generated_bindings(tokens: TokenStream2) -> syn::Result<TokenStream2>
     if !collected_methods.is_empty() {
         let struct_ident = syn::Ident::new(WRAPPER_STRUCT_NAME, Span::call_site());
         let struct_item: ItemStruct = parse_quote! {
-            /// Wrapper struct providing methods that delegate to imported interface functions.
-            #[derive(Clone, Copy, Default)]
+            /// Wrapper struct that contains all the methods from all the account component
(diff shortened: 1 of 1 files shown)


bitwalker (approved):
It looks like this implements passing accounts by reference (see my feedback from #798), which is awesome!

I mostly skimmed through this, but nothing caught my eye as an issue. Definitely like how this is shaping up! Nice work!

Issue #698, opened 2025-10-01
Typed note inputs DX


greenhat:
With the introduction of the `#[note_script]` attribute macro and the P2ID note script looks like:

https://github.com/0xMiden/compiler/blob/41926fa0c3c99aa8bb7efa9b649de79baa069a66/examples/p2id-note/src/lib.rs?plain=1#L25-L37

The next step up is the typed note inputs. Instead of `Vec<Felt>` we could return the note inputs as `T` given the deserialization from `Vec<Felt>` is implemented for `T`.

```rust
#[derive(Deserialize)]
struct MyNote {
    account_id: AccountId,
}


#[note_script]
fn run(_arg: Word) {
    let my_note = note::get_inputs_as::<MyNote>();
    ... or 
    let my_note: MyNote = note::get_inputs_as();
    ...
    // access my_note.account_id
    ...
}

```

Where `AccountId` implements the deserialization from `Vec<Felt>`. We could implement the deserialization for the Miden SDK types, but for custom types the deserialization is expected to be implemented by the user.

@bitwalker @otrho @bobbinth

bobbinth:
Thank you @greenhat! I think this will simplify working with notes quite a bit.

As the end goal (not for next iteration but maybe for one after that), I wonder if we can hide even more boilerplate behind proc macros. It would be awesome if we could get to something that looks like this:

```rust
#[note]
struct MyNote {
    recipient: AccountId,
}

#[note]
impl MyNote {

    #[entrypoint]
    pub fn run(&self, _args: Word) {
        let account_id = account::get_id(); 
        assert_eq!(self.recipient, account.id());

        ...
    }   
}
```
And in an even longer run, to something like this:
```rust
#[note]
struct MyNote {
    recipient: AccountId,
}

#[note]
impl MyNote {

    #[entrypoint]
    pub fn run(&self, account: &BasicWallet, _args: Word) {
        assert_eq!(self.recipient, account.id());

        ...
    }   
}
```
Not sure though if there are some practical limitations of proc macros which would make this difficult (or maybe impossible).

Also, cc @mmagician, @Dominik1999, and @partylikeits1983 for visibility.

bobbinth:
[...]

PR #808, opened 2025-12-05
Typed note inputs

greenhat:
Ref #698 

### Next steps:

- Create empty entities on the crates.io to publish the new serialization crates in the future;

- Implement off-chain serialization (via macros) for the types in `miden-base` that have counterparts in the on-chain Miden SDK (`AccountId`, etc.);

### Implementation details

This PR introduces crates for on-chain and off-chain serialization. 

I chose the trait names `ToFeltRepr` and `FromFeltRepr` to highlight that "felt representation" differs from the generic serialization/deserialization in a way that optimizes the on-chain deserialization code to be of minimal complexity.

The reason for two crates instead of one is different `Felt` types. For off-chain, it wraps `u64` and for on-chain, it wraps `float`.

The on-chain deserialization trait is at https://github.com/0xMiden/compiler/blob/0eb8e57ec8da335c2010c27df26099ab0c2d675e/sdk/felt-repr/onchain/src/lib.rs?plain=1#L64-L68
The off-chain serialization trait is at https://github.com/0xMiden/compiler/blob/0eb8e57ec8da335c2010c27df26099ab0c2d675e/sdk/felt-repr/offchain/src/lib.rs?plain=1#L137-L149

Although the off-chain and on-chain traits look identical the only reason for them to be in separate crates is different underlying `Felt` type in off-chain and on-chain code (see above). To have it defined only once the schema for user-defined types is discussed at https://github.com/0xMiden/compiler/issues/814.

The usage in the P2ID is at https://github.com/0xMiden/compiler/blob/0eb8e57ec8da335c2010c27df26099ab0c2d675e/examples/p2id-note/src/lib.rs?plain=1#L12 with the `AccountId` implementation derived at https://github.com/0xMiden/compiler/blob/0eb8e57ec8da335c2010c27df26099ab0c2d675e/sdk/base-sys/src/bindings/types.rs?plain=1#L9-L13

The off-chain code is at https://github.com/0xMiden/compiler/blob/0eb8e57ec8da335c2010c27df26099ab0c2d675e/tests/integration-node/src/node_tests/basic_wallet.rs?plain=1#L86 with the implementation derived temporarily through proxy (`AccountIdFeltRepr`) later the `miden-base::AccountId` will have it derived as well.

This PR implements serialization and deserialization parts only to keep the PR size reasonable. The rest of #698 (DX, macros) will be addressed in the subsequent PR.

### DIFF
--- a/tools/cargo-miden/src/commands/example_project.rs
+++ b/tools/cargo-miden/src/commands/example_project.rs
@@ -12,12 +12,21 @@ use crate::{
 /// Paired project mappings for examples that create multiple related projects
 const PAIRED_PROJECTS: &[(&str, &str)] = &[("counter-contract", "counter-note")];
 
+/// Triple project mappings for examples that create three related projects
+/// Each tuple contains (tx-script, account, note) project names
+/// Any of these names can be used to create all three projects
+const TRIPLE_PROJECTS: &[(&str, &str, &str)] =
+    &[("basic-wallet-tx-script", "basic-wallet", "p2id-note")];
+
 /// Create a new Miden example project
 #[derive(Debug, Args)]
 #[clap(disable_version_flag = true)]
 pub struct ExampleCommand {
     #[clap(help = r#"The example name to use from the compiler repository
 Available examples:
+basic-wallet          : Basic wallet account
+p2id-note             : Pay-to-ID note
+basic-wallet-tx-script: Transaction script used in basic-wallet and p2id-note
 counter-contract      : Counter contract
 counter-note          : Counter note
 fibonacci             : Fibonacci sequence calculator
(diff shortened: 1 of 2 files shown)


bitwalker:
@greenhat FYI, I appended the fix for #827 to this PR

bitwalker (approved):
LGTM!

I suspect we'll want something like the `miden` crate to host the offchain plumbing, from which it can be re-exported (like we do the onchain plumbing from `miden`), but I don't know what that would be just yet. I think in a more ideal world, the `miden` crate could be used for both, but conditional compilation would determine which set of APIs you get, since that's the crate people would intuitively reach for - but there's still some awkwardness there. Anyway, something we should keep trying to improve on - we have a lot of crates, and the number keeps growing, and that's going to impede discoverability if we don't get ahead of it a bit.

PR #812, opened 2025-12-09
Add synthetic source locations

djolertrk:
This adds sveral features/fixes for improved debugging:
    
1. Synthetic source locations: Compiler-generated code (like local
    initialization, ABI transforms) now uses SourceSpan::SYNTHETIC instead
    of UNKNOWN. This differentiates compiler-generated code from truly
    unknown locations and renders as #loc(synthetic) in output.
    
2. -Z print-masm-source-locations: New flag to print source locations
     as #loc("file":line:col) annotations in MASM output.
    
3. -Z remap-path-prefix=FROM=TO: Remap source path prefixes for resolving
     paths from DWARF when sources are in different locations (e.g., stdlib).
    
4. Use current_span to have more real source locations attached --
     currently we used Default locations a lot - for example, each
     operator would start with it.
    
5. Fix wasm component-level DWARF parsing
    
Requires miden-vm with SourceSpan::SYNTHETIC debug type.
    
    
To check locations on masm level, use:

```
$ ./bin/midenc examples/assert-debug-test/target/wasm32-unknown-unknown/release/assert_debug_test.wasm \
    --entrypoint=assert_debug_test::test_assert \
    -Ztrim-path-prefix=examples/assert-debug-test \
    -Zprint-masm-source-locations \
    --debug full --emit=masm=-
```

### DIFF
--- a/frontend/wasm/src/component/parser.rs
+++ b/frontend/wasm/src/component/parser.rs
@@ -365,7 +365,6 @@ impl<'a, 'data> ComponentParser<'a, 'data> {
         // but reference the embedded module's code. We need to:
         // 1. Copy the DWARF data to the module's debuginfo
         // 2. Store the module's base offset for address translation
-        // TODO: Add test for this!!
         if let Some(first_module) = self.static_modules.values_mut().next() {
             // Only inject if DWARF was actually parsed
             if !self.component_debuginfo.dwarf.debug_info.reader().is_empty() {
@@ -938,7 +937,6 @@ impl<'a, 'data> ComponentParser<'a, 'data> {
     /// Parses a DWARF debug section from the component.
     /// These sections are stored at the component level but contain debug info
     /// for the embedded modules.
-    /// TODO: Add tests for this!!!
     fn dwarf_section(&mut self, section: &wasmparser::CustomSectionReader<'data>) {
         let name = section.name();
         if !self.config.generate_native_debuginfo && !self.config.parse_wasm_debuginfo {
@@ -985,6 +983,45 @@ impl<'a, 'data> ComponentParser<'a, 'data> {
 
         dwarf.ranges = gimli::RangeLists::new(info.debug_ranges, info.debug_rnglists);
         dwarf.locations = gimli::LocationLists::new(info.debug_loc, info.debug_loclists);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use midenc_hir::Context;
+    use wasmparser::Validator;
+
+    use super::*;
+    use crate::supported_component_model_features;
+
+    #[test]
+    fn injects_component_level_dwarf_into_first_module() {
+        let component = wat::parse_str(
+            r#"
+            (component
+                (@custom ".debug_info" "\01\02\03")
+                (core module
+                    (func (export "f"))
+                )
+            )
+            "#,
+        )
+        .expect("component wat should compile");
+        let context = Context::default();
+        let config = WasmTranslationConfig::default();
+        let mut validator = Validator::new_with_features(supported_component_model_features());
+        let mut types = ComponentTypesBuilder::default();
+        let parser = ComponentParser::new(&config, context.session(), &mut validator, &mut types);
+
+        let parsed = parser.parse(&component).expect("component should parse");
+        let first_module = parsed
+            .static_modules
+            .values()
+            .next()
+            .expect("component should contain a core module");
+
+        assert_eq!(first_module.debuginfo.dwarf.debug_info.reader().len(), 3);
+        assert!(first_module.wasm_file.module_base_offset > 0);
     }
 }
 
--- a/frontend/wasm/src/component/parser.rs
+++ b/frontend/wasm/src/component/parser.rs
@@ -3,9 +3,11 @@
 
 // Based on wasmtime v16.0 Wasm component translation
 
+use alloc::sync::Arc;
 use std::mem;
 
 use cranelift_entity::PrimaryMap;
+use gimli::Section;
 use indexmap::IndexMap;
 use midenc_hir::{FxBuildHasher, FxHashMap};
 use midenc_session::{Session, diagnostics::IntoDiagnostic};
@@ -22,7 +24,7 @@ use crate::{
     component::*,
     error::WasmResult,
     module::{
-        module_env::{ModuleEnvironment, ParsedModule},
+        module_env::{DebugInfoData, Dwarf, ModuleEnvironment, ParsedModule},
         types::{
             EntityIndex, FuncIndex, GlobalIndex, MemoryIndex, TableIndex, WasmType,
             convert_func_type, convert_valtype,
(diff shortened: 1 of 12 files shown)


djolertrk:
NOTE: This depends on the new version miden-vm (synthetic location debug type).

bitwalker:
@djolertrk What should we do with this PR? There's been several changes since this was originally opened - not only do we have `SourceSpan::SYNTHETIC`, but we also now have `Location` as an attribute, which can represent a wider variety of location types, though we do not currently leverage this really.

We probably should bring this up to date with `next` so we can take advantage of the ability to track down missing locations.

djolertrk:
> @djolertrk What should we do with this PR? There's been several changes since this was originally opened - not only do we have `SourceSpan::SYNTHETIC`, but we also now have `Location` as an attribute, which can represent a wider variety of location types, though we do not currently leverage this really.
> 
> We probably should bring this up to date with `next` so we can take advantage of the ability to track down missing locations.

Yes! I will work on brining this on top of `next`.

frontend/wasm/src/module/func_translator.rs
```diff
@@ -298,10 +316,9 @@ fn parse_function_body<B: ?Sized + Builder>(
                     let column = loc.column.and_then(ColumnNumber::new).unwrap_or_default();
                     span = source_file
                         .line_column_to_span(line, column)
-                        .unwrap_or(SourceSpan::UNKNOWN);
-                    if !span.is_unknown() {
-                        last_valid_span = span;
-                    }
+                        .unwrap_or_else(synthetic_span);
```
bitwalker: Isn't this a case where `UNKNOWN` makes more sense? 

To me, a synthetic location is used to indicate that some IR is purely compiler-generated (i.e. doesn't have a direct correlation to something in the original source code). For example, we generate code at the start of a program to initialize the heap - this doesn't have any source provenance at all, so `SYNTHETIC` is the right choice for that - we "know" the location, but there is no actual source code to provide a span for.

Unknown locations on the other hand, indicate that we have some IR that _should_ have a source location, because it is not compiler-generated - but we don't know it, or it is unavailable.

Things are a bit murky with compiling from Wasm of course, since DWARF doesn't distinguish between "unknown" and "synthetic" locations - you either have a source location or you don't. That said, for our purposes, I _think_ it makes sense to use `UNKNOWN` in this case - since the goal is to catch when `midenc` itself is dropping source location info, or improperly marking something as unknown when it is synthetic - but we still want to know when we are legitimately missing source location info.

frontend/wasm/src/module/func_translator.rs
```diff
@@ -231,21 +241,30 @@ fn parse_function_body<B: ?Sized + Builder>(
     debug_assert_eq!(state.control_stack.len(), 1, "State not initialized");
 
     let func_name = builder.name();
-    let mut end_span = SourceSpan::UNKNOWN;
-    // Track the last valid span to use as a fallback for instructions without DWARF debug info.
-    // This is particularly useful for `unreachable` instructions that follow panic calls,
-    // where the unreachable itself has no source location but should inherit the panic call's span.
-    let mut last_valid_span = SourceSpan::UNKNOWN;
+    let mut end_span = synthetic_span();
```
bitwalker: Same note here - I think we want to know when source location info is actually unknown. Synthetic locations are semantically "known", so they aren't appropriate for things like representing missing locations.

frontend/wasm/src/translation_utils.rs
```diff
@@ -12,6 +12,16 @@ use crate::{
     error::WasmResult, module::function_builder_ext::FunctionBuilderExt, unsupported_diag,
 };
 
+/// Returns a synthetic SourceSpan for compiler-generated code.
+///
+/// This uses SourceSpan::SYNTHETIC from miden-debug-types which is identified
+/// by having an unknown source_id and both start and end set to u32::MAX.
+/// This differentiates it from UNKNOWN spans (which have start and end at 0)
+/// and indicates the code doesn't correspond to any specific user source location.
+fn synthetic_span() -> SourceSpan {
```
bitwalker: This is a duplicate of the same code in `func_translator.rs` - let's just remove both of these functions and use `SourceSpan::SYNTHETIC` directly, it doesn't make sense to turn a constant into a function call IMO

frontend/wasm/src/component/parser.rs
```diff
@@ -348,6 +360,23 @@ impl<'a, 'data> ComponentParser<'a, 'data> {
         assert!(remaining.is_empty());
         assert!(self.lexical_scopes.is_empty());
 
+        // Inject component-level DWARF debug info into the first module.
+        // In wasm components, DWARF sections are stored at the component level
+        // but reference the embedded module's code. We need to:
+        // 1. Copy the DWARF data to the module's debuginfo
+        // 2. Store the module's base offset for address translation
+        // TODO: Add test for this!!
```
bitwalker: Could we add the mentioned test as part of this PR? TODOs like this often go unaddressed, and are hard to keep track of - I suspect the test would never end up getting added

bitwalker (changes requested):
Thanks! Overall looks good, just a couple of nits/tweaks, and we can get this merged

frontend/wasm/src/component/parser.rs
djolertrk: Sure. I agree...

frontend/wasm/src/module/func_translator.rs
djolertrk: Hmm, yeah, it is `UNKNOWN` actually.

bitwalker:
@djolertrk Once you've resolved the conflicts with `next`, I can get this merged. There are some changes that impact how source locations are handled, so you'll want to carefully review conflicts related to that part of the code to ensure there aren't regressions. The main thing place where there may be conflicts here, are the changes I made to properly handle `--remap-path-prefix` so that source locations are remapped as expected. We want to load sources using absolute paths, but we don't want to encode absolute paths in debug info unless required. This is now handled correctly in `next`

djolertrk:
> @djolertrk Once you've resolved the conflicts with `next`, I can get this merged. There are some changes that impact how source locations are handled, so you'll want to carefully review conflicts related to that part of the code to ensure there aren't regressions. The main thing place where there may be conflicts here, are the changes I made to properly handle `--remap-path-prefix` so that source locations are remapped as expected. We want to load sources using absolute paths, but we don't want to encode absolute paths in debug info unless required. This is now handled correctly in `next`

It makes sense. I needed to open a new PR in debugger, so it resolves relative paths properly: https://github.com/0xMiden/miden-debug/pull/75

tests/lit/debug/inline-stubs.shtest
```diff
@@ -1,7 +1,7 @@
-;; RUN: midenc %s --entrypoint=test_felt_add --emit=hir=- -Canalyze-only 2>&1 | filecheck %s --check-prefix=HIR-ADD
-;; RUN: midenc %s --entrypoint=test_felt_assert_eq --emit=hir=- -Canalyze-only 2>&1 | filecheck %s --check-prefix=HIR-ASSERT
-;; RUN: midenc %s --entrypoint=test_felt_arithmetic --emit=hir=- -Canalyze-only 2>&1 | filecheck %s --check-prefix=HIR-ARITH
-;; RUN: midenc %s --entrypoint=inline_stubs_test::test_felt_add -Zprint-hir-source-locations --emit=hir=- -Canalyze-only 2>&1 | filecheck %s --check-prefix=HIR-LOC
+;; RUN: midenc - --entrypoint=test_felt_add --emit=hir=- -Canalyze-only < %s 2>&1 | filecheck %s --check-prefix=HIR-ADD
```
bitwalker: What's the rationale for using `stdin` here vs the usual path-based input? Is this test intending to exercise the `stdin` input path specifically?

djolertrk: it was a hack for the .wat file, but I added "*.wat" extension as accepted files, so it is being picked up properly by lit. now it does use `%s`

bitwalker (approved):

PR #817, opened 2025-12-11
feat: custom test harness

lima-limon-inc:
Note: this PR is mainly intended to be an exploration on what a custom test harness for miden could look like. Comments/feeback more than welcome!

This PR implements a custom test harness for miden rust code. The custom harness itself is comprised of two crates:
- `miden-test-harness-macros`: Crate containing the `#[miden_test]` and `#[miden_test_block]` macros; which serve as miden-test specific equivalents for `#[test]` and `mod tests` respectively.
-  `miden-test-harness-lib`: Library which general utilities and a small wrapper around `libtest_mimic`; it also re-exports the macros crate. This is intended to be the test harness' entry point.  

A `README.md` file is included in `miden-test-harness-lib` explaining the main features of the testing harness. 

A new test project was created in `tests/examples/counter` to showcase the harness (the code itself was taken from `examples/counter-contract`).

Additionally, a new subcommand was added to `cargo miden` called `test` (i.e. `cargo miden test`), which serves as a wrapper around `cargo test`. The rationale behind making `cargo miden test` a 1-1 copy of `cargo test` can be found in in these comments (https://github.com/0xMiden/compiler/pull/817#issuecomment-3653597639, https://github.com/0xMiden/compiler/pull/817#discussion_r2656128146, https://github.com/0xMiden/compiler/pull/817#issuecomment-3762180898). TLDR: If we want the harness to work with pre-existing tools (most notably IDE's that support single test execution), we need the harness to work as similarly as `cargo test` as possible.

Limitation/Places to go:
- I'm not too fond of having to add a [wrapping `mod`](https://github.com/0xMiden/compiler/pull/817/changes#diff-cfdb1ddcf929a576af15821b3dbd98a1684ffbb403b0b424fd8f2cbb6c6cae68R13) around the non-test rust code. This was added in order to omit compilation of the contract code during `cargo test`'s execution.
- Finally, since the `MockChainBuilder` is so versatile and has a lot of customization options, I struggled to automatically generate a `MockChain` and thus the test ended up [having a lot of boilerplate](https://github.com/0xMiden/compiler/pull/817/changes#diff-342d79ac2d1e87a3b978c3f32d94e286a5b26a1353639d1bfb0af75fb03efdcbR96-R115). Maybe defining sane defaults would aid in passing in a pre-built `MockChain` and thus skipping this boilerplate. 

Would love to hear your ideas!

### DIFF
--- a/tools/cargo-miden/src/commands/test.rs
+++ b/tools/cargo-miden/src/commands/test.rs
@@ -4,14 +4,12 @@ use clap::Args;
 
 /// Command-line arguments accepted by `cargo miden build`.
 ///
-/// All arguments following `build` are parsed by the `midenc` compiler's argument parser.
-/// Cargo-specific options (`--release`, `--manifest-path`, `--workspace`, `--package`)
-/// are recognized and forwarded to the underlying `cargo build` invocation.
-/// All other options are passed to `midenc` for compilation.
+/// This command is a thin wrapper around `cargo test`, forwarding all arguments
+/// to the underlying test invocation.
 #[derive(Clone, Debug, Args)]
 #[command(disable_version_flag = true, trailing_var_arg = true)]
 pub struct TestCommand {
-    /// Arguments parsed by midenc (includes cargo-compatible options).
+    /// Arguments forwarded to `cargo test`.
     #[arg(value_name = "ARG", allow_hyphen_values = true)]
     pub args: Vec<String>,
 }
@@ -30,8 +28,10 @@ impl TestCommand {
 fn test_cargo_args(cli_args: Vec<String>) -> Vec<String> {
     let mut args = vec!["test".to_string()];
 
-    args.extend(["--".into()]);
-    args.extend(cli_args);
+    if !cli_args.is_empty() {
+        args.extend(["--".into()]);
+        args.extend(cli_args);
+    }
 
     args
 }
@@ -46,7 +46,11 @@ fn run_cargo_test(spawn_args: &[String]) -> anyhow::Result<()> {
 
     cargo.args(spawn_args);
 
-    let _artifacts = crate::utils::spawn_cargo(cargo, &cargo_path)?;
+    let status = cargo.status()?;
+
+    if !status.success() {
+        std::process::exit(status.code().unwrap_or(1));
+    }
 
     Ok(())
 }


greenhat:
Also, https://github.com/rust-lang/rust-analyzer/pull/16662 test explorer is coming. Which we'd want to support as well.

test-harness/test-harness-lib/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "miden-test-harness-lib"
```
bitwalker: Let's call this crate just `miden-test-harness`

test-harness/test-harness-macros/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+    let mut input_module = parse_macro_input!(item as syn::ItemMod);
+
+    // We add an internal "use" here in order for the tests inside the `mod tests`
+    // block to use the `miden_test` macro without needing to pass the full path.
+    let internal_use = syn::parse_quote! {
+        use miden_test_harness_macros::miden_test;
+    };
+    input_module.content.as_mut().unwrap().1.insert(0, internal_use);
+
+    let module = if is_test() {
```
bitwalker: Why not just emit the module annotated with `#[cfg(test)]`? That should accomplish the same goal, no? It isn't 100% clear why this would need to be facilitated via feature flags on the macro crate (which when you take into account Cargo's feature flag unification approach, could lead to test modules being emitted even when they aren't expected - not likely, but one downside of using features for this sort of thing).

test-harness/test-harness-macros/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+    } else {
+        quote! {}
+    };
+
+    let main_function = if is_test() {
+        quote! {
+            use miden_test_harness_lib;
+
+            fn main() {
+                let args = miden_test_harness_lib::MidenTestArguments::from_args();
```
bitwalker: I'd always qualify paths absolutely, e.g. `::miden_test_harness_lib::MidenTestArguments::from_args()`, this ensures that the path resolves from the crate root, ignoring any imports in the current module. Not a super big problem in this case, but still good practice for proc macros like this.

test-harness/test-harness-macros/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+        process_arguments::<Package>(function, 1).unwrap_or_else(|err| panic!("{err}"));
+
+    let Some(package_binding_name) = found_packages_vars.first() else {
+        // If there are no variables with `Package` as its type, then don't load
+        // the `Package`.
+        return;
+    };
+
+    // This env var is set by `cargo miden test`.
+    let package_path = std::env::var("CREATED_PACKAGE").unwrap();
```
bitwalker: We should namespace all env vars that we use, e.g. `CARGO_MIDEN_CREATED_PACKAGE` (though I think it would probably be better to call it `CARGO_MIDEN_TEST_PACKAGE_PATH` to make it clearer what defines the variable, and what the value is expected to be.

test-harness/test-harness-macros/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        #[cfg(test)]
+        #input_fn
+    };
+
+    TokenStream::from(function)
+}
+
+#[proc_macro_attribute]
+pub fn miden_test_block(
```
bitwalker: Let's name this attribute `#[miden_test_suite]`, since that is effectively what it is used to define.

test-harness/test-harness-macros/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+        quote! {
+            #input_module
+        }
+    } else {
+        quote! {}
+    };
+
+    let main_function = if is_test() {
+        quote! {
+            use miden_test_harness_lib;
```
bitwalker: This line shouldn't be needed, correct?

tools/cargo-miden/src/commands/test.rs
```diff
@@ -0,0 +1,79 @@
...
+            output: BuildOutput::Masm { artifact_path },
+        }) = build.exec(OutputType::Masm)?
+        else {
+            // This should never happend since we are hardcoding the output as MASM.
+            bail!("cargo miden test requires projects to be compiled to a masm artifact.")
+        };
+
+        let spawn_args = test_cargo_args();
+        unsafe {
+            std::env::set_var("CREATED_PACKAGE", artifact_path);
```
bitwalker: See my other comment about the naming of this env variable

test-harness/test-harness-lib/src/utils.rs
bitwalker, line 1: This module appears to be unused

test-harness/test-harness-lib/src/reexport.rs
bitwalker, line 1: I'd move the contents of this module up to the root `mod.rs`, and skip the renaming of `miden_mast_package`. For the `Deserializable` trait, I'd just re-export `miden_objects`, and then import `miden_objects::utils::Deserializable` where needed, since we may need other things re-exported from that crate later.

test-harness/test-harness-lib/src/libtest.rs
```diff
@@ -0,0 +1,70 @@
+use alloc::vec::Vec;
+
+// ============================= Test function ================================
+
+/// Struct that represents a function marked with #[miden-test].
```
bitwalker: I think you meant `#[miden_test]` here

test-harness/test-harness-macros/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[lib]
+proc-macro = true
+
+[features]
+test-flag = []
+
+[dependencies]
+# Miden dependencies
+miden-test-harness-lib.workspace = true
```
bitwalker: This crate shouldn't depend on `miden-test-harness`, instead it should depend on the upstream crates it actually needs (e.g. `miden-mast-package`).

This will allow us to re-export the proc macros from `miden-test-harness` itself

test-harness/test-harness-lib/src/mod.rs
bitwalker, line 1: This crate should depend on `miden-test-harness-macros`, and re-export its macros from the crate root, e.g.:

```rust
pub use miden_test_harness_macros::{miden_test, miden_test_suite};
```

Then, downstream crates need only depend on `miden_test_harness`

examples/counter-contract/src/lib.rs
```diff
@@ -1,36 +1,117 @@
...
-use miden::{Felt, StorageMap, StorageMapAccess, Word, component, felt};
+use miden_test_harness_macros::miden_test_block;
 
-/// Main contract structure for the counter example.
-#[component]
-struct CounterContract {
-    /// Storage map holding the counter value.
-    #[storage(slot(0), description = "counter contract storage map")]
-    count_map: StorageMap,
+#[cfg(not(test))]
```
bitwalker: Instead of gating this behind `#[cfg(not(test))]`, it would make more sense to gate this behind `#[cfg(target_family = "wasm")]` - our test suite will never be built for the Wasm target, and even if it _was_, then the compilation error you ran into wouldn't be an issue anyway.

examples/counter-contract/src/lib.rs
```diff
@@ -1,36 +1,117 @@
 // Do not link against libstd (i.e. anything defined in `std::`)
-#![no_std]
-#![feature(alloc_error_handler)]
+#![cfg_attr(not(test), no_std)]
```
bitwalker: We're not actually dependent on `std` for the code we emit via `#[miden_test]`/`#[miden_test_suite]` AFAICT, so do we actually need to make this conditionally compiled?

examples/counter-contract/src/lib.rs
```diff
@@ -1,36 +1,117 @@
...
+    }
 }
 
-#[component]
-impl CounterContract {
-    /// Returns the current counter value stored in the contract's storage map.
-    pub fn get_count(&self) -> Felt {
-        let key = Word::from([felt!(0), felt!(0), felt!(0), felt!(1)]);
-        self.count_map.get(&key)
+#[miden_test_block]
```
bitwalker: I believe test discovery will work if instead of `#[miden_test_suite]` stripping out the decorated module entirely, it instead annotated it with `#[cfg(test)]`. I mentioned in my other comment that this would also remove the need for the `test_flag` feature in `miden_test_harness_macros`, but the big win here is test discovery.

examples/counter-contract/src/lib.rs
```diff
@@ -1,36 +1,117 @@
...
+    use miden_lib::account::auth::AuthRpoFalcon512;
+    use miden_objects::account::{
+        auth::AuthSecretKey, Account, AccountBuilder, AccountComponent, InitStorageData,
+    };
+
+    // This tests loads the generated package in the `foo` variable and is then
+    // printed.
+    #[miden_test]
+    fn bar(bar: Package) {
+        std::dbg!(&bar);
```
bitwalker: You can support this even in `#[no_std]` crates by simply adding `#[cfg(test)] extern crate std;` somewhere in the module (conventionally, at the top near the `#![no_std]` attribute itself, but I believe it can go anywhere (I haven't experimented with that though).

examples/counter-contract/src/lib.rs
```diff
@@ -1,36 +1,117 @@
...
+use miden_test_harness_macros::miden_test_block;
 
-/// Main contract structure for the counter example.
-#[component]
-struct CounterContract {
-    /// Storage map holding the counter value.
-    #[storage(slot(0), description = "counter contract storage map")]
-    count_map: StorageMap,
+#[cfg(not(test))]
+mod component {
```
greenhat: Since this is a user-facing example (i.e. used in `cargo miden example`) let's copy it to `tests/rust-apps-wasm/rust-sdk/` where our internal test examples live.

tools/cargo-miden/src/commands/test.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl TestCommand {
+    pub fn exec(self) -> anyhow::Result<()> {
+        let build = BuildCommand {
+            args: self.args.clone(),
+        };
+
+        let Some(CommandOutput::BuildCommandOutput {
+            output: BuildOutput::Masm { artifact_path },
+        }) = build.exec(OutputType::Masm)?
```
greenhat: We discussed that `cargo miden test` should be just a dumb `cargo test` alias, nothing more. Otherwise, the rust-analyzer lens and test discovery would not work. Unless vscode calls `cargo miden test` when the test is started via lens.

examples/counter-contract/src/lib.rs
bitwalker: I'd actually prefer we start placing these integration test examples under a new directory, `tests/examples`. We've had a longstanding TODO to re-organize and cleanup the integration tests directory structure and module organization, and we may as well start doing that with new code.

tools/cargo-miden/src/commands/test.rs
bitwalker: That isn't actually true for discovery at least - the test discovery relies purely on analysis of the macro-expanded code, which with the changes I suggested in my review, will work just like any standard test module.

That said, the `rust-analyzer` test lens is a slightly different story - I haven't looked at how it is implemented, so I don't know how specific it is about how the test gets run.

I agree with the general point though, which is that _ideally_ we can simply make `cargo miden test` an alias of `cargo test`, possibly with additional options to allow for us to provide additional Miden-specific functionality when running tests via `cargo miden`, but hopefully we won't have to require executing tests using `cargo miden test`, to make IDE integration trivial.

greenhat: I agree. The point I wanted to make was that we cannot rely on the `cargo miden test` to compile the package under test. The test harness should compile the package. We could do it in the `main` function generated by the `miden_test_block` macro.

examples/counter-contract/src/lib.rs
greenhat: Sure.

tools/cargo-miden/src/commands/test.rs
bitwalker: The one tricky detail there is going to be whether we will have enough information available when the test binary is built to be able to invoke `cargo miden build`. I _think_ we should be good, but not 100% sure. I'm guessing it would be enough to bake in the build target name when generating `main`? Might have to play with it a bit and see what works and what doesn't.

test-harness/test-harness-lib/Cargo.toml
lima-limon-inc: Changed here: https://github.com/0xMiden/compiler/pull/817/commits/94f024b4d00d6002f7f080b8308e23c2363fc021

test-harness/test-harness-macros/src/lib.rs
lima-limon-inc: Noted! I'll keep it in mind, thanks for the advice 😊 .

Addressed here: https://github.com/0xMiden/compiler/pull/817/commits/c078accd3f083bac18d8d385bf9ae774f0143ea8

test-harness/test-harness-macros/src/lib.rs
lima-limon-inc: Changed the name here: https://github.com/0xMiden/compiler/pull/817/commits/69bea1986931cd146f69bc52d7557a8bdb920678

test-harness/test-harness-lib/src/utils.rs
lima-limon-inc: Removed it here: https://github.com/0xMiden/compiler/pull/817/commits/2cdfd78585b0a4cfd75db618af9e30b0d4139357

test-harness/test-harness-lib/src/libtest.rs
lima-limon-inc: Uops! Nice catch 😊 

Changed here: https://github.com/0xMiden/compiler/pull/817/commits/dffe4b52ed50db72a3a96959a9d0166475b08d37

test-harness/test-harness-macros/Cargo.toml
lima-limon-inc: > This will allow us to re-export the proc macros from miden-test-harness itself

Noted, this approach seems more intuitive! Thanks

Changed here: https://github.com/0xMiden/compiler/pull/817/commits/462c9645b946131f7d2ceac784310697c3e6cc0b

examples/counter-contract/src/lib.rs
lima-limon-inc: Agreed! I've removed the `test_flag` feature flag in this commit: https://github.com/0xMiden/compiler/pull/817/commits/69ea6dacc647cd093ee2d267d5e02a1522c71b8b

test-harness/test-harness-macros/src/lib.rs
lima-limon-inc: Totally! It also makes the code simpler I think! Thanks ^_^

Changed here: https://github.com/0xMiden/compiler/pull/817/commits/69ea6dacc647cd093ee2d267d5e02a1522c71b8b

test-harness/test-harness-lib/src/mod.rs
lima-limon-inc: Addressed here: https://github.com/0xMiden/compiler/pull/817/commits/84331e59099e47569866bddedaf120971d09ddab

examples/counter-contract/src/lib.rs
lima-limon-inc: Also addressed here: https://github.com/0xMiden/compiler/commit/84331e59099e47569866bddedaf120971d09ddab

test-harness/test-harness-lib/src/reexport.rs
lima-limon-inc: Noted! I removed it here: https://github.com/0xMiden/compiler/pull/817/commits/69630838a742fa1af370527fd6450e03391b7fba

examples/counter-contract/src/lib.rs
lima-limon-inc: > We're not actually dependent on std for the code we emit via #[miden_test]/#[miden_test_suite] AFAICT, so do we actually need to make this conditionally compiled?

I believe the `load_package` function was generating std code, since it called `std::fs` to read the  generated`.masp` file at runtime (see [this snippet](https://github.com/0xMiden/compiler/pull/817/changes/9edc872e00fb4053ae33f350339af1fc6cd53638#diff-bfef4f60e5c20b90076b414887da0e957a03cfc8f378562b2d3033edcb9b7af0L101)).

To avoid it, now the file read is executed at macro expansion time and the generated code gets a harcoded byte array, see [this other snippet](https://github.com/0xMiden/compiler/pull/817/changes/9edc872e00fb4053ae33f350339af1fc6cd53638#diff-bfef4f60e5c20b90076b414887da0e957a03cfc8f378562b2d3033edcb9b7af0R104-R107)

This is the full commit https://github.com/0xMiden/compiler/pull/817/commits/9edc872e00fb4053ae33f350339af1fc6cd53638.

test-harness/test-harness-lib/src/reexport.rs
lima-limon-inc: Addressed in this commit: https://github.com/0xMiden/compiler/pull/817/commits/69630838a742fa1af370527fd6450e03391b7fba

test-harness/test-harness-macros/src/lib.rs
lima-limon-inc: Changed here: https://github.com/0xMiden/compiler/pull/817/commits/c4066b24b15961285d52bcf16d4b8a2afe92d576

tools/cargo-miden/src/commands/test.rs
lima-limon-inc: Also changed here: https://github.com/0xMiden/compiler/pull/817/commits/c4066b24b15961285d52bcf16d4b8a2afe92d576

test-harness/test-harness-macros/src/lib.rs
lima-limon-inc: > Why not just emit the module annotated with #[cfg(test)]? That should accomplish the same goal, no?

Addressed in this commit: https://github.com/0xMiden/compiler/pull/817/commits/69ea6dacc647cd093ee2d267d5e02a1522c71b8b

lima-limon-inc:
> We discussed that cargo miden test should be just a dumb cargo test alias, nothing more. Otherwise, the rust-analyzer lens and test discovery would not work. Unless vscode calls cargo miden test when the test is started via lens.

The commits corresponding to this change go from https://github.com/0xMiden/compiler/pull/817/commits/5b3a2574002d672da21a2716485e04765de60854 ... https://github.com/0xMiden/compiler/pull/817/commits/499407cd9b106538a5a3b331a0c16872bf6d2973.

To best way I found for the `#[miden-test]` functions to appear as runnable was by adding `#[test]` (tested in VSCode by seeing the "Run Test" icon next to the function and also on Zed) on top of the function during macro expansion. See the image:

Screenshot 2026-01-16 at 7 58 28 PM

Besides this, `cargo miden test` is simply an alias of `cargo test` (`cargo miden build` is called directly inside the generated code).

However, using `#[test]` did come with a limitation. Adding it on top of a function makes `cargo test` generate the default `libtest` test registration and execution code. Thus, if both our custom registration code (done with `inventory`) and `#[test]` are used, all the tests would be run twice. However, if we rely on the default `#[test]` registration and execution code, we can not rely use our generated `main` function as a reliable setup; since we do not control the order in which tests are executed. From my experimentation, it would seems as if the standard `libtest` main function takes precedence over our own.

Since being able to execute a test from an IDE is such a useful feature, I went with using `#[test]` over `inventory`; which did require some changes. 

Just wanted to mention this. The PR is almost ready for re-review. If I'm not mistaken, the only thing that is missing is moving the example code to `tests/examples` and fixing the conflicts.

tools/cargo-miden/src/commands/test.rs
lima-limon-inc: This change was tackled from https://github.com/0xMiden/compiler/pull/817/changes/06db878b0632a793d6c9235478900cfdae29660f to https://github.com/0xMiden/compiler/pull/817/changes/499407cd9b106538a5a3b331a0c16872bf6d2973. 

The rationale behind the change was explained here: https://github.com/0xMiden/compiler/pull/817#issuecomment-3762180898

examples/counter-contract/src/lib.rs
lima-limon-inc: Moved to `examples/counter-contract/src/lib.rs` in this commit https://github.com/0xMiden/compiler/pull/817/commits/615d142f32f900b02e01f93d8b3d4db1114bdc35.

tools/cargo-miden/src/commands/test.rs
```diff
@@ -0,0 +1,57 @@
...
+        Ok(())
+    }
+}
+
+/// Builds the argument vector for the underlying `cargo test` invocation.
+fn test_cargo_args(cli_args: Vec<String>) -> Vec<String> {
+    let mut args = vec!["test".to_string()];
+
+    // Add build-std flags required for Miden compilation
+    args.extend(["-Z", "build-std=std,core,alloc"].into_iter().map(|s| s.to_string()));
```
greenhat: Why do we need these flags here?

tools/cargo-miden/src/commands/build.rs
```diff
@@ -286,7 +286,7 @@ where
     Ok(outputs)
 }
 
-fn spawn_cargo(mut cmd: Command, cargo: &Path) -> Result<Vec<Artifact>> {
+pub(crate) fn spawn_cargo(mut cmd: Command, cargo: &Path) -> Result<Vec<Artifact>> {
```
greenhat: Seems like it's time to extract it into the `utils` module?

tests/examples/counter/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
```
greenhat, line 1: The counter contract project should be in `tests/examples/counter`, not in the `tests/examples` root.

tests/examples/counter/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+    // 55 |     #[miden_test]
+    //    |     ^^^^^^^^^^^^^
+    //    |
+    //    = help: message:
+    //            Detected that all of the following variables are `Package`s: foo, bar
+    //
+    //            #[miden_test] only supports having a single `Package` in its argument list.
+    // Uncomment to see the failure!
+    // #[miden_test]
+    // fn bing(foo: Package, bar: Package) {
```
greenhat, line 76: I wonder if it's what we'll have to think about eventually. This test suite format (in the same file as the code) fits for unit tests. But because of the context the integration tests are the first tests that will be written by the users. Usually they consist of multiple packages (account, note script, tx script, etc.).

This PR is already large so I'd address this in #876.

tests/examples/counter/Cargo.toml
lima-limon-inc: Noted, my bad. 

Renamed here: https://github.com/0xMiden/compiler/pull/817/commits/d19bd55c30a63498f1cabcd73f8dab820c590951

tools/cargo-miden/src/commands/build.rs
lima-limon-inc: Done in this commit: https://github.com/0xMiden/compiler/pull/817/commits/2cb3669ff0d74e8ffb741c7e46ddf61c076b600a

tools/cargo-miden/src/commands/test.rs
lima-limon-inc: I believe we don't anymore! Nice catch, removed here: https://github.com/0xMiden/compiler/pull/817/commits/a7385487fcf1b705caeb8a3e0ab4e3d6ec522980

lima-limon-inc:
@bitwalker @greenhat I believe this PR is ready for re-review. In the mean time, I'll start working on #876

PR #819, opened 2025-12-12
chore: bump toolchain to 1.94/nightly-2025-12-10

bitwalker:
Bumping the toolchain and Rust edition to pick up some nice changes in latest stable and 2024 edition, as well as to finally be able to get rid of the old `hashbrown` hack we needed to have in place. I've updated our dependencies as well and removed some unused ones that I found.

The main issue I encountered was that something in how Rust is compiling our programs causes an extra object file for the program to be emitted with some kind of default/stub implementation of the `#[alloc_error_handler]` required for no-std crates that link against liballoc. The problem is that these stubs have incorrect signatures, and this causes `rust-lld` to raise an error because it cannot unify the actual implementation defined in liballoc, with the stub implementation. The solution ends up being to define our own `#[alloc_error_handler]` in the crate itself, but this isn't ideal, as it requires activation of `alloc_error_handler` feature, which we cannot hide with other boilerplate emitted by our proc-macro infra.

We'll have to try and dig in deeper to understand what is actually going on there, but for now I've simply modified everything with the workaround above.

Note that `-Zbuild-std-features=panic_immediate_abort` was removed, it is now an actual panic type that can be activated with `panic = "immediate-abort"` or `-Cpanic=immediate-abort`. Use in `Cargo.toml` is still behind a nightly Cargo feature called `panic-immediate-abort`, but we're emitting `-Cpanic=immediate-abort` via `RUSTFLAGS`, so we don't actually need to use it via Cargo anyway.

### DIFF
--- a/tools/cargo-miden/src/commands/new_project.rs
+++ b/tools/cargo-miden/src/commands/new_project.rs
@@ -9,11 +9,17 @@ use toml_edit::{DocumentMut, Item, Value};
 
 use crate::template::{GenerateArgs, TemplatePath, generate};
 
-/// The tag used in checkout of the new project template.
+/// The tag used in checkout of the new contract project template (`cargo miden new --account <NAME>`, `--note`, etc) .
 ///
 /// Before changing it make sure the new tag exists in the rust-templates repo and points to the
 /// desired commit.
-const TEMPLATES_REPO_TAG: &str = "v0.23.0";
+const PROJECT_TEMPLATES_REPO_TAG: &str = "v0.23.0";
+
+/// The tag used in checkout of the new Miden project template (`cargo miden new <NAME>`)
+///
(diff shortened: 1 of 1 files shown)


greenhat:
I reverted https://github.com/0xMiden/project-template/pull/12 because it breaks the `cargo miden new` for the released cargo-miden and we have users now. I made https://github.com/0xMiden/project-template/pull/14 with the same changes instead and added the v0.6 git tag. Then, I made #821 to this branch where I added the git tag to the `cargo miden new` command. So it fetches the Miden project template from the `v0.6` git tag. After we make the next compiler release the https://github.com/0xMiden/project-template/pull/14 will be safe to merge. 

The test_all_templates_and_examples test still fails.

TLDR; merge #821 to this branch, don't merge https://github.com/0xMiden/project-template/pull/14 yet.

PR #823, opened 2025-12-16
feature(SDK): expose `miden::note::build_recipient_hash` tx kernel Rust equivalent as `Recipient::compute`

### LINKED ISSUE #818: Add `miden::note` MASM utility library binding
As of right now, the bindings inside of the `sdk/base-sys/src/bindigns/` folder do not include a file for implementing bindings for the following Miden Assembly file: https://github.com/0xmiden/miden-base/blob/next/crates/miden-lib/asm/miden/note.masm

This file holds utilities for working with notes, most crucially the `build_recipient_hash` procedure. This procedure is especially important to have a binding for, since devs will use it for creating Notes within Rust code.

We should create a binding for those utilities or implement them natively in Rust, so that people can work with them inside of Miden Rust code.

greenhat:
Close #818

### DIFF
--- a/sdk/base-sys/src/bindings/types.rs
+++ b/sdk/base-sys/src/bindings/types.rs
@@ -2,7 +2,7 @@ extern crate alloc;
 
 use alloc::vec::Vec;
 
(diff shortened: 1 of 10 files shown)


greenhat:
@bitwalker Rebased and ready for review.

PR #826, opened 2025-12-17
feat: add Word::empty() method to Miden SDK

### LINKED ISSUE #822: Add `Word::empty()` in the Miden SDK
and eliminate constructing `Word` from 4 zeroes in the existing code.

PivasDesant:
Fixes #822

I've added a `Word::empty()` method that returns a Word with all four field elements set to zero. This replaces the repetitive `Word::from([Felt::from_u32(0); 4])` pattern we had scattered across the test files.

Updated all 5 occurrences in the integration tests to use the new method.

### DIFF
--- a/sdk/stdlib-sys/src/intrinsics/word.rs
+++ b/sdk/stdlib-sys/src/intrinsics/word.rs
@@ -15,6 +15,13 @@ impl Word {
         }
     }
 
+    /// Creates a new `Word` with all four field elements set to zero.
+    pub fn empty() -> Self {
+        Self {
+            inner: (felt!(0), felt!(0), felt!(0), felt!(0)),
+        }
+    }
+
     pub fn reverse(&self) -> Word {
         // This is workaround for the https://github.com/0xMiden/compiler/issues/596 to avoid
         // i64.rotl op in the compiled Wasm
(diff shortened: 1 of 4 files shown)


sdk/stdlib-sys/src/intrinsics/word.rs
```diff
@@ -15,6 +15,13 @@ impl Word {
         }
     }
 
+    /// Creates a new `Word` with all four field elements set to zero.
+    pub fn empty() -> Self {
```
bitwalker: Let's implement this via the `Default` trait instead, which makes it more generally useful anyway.

bitwalker (changes requested):
Thanks! For future reference, we'd prefer if unsolicited PRs refer back to a previously reported issue in the tracker, to avoid people working on things that we are already working on, or that might be wasted effort (e.g. updating code that is being removed). In general we close/reject PRs that don't follow that process.

I've left a comment here with a change I'd like to make before this is merged, other than that, looks good to me.

bitwalker:
Closing as this is no longer applicable here, since the definition of this type now lives in `miden-field` (part of the crypto repo), and it now implements `Default`.

PR #810, opened 2025-12-08
test: Add test for debug locations

djolertrk:
This demonstrates that simple "assert-based debugging" does not work.
The reason is that there is no source locations attached to `ub.unreachable`.

It can be confirmed by using following commands:

```
  # Build the example
  cargo build --release --target wasm32-unknown-unknown \
    --manifest-path examples/assert-debug-test/Cargo.toml

  # Check HIR with source locations  
  ./bin/midenc examples/assert-debug-test/target/wasm32-unknown-unknown/release/assert_debug_test.wasm \
    --entrypoint=assert_debug_test::test_assert \
    -Ztrim-path-prefix=examples/assert-debug-test \
    -Zprint-hir-source-locations \
    --debug full --emit=hir=-
```

And to run the integration test:

```
cargo test --release -p miden-integration-tests debug_source_locations -- --nocapture
```

This addresses https://github.com/0xMiden/compiler/issues/806.

PR #836, opened 2025-12-23
Fix HIR and MASM handling in `--emit` option, add internal group and folder path support

### LINKED ISSUE #724: midenc option `--emit hir=./target` does not emit when passed along with `-o`
Discovered while working on #723

When running `cargo miden build --release --emit hir=./target` the `--emit` option is [overridden](https://github.com/0xMiden/compiler/blob/e6d36e3b8d4e84912c1861f41382500231cb615c/midenc-compile/src/compiler.rs?plain=1#L480-L486) by `-o`. The run finishes with error `compilation was canceled by user` but the Miden package is created.

The `midenc` arguments passed in this case are:
```
    "--output-dir",
    "/Users/green/src/miden/compiler/examples/basic-wallet/target/miden/release",
    "-o",
    "/Users/green/src/miden/compiler/examples/basic-wallet/target/miden/release/basic_wallet.masp",
    "--verbose",
    "--target",
    "rollup:account",
    "--lib",
    "--emit",
    "hir=./target",
```

### LINKED ISSUE #816: IR dumping infrastructure (stop expect tests abuse)
## Problem

We've been using the `expect_test` infrastructure in ways it wasn't designed for. Instead of using it to catch unexpected changes in compiler output, we use it primarily for **introspection** into intermediate representations (WAT, HIR, MASM) during development. I'm the biggest offender. It's just so easy to glance through the WAT/HIR/MASM when I need to.

This creates several issues:

1. **PR noise**: When IR output changes (even legitimately), dozens of expected files need regeneration. This clutters diffs and makes code review harder.

2. **CI friction**: Tests fail when expected files are outdated, requiring us to run `UPDATE_EXPECT=1`, regenerate files, and update the PR—even when the changes are intentional.

3. **False sense of testing**: Typically I regenerate expected files without carefully reviewing the changes. The tests aren't catching bugs; they're just creating busywork.

## Proposed Solution

Replace expect-based IR introspection with an **on-demand dumping infrastructure**, inspired by [rustc's MIR dumping approach](https://rustc-dev-guide.rust-lang.org/mir/debugging.html).

### Design

Add the the `--dump-ir` family of options similar to `rustc` behind nightly `-Z` in `midenc`. They should be available to set via env vars as well.
The `--dump-ir` will be simple boolean for now and extended to full-fledged filtering abilities similar to the `rustc` (see above), like the `--dump-ir='foo & sparse-constant-propagation'` filter only functions containing `foo` in their name and dumping IR before and after the `sparse-constant-propagation` pass.

Output goes to `<workspace_root>/ir_dump/` (`.gitignore`d) with files named `<artifact_name>.<ext>`:
- `my_test.wat` - WebAssembly text format
- `my_test.hir` - High-level IR (adding prefix and suffix for encoding filters for function, pass, etc.)
- `my_test.masm` - Miden Assembly

The `--dump-ir-subdir=my_folder` option will create a `ir_dump/my_folder` folder and write files there. This should allow easy IR diffing while working on the compiler code. Furthermore, this works well with an AI agent (I tested CC with Opus) and is used in its feedback loop.

This would eliminate the issues I mentioned above. The only downside I see from ditching the current expect-test-heavy approach is that PR reviewers will not see how changes to the compiler code affect the produced HIR/MASM code. In the near term the PR author can post generate and post the diff as a PR comment if requested. In the future we could have a special CI job that automates posting these diffs.

I'm test-driving this design in https://github.com/0xMiden/compiler/pull/808.

### My use cases

I list here my use cases with the existing expect tests and how they'd look like with new dumping infra:

#### 1. Track the WAT/HIR/MASM while changing the compiler code.
In this case, I'm running tests in the file watcher that runs them automatically when I edit the .rs files. Prepend the watcher shell command with `DUMP_IR=1` and open the WAT/HIR/MASM files in the editor.

#### 2. Diff the WAT/HIR/MASM while changing the compiler code.

2.1 Establish the baseline with `DUMP_IR=1`.
2.2 Dump with `DUMP_IR_SUBDIR=mydir` and show the diff in the editor.

If I forgot to establish the baseline before changes, do `git stash`, dump with `DUMP_IR_SUBDIR=baseline`, `git stash pop` and dumb with `DUMP_IR=1` after code edits.

### Migration Path

1. Implement dumping infrastructure alongside existing expect tests.

2. Audit existing expect tests:
   - Keep tests that verify **specific, stable** compiler behavior
   - Remove tests that are purely for introspection
   - Convert remaining expect tests to proper behavioral tests where possible.

## Future Enhancements

- **Graphviz output**: Visual CFG representations for HIR like `rustc`

@bitwalker  @otrho  What do you think?

greenhat:
Close #724 
Close #816 

This PR:
- Fixes `--emit=hir` and `masm` and preserves Miden package output location;
- Adds the `--emit=wat[=<path>]` option to emit wat files;
- Adds the `--emit=inter[=<path>]` option to emit wat, hir, masm files;
- Pass all the `cargo miden build` options to the `midenc` (instead of only `--emit`);
- Adds `MIDENC_EXPAND=[<path>]` env var handling in the compiler test suite to emit Rust code with expanded macros;

### DIFF
--- a/midenc-compile/src/stages/codegen.rs
+++ b/midenc-compile/src/stages/codegen.rs
@@ -10,38 +10,9 @@ use midenc_codegen_masm::{
     },
 };
 use midenc_hir::{interner::Symbol, pass::AnalysisManager};
-use midenc_session::{Emit, OutputType, Writer};
+use midenc_session::OutputType;
 
 use super::*;
-
-/// A wrapper that allows emitting a synthetic MASM output representing the full [MasmComponent].
-///
-/// The format matches the [core::fmt::Display] implementation for [MasmComponent], which is
-/// defined in `codegen/masm/src/artifact.rs`.
-struct MasmComponentEmit<'a>(&'a MasmComponent);
-
-impl Emit for MasmComponentEmit<'_> {
-    fn name(&self) -> Option<midenc_hir::interner::Symbol> {
-        None
-    }
-
-    fn output_type(&self, _mode: OutputMode) -> OutputType {
-        OutputType::Masm
-    }
-
-    fn write_to<W: Writer>(
-        &self,
-        mut writer: W,
-        mode: OutputMode,
-        _session: &Session,
-    ) -> anyhow::Result<()> {
-        if mode != OutputMode::Text {
-            anyhow::bail!("masm emission does not support binary mode");
-        }
-        writer.write_fmt(format_args!("{}", self.0))?;
-        Ok(())
-    }
-}
 
 pub struct CodegenOutput {
     pub component: Arc<MasmComponent>,
@@ -99,8 +70,7 @@ impl Stage for CodegenStage {
         }
 
         if session.should_emit(OutputType::Masm) {
-            let artifact = MasmComponentEmit(&masm_component);
-            session.emit(OutputMode::Text, &artifact).into_diagnostic()?;
+            session.emit(OutputMode::Text, masm_component.as_ref()).into_diagnostic()?;
         }
 
         Ok(CodegenOutput {
--- a/midenc-compile/src/stages/parse.rs
+++ b/midenc-compile/src/stages/parse.rs
@@ -1,16 +1,14 @@
 #[cfg(feature = "std")]
+use alloc::borrow::Cow;
+#[cfg(feature = "std")]
 use alloc::string::ToString;
-#[cfg(feature = "std")]
-use alloc::{borrow::Cow, string::String};
 use alloc::{format, rc::Rc, sync::Arc};
-#[cfg(feature = "std")]
-use core::fmt;
 
 use miden_assembly::utils::Deserializable;
 #[cfg(feature = "std")]
 use miden_assembly::utils::ReadAdapter;
 #[cfg(feature = "std")]
-use midenc_session::{Emit, OutputType, Writer};
+use midenc_frontend_wasm::{WatEmit, wasm_to_wat};
 #[cfg(feature = "std")]
 use midenc_session::{FileName, Path};
 use midenc_session::{
@@ -19,34 +17,6 @@ use midenc_session::{
 };
 
 use super::*;
-
-/// A wrapper that emits WebAssembly text format (WAT).
-#[cfg(feature = "std")]
-struct WatEmit<'a>(&'a str);
-
-#[cfg(feature = "std")]
-impl Emit for WatEmit<'_> {
-    fn name(&self) -> Option<midenc_hir::interner::Symbol> {
-        None
-    }
-
-    fn output_type(&self, _mode: OutputMode) -> OutputType {
-        OutputType::Wat
-    }
-
-    fn write_to<W: Writer>(
-        &self,
-        mut writer: W,
-        mode: OutputMode,
-        _session: &Session,
-    ) -> anyhow::Result<()> {
-        if mode != OutputMode::Text {
-            anyhow::bail!("wat emission does not support binary mode");
-        }
-        writer.write_fmt(format_args!("{}", self.0))?;
-        Ok(())
-    }
-}
 
 /// This represents the output of the parser, depending on the type of input that was parsed/loaded.
 pub enum ParseOutput {
@@ -171,7 +141,7 @@ impl Stage for ParseStage {
 impl ParseStage {
     #[cfg(feature = "std")]
     fn emit_wat_for_wasm_input(&self, input: &InputType, session: &Session) -> CompilerResult<()> {
-        if !session.should_emit(OutputType::Wat) {
+        if !session.should_emit(midenc_session::OutputType::Wat) {
             return Ok(());
         }
 
@@ -184,7 +154,9 @@ impl ParseStage {
             InputType::Stdin { input, .. } => Cow::Borrowed(input),
         };
 
-        let wat = wasm_to_wat(wasm_bytes.as_ref())?;
+        let wat = wasm_to_wat(wasm_bytes.as_ref())
+            .into_diagnostic()
+            .wrap_err("failed to convert wasm to wat")?;
         let artifact = WatEmit(&wat);
         session
             .emit(OutputMode::Text, &artifact)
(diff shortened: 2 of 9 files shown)


tests/integration/src/compiler_test.rs
```diff
@@ -1029,6 +1033,90 @@ fn wasm_to_wat(wasm_bytes: &[u8]) -> String {
...
 
+/// Run `cargo expand` for the given Cargo test fixture, and write the expanded Rust code to disk if
+/// `MIDENC_EXPAND[=<path>]` is set.
+///
+/// When `MIDENC_EXPAND` is set with an empty value, the expanded output is written to the current
+/// working directory. When set to `1`, it is treated as enabled and also defaults to the current
+/// working directory. When set to a non-empty value other than `1`, it is treated as the output
+/// directory.
+fn maybe_dump_cargo_expand(test: &CargoTest, rustflags_env: Option<&str>) {
+    let Some(value) = std::env::var_os("MIDENC_EXPAND") else {
```
bitwalker: I know it's more verbose, but could we use `MIDENC_EMIT_MACRO_EXPAND` for this? For a couple of reasons:

* It keeps our environment variables a bit more structured, and eases future additions.
* It is much clearer what the purpose of it is, and it's relationship to `--emit`
* LLMs can probably better understand it. Which isn't a priority for me, but doesn't hurt.

midenc-session/src/outputs.rs
```diff
@@ -64,21 +100,29 @@ impl OutputType {
...
             OutputType::Masm,
             OutputType::Mast,
             OutputType::Masl,
             OutputType::Masp,
         ]
     }
+
+    /// Returns the subset of [OutputType] values considered "intermediate" for convenience
+    /// emission (WAT, HIR, MASM).
+    pub fn inter() -> [OutputType; 3] {
```
bitwalker: *Nit:* I'd prefer just `ir` for this (since all of these are intermediate representations, it's better shorthand, and easier to remember than "inter" IMO).

bitwalker: I'd make the return type `&'static [OutputType]` here, and mark it `const`, that way we don't have to hardcode the size of the array in the signature, and it's just as useful at compile-time as an array.

midenc-session/src/outputs.rs
```diff
@@ -418,6 +545,12 @@ pub enum OutputTypeSpec {
     All {
         path: Option<OutputFile>,
     },
+    /// Emit a common set of intermediate artifacts (WAT, HIR, MASM) into a directory.
+    ///
+    /// The directory path is interpreted relative to the session working directory.
+    Inter {
```
bitwalker: I'd suggest we name this variant `Subset`, and definite it like so:

```rust
enum OutputTypeSpec {
    // ...
    Subset {
        output_types: SmallVec<[OutputType; 3]>,
        path: Option<OutputFile>,
    },
    // ...
}
```

This makes `OutputTypeSpec` a bit more versatile, and doesn't bake into it the assumption that emitting multiple outputs has to be just intermediate representations, but any combination of two or more output types.

midenc-session/src/outputs.rs
```diff
@@ -451,11 +584,13 @@ impl clap::builder::TypedValueParser for OutputTypeParser {
         Some(Box::new(
             [
                 PossibleValue::new("ast").help("Abstract Syntax Tree (text)"),
+                PossibleValue::new("wat").help("WebAssembly text format (text)"),
                 PossibleValue::new("hir").help("High-level Intermediate Representation (text)"),
                 PossibleValue::new("masm").help("Miden Assembly (text)"),
                 PossibleValue::new("mast").help("Merkelized Abstract Syntax Tree (text)"),
                 PossibleValue::new("masl").help("Merkelized Abstract Syntax Tree (binary)"),
                 PossibleValue::new("masp").help("Miden Assembly Package Format (binary)"),
+                PossibleValue::new("inter").help("WAT + HIR + MASM (text, optional directory)"),
```
bitwalker: Per my other comment, let's name this value `ir` instead of `inter`

midenc-compile/src/compiler.rs
```diff
@@ -537,7 +537,11 @@ impl Compiler {
 
         // Initialize output types
         let mut output_types = OutputTypes::new(self.output_types).unwrap_or_else(|err| err.exit());
-        if output_types.is_empty() {
+        let has_final_output = output_types
+            .keys()
+            .any(|ty| matches!(ty, OutputType::Mast | OutputType::Masl | OutputType::Masp));
```
bitwalker: We should remove the `Masl` output type - it is no longer useful, and I'm in the process of removing `Library` entirely from the assembler

midenc-compile/src/stages/parse.rs
```diff
@@ -14,6 +20,34 @@ use midenc_session::{
 
 use super::*;
 
+/// A wrapper that emits WebAssembly text format (WAT).
```
bitwalker: *Nit:* Probably better to define this in the `midenc-frontend-wasm` crate, and it'll reduce the clutter in this file as an extra benefit.

midenc-compile/src/stages/codegen.rs
```diff
@@ -10,10 +10,39 @@ use midenc_codegen_masm::{
...
-use midenc_session::OutputType;
+use midenc_session::{Emit, OutputType, Writer};
 
 use super::*;
 
+/// A wrapper that allows emitting a synthetic MASM output representing the full [MasmComponent].
+///
+/// The format matches the [core::fmt::Display] implementation for [MasmComponent], which is
+/// defined in `codegen/masm/src/artifact.rs`.
+struct MasmComponentEmit<'a>(&'a MasmComponent);
```
bitwalker: Similarly to `WasmEmit`, let's move this to the `midenc-codegen-masm` crate.

bitwalker (changes requested):
Looks good! I have a few minor things I'd like to tweak, but overall I think this should do nicely 👍

tests/integration/src/compiler_test.rs
greenhat: Sure. Done.

midenc-session/src/outputs.rs
greenhat: Done. 
Funnily enough, initially I thought about naming it `ir` but then I decided to go with `inter` to emphasize that this is a group of multiple intermediate formats and not the HIR representation.

greenhat: Done

midenc-session/src/outputs.rs
greenhat: Good idea! Done.

midenc-compile/src/compiler.rs
greenhat: Done.

midenc-compile/src/stages/parse.rs
greenhat: Done.

midenc-compile/src/stages/codegen.rs
greenhat: Done.

greenhat:
Don't merge before https://github.com/0xMiden/compiler/pull/810 . I removed `merge_midenc_flags` in this PR and it is changed in #810. I need to port changes first.

PR #842, opened 2025-12-28
feat: document MASM output option in cargo-miden

### LINKED ISSUE #394: Allow printing the MASM code in `cargo-miden`
### Why

1. The users might want to check out the MASM code of the compiled Miden package for various purposes (debugging, re-use, curiosity, etc.);
2. To compile MASM in the integration tests using the `cargo-miden` (instead of calling the driver);

### How
Clone `midenc` options for pretty-printing the MASM(IR) in `cargo-miden`.

avorylli:
Document the --emit masm option in cargo-miden build command.

The --emit masm option was already supported but lacked documentation.
Added examples showing how to emit MASM code to files, stdout, or output directory.

Fixed #394

### DIFF
--- a/tools/cargo-miden/src/commands/build.rs
+++ b/tools/cargo-miden/src/commands/build.rs
@@ -235,8 +235,18 @@ fn build_cargo_args(cargo_opts: &CargoOptions) -> Vec<String> {
 ///
 /// Only the `--emit` option is merged from user input. All other options are
 /// determined by the detected target environment and project type.
+///
+/// This function enables users to request intermediate compiler artifacts,
+/// such as MASM code, by passing `--emit` options that are forwarded to `midenc`.
+/// For example:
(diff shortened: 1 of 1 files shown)


bitwalker:
The `--emit` option is already documented in `midenc-session` where all of the compiler options are defined. While it doesn't hurt to expand the comment as you've done here, it isn't really an appropriate place to document it, as it isn't user facing, and it is better for the documentation on compiler options to be centralized.

Also, this PR wouldn't fix the issue you've linked, as it doesn't actually implement the change (it has already been implemented actually, so we should close it anyway, but we prefer that closing of issues be associated with the actual fix).

I'm going to close this however, because as a rule, we don't accept trivial documentation PRs without an associated issue on the tracker that requests the change. If you have a suggestion for something you'd like to submit a patch for, please open an issue, so we can discuss how best to tackle it, and avoid duplication of effort by assigning it to you.

PR #843, opened 2025-12-29
Inline stubs

djolertrk:
Do it in two phases:

1) Detect and register all linker stubs first (ensures that when we translate function bodies in pass  all stubs are known and calls to them will be inlined regard less of function ordering in the WASM file)
2) Do the inlining


bitwalker:
@djolertrk Is this ready? Looks like you just need to update the expect tests that are failing (since the handling of stubs has changed), but once that is done, I'm guessing there isn't much more planned for this PR correct?

frontend/wasm/src/module/build_ir.rs
```diff
@@ -115,24 +118,57 @@ pub fn build_ir_module(
...
         let func_name = parsed_module.module.func_name(func_index).as_str();
 
         let function_ref =
             module_state.module_builder.get_function(func_name).unwrap_or_else(|| {
                 panic!("cannot build {func_name} function, since it is not defined in the module.")
             });
-        // If this is a linker stub, synthesize its body to exec the MASM callee.
-        // Note: Intrinsics and Miden ABI (SDK/stdlib) calls are expected to be
-        // surfaced via such linker stubs rather than as core-wasm imports.
-        if maybe_lower_linker_stub(function_ref, &body_data.body, module_state)? {
```
greenhat, line 133: `maybe_lower_linker_stub` handles cases where the stub is transformed to call the stdlib/tx_kernel function. See for example https://github.com/0xMiden/compiler/blob/e8de1aefc40383b3240d8134a1273383760ae5f7/tests/integration/expected/examples/p2id.wat?plain=1#L707-L715 transformed into https://github.com/0xMiden/compiler/blob/e8de1aefc40383b3240d8134a1273383760ae5f7/tests/integration/expected/examples/p2id.hir?plain=1#L1253-L1276  

I don't see them handled anywhere after the deletion of this call.

greenhat (approved):
Look good!

greenhat:
As a side note, github does not send a notification on PR status change (draft -> ready) so when I get a notification for a new commit I'm not sure if it's ready for review. Re-requesting review or ping in a comment is appreciated since it ends up as a notification in my inbox.

PR #849, opened 2025-12-31
test: Update test for debug locations

djolertrk:
After the fix in the debugger, it is able to find proper location.

### DIFF
--- a/tests/integration/src/rust_masm_tests/debug_source_locations.rs
+++ b/tests/integration/src/rust_masm_tests/debug_source_locations.rs
@@ -1,46 +1,22 @@
 //! Tests that verify debug source location information is correctly preserved
 //! from Rust source code through to MASM compilation and execution.
 
-use std::{
-    panic::{self, AssertUnwindSafe},
-    path::PathBuf,
-};
+use std::panic::{self, AssertUnwindSafe};
 
 use miden_core::Felt;
 use midenc_frontend_wasm::WasmTranslationConfig;
 
 use crate::{CompilerTest, testing::executor_with_std};
 
-/// Get the absolute path to the assert-debug-test fixture.
-fn get_test_fixture_path() -> PathBuf {
-    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR")
-        .unwrap_or_else(|_| std::env::current_dir().unwrap().to_str().unwrap().to_string());
-    PathBuf::from(manifest_dir)
-        .parent()
-        .unwrap()
-        .join("rust-apps-wasm")
-        .join("rust-sdk")
-        .join("assert-debug-test")
-}
-
 #[test]
 fn test_rust_assert_macro_source_location_with_debug_executor() {
     let config = WasmTranslationConfig::default();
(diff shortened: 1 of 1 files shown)


tests/integration/src/rust_masm_tests/debug_source_locations.rs
```diff
@@ -96,24 +96,39 @@ fn test_rust_assert_macro_source_location_with_debug_executor() {
                 "Panic message should indicate assertion failure. Got: {panic_message}"
             );
 
-            // Check if source location info is present
-            let has_source_file =
-                panic_message.contains("lib.rs") || panic_message.contains("src/");
+            // Check if source location info is present - the assert! macro is on line 26
+            let has_source_file = panic_message.contains("assert-debug-test/src/lib.rs");
```
greenhat: I believe this is a good fit for the `expect!` macro. See https://github.com/0xMiden/compiler/blob/e8de1aefc40383b3240d8134a1273383760ae5f7/tests/integration/src/rust_masm_tests/examples.rs?plain=1#L43-L61 for an example.

greenhat (changes requested):
The test is red. :)

tests/integration/src/rust_masm_tests/debug_source_locations.rs
```diff
@@ -71,49 +50,58 @@ fn test_rust_assert_macro_source_location_with_debug_executor() {
...
+        };
 
-            // FIXME: Currently source locations show <unavailable> in stack traces.
-            // This test documents the current behavior.
-            eprintln!("SUCCESS: Assertion correctly failed when x=50 <= 100");
+        // The panic message should contain the source location.
+        // Extract just the source location ("assert-debug-test/src/lib.rs:26:13")
+        let panic_message = strip_ansi_codes(&panic_message);
+        let source_location = extract_source_location(&panic_message);
+        expect![[r#"assert-debug-test/src/lib.rs:26:13"#]].assert_eq(&source_location);
```
greenhat: OTOH, if we have to do so much cleaning to use `expect!` maybe `panic_message.contains("assert-debug-test/src/lib.rs:26:13");` would be enough?

greenhat (changes requested):
Thank you! I have second thoughts about `expect!`.

tests/integration/src/rust_masm_tests/debug_source_locations.rs
djolertrk: It is much more straightforward.

djolertrk:
> Thank you! Almost there. Let's slim it down to ease the maintenance:
> 
> 1. No need for `get_test_fixture_path()`
> 2. No need for that much `eprtintln!`;
> 3. Ideally, one assertion on `panic_message.contains("assert-debug-test/src/lib.rs:26:13");` should be enough. If/when line number changes we would rather not update it in 8 places. If the assertion fails let's print the whole `panic_message` (`dbg!(&panic_message);`) to help our future selves.

Okay. I agree. I have used those extra prints during debugging...

PR #865, opened 2026-01-07
Unified `Felt` type PoC

### LINKED ISSUE #861: PoC for `miden-felt` crate with this unified `Felt`
The cheap way to test this idea would be to make a `miden-felt` crate with this unified `Felt` (float for Wasm and wrap `miden-core::Felt` for other targets) type and use it in `miden` (SDK, on-chain) and revamped unified Felt serialization introduced in #808.

_Originally posted by @greenhat in https://github.com/0xMiden/compiler/discussions/853#discussioncomment-15416072_

greenhat:
Close #861 

I deem the PoC a success. Both off-chain and on-chain serialization tests are passing.

This PR:
- Introduces a unified `Felt` type defined in the new `miden-felt` crate (`sdk/felt`).
- Merges on-chain and off-chain Felt serialization into one crate.

The `Felt` is defined at: https://github.com/0xMiden/compiler/blob/19df630f02e9c06e196bc2d73030b6768e52eaea/sdk/felt/src/lib.rs?plain=1#L22-L38

For non-wasm32 targets `Felt` is a wrapper around `miden_core::Felt`.
The `miden-felt` crate is as minimalistic as possible but still due to the orphan rule, carries a significant amount of code.

Next steps after merging this PR and how we could migrate to the unified `Felt`:

 1. The `miden-felt` crate should become a dependency as deep in our dependency tree as possible. Either below `winterfell/p3` or at least below `miden-core`. This `Felt` has to be the canonical felt type throughout our ecosystem.
 2. All end-user-facing types (e.g. `Asset`, `AccountId`, etc.) that might be used in the on-chain code should live in a separate new minimalistic crate (`miden-sdk-types`?). So that everyone uses the same type(e.g. `Asset`) both in the on-chain and off-chain code. This crate should be below the `miden-base` crates.

While working on unification I found a bug and fixed it:
- Change `u64` encoding to 2 felts (`lo u32`, then `hi u32`);

### DIFF
--- a/sdk/felt/src/lib.rs
+++ b/sdk/felt/src/lib.rs
@@ -8,6 +8,8 @@
 #![no_std]
 #![deny(warnings)]
 
+use core::{fmt, hash::Hash};
+
 /// The field modulus, `2^64 - 2^32 + 1`.
 pub const MODULUS: u64 = 0xffff_ffff_0000_0001;
 
@@ -16,6 +18,50 @@ pub const MODULUS: u64 = 0xffff_ffff_0000_0001;
 pub enum FeltError {
     /// The provided value was not a valid canonical felt.
     InvalidValue,
+}
+
+/// A crate-local trait capturing the API surface shared by all `Felt` representations.
+///
+/// This is used to ensure the on-chain and off-chain implementations don't drift in the common
+/// "core" operations, and that required operator traits are implemented consistently.
+pub(crate) trait FeltImpl:
+    Copy
+    + Clone
+    + fmt::Debug
+    + fmt::Display
+    + Eq
+    + Ord
+    + Hash
+    + core::ops::Add<Output = Self>
+    + core::ops::Sub<Output = Self>
+    + core::ops::Mul<Output = Self>
+    + core::ops::Div<Output = Self>
+    + core::ops::Neg<Output = Self>
+    + core::ops::AddAssign
+    + core::ops::SubAssign
+    + core::ops::MulAssign
+    + core::ops::DivAssign
+{
+    /// Creates a `Felt` from `value` without range checks.
+    fn from_u64_unchecked(value: u64) -> Self;
+
+    /// Creates a `Felt` from a `u32` value.
+    fn from_u32(value: u32) -> Self;
+
+    /// Returns the canonical `u64` value of this felt.
+    fn as_u64(self) -> u64;
+
+    /// Returns true if this felt is odd.
+    fn is_odd(self) -> bool;
+
+    /// Returns `self^-1`. Fails if `self = 0`.
+    fn inv(self) -> Self;
+
+    /// Returns `2^self`. Fails if `self > 63`.
+    fn pow2(self) -> Self;
+
+    /// Returns `self^other`.
+    fn exp(self, other: Self) -> Self;
 }
 
 #[cfg(all(target_family = "wasm", miden))]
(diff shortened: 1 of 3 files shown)


sdk/felt/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+pub const MODULUS: u64 = 0xffff_ffff_0000_0001;
+
+/// Errors returned by [`Felt::new`].
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum FeltError {
+    /// The provided value was not a valid canonical felt.
+    InvalidValue,
+}
+
+#[cfg(target_arch = "wasm32")]
```
bitwalker: It's better to use `target_family = "wasm"` here (and elsewhere that we gate on the Wasm targets).

sdk/felt/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+#[repr(transparent)]
+#[derive(Copy, Clone, Debug)]
+/// A `Felt` represented as an on-chain felt.
+pub struct Felt {
```
bitwalker: I would move the actual struct definitions into their respective submodules, and then conditionally re-export the `Felt` type from here. That will keep all of the target-specific details in those modules.

sdk/felt/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// The backing type is `f32` which will be treated as a felt by the compiler.
+    /// We're basically hijacking the Wasm `f32` type and treat as felt.
+    pub inner: f32,
```
bitwalker: This should probably be `pub(crate)` no?

sdk/felt/src/native.rs
```diff
@@ -0,0 +1,203 @@
+//! Off-chain implementation of [`crate::Felt`].
+
+use miden_core::{Felt as CoreFelt, FieldElement};
+
+use crate::{Felt, FeltError, MODULUS};
+
+impl Felt {
```
bitwalker: I wonder if we might not want to define a crate-local trait for all of the APIs that are shared between both representations, in order to catch when we've missed a function somewhere. It would probably also allow sharing implementation details between the two which aren't target-specific.

The functions defined on the `Felt` type itself would then just delegate to the trait implementation.

bitwalker (approved):
This looks great! Definitely an improvement to be able to unify things around a canonical definition.

I've left a few comments on things we likely want to change before landing/moving upstream, but I'll summarize here:

1. Use `target_family = "wasm"` rather than `target_arch = "wasm32"` for target-gating implementation details/features. It doesn't make a practical difference at the moment, but it is the recommended way to do feature-gating for Wasm targets (as well as other targets more generally), and keeps things consistent with how we usually do target gating throughout Miden's repos.
2. Related to 1, but we may need to add a build script that checks for an environment variable set by `cargo-miden`/`midenc` unique to targeting the Miden VM via Wasm. The build script would then emit a custom `cfg` that is present based on whether we are targeting the Miden VM, or a real Wasm runtime, something like `#[cfg(miden)]`. We could then gate the Miden representation behind `#[cfg(all(target_family = "wasm", miden))]`.
3. Provide some kind of way for us to check at compile-time that we haven't missed functionality that should be present across all targets. My feeling is that a trait that defines all of the shared functions would accomplish this - the trait impl would go in the target-specific submodule, and the `Felt` functions/methods could be defined at the crate root and simply delegate to the trait. This doesn't preclude each target from having extra functionality, but ensures the expected common functionality doesn't drift out of sync.

bobbinth:
Looks great! I do have a couple of questions:

> The `miden-felt` crate should become a dependency as deep in our dependency tree as possible. Either below `winterfell/p3` or at least below `miden-core`. This `Felt` has to be the canonical felt type throughout our ecosystem.

I'm assuming that this means that Miden VM and everything downstream would use this as type as `Felt`, right? If so, how do we deal with the fact that we need to compile `miden-vm` and other crates to WASM for things like Web SDK etc. Basically, if we use `#[cfg(target_arch = "wasm32")] ` to re-define `Felt` to be a `f32`, then this would also leak any time we'd want to compile `miden-vm` to WASM as well. Or maybe I'm missing something?

Would another option be to use some dedicated feature (e.g., `compiler_compat`) to gate `f32`-based felt?

> All end-user-facing types (e.g. `Asset`, `AccountId`, etc.) that might be used in the on-chain code should live in a separate new minimalistic crate (`miden-sdk-types`?). So that everyone uses the same type(e.g. `Asset`) both in the on-chain and off-chain code. This crate should be below the `miden-base` crates.

I was thinking about having a crate like this, mostly for centralizing all serialization logic (I was planning to re-use `miden-objects` for this - since this name is now available - but `miden-sdk-types` doesn't sound too bad either). But I wasn't planning to put a lot of functionality into that crate. I'm guessing the value of this crate for the compiler would be that each such type would have some set of utility methods defined that would make it possible to manipulate the type in on-chain code as well, right?

sdk/felt/src/lib.rs
greenhat: No. It will break the bindings generated by the `wit-bindgen`. The type remapping is expected to bring a type that is the same shape as the one generated from WIT. And since in WIT there is no struct tuple support the only way to define a newtype is like this.

I'm adding the comment explaining this.

sdk/felt/src/native.rs
greenhat: Sounds interesting but since we have to implement many traits from `core::ops` (`Add`, `Sub`, etc.) anyway, there will always be room for on-chain/off-chain to drift apart.
This should not stop us though. The only question I have is if we want the on-chain and off-chain API to be the same. So far my gut feeling is that the answer is "yes", but one counter example is enough to blow up this idea.

sdk/felt/src/lib.rs
greenhat: Done in eda5e3cf46c08b200e0a2d89cb89c9679d42b30e.

sdk/felt/src/native.rs
bitwalker: The trait mentioned would only need to handle the overlap between on-chain/off-chain implementations, so it isn't so much that some drift isn't acceptable/expected, but rather that we'll want to ensure that the common "core" is checked in some way.

Regarding certain traits that are also shared in common (namely `Add` and such), we can make those super traits of whatever this common core trait would be, so that they are also captured by this (and we can't forget to implement them). The main trick is determining which traits need to be covered this way (I'd assume mostly arithmetic and basic things like Debug/Display/Copy/Clone/PartialEq/Eq/PartialOrd/Ord/Hash).

greenhat: Done in https://github.com/0xMiden/compiler/pull/865/commits/58c96052f1fa1d2b0446cbf6c8e41ed4f8bb9df4

bitwalker (approved):

PR #869, opened 2026-01-13
Rewrite uses of spilled values in any nested regions of `op`

### LINKED ISSUE #831: `invalid stack offset for movup: 16 is out of range` compilation error
The assertion triggered at 
https://github.com/0xMiden/compiler/blob/920c47a3ccb82bc35a31e29dbba8fd071385a522/codegen/masm/src/emit/mod.rs?plain=1#L680

When compiling the code at https://github.com/inicio-labs/miden-swapp/blob/05c0ce4a5ce1d0490b3923932b1414c05d777fbe/contracts/swapp-note/src/lib.rs

The assertion is triggered by `create_swapp_note` flattened argument size at 15 felts. Decreasing flattened argument size **does not** trigger the assertion. Confirmed in https://github.com/inicio-labs/miden-swapp/pull/1

Log:
```
[TRACE codegen] starting with stack: [
      0 => Operand { word: [U32], operand: Value(v889) },
      1 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v838) },
      2 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v831) },
      3 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v824) },
      4 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v817) },
      5 => Operand { word: [I32], operand: Value(v810) },
      6 => Operand { word: [Felt], operand: Value(v806) },
      7 => Operand { word: [Felt], operand: Value(v804) },
      8 => Operand { word: [Felt], operand: Value(v803) },
      9 => Operand { word: [Felt], operand: Value(v414) },
      10 => Operand { word: [I1], operand: Value(v280) },
      11 => Operand { word: [I32], operand: Value(v2549) },
      12 => Operand { word: [Felt], operand: Value(v2552) },
      13 => Operand { word: [Felt], operand: Value(v2553) },
  ]
[TRACE codegen] stack after scheduling: [
      0 => Operand { word: [U32], operand: Value(v889) },
      1 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v838) },
      2 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v831) },
      3 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v824) },
      4 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v817) },
      5 => Operand { word: [I32], operand: Value(v810) },
      6 => Operand { word: [Felt], operand: Value(v806) },
      7 => Operand { word: [Felt], operand: Value(v804) },
      8 => Operand { word: [Felt], operand: Value(v803) },
      9 => Operand { word: [Felt], operand: Value(v414) },
      10 => Operand { word: [I1], operand: Value(v280) },
      11 => Operand { word: [I32], operand: Value(v2549) },
      12 => Operand { word: [Felt], operand: Value(v2552) },
      13 => Operand { word: [Felt], operand: Value(v2553) },
  ]
[TRACE codegen] dropping unused operands at: hir.store v892, v2552;
[TRACE codegen] found unused operands [] with constraints []
[TRACE codegen] scheduling operands for hir.store v892, v2552;
[TRACE codegen] v892 is live at/after entry: false
[TRACE codegen] v2552 is live at/after entry: false
[TRACE codegen] starting with stack: [
      0 => Operand { word: [Ptr(PointerType { addrspace: Byte, pointee: Felt })], operand: Value(v892) },
      1 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v838) },
      2 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v831) },
      3 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v824) },
      4 => Operand { word: [Array(ArrayType { ty: U8, len: 4 }), Array(ArrayType { ty: U8, len: 4 })], operand: Value(v817) },
      5 => Operand { word: [I32], operand: Value(v810) },
      6 => Operand { word: [Felt], operand: Value(v806) },
      7 => Operand { word: [Felt], operand: Value(v804) },
      8 => Operand { word: [Felt], operand: Value(v803) },
      9 => Operand { word: [Felt], operand: Value(v414) },
      10 => Operand { word: [I1], operand: Value(v280) },
      11 => Operand { word: [I32], operand: Value(v2549) },
      12 => Operand { word: [Felt], operand: Value(v2552) },
      13 => Operand { word: [Felt], operand: Value(v2553) },
  ]
[TRACE midenc_codegen_masm::opt::operands::solver] an initial solution of size 2 was found using tactic TwoArgs

thread 'main' (339581351) panicked at codegen/masm/src/emit/mod.rs:680:20:
invalid stack offset for movup: 16 is out of range
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/c61a3a44d1a5bee35914cada6c788a05e0808f5b/library/std/src/panicking.rs:689:5
   1: core::panicking::panic_fmt
             at /rustc/c61a3a44d1a5bee35914cada6c788a05e0808f5b/library/core/src/panicking.rs:80:14
   2: midenc_codegen_masm::emit::movup_from_offset
             at /Users/green/src/miden/compiler5/codegen/masm/src/emit/mod.rs:680:20
   3: <midenc_codegen_masm::emit::OpEmitter>::movup
             at /Users/green/src/miden/compiler5/codegen/masm/src/emit/mod.rs:353:23
   4: <midenc_codegen_masm::opt::operands::solver::OperandMovementConstraintSolver>::solve_and_apply
             at /Users/green/src/miden/compiler5/codegen/masm/src/opt/operands/solver.rs:341:37
   5: <midenc_codegen_masm::emitter::BlockEmitter>::schedule_operands
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:305:24
   6: <midenc_dialect_hir::ops::mem::Store as midenc_codegen_masm::lower::lowering::HirLowering>::schedule_operands
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/lowering.rs:79:14
   7: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inst
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:117:14
   8: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inline
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:69:18
   9: midenc_codegen_masm::lower::utils::emit_if
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/utils.rs:46:26
  10: <midenc_dialect_scf::ops::If as midenc_codegen_masm::lower::lowering::HirLowering>::emit
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/lowering.rs:154:9
  11: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inst
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:123:14
  12: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inline
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:69:18
  13: midenc_codegen_masm::lower::utils::emit_if
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/utils.rs:46:26
  14: <midenc_dialect_scf::ops::If as midenc_codegen_masm::lower::lowering::HirLowering>::emit
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/lowering.rs:154:9
  15: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inst
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:123:14
  16: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inline
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:69:18
  17: midenc_codegen_masm::lower::utils::emit_if
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/utils.rs:23:22
  18: midenc_codegen_masm::lower::utils::emit_binary_search
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/utils.rs:158:21
  19: <midenc_dialect_scf::ops::IndexSwitch as midenc_codegen_masm::lower::lowering::HirLowering>::emit
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/lowering.rs:379:20
  20: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inst
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:123:14
  21: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inline
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:69:18
  22: midenc_codegen_masm::lower::utils::emit_if
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/utils.rs:46:26
  23: <midenc_dialect_scf::ops::If as midenc_codegen_masm::lower::lowering::HirLowering>::emit
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/lowering.rs:154:9
  24: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inst
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:123:14
  25: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inline
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:69:18
  26: midenc_codegen_masm::lower::utils::emit_if
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/utils.rs:46:26
  27: <midenc_dialect_scf::ops::If as midenc_codegen_masm::lower::lowering::HirLowering>::emit
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/lowering.rs:154:9
  28: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inst
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:123:14
  29: <midenc_codegen_masm::emitter::BlockEmitter>::emit_inline
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:69:18
  30: <midenc_codegen_masm::emitter::BlockEmitter>::emit
             at /Users/green/src/miden/compiler5/codegen/masm/src/emitter.rs:37:14
  31: <midenc_codegen_masm::lower::component::MasmFunctionBuilder>::build
             at /Users/green/src/miden/compiler5/codegen/masm/src/lower/component.rs:594:32

  ...

```

greenhat:
Close #831. 

This PR:

1. Rewrites uses of spilled values in any nested regions of op.

This is required for correctness when spills/reloads are inserted into regions nested under a
CFG block (e.g. structured control flow like `scf.if`), as those uses are otherwise ignored by
the current CFG rewrite pass.

2. Adds `merge_op_nested_region_uses` to collect unsatisfied uses of spilled values from any nested regions under an op.
  - For `RegionBranchOpInterface`, use successor regions to only include regions reachable from the parent.
  - For other region-bearing ops, conservatively scan each single-block region and merge its uses.

3. Deduplicate the per-op use/def tracking between the CFG and single-block rewrite paths by reusing `find_inst_uses` in the single-block traversal, reducing divergence and preventing the class of regressions 

4. `RegionBranchOpInterface` successors are now handled transitively so nested regions reachable through other regions contribute uses during CFG rewrite (`scf.while` `after` case).

I recommend reviewing this PR on a per-commit basis. It starts with the original fix and then generalizing the solution.


tests/integration/src/rust_masm_tests/rust_sdk/mod.rs
bitwalker, line 1: This is a good example of something we should use `lit` for. If you move this reproducer crate to `tests/lit/issue831`, and in that directory add two files:

1. `lit.suite.toml`:

```toml
name = "issue831"
patterns = ["*.masm"]
working_dir = "../../../"

[format.shtest]
```

2. `test.masm`:

```
;; RUN: cargo miden build --release --manifest-path %S/test-project/Cargo.toml --emit=masm=- 2>&1 | filecheck %s

;; CHECK-NOT: invalid stack offset for movup: 16 is out of range
;; CHECK: proc {{.*}}create_swapp_note{{.*}}
```

This test will fail if the output of `cargo miden build` contains the error message, and succeeds so long as we observe the expected procedure definition in the emitted MASM (that part may need to be tweaked, I just threw that in as an example). The `CHECK:` directive here just acts as a positive test to avoid false positives due to compilation failing for a reason other than the one related to issue 831.

dialects/hir/src/transforms/spill/expected/materialize_spills_nested_scf_while_after_region_before.hir
bitwalker, line 1: I'd like to use `lit` for these sort of tests, to be resilient against other unrelated IR changes breaking the test, and crucially, to avoid us ignoring a regression because of expect test churn. 

Unfortunately, for something like this, we really would benefit from having HIR parsing implemented, because to test the thing we're trying to test here, we currently need to find a Rust/WAT input that produces suitable "before" IR, and then use `midenc`/`cargo miden build` with `--print-ir-after-pass=transform-spills` set to assert things about the post-spills HIR. Once the `--emit` infra changes we discussed are implemented, we could write this as a `lit` test I think (so long as we can contrive Rust/WAT input that produces something close enough to the HIR constructed by hand here).

Anyway, I bring this up not because this needs to be written as a `lit` test in this PR (unless you think you can make it work with what we have on hand today), but because for testing compiler passes, using expect tests is going to be a maintenance headache at best, and will be hard to review (both in terms of initial review, and review when expected outputs change for some reason).

One thing I think would be a useful addition to the `filecheck` component of `litcheck` though, would be an API that lets us do something like:

```rust
filecheck!(post_spills_hir.to_string(), "
; CHECK: ...
; CHECK: ...
");
```

In other words, apply `filecheck` rules to any arbitrary string. This actually does exist in the crate already, but not wrapped up in a nice macro interface like that. We could then avoid the need to parse HIR at all (as well as the overhead of managing external processes), while still benefiting from `lit`-style tests. I'll make a note to do that ASAP.

bitwalker: As a follow up to this, I've actually published a new release of all the `litcheck` crates (0.3.0) that provides this. See `litcheck_filecheck::filecheck!` for details.

hir-transform/src/spill.rs
```diff
@@ -538,6 +499,112 @@ fn rewrite_cfg_spills(
...
+
+        let region_used = collect_region_uses(entry, analysis);
+        for (value, users) in region_used {
+            used.entry(value).or_default().extend(users.iter().copied());
+        }
+    }
+}
+
+/// Collect the set of unsatisfied uses of spilled values at the entry of `block_ref`.
+fn collect_region_uses(
```
bitwalker: *Nit:* I find the name of this a bit confusing - it's collecting uses from a single block, not a region. It just so happens that this is called from another function which is ensuring that regions it processes are single-block regions.

I'd just call this `collect_block_uses` since that's the unit of IR it operates on.

hir-transform/src/spill.rs
```diff
@@ -538,6 +499,112 @@ fn rewrite_cfg_spills(
...
+        merge_nested_region_uses(branch, used, analysis);
+        return;
+    }
+
+    // Fallback for ops that contain regions but do not implement RegionBranchOpInterface.
+    //
+    // We conservatively treat each region as reachable from the parent, and only attempt to
+    // collect uses from single-block regions (multi-block nested regions require SSA reconstruction
+    // within the nested CFG).
+    for region in op.regions().iter() {
```
bitwalker, line 543: *Nit:* We should skip this for ops which implement `IsolatedFromAbove`, since by definition such ops cannot reference SSA values from outer scopes.

As far as I'm aware, we don't have any ops with regions that don't implement either `RegionBranchOpInterface` or `IsolatedFromAbove`, so I suspect that in practice we wouldn't ever hit this. That said, if the op doesn't implement either of those traits, then this is _probably_ sane. Since we don't have an example of something that meets the criteria, it's a bit tough to say, but I'm comfortable with it for now.

hir-transform/src/spill.rs
```diff
@@ -538,6 +499,112 @@ fn rewrite_cfg_spills(
...
+        let region = region.as_region_ref();
+        let region_borrowed = region.borrow();
+
+        if region_borrowed.is_empty() {
+            continue;
+        }
+        if !region_borrowed.has_one_block() {
```
bitwalker, line 550: We could probably handle multi-block regions using normal SSA CFG semantics, but I think this is a fine limitation to impose to keep the complexity down. For the time being, we don't have any such ops anyway.

hir-transform/src/spill.rs
```diff
@@ -538,6 +499,112 @@ fn rewrite_cfg_spills(
     rewrite_spill_pseudo_instructions(context, analysis, interface, Some(dominfo))
 }
 
+/// Rewrite uses of spilled values in `op` and any nested regions of `op`.
+///
+/// This is required for correctness when spills/reloads are inserted into regions nested under a
+/// CFG block (e.g. structured control flow like `scf.if`), as those uses are otherwise ignored by
+/// the CFG rewrite pass.
+fn find_inst_uses_including_nested_regions(
```
bitwalker: *Nit:* Is there any reason not to just name this `find_inst_uses`? My understanding is we're just addressing the issue that the old implementation of `find_inst_uses` was not visiting the regions of ops in a given block.

dialects/hir/src/transforms/spill/expected/materialize_spills_nested_scf_while_after_region_before.hir
bitwalker: The biggest benefit I can see already to using `lit`-style tests for this, is it is clear exactly what we're expecting to see change. It is very hard to get that from these expect files unfortunately.

Using `filecheck`, we can do:

```rust
filecheck!(after, "
; COM: Spill of v5
; CHECK:      hir.store_local v5 #[local = lv[[L:r[0-9]+]]]
; CHECK-NEXT: hir.exec @test/example

; COM: First reload of v5
; CHECK-LABEL: ^block3
; CHECK:            v[[R1:r[0-9]+]] = hir.load_local  : ptr<element, u128> #[local = lv[[L]]];
; CHECK-NEXT:       hir.store v[[R1]], v{{\d+}}

; COM: Second reload of v5
; CHECK-LABEL: ^block4
; CHECK:            v[[R2:r[0-9]+]] = hir.load_local  : ptr<element, u128> #[local = lv[[L]]];
; CHECK-NEXT:       hir.store v[[R2]], v{{\d+}}
");
```

The other nice thing here, is that we can make these tests resilient against minor things changing the SSA value names, as shown above.

bitwalker (approved):
I think this looks good! I have a couple of nits, but not really anything significant.

The biggest thing here is the testing aspect, I think it'd be worth making use of the new `filecheck!` macro I implemented in `litcheck_filecheck`, to allow us to write more precise tests, but if you'd rather do that in a separate PR, that's fine too.

tests/integration/src/rust_masm_tests/rust_sdk/mod.rs
greenhat: Thank you for nudging me to the lit tests! Nice! Since the lit tests are broken ATM (see #874) I created #875 for this and left some comments there.

dialects/hir/src/transforms/spill/expected/materialize_spills_nested_scf_while_after_region_before.hir
greenhat: > I'd like to use `lit` for these sort of tests, to be resilient against other unrelated IR changes breaking the test, and crucially, to avoid us ignoring a regression because of expect test churn.

Old habits die slowly. I put expect check there for introspection only. The check is performed by these asserts at the end of the test: https://github.com/0xMiden/compiler/blob/00dd5228c0c611b21aedb54cd4a1b5b4a1b2582c/dialects/hir/src/transforms/spill/tests.rs?plain=1#L436-L437. The `filecheck!` will fit nicely here. I'll do it in this PR.


> Unfortunately, for something like this, we really would benefit from having HIR parsing implemented, because to test the thing we're trying to test here, we currently need to find a Rust/WAT input that produces suitable "before" IR, and then use `midenc`/`cargo miden build` with `--print-ir-after-pass=transform-spills` set to assert things about the post-spills HIR. Once the `--emit` infra changes we discussed are implemented, we could write this as a `lit` test I think (so long as we can contrive Rust/WAT input that produces something close enough to the HIR constructed by hand here).

I considered doing it as a lit test but decided that the showstopper is the absence of the HIR parser. I would like to write the input as IR rather than Rust code.

greenhat: Done in 5d2dd782c0af5fba9fe6950dd3a2044272c0ef08

hir-transform/src/spill.rs
greenhat: Done in 55a29a074ee1eb7d71a3720ec6c4ebaa04140fe4

hir-transform/src/spill.rs
greenhat: Done in 55a29a074ee1eb7d71a3720ec6c4ebaa04140fe4

hir-transform/src/spill.rs
greenhat: Done in 55a29a074ee1eb7d71a3720ec6c4ebaa04140fe4

greenhat:
I addressed all the notes. Please do another round.

dialects/hir/src/transforms/spill/expected/materialize_spills_nested_scf_while_after_region_before.hir
greenhat: I could not make fancy `CHECK`s (with capturing) work, so I resorted to the simple `CHECK` and filed https://github.com/bitwalker/litcheck/issues/1

bitwalker: I'm looking into the issue with capturing - some of my syntax was a bit off (I was writing it from memory, and it's been awhile 😅), but there does seem to be an issue with variable scoping, which prevents the captures from working properly.

Once that is fixed, I'll push a new release and we can try again.

PR #882, opened 2026-01-16
fix: operands order in `le` and `lt` `Felt` op bindings

### LINKED ISSUE #175: Missing integration (semantic) tests for comparison ops for Felt intrinsics
Due to the missing comparison ops implementation for the Rust `Felt` type.

### LINKED ISSUE #868: Check `lt` and `gt` felt ops
Mentioned elsewhere:

> `lt` and `gt` operator doesn't give correct output , I used  `.as_u64()` over the felt to resolve this
> ```rust
>     {
>         // Validate input: input_amount must not exceed requested_asset_total
>         let is_valid = if total_input_amount <= requested_asset_total {
>             felt!(1)
>         } else {
>             felt!(0)
>         };
> 
>         assert_eq(is_valid, felt!(1));
>     }
> ```

greenhat:
Close #868 
Close #175

### DIFF
--- a/sdk/field/src/wasm32.rs
+++ b/sdk/field/src/wasm32.rs
@@ -255,12 +255,12 @@ impl PartialOrd for Felt {
 
     #[inline(always)]
     fn lt(&self, other: &Self) -> bool {
-        unsafe { extern_lt(*other, *self) != 0 }
+        unsafe { extern_lt(*self, *other) != 0 }
     }
 
     #[inline(always)]
     fn le(&self, other: &Self) -> bool {
-        unsafe { extern_le(*other, *self) != 0 }
+        unsafe { extern_le(*self, *other) != 0 }
     }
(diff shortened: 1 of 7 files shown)


sdk/field/src/wasm32.rs
```diff
@@ -255,12 +255,12 @@ impl PartialOrd for Felt {
 
     #[inline(always)]
     fn lt(&self, other: &Self) -> bool {
-        unsafe { extern_lt(*other, *self) != 0 }
+        unsafe { extern_lt(*self, *other) != 0 }
```
bitwalker, line 258: Ah, I see, we were reversing the operand order in the Rust bindings, but since the compiler already does the reversal, we ended up with the operands in the wrong order. Good catch!

bitwalker (approved):

PR #884, opened 2026-01-16
Stack scheduling fails on 16-felt function arguments calls when other live values are present

### LINKED ISSUE #872: Stack scheduling fails on 16-felt function arguments calls when other live values are present
Discovered while fixing #831

greenhat:
Close #872

Fix invalid MASM stack access in the operand-movement solver by making copy/materialization and tactic selection aware of the 16-felt stack window.

Changes in this PR:
- Reject candidate solutions that would require MASM to address beyond the first 16 field elements.
- ~~Improve the `Linear` operand-movement tactic to keep move-constrained operands within the addressable window while materializing required copies.~~  I extracted the logic from `Linear` into the `LinearStackWindow` tactic. A wrapper around `Linear`. 
- Exhaust tactics before stopping the solver on fuel.
- Add regression coverage (integration test for #831) plus focused unit tests in `linear.rs` that document how the tactic resolves cycles, evicts non-expected operands, and materializes copies safely.
- Regression test for the case of 4 words + 1 felt (17 felts total) function arguments where the words are passed by pointer.

### DIFF
--- a/codegen/masm/src/opt/operands/tactics/linear_stack_window.rs
+++ b/codegen/masm/src/opt/operands/tactics/linear_stack_window.rs
@@ -1,6 +1,9 @@
 use super::*;
 use crate::opt::operands::MASM_STACK_WINDOW_FELTS;
 
+/// Log target for this tactic and its helpers.
+const LOG_TARGET: &str = "codegen:scheduling:linear-windowed";
+
 /// Represents the portion of the operand stack which is directly addressable by MASM stack
 /// manipulation instructions.
 #[derive(Debug, Copy, Clone)]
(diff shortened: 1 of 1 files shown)


greenhat:
@bitwalker I extracted the logic from `Linear` into the `LinearStackWindow` tactic. A wrapper around `Linear`. Added tests and refactored a bit. Now it's ready for review.

codegen/masm/src/opt/operands/tactics/linear_stack_window.rs
```diff
@@ -0,0 +1,366 @@
...
+                }
+                _ => {}
+            }
+        }
+
+        let Some((source_at, expected)) = next_copy else {
+            break;
+        };
+
+        log::trace!("materializing copy of {expected:?} from index {source_at} to top of stack",);
```
bitwalker: We should make sure to add targets to all of our logging, e.g. `"codegen:scheduling:linear-windowed"` in this case. We still have a lot of places without targets specified, but all new ones should, and we can fix the old ones as we make changes in those areas of the codebase.

greenhat: Thanks! Fixed and added more logging in 7d2b3885c69c1af91d075511a1d7d4563a8bc81f.

PR #888, opened 2026-01-20
`i64.rotl` and `i64.rotr` with 32 bit rotation rewrite

### LINKED ISSUE #596: `#[repr(C, align(16))]` triggers unwanted `u64` Wasm ops on felt
Discovered in #589 

To reproduce, remove the `#[inline]` attribute on `hash_elements` at https://github.com/0xMiden/compiler/blob/8c04086a2c8a5af89f765e5e94ea84abab0eea5e/sdk/stdlib-sys/src/stdlib/crypto/hashes.rs?plain=1#L204.

The `Word::reverse()` (implemented at https://github.com/0xMiden/compiler/blob/8c04086a2c8a5af89f765e5e94ea84abab0eea5e/sdk/stdlib-sys/src/intrinsics/word.rs?plain=1#L18-L22) call on the last line of the function will be compiled to two `u64.rotl` Wasm ops, which fail on felts.

Removing the `#[repr(C, align(16))]` makes this issue go away, and the test suite stays green.

### LINKED ISSUE #664: Remove the `Word::reverse` workaround for #596
### Why
To avoid the awful "workaround" - https://github.com/0xMiden/compiler/blob/d6a13ea66fa563c62a72cfd9d041d5aedda12488/tests/integration/expected/examples/basic_wallet.wat?plain=1#L226-L283

### How
Either after the https://github.com/0xMiden/compiler/issues/596 is done or by calling [`std::word::reverse`](https://github.com/0xPolygonMiden/miden-vm/blob/583fbaf7863e7b4cd15f484c3ae73795691de74c/stdlib/asm/word.masm?plain=1)

greenhat:
Close #596 
Close #664 

This PR:

- Add `arith` canonicalization pattern to rewrite `i64`/`u64` `rotl`/`rotr` by `32` into a swap of the two 32-bit limbs via `arith.split`/`arith.join` using `felt` limbs (preserves any extra bits).
- Generalize `arith.join`/`arith.split` to be type-directed.
- Add support in the spill analysis to handle cases where block arguments/results exceed the K.
- Relax `OperandStack::rename` to allow >16 deep indices.

### DIFF
--- a/codegen/masm/src/stack.rs
+++ b/codegen/masm/src/stack.rs
@@ -327,7 +327,6 @@ impl OperandStack {
     /// Renames the `n`th operand from the top of the stack to `value`
     ///
     /// The type is assumed to remain unchanged
-    /// NOTE: `n` can be > 16
     pub fn rename(&mut self, n: usize, value: ValueRef) {
         let len = self.stack.len();
         assert!(n < len, "invalid operand stack index ({n}), only {len} operands are available");


dialects/arith/src/builders.rs
```diff
@@ -499,18 +499,85 @@ pub trait ArithOpBuilder<'f, B: ?Sized + Builder> {
...
+    {
+        let op_builder = self.builder_mut().create::<crate::ops::Join, (A, Type)>(span);
+        let op = op_builder(limbs, ty)?;
         Ok(op.borrow().result().as_value_ref())
     }
 
-    fn split(&mut self, n: ValueRef, span: SourceSpan) -> Result<(ValueRef, ValueRef), Report> {
+    /// Join 2 limbs into a single value of type `ty`.
+    ///
+    /// Limbs are ordered from most-significant to least-significant.
```
bitwalker, line 515: This will change when we move to miden-vm v0.21, just FYI - everything will be little-endian there.

dialects/arith/src/builders.rs
```diff
@@ -499,18 +499,85 @@ pub trait ArithOpBuilder<'f, B: ?Sized + Builder> {
...
+    fn join4(
+        &mut self,
+        limbs: [ValueRef; 4],
+        ty: Type,
+        span: SourceSpan,
+    ) -> Result<ValueRef, Report> {
+        self.join(limbs, ty, span)
+    }
+
+    /// Split `n` into limbs of type `limb_ty`, ordered from most-significant to least-significant.
```
bitwalker, line 538: Same here

codegen/masm/src/lower/lowering.rs
```diff
@@ -1210,10 +1210,53 @@ impl HirLowering for arith::Cto {
...
+
+        let args = self.required_operands();
+        if args.is_empty() {
+            return Ok(());
+        }
+
+        let mut constraints = emitter.constraints_for(op, &args);
+        let mut args = args.into_smallvec();
+
+        // For `i128`/`u128` we use a different stack order for 64-bit limbs.
```
bitwalker, line 1224: This will be consistent after v0.21 as well

codegen/masm/src/stack.rs
```diff
@@ -327,8 +327,12 @@ impl OperandStack {
     /// Renames the `n`th operand from the top of the stack to `value`
     ///
     /// The type is assumed to remain unchanged
+    /// NOTE: `n` can be > 16
```
bitwalker: Can it? AFAIK we assert everywhere in the emitter that would try to specify an operand index > 16. I can't recall if we do so in this stack datastructure directly (since it abstracts over the actual size of the Miden operand stack, and is only considering size in logical operands).

We should probably be stricter and enforce the size against the underlying element size of the stack (and I know we do in some places, just not consistently), but I'm not sure if we rely on being able to overflow or not anywhere (we shouldn't be, but I'm not 100% on that).

dialects/arith/src/builders.rs
greenhat: Thank you for the confirmation/heads up! I was thinking about it as a temporary solution.

codegen/masm/src/stack.rs
greenhat: No, it cannot. I botched the comment. The `n` is less than 16, but the stack size is > 16. The reason I removed the assert was that I'm witnessing in https://github.com/0xMiden/compiler/blob/10f67419fcce17ebe33364008b655a18a592bea7/tests/integration/expected/examples/auth_component_rpo_falcon512.hir?plain=1#L234 that although `n` is zero the stack size is 17. The stack is filled mostly with the block80 and block82 params. `v927` is short-lived, so although the stack grows > 16, there are no access ops below 16 elements.

PR #889, opened 2026-01-20
refactor(sdk): make some constructors `const fn`

### LINKED ISSUE #784: Make SDK type constructors `const fn` where possible
See the `key` definition in the counter contract code at https://github.com/0xMiden/project-template/blob/61878235de3d73345dbb43f6f27a788423e111a5/contracts/counter-account/src/lib.rs?plain=1#L24. It should be a `const`.

mooori:
Closes #784, though there might be follow ups as described below.

These are all cases I found where constructors can be made `const fn` without breaking APIs or adding new constructors. It's sufficient to use a `const fn` constructor in [the code](https://github.com/0xMiden/project-template/blob/61878235de3d73345dbb43f6f27a788423e111a5/contracts/counter-account/src/lib.rs?plain=1#L24) mentioned in #784 when using `Word::new`.

### Potential follow-ups for more `const fn` constructors

- Adding new constructors due to:
  - [E0379](https://doc.rust-lang.org/error_codes/E0379.html): Some constructors are trait methods of `From<T>, FeltImpl, ...` and can therefore not be `const fn`.
  - Some constructors perform checks and return a `Result`.
  - Newly added `const fn` constructors could be outside of traits or be `*_unchecked()` versions.
- Breaking APIs, for example make `Asset::new` take `word: Word` instead of `word: impl Into<Word>`

My gut feeling is that these follow-ups would be worth it only if there's use case that needs them or a speedup.

## TODO

- [ ] Open issue in `project-template` repo to use `const Word::new` once it uses an sdk version which has `const fn Word::new`.

### DIFF
--- a/examples/counter-contract/src/lib.rs
+++ b/examples/counter-contract/src/lib.rs
@@ -9,6 +9,8 @@
 
 use miden::{Felt, StorageMap, StorageMapAccess, Word, component, felt};
 
+const STORAGE_KEY: Word = Word::new_const(0, 0, 0, 1);
+
 /// Main contract structure for the counter example.
 #[component]
 struct CounterContract {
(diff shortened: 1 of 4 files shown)


sdk/stdlib-sys/src/intrinsics/crypto.rs
```diff
@@ -17,9 +17,9 @@ pub struct Digest {
 impl Digest {
     /// Creates a new `Digest` from a `[Felt; 4]` array.
     #[inline]
-    pub fn new(felts: [Felt; 4]) -> Self {
+    pub const fn new(felts: [Felt; 4]) -> Self {
         Self {
-            inner: Word::from(felts),
+            inner: Word::new(felts),
```
mooori, line 22: `Word::from` is a trait method and therefore can't be const.

greenhat (changes requested):
Good thinking! I agree. 

To check the DX let's convert https://github.com/0xMiden/compiler/blob/f71f1743438476685a10184d897934929d7dc965/examples/counter-contract/src/lib.rs?plain=1#L24 to `const`.
Also, let's make `Digest` as a constant definition at 
https://github.com/0xMiden/compiler/blob/f71f1743438476685a10184d897934929d7dc965/tests/rust-apps-wasm/rust-sdk/issue-invalid-stack-offset-movup/src/lib.rs?plain=1#L123-L128

PR #891, opened 2026-01-22
feat(sdk)!: assert `value <= Felt::M` in `Felt::from_u64_unchecked`

mooori:
[Reference](https://github.com/0xMiden/compiler/pull/889#issuecomment-3782884359): Adds on `assert` to avoid silently truncating in `Felt::from_u64_unchecked(value)` if `value > Felt::M`. If there's code which hits this case, it'll now panic instead of silently pass.

`rustc` gets quite far with verifying and removing asserts at comptime, see below.

### Why no changes in `tests/integration/expected`

If `value` is known at comptime, the assert does not show up in wasm, even for something like:

https://github.com/0xMiden/compiler/blob/15d5cc0ead5f6df661b0f05a253111309cfed025/tests/integration/src/rust_masm_tests/abi_transform/advice_map.rs#L34

Apparently `rustc` figures out that the assert cannot trigger because:

```python
# u64::MAX / 4 < Felt::M
18446744073709551615 / 4 < 2**64 - 2**32 + 1 # => true
```

### Wasm/wat generated for the `assert`

```Rust
#[unsafe(no_mangle)]
pub fn entrypoint(a: u64, b: u64) -> u64 {
    let x = Felt::from_u64_unchecked(a + b);
    x.as_u64()
}
```

compiled to wat:

```wat
(func $entrypoint (;2;) (type 2) (param i64 i64) (result i64)
    block ;; label = @1
      local.get 1
      local.get 0
      i64.add
      local.tee 0
      i64.const -4294967294
      i64.lt_u
      br_if 0 (;@1;)
      unreachable
    end
    local.get 0
    call $intrinsics::felt::from_u64_unchecked
    call $intrinsics::felt::as_u64
  )
```

This corresponds to expectations: one comparison that triggers a trap if `value > Felt::M`.

Since there are no changes in `expected/`: is above manual check of generated wasm still sufficient or should we add a test?


sdk/sdk/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### BREAKING
 - Require `&mut` in mutating methods of the account storage;
+- Raise a panic in `Felt::from_u64_unchecked` if the value is larger than `Felt::M` #891
```
greenhat: I appreciate you updated the changelog! I don't think it's a breaking change though.

PR #894, opened 2026-01-23
feat(sdk): add `Word` constructor with friendlier DX

mooori:
[Reference](https://github.com/0xMiden/compiler/pull/889#issuecomment-3782884359)

Adds a new `Word` constructor: `Word::from_u64_unchecked([0, 0, 0, 1])`

Previously: `Word::from([felt!(0), felt!(0), felt!(0), felt!(1)])`


sdk/stdlib-sys/src/intrinsics/word.rs
```diff
@@ -15,6 +15,22 @@ impl Word {
         }
     }
 
+    /// Converts each of the `values` to `Felt`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if a value is larger than `Felt::M`.
```
mooori, line 22: Will be true after #891 is merged.

sdk/stdlib-sys/src/intrinsics/word.rs
```diff
@@ -15,6 +15,22 @@ impl Word {
...
+    pub fn from_u64_unchecked(values: [u64; 4]) -> Self {
```
bitwalker: If this is represented as `pub fn from_u64_unchecked(a: u64, b: u64, c: u64, d: u64) -> Self`, then a considerable amount of overhead in the generated code can be avoided.

By using an array, the generated Wasm gets a pointer and a length to the array data, and must compute the pointer to each u64 element and dereference it to get it on the operand stack, before it can pass it to `Felt::from_u64_unchecked`.

If you use plain arguments, then each `u64` argument gets passed directly to `Felt::from_u64_unchecked`, and there is a decent chance that the optimizer will be able to elide all of the u64-related overhead when the inputs to `Word::from_u64_unchecked` are literal values, and even if it's a mix of literal and non-literal values, some overhead can be optimized away.

Additionally, if both the `Word::from_u64_unchecked` and `Felt::from_u64_unchecked` functions are marked `#[inline(always)]`, the optimizer will be able to make use of local sparse-conditional constant propagation and integer range analysis to help elide the assertion overhead in `Felt::from_u64_unchecked` even in cases where the values aren't literal. For example, if a conditional like `if v < u32::MAX as u64` guards a call to `Felt::from_u64_unchecked(v)`, then the optimizer can see that an assertion in `Felt::from_u64_unchecked` like `assert!(v < Felt::M)` cannot possibly fail along that control flow path.

So, all that said, I think we should avoid using an array for this function, and use individual parameters. What do you think @greenhat?

greenhat: Yes. I agree. The array is not worth it.

mooori: Done and the diff in `expected/` shows the overhead disappears :)

In [word.rs](https://github.com/0xMiden/compiler/blob/6e14a767d504be5d9bd47d0f316effe10e53bde4/sdk/stdlib-sys/src/intrinsics/word.rs) there are more constructors taking `[Felt; 4]`. Should they remain as is or can we at some point try to see if that could be optimized too? Would probably be a breaking change for all sdk users.

greenhat: > In [word.rs](https://github.com/0xMiden/compiler/blob/6e14a767d504be5d9bd47d0f316effe10e53bde4/sdk/stdlib-sys/src/intrinsics/word.rs) there are more constructors taking `[Felt; 4]`. Should they remain as is or can we at some point try to see if that could be optimized too? 

Do you mean `From` implementations? We can remove them, or use tuple there instead of an array.

> Would probably be a breaking change for all sdk users.

We are still at the stage where we can introduce breaking changes easily since we have only a few users.

bitwalker: The conversions probably don't matter as much here, i.e. it's fine to leave them IMO. However, we could investigate if tuples have less overhead - since they get compiled as structs, it's possible that the calling convention permits passing the individual fields by value - but I kinda expect it doesn't do that, and will be basically the same as using an array.

mooori: > Do you mean From implementations?

In addition there's also `Word::new(word: [Felt; 4])`

PR #895, opened 2026-01-23
debuginfo: Use effective spans on ub.unreachable

djolertrk:


frontend/wasm/src/module/func_translator.rs
```diff
@@ -268,6 +272,9 @@ fn parse_function_body<B: ?Sized + Builder>(
                     let line = loc.line.and_then(LineNumber::new).unwrap_or_default();
                     let column = loc.column.and_then(ColumnNumber::new).unwrap_or_default();
                     span = source_file.line_column_to_span(line, column).unwrap_or_default();
+                    if span != SourceSpan::default() {
```
bitwalker: For comparisons like this, it's preferable to use `SourceSpan::UNKNOWN` explicitly, or `span.is_unknown()` since it is clearer what the intent is. The former also avoids some function call overhead, though that's more of a bonus than the goal.

djolertrk: Addressed, thanks!

PR #896, opened 2026-01-23
feat: rework test-harness for integration tests

lima-limon-inc:
Tackles #981 

This PR aims to remove some functionality from the compiler's testing crate and rely on `miden-protocol`'s testing crate.
This includes the following changes:
- Removed the `create_note_from_package` function. This now relies on `miden-protocol`'s `NoteBuilder` (which got `Package` support added into it in this [PR](https://github.com/0xMiden/protocol/pull/2502)).
- Removed the `account_component_from_package` function which was replaced with `AccountoComponent::from_package`.
- Now, the `compile_rust_package` method calls `miden build` instead of the compiler's library.
- Removed `build_existing_basic_wallet_account_builder` function. `MockChainBuilder::add_existing_account_from_components` which was also added in [PR](https://github.com/0xMiden/protocol/pull/2502)
- Removed `NoteCreationConfig` since it was superseded by `NoteBuilder`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -318,14 +318,6 @@ jobs:
             cargo install cargo-make --force
           fi
 
-      # Some integration tests use midenup to compile the example projects
-      - name: Install midenup
-        run: |
-          if ! miden help 2>/dev/null; then
-            cargo install --git https://github.com/0xMiden/midenup
-            midenup init
-          fi
-
       - name: Test
         run: |
           cargo make test -E 'package(midenc-integration-network-tests)'
--- a/tests/integration-network/Cargo.toml
+++ b/tests/integration-network/Cargo.toml
@@ -13,7 +13,7 @@ publish = false
 
 [dependencies]
 # miden-client = { version = "0.14", features = ["std", "tonic", "testing"] }
-miden-client = { git = "https://github.com/0xMiden/miden-client", branch = "release/v0.14.0-beta" }
+miden-client = { git = "https://github.com/0xMiden/miden-client", branch = "release/v0.14.0-beta", features = ["std", "testing"] }
 miden-core.workspace = true
 miden-protocol = { workspace = true, features = ["std", "testing"] }
 miden-standards = { workspace = true, features = ["std"] }
--- a/tests/integration-network/src/mockchain/helpers.rs
+++ b/tests/integration-network/src/mockchain/helpers.rs
@@ -3,36 +3,33 @@
 use std::{future::Future, sync::Arc};
 
 use miden_client::{
-    account::{
-        component::{BasicWallet, InitStorageData},
-        AccountStorage, StorageSlotName,
-    },
+    Word,
+    account::component::{BasicWallet, InitStorageData},
     asset::FungibleAsset,
     auth::AuthSecretKey,
     crypto::FeltRng,
-    note::{
-        Note, NoteAssets, NoteMetadata, NoteRecipient, NoteScript, NoteStorage, NoteTag, NoteType,
-    },
+    note::{Note, NoteType},
     transaction::RawOutputNote,
-    Deserializable, Word,
 };
-use miden_core::{crypto::hash::Rpo256, Felt, FieldElement};
+use miden_core::Felt;
 use miden_integration_tests::CompilerTestBuilder;
 use miden_mast_package::Package;
 use miden_protocol::{
     account::{
         Account, AccountBuilder, AccountComponent, AccountComponentMetadata, AccountId,
-        AccountStorage, AccountStorageMode, AccountType, StorageMap, StorageMapKey, StorageSlot,
-        StorageSlotName,
+        AccountStorage, AccountStorageMode, AccountType, StorageSlot, StorageSlotName,
     },
     asset::Asset,
     note::PartialNote,
     transaction::{TransactionMeasurements, TransactionScript},
 };
-use miden_standards::account::interface::{AccountInterface, AccountInterfaceExt};
+use miden_standards::{
+    account::interface::{AccountInterface, AccountInterfaceExt},
+    testing::note::NoteBuilder,
+};
 use miden_testing::{MockChain, TransactionContextBuilder};
 use midenc_frontend_wasm::WasmTranslationConfig;
-use rand::{rngs::StdRng, SeedableRng};
+use rand::{SeedableRng, rngs::StdRng};
 
 /// Converts a value's felt representation into `miden_core::Felt` elements.
 pub(super) fn to_core_felts(value: &AccountId) -> Vec<Felt> {
@@ -156,7 +153,7 @@ pub(super) fn build_asset_transfer_tx(
     let output_note = NoteBuilder::new(sender_id, rng)
         .serial_number(serial_num)
         .package((*p2id_note_package).clone())
-        .note_inputs(to_core_felts(&recipient_id))
+        .note_storage(to_core_felts(&recipient_id))
         .unwrap()
         .add_assets([asset])
         .tag(0)
@@ -174,7 +171,7 @@ pub(super) fn build_asset_transfer_tx(
     let recipient_digest: [Felt; 4] = output_note.recipient().digest().into();
     commitment_input.extend(recipient_digest);
 
-    let asset_elements = miden_protocol::asset::Asset::from(asset).as_elements();
+    let asset_elements = asset.as_elements();
     commitment_input.extend(asset_elements);
     // Ensure word alignment for `adv_load_preimage` in the tx script.
     commitment_input.extend([Felt::ZERO, Felt::ZERO]);
@@ -239,13 +236,14 @@ pub(super) fn build_existing_counter_account_builder_with_auth_package(
     auth_storage_slots: Vec<StorageSlot>,
     seed: [u8; 32],
 ) -> AccountBuilder {
-    let supported_types = BTreeSet::from_iter([AccountType::RegularAccountUpdatableCode]);
+    let metadata =
+        AccountComponentMetadata::new("auth", [AccountType::RegularAccountUpdatableCode]);
     let auth_component = AccountComponent::new(
         auth_component_package.unwrap_library().as_ref().clone(),
         auth_storage_slots,
+        metadata,
     )
-    .unwrap()
-    .with_supported_types(supported_types);
+    .unwrap();
 
     AccountBuilder::new(seed)
         .account_type(AccountType::RegularAccountUpdatableCode)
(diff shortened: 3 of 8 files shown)


greenhat:
> I have just pushed the code that supports `#[miden_test(help)]`. When used, it will produce the following output:
> 
> ```rust
> --> tests/integration-network/src/mockchain/basic_wallet.rs:20:1
>  |
>  | #[miden_test(help)]
>  | ^^^^^^^^^^^^^^^^^^^
>  |
>  = help: message:
>    <documentation string>
> ```
> 
> Where `<documentation string>` is the documentation for every supported attribute (and its corresponding recognized fields). The list of supported struct is none other than all the enum variants present in `RecognizedAttrsBuilder`.
> 
> A nifty detail is that it uses each struct's doc-comments to generate the help message. Which means that it remains consistent and up to date with the generated `cargo-doc` webpage.

Nice! I have not thought about calling a macro to get help.

lima-limon-inc:
> (...)
> 
> I consider the following to be the extra price that the developers would pay for macros on top of Mockchain (which is not going anywhere):
> 
>     1. Figuring out, understanding and remembering the entirely new for them macro parameters system.
> 
>     2. No discoverability, no autocomplete, and no parameters help in an IDE.
> 
>     3. Limited LSP; no refactoring (renaming `alice_account` or `wallet_package` would not work), limited "find references" and "goto definition".
> 
>     4. Deciphering runtime errors encountered in the macros (wrong path, no .masp file, etc.).
> 
>     5. LLMs are struggling with the unknown macro params system.
> 
> With all that in mind I think that we should consider that macros might not be worth pursuing and we might need to change the direction. The alternatives I see:

I see, those are some great points! Thank you very much for the thorough explanation :blush: @greenhat 

>
>     1. Some hybrid (declarative/imperative) API on top of MockChain. @bitwalker Like the original `Scenario` but with imperative escape hatches.

I have been working on a rework considering your comments and also taking as inspiration the [`Scenario`](https://github.com/0xMiden/compiler/commit/263ae570) struct that you mentioned.

Firstly, the `Scenario` struct is broken into multiple "steps" structs (à la typestate pattern). Every step struct has a `next` method which turns one struct into the next. These can be chained to form fairly idiomatic code, like so:

``` rust
    let scenario = ScenarioBuilder::new()
        .next()
        .add_package("../../examples/basic-wallet", "wallet")
        .add_package("../../examples/p2id-note", "p2id-note")
        .add_package("../../examples/basic-wallet-tx-script", "tx-script")
        .next()
        .add_faucet(FaucetAccountConfig::default().with_alias("faucet"))
        (...)
```

Currently there are 4 of these: `ScenarioBuilder` (the entry point), `ScenarioPackageSetup` (used to compile packages and save them under an alias), `ScenarioAccountSetup` (adds accounts to the mockchain) and `Scenario` final step where the `MockChainBuilder` is built.
In case a more detailed/peculiar scenario needs to be tested, every one of these structs contains an escape hatch to the underlyling `MockChainBuilder`. Thus, if needed, at any point one can access it via `get_mock_chain(&mut self) -> &mut MockChainBuilder`, and manipulate it as needed. These two mechanisms are not mutually exclusive,since the `Scenario` structs are wrappers around the actual `MockChainBuilder`.

Additionally, similar to the previous macro PR, we take a "default if omitted" approach: if a value is not specified, we use a "sane" default. This is achieved with the `WalletAccountConfig` and `FaucetAccountConfig` structs.

I have just pushed an initial implementation of this approach, and used it in the `test_basic_wallet_p2id` test to showcase what it might look like. Here's what it currently looks like: https://github.com/0xMiden/compiler/pull/896/changes/133c6691be265495dd909e64e5080dc0228b485f#diff-4396fed87abb04dd237f9628c6d8d2f9a7a1cd701e57248648f1695552514b6cR22-R33.

What do you think of this new approach? Does it better tackle your points? Or should I change course in some/all aspects? Thanks in advance!

Would love to hear your thoughts! 😊

tests/integration-network/src/mockchain/helpers.rs
```diff
@@ -360,3 +391,131 @@ pub(super) fn build_counter_account_with_rust_rpo_auth(
 
     (account, secret_key)
 }
+
+pub trait ComponentFromProject {
+    fn build_project(project_path: &str) -> Arc<Package> {
+        let output = std::process::Command::new("miden")
```
greenhat: Let's check if `miden` is available and if not, print an instruction (or link to) on how to install it.

lima-limon-inc: > Let's check if `miden` is available and if not, print an instruction (or link to) on how to install it.

Great point, I'll add a validation!

Edit: here's the validation: https://github.com/0xMiden/compiler/pull/896/commits/c1f3086b391f494934483bbfcb933cab6db6c431

tests/integration-network/Cargo.toml
```diff
@@ -12,7 +12,7 @@ edition.workspace = true
 publish = false
 
 [dependencies]
-miden-client = { version = "0.13", features = ["std", "tonic", "testing"] }
+miden-client = { branch = "fabrizioorsi/miden-testing-compiler-backport", git = "https://github.com/lambdaclass/miden-client.git", features = ["std", "tonic", "testing"] }
```
lima-limon-inc: This branch contains the changes made in [this PR](https://github.com/0xMiden/protocol/pull/2502), only backported to version 0.13 of the miden client.

.github/workflows/ci.yml
```diff
@@ -317,6 +317,14 @@ jobs:
           if ! cargo make --version 2>/dev/null; then
             cargo install cargo-make --force
           fi
+
+      # Some integration tests use midenup to compile the example projects
+      - name: Install midenup
+        run: |
+          if ! miden help 2>/dev/null; then
+            cargo install --git https://github.com/0xMiden/midenup
+          fi
```
lima-limon-inc: Currently, accessing the `miden` interface requires further intervention by adding some variables to the system's `$PATH`. This is not ideal for cases like this I believe.

lima-limon-inc: I'll create an issue on `midenup` to track this: https://github.com/0xMiden/midenup/issues/165

tests/integration-network/src/mockchain/helpers.rs
```diff
@@ -51,105 +50,45 @@ pub(super) fn block_on<F: Future>(future: F) -> F::Output {
...
 // ================================================================================================
 
-/// Helper to compile a Rust package to a Miden `Package`.
-pub(super) fn compile_rust_package(package_path: &str, release: bool) -> Arc<Package> {
-    let config = WasmTranslationConfig::default();
-    let mut builder = CompilerTestBuilder::rust_source_cargo_miden(package_path, config, []);
-
-    if release {
-        builder.with_release(true);
+pub(super) fn compile_rust_package(project_path: &str, release: bool) -> Arc<Package> {
```
greenhat, line 55: This function is intended to be used by the users (smart contract developers in their tests). And I believe I saw some discussion of how to implement the package compilation in the `miden-testing`. Or am I missing something?

@lima-limon-inc For the compiler test suite we should use the previous version of the `compile_rust_package` that calls `CompilerTestBuilder`. This should eliminate the need for the CI job above.

lima-limon-inc: > And I believe I saw some discussion of how to implement the package compilation in the `miden-testing`. Or am I missing something?

If I'm not mistaken, the only mention of package compilation in `miden-testing`/`miden-protocol` was this comment by @PhilippGackstatter in https://github.com/0xMiden/compiler/issues/981#issuecomment-3949988737.

> PhilippGackstatter said:
> 
> I'm not sure what this will look like, but if these are compiler specifics, isn't this better suited to be an extension trait of MockChainBuilder instead of integrating the need for midenup into miden-testing (but maybe I'm imagining the wrong thing)?

Edit: I've adressed your comment in this commit: https://github.com/0xMiden/compiler/pull/896/commits/4c3d1ea13d8f26a04ba0b500deb5f650ee9cb8d8. 

As an aside, currently the `Cargo.toml` file points to a branch I created as a backport. I believe that we can update it as soon as this PR https://github.com/0xMiden/protocol/pull/2502 gets merged, right?

.github/workflows/ci.yml
greenhat: I don't think we need it anymore after we switched to using `cargo-miden` lib for package compilation.

tests/integration-network/Cargo.toml
greenhat: This is tricky. https://github.com/0xMiden/protocol/pull/2502 will end up in the next branch and in the v0.14 version. The migration to v0.14 is done in #995 so we need to rebase the this branch onto #995 after https://github.com/0xMiden/protocol/pull/2502 is merged.

tests/integration-network/src/mockchain/helpers.rs
greenhat: > > And I believe I saw some discussion of how to implement the package compilation in the `miden-testing`. Or am I missing something?
> 
> If I'm not mistaken, the only mention of package compilation in `miden-testing`/`miden-protocol` was this comment by @PhilippGackstatter in [#981 (comment)](https://github.com/0xMiden/compiler/issues/981#issuecomment-3949988737).

Thanks. I made https://github.com/0xMiden/protocol/issues/2611 for the it.

> Edit: I've adressed your comment in this commit: [4c3d1ea](https://github.com/0xMiden/compiler/commit/4c3d1ea13d8f26a04ba0b500deb5f650ee9cb8d8).

Thanks!

> As an aside, currently the `Cargo.toml` file points to a branch I created as a backport. I believe that we can update it as soon as this PR [0xMiden/protocol#2502](https://github.com/0xMiden/protocol/pull/2502) gets merged, right?

Well it's a bit more tricky. See https://github.com/0xMiden/compiler/pull/896/changes#r2944522824

greenhat (changes requested):
Almost there! Just a minor note about the new CI job.

greenhat:
@lima-limon-inc After this branch is merged please do https://github.com/0xMiden/project-template/issues/27.

tests/integration-network/Cargo.toml
lima-limon-inc: > This is tricky. [0xMiden/protocol#2502](https://github.com/0xMiden/protocol/pull/2502) will end up in the next branch and in the v0.14 version. The migration to v0.14 is done in [#995](https://github.com/0xMiden/compiler/pull/995) so we need to rebase the this branch onto [#995](https://github.com/0xMiden/compiler/pull/995) after [0xMiden/protocol#2502](https://github.com/0xMiden/protocol/pull/2502) is merged.


Noted! With [0xMiden/protocol#2502](https://github.com/0xMiden/protocol/pull/2502) now having being merged, I'll rebase this pr into `pr/migrate-to-v0.21.1`. I'll push the changes ASAP.

PR #902, opened 2026-01-28
test(sdk): setup negative tests for macro expansion

### LINKED ISSUE #900: test: Setup negative macro tests
### What should be done?

Add tests that use macros in a wrong way and verify macro expansion throws the expected error.

### How should it be done?

- easy to use
- be documented / have example

### When is this task done?

Negative tests can be added.

### Additional context

A use case will be e.g. #803: It can use such tests to verify having a field with non-whitelisted type in a `#[component]` struct throws the expected error.

mooori:
Closes #900

The diff contains a `README.md` which describes the setup and workflow.

### Implementation details

- Uses plain `cargo/rustc` instead of the miden toolchain or the compiler test harness.
  - To simplify it, skips `#![no_std], ...`

This is based on the assumption that these tests should just deal with macro expansion, not the steps following it.

### DIFF
--- a/sdk/base-macros/tests/README.md
+++ b/sdk/base-macros/tests/README.md
@@ -8,9 +8,8 @@ When tests are run, the Rust source files in `./errors/` are compiled. For `./er
 
 **Adding new source files**:
 
-- Add `./errors/your_file.rs` and run tests. This will fail and produce `wip/your_file.stderr`.
-- Inspect `your_file.stderr`. If it contains the expected error, move it next to `your_file.rs`.
-- Run tests again, now they should pass.
+- Add `./errors/your_file.rs` and run tests with the environment variable `TRYBUILD=overwrite`.
+- This adds `your_file.stderr`. Inspect it and commit if it contains the expected error.
 
 **Updating `stderr` files**:
 


sdk/base-macros/tests/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+These tests verify that incorrect macro usage leads to the expected errors during expansion.
+
+When tests are run, the Rust source files in `./errors/` are compiled. For `./errors/foo.rs`, the expected compilation error is contained in `./errors/foo.stderr`.
+
+## Workflow
+
+**Adding new source files**:
+
+- Add `./errors/your_file.rs` and run tests. This will fail and produce `wip/your_file.stderr`.
```
greenhat: I'd personally prefer the "alternative" (from https://github.com/dtolnay/trybuild):

>Alternatively, run cargo test with the environment variable TRYBUILD=overwrite to skip the wip directory and write all compiler output directly in place. You'll want to check git diff afterward to be sure the compiler's output is what you had in mind.

We should at least mention it here.

sdk/base-macros/tests/README.md
```diff
@@ -0,0 +1,26 @@
+# Negative macro expansion tests
+
+These tests verify that incorrect macro usage leads to the expected errors during expansion.
```
greenhat, line 3: I think we should not limit ourselves only to the negative tests. I'm pretty sure we'd use `t.pass` as well.

greenhat (approved):
Nice job! Only a couple of nits.

sdk/base-macros/tests/README.md
mooori: True, changed it to recommend that way.

sdk/base-macros/tests/README.md
mooori: Yes it would be nice to use `t.pass` as well! I've tried it and unfortunately things get tricky here.

When `t.compile_fail` encounters an error during macro expansion, it stops there. So it's possible to use plain `cargo/rustc`.

In contrast, for `t.pass` trybuild expects to compile all the way successfully. It only has a [minimal API](https://docs.rs/trybuild/1.0.114/trybuild/struct.TestCases.html#method.new) and doesn't allow setting `RUSTFLAGS` *only* for some tests ([ref](https://github.com/dtolnay/trybuild/issues/183), [ref](https://github.com/dtolnay/trybuild/issues/315)).

So it seems like there's no trivial way to set [any of these flags](https://github.com/0xMiden/compiler/blob/63ef6d0840d540ba28f7c27565212c97fd7314f4/tests/integration/src/compiler_test.rs#L192-L205) only for trybuild related tests. But without it I suppose account/contract related code can't compile?

Given this situation, maybe let's stick with existing test setup for the happy-path and use trybuild only for negative macro expansion tests?

greenhat: Oh, that is unfortunate. Let's stick with the negative tests then.

bitwalker (changes requested):
We should use lit tests for this, like so:

1. Add a new subdirectory to `tests/lit` called `sdk`
2. Add the following content to `tests/lit/sdk/lit.suite.toml`:
    ```toml
    name = "sdk"
    patterns = ["*.rs"]
    working_dir = "../../../"
      
    [format.shtest]
    ```
3. Add the following content to `tests/lit/sdk/component-macro-invalid-argument-error.rs`:
    ```rust
    //! ```cargo
    //! [dependencies]
    //! miden-base-macros = { path = "sdk/base-macros" }
    //! ```
    //!
    //! RUN: cargo +nightly -Zscript %s | filecheck %s
    use miden_base_macros::component;

    #[component(foo)]
    struct Contract;
 
    fn main() { }

    // CHECK: error: #[component] does not accept arguments
    // CHECK-NEXT: --> {{.*\.rs}}:[[@LINE-6]]:1
    // CHECK-NEXT:             |
    // CHECK-NEXT: [[@LINE-8]] | #[component(foo)]
    // CHECK-NEXT:             | ^^^^^^^^^^^^^^^^^
    ```
4. Add the `tests/lit/sdk` directory to the set of directories in the `test-lit` task in `Makefile.toml`

Adding new tests can then be done by simply adding another `.rs` file to that directory with the appropriate checks.

There a couple of reasons to prefer this approach IMO:

* We already have lit/filecheck as part the compilers test tooling (via `litcheck`)
* We avoid adding another way of defining and running tests. We already have quite a variety of different tools in this department, and I'd like to keep things simpler
* By using `lit`, we can test for just the parts of the output that we actually care about (i.e. we aren't using stderr snapshots, which are brittle). We can also express relationships between the source code and the output in ways that are not possible with other test tools
* It provides a helpful example for using lit-style tests for user-facing diagnostics and output, which will aid those who wish to contribute such tests in the future.

What are your thoughts @greenhat / @mooori?

sdk/base-macros/tests/README.md
bitwalker: This is solvable using lit-style tests - another reason to go that route for this sort of thing IMO

bitwalker:
We should merge #905 before this, that way this PR can build on the new lit testing structure and CI

PR #905, opened 2026-01-30
test: clean up lit/filecheck tests, run them in ci

bitwalker:
* Separates the lit test task out of the default "test" task
* Adds new test-all task to run both Rust and lit tests
* Adds a new CI step to run the lit tests if all Rust unit tests pass
* Cleans up old/outdated lit tests
* Reorgnize existing lit tests into a better structure that lets them all be run with a single `lit run -v tests/lit` command
* Configure substitutions in lit test suite so that the working directory can be outside the workspace directory, to keep build artifacts out of the source tree.
* Add a new Makefile task for running lit manually


djolertrk (approved):
Nice! LGTM. Thanks.

Makefile.toml
```diff
@@ -31,7 +31,29 @@ MIDENC_BUILD_PROFILE = "debug"
 MIDENC_BUILD_PROFILE = "release"
 
 [tasks.init]
-run_task = "print-env"
+script = ['''
+#!@duckscript
```
greenhat, line 35: Any particular reason to prefer duckscript over bash(awk/sed)?

bitwalker: It works cross-platform, whereas `sh` is only going to work on *NIX systems. If we want to support Windows builds, we can't be using `bash`/`sh` scripts.

I'm not a fan of it personally, but it works well enough for the few little things we need it for in our `Makefile.toml`. If I had my wish, they'd have baked in a Lua interpreter instead, but it is what it is I guess.

PR #908, opened 2026-02-02
test(sdk): setup negative lit tests for macro expansion

### LINKED ISSUE #900: test: Setup negative macro tests

### LINKED ISSUE #909: build: reduce rebuild times of lit sdk tests
### What should be done?

The sdk lit tests introduced in #908 trigger time intensive rebuilds even after changing a comment in files in `tests/lit/sdk/*.rs`. This slows down working on these tests significantly.

This seems to be expected behavior for `-Zscript`: IIUC it uses a target directory derived from the file hash. So any changes triggers a full rebuild, including all the `miden-*` crates, which takes quite some time.

### How should it be done?

Some ideas:

- Avoid using `-Zscript` and embedded manifests.
  - Perhaps by making `tests/lit/sdk` a crate and place tests in different bin targets, i.e. `tests/lit/sdk/bin/*.rs`. Then change the run command to something like `cargo check --bin ...`.
- Make sdk lit tests use a `CARGO_TARGET_DIR` independent of the file hash. Though there seems to be no clean way to do that only for a subdirectory of a workspace, given that lit tests are executed from the workspace root.

### When is this task done?

Changing lit checks doesn't trigger rebuilding all dependencies.

### Additional context

_No response_

mooori:
Closes #900.
Closes #909

Reworks to #902 to use lit tests as outlined in [this comment](https://github.com/0xMiden/compiler/issues/909#issuecomment-3836268384).

### Test location

Places sdk tests outside of the sdk's crate. I suppose sdk crates aren't released without testing the whole workspace, so this shouldn't be an issue.

### DIFF
--- a/codegen/masm/src/emit/mem.rs
+++ b/codegen/masm/src/emit/mem.rs
@@ -5,7 +5,7 @@ use midenc_hir::{
 };
 
 use super::{OpEmitter, masm};
-use crate::lower::NativePtr;
+use crate::{OperandStack, lower::NativePtr};
 
 /// Allocation
 impl OpEmitter<'_> {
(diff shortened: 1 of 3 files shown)


tests/lit/sdk/sdk-project/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "sdk-project"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+miden-base-macros = {path = "../../../../sdk/base-macros"}
```
mooori, line 7: Using a path like `sdk/base-macros` hasn't worked for me.

tests/lit/sdk/component-macro-invalid-argument.stderr
```diff
@@ -0,0 +1,7 @@
+// RUN: env CARGO_TARGET_DIR=%target_dir/lit/sdk-project cargo run --bin component-macro-invalid-argument --manifest-path %S/sdk-project/Cargo.toml 2>&1 | filecheck %s
+
+// CHECK: error: #[component] does not accept arguments
+// CHECK-NEXT:  --> {{.*\.rs}}:{{[0-9]+}}:{{[0-9]+}}
+// CHECK-NEXT:   |
+// CHECK-NEXT: {{[0-9]+}} | #[component(foo)]
```
mooori: Does `litcheck` allow capturing a line number in 4 and then re-using it in 6?

bitwalker: If you want to reference the line number of the check file, you can get that with `[[@LINE]]`, and it supports offsets if you want to refer to lines around it, e.g. `[[@LINE - 2]]`.

If you want to capture some text from the input file as a variable, and then use it as a substitution pattern later, there are two types of variables: numeric and textual. The former supports various number formats and can be used in expressions - the latter is purely capturing text:

* Numeric variables, for a simple decimal number, can be captured with `[[#%d,VAR:]]`, and then used later in the file with `[[#VAR]]`
* String variables can be captured with `[[VAR:<pattern>]]` where `<pattern>` is a regular expression, and used later in the file with `[[VAR]]`

See https://llvm.org/docs/CommandGuide/FileCheck.html for more details on this. At some point I plan on adding my own documentation, because LLVM's is pretty spread out, but for now you can pretty much use the LLVM lit/FileCheck docs as-is.

tests/lit/sdk/sdk-project/Cargo.toml
bitwalker: This is likely due to the working directory of the suite being in a temporary directory by default, so I'm not sure if the path is relative to the Cargo manifest, or the working directory, but I suspect it may be the latter.

You can force Cargo to set its working directory with `-C PATH` (when building with nightly anyway, which we are), so from a `RUN` line you can do something like `cargo -C %S build ...` where `%S` is a substitution that `litcheck` will expand to the directory of the source file containing the `RUN` line.

Alternatively, one can define a nested `lit.suite.toml` in a directory, with overrides for specific options you want to change, such as `working_directory` - but then you may end up with transient artifacts in that directory, so use with care.

bitwalker: Looking at the path here, I think you are missing one `../`, it looks to me like this would try to expand to `<workspace root>/tests/sdk/base-macros`

mooori: > Using a path like sdk/base-macros hasn't worked for me.

Oh, I should have been more precise: It was meant to say a path relative to the workspace root like `sdk/base-macros` didn't work for me.

> Looking at the path here, I think you are missing one `../`

The current path `../../../../sdk/base-macros` works when I run `cargo make test-lit` from the workspace root. It's a bit lengthy and brittle, but I suppose these projects are rarely moved. The only less brittle option I can think of would be `sdk/base-macros` from above.

tests/lit/sdk/component-macro-invalid-argument.stderr
mooori: Thank you for the explanation. Used it to capture `ERRLINE` in 7302943.

tests/lit/sdk/sdk-project/Cargo.toml
bitwalker: Yeah there's not much that can be done about that aspect unfortunately, but as you noted, these rarely move, so it's not an issue in practice

PR #918, opened 2026-02-04
feat(sdk): check type of storage fields during macro expansion

### LINKED ISSUE #803: Allow only account storage fields in the account (`#[component]`) struct
### Why

Putting some non-storage (`u32`, or any other type) field in the account `struct` will surprise the user regarding the persistence. For example, writing a value in an account code and expecting the "stored" value when called from the script will not work.

### How

Forbid non-storage field types in the `#[component]` macro with an error message explaining why and providing tips.

mooori:
...to emit a user-friendly error, which is shown in the `stderr` file used in the lit test.

Closes #803 
Depends on #908 for testing


greenhat:
> There is another code path checking that a type is either `StorageMap` or `Value` [here](https://github.com/0xMiden/compiler/blob/7f372bb63b01fc1843291cd0d70e81ca789853e6/sdk/base-macros/src/account_component_metadata.rs#L61). Should this stay as is or should it also use `typecheck_storage_field`?

Good call! Let's use `typecheck_storage_field` there.

PR #922, opened 2026-02-08
wip: fix various compilation issues and codegen blockers

bitwalker:
There are still two broken tests related to operand stack overflow, but in general this branch addresses a lot of issues. I'm hoping to have the other two tests working by tomorrow sometime.

Both the failing tests are due to unscheduable operand stack problems. Both are caused by the presence of values assigned deep in a nested loop structure, and so far always unused in the loop itself.

This is an artifact of the structured control flow transformation needing to return these values as results from the loop operation, leading to an excessive number of operands passed and returned from each structured op and its terminators.

The best way to solve this IMO is to identify these values and convert them to load/stores to a local. This would remove the need to propagate them around, while preserving the semantics. It will also have the effect of improving the efficiency of the generated code.

I am going to prototype this tonight/tomorrow and follow up with my findings.

### DIFF
--- a/codegen/masm/Cargo.toml
+++ b/codegen/masm/Cargo.toml
@@ -42,5 +42,5 @@ thiserror.workspace = true
 # Use local paths for dev-only dependency to avoid relying on crates.io during packaging
 midenc-expect-test = { path = "../../tools/expect-test" }
 proptest.workspace = true
-midenc-log.workspace = true
+midenc-log = { path = "../../midenc-log" }
 itertools = "0.14.0"
--- a/dialects/hir/Cargo.toml
+++ b/dialects/hir/Cargo.toml
@@ -25,8 +25,8 @@ midenc-hir-transform.workspace = true
 
 [dev-dependencies]
 # Use local paths for dev-only dependency to avoid relying on crates.io during packaging
-midenc-dialect-scf.workspace = true
+midenc-dialect-scf = { path = "../scf" }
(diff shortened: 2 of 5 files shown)


bitwalker:
This PR fixes all outstanding operand stack overflow issues.

With the frontend changes made, we'll want to write an optimization pass to remove uses of procedure locals that can replaced with just pure SSA values, for example many of the instruction tests now emit a pair of stores immediately on entry, immediately followed by a pair of loads for the op being tested - these could be trivially erased so that the function arguments are used directly. This is basically a mem2reg-type transformation, but narrower, since we'd be focusing on procedure locals only.

dialects/hir/src/transforms/local2reg.rs
```diff
@@ -0,0 +1,233 @@
...
+    Backward, CallOpInterface, EntityMut, FxHashMap, Op, OperationName, OperationRef, ProgramPoint,
+    RawWalk, RegionBranchOpInterface, Report, Rewriter, SmallVec, Symbol, TraceTarget, ValueRef,
+    dialects::builtin::{Function, LocalVariable},
+    pass::{Pass, PassExecutionState, PostPassStatus},
+    patterns::{RewriterImpl, TracingRewriterListener},
+    traits::BranchOpInterface,
+};
+
+use crate::{LoadLocal, StoreLocal};
+pub struct Local2Reg;
```
greenhat, line 13: It'd be nice to have a description of what the `Local2Reg` pass does.

PR #928, opened 2026-02-10
fix(examples): remove unused dependency

mooori:
Removes compilation warnings due to `native_account` not being used (anymore).


mooori:
`CodeQL` action failed as it ran out of disk space, might be GH issue.

PR #934, opened 2026-02-10
perf: cache stack_size in ValueOrAlias to avoid repeated virtual dispatch

sashass1315:
`ValueOrAlias::stack_size()` called `self.value.borrow().ty().size_in_felts()` on every invocation — this involves dynamic borrow tracking (Cell read/write), vtable indirect call through `dyn Value`, and `BorrowRef` drop overhead. The method is called ~50-100 times per instruction in the operand movement solver and hundreds of times in spill analysis, making it a hot path.

Since the type (and therefore stack size) never changes after value creation, we cache `size_in_felts()` as a u8 field in `ValueOrAlias` at construction time — reusing the `borrow()` that already exists for caching `value_id`. The field fits in existing struct padding so `sizeof(ValueOrAlias) `stays the same (24 bytes). All 30+ call sites across solver and spill analysis benefit automatically.

### DIFF
--- a/hir/src/ir/value/aliasing.rs
+++ b/hir/src/ir/value/aliasing.rs
@@ -17,22 +17,30 @@ pub struct ValueOrAlias {
     /// on the operand stack as a unique value. The alias identifier is usually generated from a
     /// counter of the unique instances of the value, but can be any unique integer value.
     alias_id: u8,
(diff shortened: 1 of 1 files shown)


bitwalker (approved):

bitwalker:
I'll look at the CI issue, seems like nextest made some kind of change that requires '--locked` now when installing

PR #938, opened 2026-02-11
chore: update new project git tags (with SDK v0.10)

greenhat:


tools/cargo-miden/src/commands/new_project.rs
```diff
@@ -13,13 +13,13 @@ use crate::template::{GenerateArgs, TemplatePath, generate};
...
 /// desired commit.
-const PROJECT_TEMPLATES_REPO_TAG: &str = "v0.26.0";
+const PROJECT_TEMPLATES_REPO_TAG: &str = "v0.27.0";
 
 /// The tag used in checkout of the new Miden project template (`cargo miden new <NAME>`)
 ///
 /// Before changing it make sure the new tag exists in the rust-templates repo and points to the
 /// desired commit.
-const MIDEN_PROJECT_TEMPLATE_REPO_TAG: &str = "v0.8";
+const MIDEN_PROJECT_TEMPLATE_REPO_TAG: &str = "v0.9";
```
mooori, line 22: Could you point me to that tag? Just for my understanding of how things are connected. On a quick search in [rust-template](https://github.com/0xMiden/rust-templates) I couldn't find it.

greenhat: Sure. It's the second commit in https://github.com/0xMiden/project-template/pull/16/commits.

greenhat: Oh, it's in a different repo. The one where the "workspace" lives.

mooori: > Sure. It's the second commit in https://github.com/0xMiden/project-template/pull/16/commits

The commit bumps `9->10` vs the diff in the PR here `8->9`. Seems like it was 1 behind before too, so I suppose it stays consistent

greenhat: Yeah, the SDK versioning differs from the git tags in templates. The git tags have to be bumped every time we change the templates so they run away faster.

PR #944, opened 2026-02-11
Trigger v0.7 compiler release (take 3)

greenhat:


greenhat:
I botched adding the runner cleanup in #943 and made #945 to fix it

PR #959, opened 2026-02-13
test: add tests for `overflowing` arith

### LINKED ISSUE #912: Test integer checked and overflowing arith ops
From https://midengroup.slack.com/archives/C08U8ANH5N3/p1770050619085179

Our test suite for arith ops at  
https://github.com/0xMiden/compiler/blob/ccdb5639a7969d9e98ab801d383ea3e38243ac0d/tests/integration/src/rust_masm_tests/instructions.rs?plain=1#L232-L298
should cover checked and overflowing ops as well.

mooori:
Addresses the `overflowing` part of #912.

Even though it doesn't fully resolve #912, I'd like to put this PR up for review to get feedback on the approach of using a generic function instead of macros.

### Generated files in `expected/`

Should they be generated and committed for these tests? For now I did that to be consistent with existing tests.

For passing tests I did not review these files, assuming matching `rust_out` and `vm_out` is sufficient. Though the generated files can be very helpful to get an idea of generated code and inspect it in case of failures. So another option could be writing these files only if specifically requested or when an error occurs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,7 +383,7 @@ dependencies = [
 
 [[package]]
 name = "cargo-miden"
-version = "0.7.0"
+version = "0.7.1"
 dependencies = [
  "anyhow",
  "cargo_metadata",
@@ -2540,7 +2540,7 @@ dependencies = [
 
 [[package]]
 name = "miden-integration-tests"
-version = "0.7.0"
+version = "0.7.1"
 dependencies = [
  "anyhow",
  "blake3",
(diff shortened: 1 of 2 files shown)


greenhat:
I bet the reason is `rodata` being different due to the file path difference that is in it - https://github.com/0xMiden/compiler/actions/runs/22096960736/job/63856846720?pr=959#step:7:816. We fix it with `trim-path` cargo option (see example at https://github.com/0xMiden/compiler/blob/0570d5919e27ae072d148c515e1c418038fb8cd5/tests/integration/src/compiler_test.rs?plain=1#L580) but in this case you're using `CompilerTest::rust_fn_body` that calls rustc directly. So you can either find `trim-path` equivalent in rustc options or switch to `CompilerTest::rust_fn_body_with_*` that uses cargo.

Issue #955, opened 2026-02-12
Release job creates a new release PR after merging the release PR


greenhat:
The #953 PR was (erroneously) created by the https://github.com/0xMiden/compiler/actions/runs/21938641353/job/63358459348#step:5:708 job triggered by the merging of #952.

It was the second time. Before that, the merging of #947 triggered the release job that created #948.

Issue #954, opened 2026-02-12
Release build artifacts failed to upload


greenhat:
https://github.com/0xMiden/compiler/actions/runs/21938641353/job/63359786464

I suspect the git tag. In the job log:
```
  gh release upload ${RELEASE_TAG} midenc-aarch64-apple-darwin
  gh release upload ${RELEASE_TAG} cargo-miden-aarch64-apple-darwin
  shell: /bin/bash -e {0}
  env:
    RELEASE_TAG: midenc-v0.7.0
```

But the actual git tag is `0.7.0` - https://github.com/0xMiden/compiler/releases/tag/0.7.0

PR #974, opened 2026-02-19
Trigger v0.7.1 compiler release

greenhat:
To test fixes for #954 and #955

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -78,7 +78,7 @@ jobs:
           echo "${RELEASES}" | jq -rM
           echo "=================="
           # `cargo-miden` owns the GitHub release/tag (see `release-plz.toml`), and it uses an
-          # unprefixed tag name like `0.7.0`. Other crates may report prefixed tags (e.g.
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -77,9 +77,12 @@ jobs:
           echo "=================="
           echo "${RELEASES}" | jq -rM
           echo "=================="
-          release_tag=$(echo "${RELEASES}" | jq -r '.[] | select(.package_name == "midenc" or .package_name == "cargo-miden") | .tag' | head -n1)
+          # `cargo-miden` owns the GitHub release/tag (see `release-plz.toml`), and it uses an
+          # unprefixed tag name like `v0.7.0`. Other crates may report prefixed tags (e.g.
+          # `midenc-v0.7.0`) which do not correspond to an actual GitHub release.
+          release_tag=$(echo "${RELEASES}" | jq -r '.[] | select(.package_name == "cargo-miden") | .tag' | head -n1)
```
mooori, line 83: Might add a sorting step if there's a chance of having multiple matches for the `select`.

greenhat: Good point! It should be only one package named `cargo-miden`.

Issue #971, opened 2026-02-17
Add `wasm` dialect


mooori:
...as described in [this comment](https://github.com/0xMiden/compiler/issues/960#issuecomment-3898026055).

Use the `wasm` dialect to resolve:

- [x] #960 
- [ ] #962 
- [x] #966 
- [x] #967

Aims of adding the `wasm` dialect are:

- Address correctness issues, for example the ones linked above.
- Generate more efficient Masm: currently generated Masm can be suboptimal if the existing dialects are not well suited to represent Wasm semantics.

bitwalker:
Implemented in #986

PR #986, opened 2026-02-24
feat: add  wasm dialect

mooori:
Sets up the `wasm` dialect and adds the `I32Extend8S` op to it. See #971 for context and motivation.

### DIFF
--- a/codegen/masm/src/lower/lowering.rs
+++ b/codegen/masm/src/lower/lowering.rs
@@ -1349,12 +1349,11 @@ impl HirLowering for builtin::GlobalSymbol {
 
 impl HirLowering for wasm::I32Extend8S {
     fn emit(&self, emitter: &mut BlockEmitter<'_>) -> Result<(), Report> {
-        // Modify the logical stack without emitting ops, so `sext` treats the operand as I8.
-        let operand = emitter.stack.pop().expect("operand stack is empty");
-        assert_eq!(operand.ty(), Type::I32);
-        emitter.stack.push(Type::I8);
-
+        // We're sign-extending an I8 contained in an I32 operand. Wasm does not specify
+        // the contents of the upper 24 bits. However the `sext` instruction requires them to be
+        // zero, so we truncate to meet that requirement.
         let mut inst_emitter = emitter.inst_emitter(self.as_operation());
+        inst_emitter.trunc(&Type::I8, self.span());
         inst_emitter.sext(&Type::I32, self.span());
 
         Ok(())
--- a/tests/integration/src/codegen/wasm.rs
+++ b/tests/integration/src/codegen/wasm.rs
@@ -71,6 +71,7 @@ fn test_i32_extend8_s() {
             &[Felt::from(input)],
             context.session(),
             |trace| {
+                dbg!(format!("{:b}", input));
                 let outputs = trace.outputs().as_int_vec();
                 assert_single_output(expected_out as u64, outputs);
                 Ok(())
@@ -87,5 +88,5 @@ fn assert_single_output(expected: u64, outputs: Vec<u64>) {
         "expected all elements after first to be zero, but got {outputs:?}"
     );
 
-    assert_eq!(outputs[0], expected);
+    assert_eq!(outputs[0], expected, "actual {:#b}, expected {:#b}", outputs[0], expected);
 }


dialects/wasm/src/ops.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[operand]
+    operand: AnySignedInteger,
+    #[result]
+    result: AnySignedInteger,
+}
+
+has_no_effects!(I32Extend8S);
+
+impl InferTypeOpInterface for I32Extend8S {
+    // TODO type check via trait based mechanisms during op construction instead of here
```
mooori: Should we allow operands smaller than 32 bits? Can't happen when translating from Wasm, but it might be handy if the Op is constructed in some other way.

codegen/masm/src/lower/lowering.rs
```diff
@@ -1345,3 +1346,12 @@ impl HirLowering for builtin::GlobalSymbol {
         Ok(())
     }
 }
+
+impl HirLowering for wasm::I32Extend8S {
+    fn emit(&self, emitter: &mut BlockEmitter<'_>) -> Result<(), Report> {
+        let mut inst_emitter = emitter.inst_emitter(self.as_operation());
+        inst_emitter.trunc(&Type::I8, self.span());
```
bitwalker, line 1356: I think you can replace this with just a simple stack pop (with an assert that the type of the popped value was I32) + `push(Type::I8)` so that the operand stack state holds the expected types, but we don't actually emit any code for the type change, since semantically we don't need to truncate, you're just doing that here to get the types to line up.

dialects/wasm/src/ops.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn infer_return_types(&mut self, _context: &Context) -> Result<(), Report> {
+        let operand_ty = self.operand().as_value_ref().borrow().ty().clone();
```
bitwalker: It's not necessary to check types like this here, the return type is always the same

dialects/wasm/src/ops.rs
bitwalker: We should require the input to be I32 always for Wasm ops that operate on 32-bit values

dialects/wasm/src/ops.rs
```diff
@@ -0,0 +1,92 @@
...
+// https://github.com/WebAssembly/spec/blob/main/proposals/sign-extension-ops/Overview.md
+// TODO implement eval?
+#[operation(
+    dialect = WasmDialect,
+    traits(UnaryOp),
+    implements(UnaryOp, InferTypeOpInterface, MemoryEffectOpInterface, Foldable)
+)]
+pub struct I32Extend8S {
+    #[operand]
+    operand: AnySignedInteger,
```
bitwalker: This type constraint should be I32, same with the result type constraint

codegen/masm/src/lower/lowering.rs
```diff
@@ -1345,3 +1346,14 @@ impl HirLowering for builtin::GlobalSymbol {
         Ok(())
     }
 }
+
+impl HirLowering for wasm::I32Extend8S {
+    fn emit(&self, emitter: &mut BlockEmitter<'_>) -> Result<(), Report> {
+        let mut inst_emitter = emitter.inst_emitter(self.as_operation());
+        let operand = inst_emitter.pop().expect("operand stack is empty");
```
bitwalker: You need to manipulate the logical operand stack if the emitter directly here (i.e. via `stack_mut`), or this will emit actual stack manipulation instructions, which we don't want in this case.

greenhat (commented):
Besides Paul's notes LGTM!

codegen/masm/src/lower/lowering.rs
mooori: Changed it to modify the `OperandStack` in 5362735.

mooori: Actually sign extension might not work as expected when changing the type to I8 only on the logical stack as in 5362735. Since there are multiple things in flight, it's getting difficult to keep tabs on everything. So I would like to suggest the following:

1. For now, land the initial wasm dialect with `inst_emitter.trunc(I8)` followed by `inst_emitter.sext(I32)` in the `I32Extend8S` lowering. As I understand it, that's inefficient but correct.
2. Open a follow-up issue to optimize the `wasm::I32Extend8S` lowering by getting rid of the `trunc`.
3. Fix the encoding of smallints: [miden-debug-i32](https://github.com/0xMiden/miden-debug/issues/31)

I've assembled these changes locally and then the overflowing arith tests pass for `i8` (along with all other tests). Proceeding from there would be easier, I think.

bitwalker: I'd like to try and pin down the cause of the issue before we merge - there shouldn't be a difference in observable behavior between emitting `trunc` and just "casting" the type of the value on top of the operand stack to i8 and sign-extending, if we assume that the lowering from the frontend is correct. The fact that something isn't working means there is a bug somewhere being papered over by emitting the `trunc` - I'd like us to to dig a bit deeper before we merge to figure that out.

mooori: @bitwalker could you take a look at the test added in e954808? I think it shows that:

- The lowering for `wasm::I32Extend8S` is correct, because `test_i32_extend8_s` passes when the lowering is implemented as follows with `trunc` instruction

```rust
impl HirLowering for wasm::I32Extend8S {
        let mut inst_emitter = emitter.inst_emitter(self.as_operation());
        inst_emitter.trunc(&Type::I8, self.span());
        inst_emitter.sext(&Type::I32, self.span());
        Ok(())
    }
}
```

- However `test_i32_extend8_s` fails when `trunc` is *not* emitted and instead only the type of the operand on the logical stack is modified, as in 5362735.

This should narrow down the problem to "something is wrong when only modifying the operand stack". I'll hold of opening an issue for it, as it's my first time using the `builder` pattern to construct a package.

mooori: It looks like the following is happening:

- The lowering of `I32Extend8S` goes through [`fn sext_smallint`](https://github.com/0xMiden/compiler/blob/3557bec3d225d4a2097e2fe160a0d3df892e0615/codegen/masm/src/emit/smallint.rs#L101). It assumes the operand is a valid 8-bit int (out of range bits are 0) and it's OR'ing with the input value.
- This may not always hold when going through `Rust + Wasm`, for instance because `i8` function arguments get sign-extended to `i32`, [see this issue](https://github.com/0xMiden/miden-debug/issues/31#issuecomment-3950679909).
- Wasm's `i32.extend8s` has an `i32` operand and just treats it's lowest 8 bits as `i8`, upper bits need not be zero.

So `i32.extend8_s` on operand `0b1_0000_0000` should result in `0`. However going through `sext_smallint` it results in `0b1_0000_0000`.

Thoughts regarding potential fixes:

- Modify `sext_smallint` to handle this
- Handle [Wasm's sign extension](https://github.com/WebAssembly/spec/blob/main/proposals/sign-extension-ops/Overview.md) differently

PR #987, opened 2026-02-27
feature: replaced `Value` with typed `Storage<T>` and `StorageMap` as `StorageMap<K, V>`

### LINKED ISSUE #789: Typed account storage definition and initialization
Account storage defined in Rust should produce a Wasm code that conveys type description, display format and initialization logic.
To achieve it, first we need to introduce typed storage, where a Rust type used for typed storage values and typed storage maps (key and value) needs to be (de)serializable to/from `Word`.

greenhat:
Close #789 

- `StorageMap` became `StorageMap<K,V>` and `Value` became `Storage<T>` where `K`, `V` and `T` have to be convertible to and from `Word`.

Here is an example:

```rust
/// An example account demonstrating storage value and map usage.
#[component]
struct MyAccount {
    /// Public key authorized to update the stored asset quantities.
    #[storage(description = "owner public key")]
    owner_public_key: Storage<Word>,

    /// A map from asset identifier to quantity held by the account.
    #[storage(description = "asset quantity map")]
    asset_qty_map: StorageMap<Asset, Felt>,
}
```

The `Word` convertibility is implemented with the `WordKey` and `WordValue` traits using the `TryFrom/TryInto` to highlight that the serialization is fallible. The error unwrapped in the account storage methods and surfaced as a panic to keep the account storage API simple.

### DIFF
--- a/tests/integration-network/src/mockchain/counter_contract.rs
+++ b/tests/integration-network/src/mockchain/counter_contract.rs
@@ -80,7 +80,7 @@ pub fn test_counter_contract() {
         .build_tx_context(counter_account.clone(), &[counter_note.id()], &[])
         .unwrap();
     let tx_measurements = execute_tx(&mut chain, tx_context_builder);
-    expect!["11916"].assert_eq(note_cycles(&tx_measurements, counter_note.id()));
+    expect!["17535"].assert_eq(note_cycles(&tx_measurements, counter_note.id()));
 
     // The counter contract storage value should be 2 after the note is consumed (incremented by 1).
     assert_counter_storage(
--- a/tests/integration-network/src/mockchain/counter_contract_no_auth.rs
+++ b/tests/integration-network/src/mockchain/counter_contract_no_auth.rs
@@ -104,7 +104,7 @@ pub fn test_counter_contract_no_auth() {
         .unwrap();
     let tx_measurements = execute_tx(&mut chain, tx_context_builder);
     expect!["2264"].assert_eq(auth_procedure_cycles(&tx_measurements));
-    expect!["11916"].assert_eq(note_cycles(&tx_measurements, counter_note.id()));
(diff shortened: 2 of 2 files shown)


bitwalker (approved):
I'm marking this approved, but I really think we should be making serialization code fallible - so far the existing serialization code has been using `From`/`Into`, and there's really no reason not to be using `TryFrom`/`TryInto` instead - deserialization especially is inherently fallible, and being able to handle that cleanly is important IMO.

Anyway, we have an open issue to move to `TryFrom` elsewhere, so I suppose we can add this to that issue, but we should probably stop adding to that technical debt sooner rather than later IMO.

examples/auth-component-rpo-falcon512/src/lib.rs
```diff
@@ -19,7 +19,7 @@ struct AuthComponent {
         description = "owner public key",
         type = "miden::standards::auth::falcon512_rpo::pub_key"
     )]
-    owner_public_key: Value,
+    owner_public_key: Storage<Word>,
```
PhilippGackstatter, line 22:
```suggestion
    owner_public_key: StorageValue<Word>,
```

Question: Should this indicate that the `Word` is stored in a value-type storage slot? I think this could be a bit more consistent with other slot types, e.g.:
- `StorageValue<AccountId>`
- `StorageMap<Word, Felt>`
- `StorageList<AccountId, 5>` (potential future type)

bobbinth: btw, looking at this now, I'm convinced that we should go with `StorageArray` rather than `StorageList`.

greenhat: I agree `StorageValue` sounds better. I made https://github.com/0xMiden/compiler/issues/1014 for it.

PR #988, opened 2026-02-27
Make deserialization return `Result` instead of panicking

### LINKED ISSUE #898: Use `TryFrom<Error = Infallible>` instead of `From` in deserialization derivation from felts
In general it will be infallible, but it still makes sense to encode that as `TryFrom<Error = Infallible>` then, just to make it clear that it can be fallible if you need it to be.

greenhat:
Close #898 

Use `TryFrom` instead of `From` in the `FromFeltRepr` `derive` macro.
Make `FeltReader::read()` return `FeltReprResult<Felt>` and `FromFeltRepr::from_felt_repr` to return `FeltReprResult<Self>`.

### DIFF
--- a/tests/integration-network/src/mockchain/basic_wallet.rs
+++ b/tests/integration-network/src/mockchain/basic_wallet.rs
@@ -87,7 +87,7 @@ pub fn test_basic_wallet_p2id() {
         chain.build_tx_context(alice_id, &[p2id_note_mint.id()], &[]).unwrap();
     let tx_measurements = execute_tx(&mut chain, consume_tx_context_builder);
     expect!["2983"].assert_eq(prologue_cycles(&tx_measurements));
-    expect!["23847"].assert_eq(note_cycles(&tx_measurements, p2id_note_mint.id()));
+    expect!["23965"].assert_eq(note_cycles(&tx_measurements, p2id_note_mint.id()));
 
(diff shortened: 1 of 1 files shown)


sdk/field-repr/repr/src/lib.rs
```diff
@@ -16,6 +16,104 @@ pub use miden_field_repr_derive::DeriveFromFeltRepr as FromFeltRepr;
 /// Re-export `DeriveToFeltRepr` as `ToFeltRepr` for `#[derive(ToFeltRepr)]` ergonomics.
 pub use miden_field_repr_derive::DeriveToFeltRepr as ToFeltRepr;
 
+/// Error returned when decoding a type from its felt representation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FeltReprError {
```
bitwalker, line 22: I'd mark this enum as `#[non_exhaustive]` so that we can add new variants to it without breaking downstream code.

bitwalker (approved):
LGTM - just one small tweak to the error type, but otherwise merge when ready!

sdk/field-repr/repr/src/lib.rs
greenhat: Sure. Done in https://github.com/0xMiden/compiler/pull/988/commits/b7bfccb276b41026f4a45adefe64210bbb2d13c5

PR #991, opened 2026-03-03
test: `Felt` constructor edge cases

### LINKED ISSUE #931: test: `Felt` construction edge cases
**Purpose**: Verify edge cases are handled correctly

**Approach**:

- In Rust: Construct `Felt` from `{Felt::M, Felt::M + 1, u64::MAX}`
- Compile to masm and execute it
- Check the result to verify conversion was handled correctly

mooori:
Closes #931

### DIFF
--- a/tests/integration/src/rust_masm_tests/constructors.rs
+++ b/tests/integration/src/rust_masm_tests/constructors.rs
@@ -1,0 +1,56 @@
+use miden_core::{Felt, StarkField};
+use miden_debug::ToMidenRepr;
+use midenc_frontend_wasm::WasmTranslationConfig;
+
+use crate::{CompilerTest, testing::eval_package};
+
+#[test]
+fn test_felt_construction_edge_cases() {
+    // Return `0` zero as sentinel value for failure of constructing the Felt.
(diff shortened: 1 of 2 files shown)


greenhat (changes requested):
Nice! I'm sorry I should have articulated it more, but after our "wrapping nature of the felt" discussion on slack I made the `Felt::new` on miden target in the `miden-crypto` to behave like the native felt (wrapping). So in the test we should check that `Felt:new` compiled to the miden target is semantically equivalent to the one compiled to the default target. However, this will be possible after we migrate to the VM v0.21 with the "unified" `Felt` from the `miden-crypto` repo.

bitwalker:
@greenhat @mooori What's the status of this PR?

greenhat:
In the upcoming `miden-field` the `Felt::new` contsructor becomes checkend again, so we should revieve this PR when we migrate to it. See https://github.com/0xMiden/crypto/blob/990e85fab4b843f154290c792a9995a59e254abe/miden-field/src/wasm_miden.rs?plain=1#L97-L99

mooori:
#931 is no longer relevant

PR #992, opened 2026-03-03
fix(test): sign-extend `i8, i16` for Wasm ABI

mooori:
This change allows unlocking overflowing arith tests for `i8`.

Depends on a change in `miden-debug` so I suppose it can be merged only after [this PR](https://github.com/0xMiden/miden-debug/pull/32) was merged.


bitwalker:
The latest version of `miden-debug` has been published to crates.io with the fix that unblocks this PR

Issue #982, opened 2026-02-24
Use MASM attribute to mark authentication procedure in the authentication component (instead of "auth_" prefix)


greenhat:
After https://github.com/0xMiden/protocol/issues/2501 is implemented.

Issue #111, opened 2024-01-29
Full Wasm CM types support


greenhat:
### Complete complex types support

Wasm CM types are translated to IR `Type` variants. Add corresponding IR `Type` variants for the missing Wasm CM [types](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#items-type) (`list`, `variant`, `enum`, `result`, `option`).

Implement the following Wasm CM types support:
* `enum`
* `result`
* `option`
* `flags`

PR #995, opened 2026-03-04
Migrate to VM v0.22 and protocol v0.14

### LINKED ISSUE #970: Migrate to VM v0.21

### LINKED ISSUE #1011: Rename `active_note::get_inputs` to `get_storage`
To be in line with MASM procedure names.
Defined at sdk/base-sys/src/bindings/active_note.rs

greenhat:
Close #970 
Close #1011 

This PR migrates the compiler and SDK stack to the newer Miden VM/runtime APIs, centered around `miden-vm` `0.22` and the corresponding protocol/base updates.

On the compiler side, this updates wasm lowering/codegen to match the new VM conventions, including the generalized sign-extension op, corrected little-endian limb handling for 64-bit values, and removal of legacy intrinsics in favor of the newer core/stdlib entry points.

The Miden SDK public API changes in this PR are:
- `Asset` changes from a single-word layout (`inner: Word`) to a two-word layout (`key: Word`, `value: Word`), and `Asset::new` now takes `(key, value)`.
- `asset::build_fungible_asset` / `asset::build_non_fungible_asset` are renamed to `asset::create_fungible_asset` / `asset::create_non_fungible_asset`.
- `active_note::get_inputs()` is renamed to `active_note::get_storage()`.
- `input_note::get_inputs_info()` is renamed to `input_note::get_storage_info()`, and `InputNoteInputsInfo` becomes `InputNoteStorageInfo { commitment, num_storage_items }`.
- `active_account` gains `get_asset` and `get_initial_asset` for direct vault lookups by asset key.
- `AccountId`, `Tag`, `NoteIdx`, and `NoteType` now convert to/from `Word`, and `StorageSlotId` adds `to_suffix_prefix()`.
- The top-level SDK re-export `Felt` and `Word` from `miden_field`(`miden-core`).

Triggered by the migration changes in this PR:
- Generalize wasm sign-extension lowering/codegen and add explicit coverage for the new `i32`/`i64` sign-extension cases.
- Fix 64-bit limb ordering in MASM memory/int codegen and related evaluation/codegen tests.
- Add the missing `cabi_realloc` runtime export required by generated component bindings when values are passed indirectly (Triggered by the `Asset` becoming two word size).

This PR is quite large. I suggest reviewing it on a per-commit basis, skipping not-important commits.

TODO:
- [x] switch to the published `miden-debug` 
- [ ] switch to the published `miden-client` in the network tests

`miden-clien` migration PR - https://github.com/0xMiden/miden-client/pull/1891

### DIFF
--- a/tests/integration/src/rust_masm_tests/rust_sdk/macros.rs
+++ b/tests/integration/src/rust_masm_tests/rust_sdk/macros.rs
@@ -27,8 +27,7 @@ fn component_macros_account_and_note() {
     };
 
     assert!(
-        panic_message.contains("not yet implemented")
-            && panic_message.contains("indirect pointer parameters"),
+        panic_message.contains("not yet implemented"),
         "unexpected panic message: {panic_message}"
     );
 


frontend/wasm/src/component/lift_exports.rs
```diff
@@ -30,14 +33,22 @@ pub fn generate_export_lifting_function(
...
-        flatten_function_type(&context, &export_func_ty, CallConv::CanonLift).map_err(|e| {
+        flatten_function_type(&context, &export_func_ty, CanonicalAbiMode::Lift).map_err(|e| {
             let message = format!(
                 "Component export lifting generation. Signature for exported function \
                  {core_export_func_path} requires flattening. Error: {e}"
             );
             diagnostics.diagnostic(Severity::Error).with_message(message).into_report()
         })?;
 
+    if cross_ctx_export_sig_flat.params().iter().any(|param| param.ty.is_pointer()) {
```
huitseeker, line 46: This turns a previously supported `#[component]` export shape into a hard failure IIUC?

frontend/wasm/src/component/flat.rs
```diff
@@ -40,6 +50,7 @@ pub fn flatten_type(context: &Rc<Context>, ty: &Type) -> Result<Vec<AbiParam>, C
         }
         Type::F64 => return Err(CanonicalTypeError::Reserved(ty.clone())),
         Type::Felt => vec![AbiParam::new(Type::Felt)],
+        Type::Enum(enum_ty) => flatten_type(context, enum_ty.discriminant())?,
```
huitseeker: Do flattening and wrapper generation now disagree on enums?

It looks like any import/export that contains an enum inside a transformed struct/result will get through flattening and then fail during wrapper generation. Should we either teach `load`/`store` to handle enums the same way, or keep rejecting enums here until that path exists?

codegen/masm/src/emit/unary.rs
```diff
@@ -106,8 +106,23 @@ impl OpEmitter<'_> {
         match (&src, dst) {
             // If the types are equivalent, it's a no-op, but only if they are integers
             (src, dst) if src == dst => (),
-            // Zero-extending a u64 to i128 simply requires pushing a 0u64 on the stack
-            (Type::U64, Type::U128 | Type::I128) => self.push_u64(0, span),
+            // Zero-extending a u64 to u128/i128 requires adding an extra 0u64 *above* the
+            // existing u64 in significance, i.e. below it on the operand stack (LE limb order).
```
bitwalker:
```suggestion
            // Zero-extending a u64 to u128/i128 requires us to provide the most-significant bits of the 
            // resulting 128-bit value. This is acheived by pushing a `0u64` value _below_ the current u64 
            // value on the operand stack, as the limb order of multi-limb integer values is little-endian

```

*NIt:* I find the wording of this awkward, this is how I'd phrase it instead

codegen/masm/src/lower/lowering.rs
```diff
@@ -470,7 +470,7 @@ impl HirLowering for hir::AssertEq {
 
 impl HirLowering for hir::Breakpoint {
     fn emit(&self, emitter: &mut BlockEmitter<'_>) -> Result<(), Report> {
-        emitter.emit_op(masm::Op::Inst(Span::new(self.span(), masm::Instruction::Breakpoint)));
+        emitter.emit_op(masm::Op::Inst(Span::new(self.span(), masm::Instruction::Nop)));
```
bitwalker: Let's remove this operation entirely from the `hir` dialect

frontend/wasm/src/component/flat.rs
```diff
@@ -12,6 +12,16 @@ use midenc_hir::{
     dialects::builtin::attributes::{AbiParam, Signature},
 };
 
+/// Identifies whether canonical ABI flattening is being performed for an exported component
+/// function or for a lowered import wrapper.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum CanonicalAbiMode {
```
bitwalker, line 17: I would clarify that what's actually happening here, IMO the terminology should work like this:

* `Lift` - means we are "lifting" a core Wasm function into the Canonical ABI (i.e. we're synthesizing a `CallConv::ComponentModel` function that will be exported, which internally calls the underlying function). The implementation of that function is "lowering" (flattening) arguments into the underlying ABI, and "lifting" results into the Canonical ABI.
* `Lower` - means we are "lowering" a Canonical ABI function into the underlying ABI (i.e. we're synthesizing a `CallConv::Wasm` function to call the Canonical ABI function). The implementation of that function is "lifting" arguments into the Canonical ABI, and "lowering" (flattening) results into the underlying ABI.

Better yet, it might make more sense to just change the names here to `Export` and `Import` to make it clearer what semantic action is being performed, and just document the WAT syntax it maps to (i.e. `(canon lift)` or `(canon lower)`).

Regardless, the terminology and semantics of this is very subtle, and I'd like to make it crystal clear what they map to.

frontend/wasm/src/component/flat.rs
bitwalker: Agreed, this flattens the enum type to its discriminant type here (so we should at least have `assert!(enum_ty.is_c_like())` to catch the as-yet-unimplemented full breadth of the `enum` type in the Canonical ABI) - but then the `load` and `store` implementations in `canon_abi_utils` return an error, rather than matching the behavior here (i.e. by assuming that the enum is C-like, and thus that we're loading/storing the corresponding discriminant type).

frontend/wasm/src/component/lift_exports.rs
bitwalker: I believe this is handling the case where specific heap-allocated types that are valid in the Canonical ABI (such as `list<T>`), but which we do not yet support, are being passed across the component boundary. In the Wasm Component Model, it is left to the Wasm runtime to copy those types from the memory of the caller component to the memory of the callee component - and for our purposes we must emit code which uses the advice provider for that. IIUC, this is just raising a diagnostic to notify the user that support for such types are not yet implemented.

I do think the diagnostic message is misleading though (unless I'm mistaken about why this check is being performed here - but we can't ever allow arbitrary references to caller memory to be passed across a component boundary, with the exception I noted above). It might be better to elaborate on what specifically the issue is (i.e. is it a matter of using something like `list<T>` when we haven't implemented support for that Canonical ABI type yet - or is it an attempt to pass a pointer across the component boundary, which we will never support)

frontend/wasm/src/component/lift_exports.rs
```diff
@@ -49,7 +60,8 @@ pub fn generate_export_lifting_function(
     // IMPORTANT: Restrict this rename to the authentication interface only.
     // We do this by matching the exact WIT name `auth-procedure` instead of
     // rewriting arbitrary names that merely start with `auth-`.
-    let export_func_ident = if export_func_name == "auth-procedure" {
+    let is_auth_procedure = export_func_name == "auth-procedure";
```
bitwalker, line 65: Maybe we should have the SDK macros emit some kind of attribute map in a custom section of the Wasm binary that the compiler can look for to transfer Rust attributes we care about to attributes in the IR, and ultimately to procedure attributes in MASM.

Something for a future PR in any case, just wanted to note it here

frontend/wasm/src/translation_utils.rs
```diff
@@ -121,6 +121,7 @@ pub fn emit_zero<B: ?Sized + Builder>(
         Type::U64 => builder.u64(0, SourceSpan::default()),
         Type::F64 => builder.f64(0.0, SourceSpan::default()),
         Type::Felt => builder.felt(Felt::ZERO, SourceSpan::default()),
+        Type::Enum(enum_ty) => emit_zero(enum_ty.discriminant(), builder, diagnostics)?,
```
bitwalker: We need to make sure we assert `is_c_like` on these everywhere that we handle them like this (i.e. by treating the enum as its discriminant type), so that we can catch any misses later when we flesh out `enum` support for the Canonical ABI

Cargo.toml
```diff
@@ -172,6 +170,13 @@ miden-test-harness-macros = { path = "test-harness/test-harness-macros" }
 #miden-processor = { path = "../miden-vm/processor" }
 #miden-mast-package = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
 #miden-mast-package = { path = "../miden-vm/package" }
+# miden-protocol = { git = "https://github.com/0xMiden/protocol", rev = "a53bbe2209f506df87876c8b9c9a1730214f456b" }
+# miden-standards = { git = "https://github.com/0xMiden/protocol", rev = "a53bbe2209f506df87876c8b9c9a1730214f456b" }
+miden-debug = { git = "https://github.com/0xMiden/miden-debug", rev = "b1bbfef59a08674d1803f5b7b7f2ce5c98490dfd" }
```
bitwalker: We can comment this back out in favor of the `0.6.0` release

greenhat: Thank you for the prompt release! Done in 3310e5b92df18f9ebbd52c61caeb81ac1c3f3df9

frontend/wasm/src/component/lift_exports.rs
greenhat: I changed the error message in da45c325aaf306a875faae2e6e5080e8364a66be. 
Yes, this is about passing the call parameters through the advice provider which we are not yet supporting.

frontend/wasm/src/component/lift_exports.rs
greenhat: Yes. This is my plan for https://github.com/0xMiden/compiler/issues/982.

frontend/wasm/src/component/flat.rs
greenhat: Good catch! Done in b45ba94bd4a1f7ddd1e21be67ae9b590426ef453. We have https://github.com/0xMiden/compiler/issues/111 where I plan to implement the enum support in full.

codegen/masm/src/emit/unary.rs
greenhat: Done in 9cf1a2795d6a57e905bcb17f8c2c0787ec5172cd

codegen/masm/src/lower/lowering.rs
greenhat: Done in c0909b788f2c5614e769b33e268e23c3dbfa06a7

frontend/wasm/src/component/flat.rs
greenhat: Thank you! Done in 721e8969b407732fe516444a7df7bced359606ca

frontend/wasm/src/translation_utils.rs
greenhat: Done in b45ba94bd4a1f7ddd1e21be67ae9b590426ef453

Issue #966, opened 2026-02-17
Masm for Rust's signed `overflowing_div` can produce wrong result


mooori:
### Packages versions

`next` branch

### Bug description

For example `(-1i8).overflowing_div(-2i8)`:

```
Rust result: (0, false)
Rust result as bytes: [0, 0] // expected
Vm out bytes: [1, 0] // actual
```

**Rust source**

```rust
pub fn f(a: i8, b: i8) -> (i8, bool) {
    a.overflowing_div(b)
}
```

**Potential explanation**

Based on the generated `test_overflowing_div_i8.hir`:

- The generated hir entrypoint has `i32` arguments for what are originally `i8` operands
- Perhaps the operands have been zero-extended instead of sign-extended
  - `-1i8 = 0xFF`: zext yields 255i32, sext yields -1i32
  - `-2i8 = 0xFE`: zext yields 254i32, sext yields -2i32
  - `255/254 = 1` instead of `-1/-2 = 0`

### How can this be reproduced?

- Checkout #959
- Unignore `test_overflowing_div_i8` in `rust_masm_tests/instructions.rs`
- Execute it with:

```
UPDATE_EXPECT=1 cargo nextest run -p miden-integration-tests -- rust_masm_tests::instructions::test_overflowing_div_i8
```

Issue #960, opened 2026-02-13
bug: incorrect lowering of wasm's `i32.extend8_s`, `i32.extend16_s`


mooori:
### Packages versions

`next` branch

### Bug description

Incorrect lowering of wasm's `i32.extend8_s`, `i32.extend16_s` leads to masm/vm misreporting overflow for Rust's `overflowing` arith on `i8, i16`:

```rust
fn main(a: i8, b: i8) -> (i8, bool) {
  a.overflowing_add(b)
}

// main(-78, -51) = (127, true)
// converted to byte array: [127, 1]
```

Compiling above to masm and executing it yields `[127, 0]` instead of expected `[127, 1]`. The annotated `hir` below highlights the issue.

Likely related to

https://github.com/0xMiden/compiler/blob/0570d5919e27ae072d148c515e1c418038fb8cd5/frontend/wasm/src/code_translator/mod.rs#L319-L322

given that the generated wat is:

```wat
(module $test_rust_b50a04860f9143be69ba02bd5cc8a27d1c38c745668f3e196a1af676a4214060.wasm
  (type (;0;) (func (param i32 i32 i32)))
  (memory (;0;) 16)
  (global $__stack_pointer (;0;) (mut i32) i32.const 1048576)
  (global (;1;) i32 i32.const 1048576)
  (global (;2;) i32 i32.const 1048576)
  (export "memory" (memory 0))
  (export "entrypoint" (func $entrypoint))
  (export "__data_end" (global 1))
  (export "__heap_base" (global 2))
  (func $entrypoint (;0;) (type 0) (param i32 i32 i32)
    local.get 0
    local.get 1
    local.get 2
    i32.add
    local.tee 2
    i32.store8
    local.get 0
    local.get 2
    i32.extend8_s
    local.get 2
    i32.ne
    i32.store8 offset=1
  )
)
```

### How can this be reproduced?

- Checkout #959 and unignore `test_overflowing_add_i8`
- Run it with `UPDATE_EXPECT=1 cargo nextest run -p miden-integration-tests -- rust_masm_tests::instructions::test_overflowing_add_i8`
- Outcome
  - Test should fail and print failure info
  - Produces `test_overflowing_add_i8.hir` (and other expected files)

Same steps can be used for other tests in `test_overflowing_*_i(8|16)`.

### Relevant log output

```shell
// annotations after `//`
// tests/integration/expected/test_overflowing_add_i8.hir
[...]

Issue #993, opened 2026-03-04
dialects(wasm): implement sign extension operators


mooori:
### What should be done?

[These ops](https://github.com/WebAssembly/spec/blob/main/proposals/sign-extension-ops/Overview.md) interpret lower bits of the operand as the value of a smaller type and sign extend it. Motivation:

- Their semantics might not be captured correctly in the `arith` dialect (#960)
- Handling them in the `wasm` dialect allows generating more efficient code

**Ops**

- [x] i32.extend8_s: extend a signed 8-bit integer to a 32-bit integer #986
- [ ] i32.extend16_s: extend a signed 16-bit integer to a 32-bit integer
- [ ] i64.extend8_s: extend a signed 8-bit integer to a 64-bit integer
- [ ] i64.extend16_s: extend a signed 16-bit integer to a 64-bit integer
- [ ] i64.extend32_s: extend a signed 32-bit integer to a 64-bit integer

PR #998, opened 2026-03-05
feat: lower `i32.extend16_s` through wasm dialect

mooori:
Context: #993 
Closes: #960
Closes: #966
~Depends on: #992~ merged in the meantime

- Renames `I32Extend8S` to `I32ExtendS` and adds `src_ty` attribute
  - To have one op in the `wasm` dialect that can handle both `i32.extend8_s, i32.extend16_s` 
- Lowers `i32.extend16_s` through `wasm` dialect instead of generic `sext`
- Unignores overflowing arith tests for `i16`

### DIFF
--- a/dialects/wasm/src/ops.rs
+++ b/dialects/wasm/src/ops.rs
@@ -2,7 +2,7 @@ use alloc::format;
 
 use midenc_hir::{
     attributes::IntegerLikeAttr,
-    derive::{EffectOpInterface, OpPrinter, operation},
+    derive::{EffectOpInterface, OpParser, OpPrinter, operation},
     dialects::builtin::attributes::TypeAttr,
     effects::MemoryEffectOpInterface,
     matchers::Matcher,
@@ -12,13 +12,13 @@ use midenc_hir::{
 
 use crate::WasmDialect;
 
-/// Interprets an `i32` operand as a value of the given source type and sign-extends it to `I32`.
+/// Interprets an `i32` operand as a value of the given source type and sign-extends it to `i32`.
 ///
 /// Handles the following Wasm instructions:
 ///
 /// - `i32.extend8_s`
 /// - `i32.extend16_s`
-#[derive(EffectOpInterface, OpPrinter)]
+#[derive(EffectOpInterface, OpPrinter, OpParser)]
 #[operation(
     dialect = WasmDialect,
     traits(UnaryOp),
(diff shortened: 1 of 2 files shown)


dialects/wasm/src/ops.rs
```diff
@@ -9,36 +12,65 @@ use midenc_hir::{
 
 use crate::WasmDialect;
 
+/// Interprets the operand as value its logical type and sign-extends it to `I32`.
```
bitwalker:
```suggestion
/// Interprets an `i32` operand as a value of the given logical type and 
/// sign-extends it to `I32`.
```

dialects/wasm/src/ops.rs
```diff
@@ -9,36 +12,65 @@ use midenc_hir::{
...
+///
+/// Handles the following Wasm instructions:
+///
+/// - `i32.extend8_s`
+/// - `i32.extend16_s`
 #[derive(EffectOpInterface, OpPrinter)]
```
bitwalker: This needs to derive `OpParser` as well since you are deriving `OpPrinter`

dialects/wasm/src/ops.rs
```diff
@@ -9,36 +12,65 @@ use midenc_hir::{
...
 #[operation(
     dialect = WasmDialect,
     traits(UnaryOp),
     implements(UnaryOp, InferTypeOpInterface, MemoryEffectOpInterface, Foldable, OpPrinter)
 )]
-pub struct I32Extend8S {
+pub struct I32ExtendS {
     #[operand]
     operand: Int32,
+    /// Valide source types are `Type::I8` and `Type::I16`.
```
bitwalker:
```suggestion
    /// Valid source types are `Type::I8` and `Type::I16`. This is verified by
    /// `InferTypeOpInterface`
```

dialects/wasm/src/ops.rs
```diff
@@ -9,36 +12,65 @@ use midenc_hir::{
...
+    src_ty: TypeAttr,
     #[result]
     result: Int32,
 }
 
-impl InferTypeOpInterface for I32Extend8S {
+impl I32ExtendS {
+    /// Interprets `x` as value of the source type and sign-extends it to `i32`. Returns `None` if
+    /// the operations source type is invalid.
+    pub fn sext_from_src(&self, x: i32) -> Option<i32> {
```
bitwalker: I think it's fine to have this just return `i32` and panic if the source type is invalid - the contract for this operation dictates that the source type must be of specific valid types, and that will also be verified when the op is constructed, so it should never be the case that the source type is invalid to begin with. If it is, that's an internal compiler bug that we want to catch immediately with a panic IMO.

PR #1004, opened 2026-03-12
fix: unaligned `memcpy`

### LINKED ISSUE #1003: Fix byte-version of `memcpy`
Discovered in #985

greenhat:
Close #1003

## Summary

- factor the counted loop setup used by `memcpy`/`memset` and add regression coverage for aligned and unaligned byte copies/sets, zero-length operations, and unaligned `u16`/`i16` memory accesses
- tighten MASM memory emission for byte-oriented `memcpy` fast paths so they only convert byte pointers to element addresses when the inputs are word-aligned
- handle unaligned 16-bit loads and stores that cross a 32-bit element boundary by routing `offset == 3` through the split-word intrinsics while preserving the existing within-element path for `offset <= 2`

I suggest reviewing on a per-commit basis skipping non-interesting commits (refactors, etc.).

### DIFF
--- a/codegen/masm/src/emit/mem.rs
+++ b/codegen/masm/src/emit/mem.rs
@@ -69,7 +69,9 @@ impl OpEmitter<'_> {
                     "expected a 16-byte-aligned byte pointer for the word-copy fast path",
                     span,
                 ),
-                masm::Instruction::U32WrappingMulImm(4.into()),
+                // `u32widening_mul` leaves `[lo, hi]` on the stack; assert on `hi` and keep `lo`.
+                masm::Instruction::U32WideningMulImm(4.into()),
+                masm::Instruction::Swap1,
                 Self::assertz_with_message_inst(
                     "word-copy fast path element address conversion overflowed",
                     span,
@@ -720,6 +722,7 @@ impl OpEmitter<'_> {
         body_emitter.emit_all(
             [
                 masm::Instruction::U32WideningMadd, // [value_size * i + dst, i, dst, count, value]
+                masm::Instruction::Swap1,
                 Self::assertz_with_message_inst(
                     "memset destination address computation overflowed",
                     span,
@@ -905,6 +908,7 @@ impl OpEmitter<'_> {
                         masm::Instruction::Swap2,
                         // Compute the corrected count
                         masm::Instruction::U32WideningMulImm(factor.into()),
+                        masm::Instruction::Swap1,
                         Self::assertz_with_message_inst(
                             "memcpy word-copy fast path element count overflowed",
                             span,
@@ -950,6 +954,7 @@ impl OpEmitter<'_> {
         body_emitter.emit_all(
             [
                 masm::Instruction::U32WideningMadd,
+                masm::Instruction::Swap1,
                 Self::assertz_with_message_inst(
                     "memcpy destination address computation overflowed",
                     span,
@@ -963,6 +968,7 @@ impl OpEmitter<'_> {
         body_emitter.emit_all(
             [
                 masm::Instruction::U32WideningMadd,
+                masm::Instruction::Swap1,
                 Self::assertz_with_message_inst(
                     "memcpy source address computation overflowed",
                     span,


codegen/masm/intrinsics/mem.masm
```diff
@@ -167,26 +167,44 @@ pub proc load_sw # [addr, offset]
...
     end
 end
 
+# Load a 16-bit integer from the given native pointer tuple.
+#
+# A native pointer tuple consists of an element address where the data begins, and a byte offset,
+# which is the offset of the first byte, in the 32-bit representation of that element.
+#
+# Stack transition: [addr, offset] -> [value]
+pub proc load_u16
```
bitwalker:
```suggestion
pub proc load_u16(addr: ptr<felt, addrspace(felt)>, offset: u8) -> u16
```

Let's make sure to add type signatures to all new procedures

codegen/masm/src/emit/mem.rs
```diff
@@ -9,6 +9,84 @@ use crate::{OperandStack, lower::NativePtr};
 
 /// Allocation
 impl OpEmitter<'_> {
+    /// Emit the loop header for a counted `while.true` loop.
+    ///
+    /// The caller provides the `dup` instruction needed to bring `count` to the top of the stack
```
bitwalker: I think it would help to clarify a bit _why_ this is up to the caller, and perhaps provide an example. IMO, it would make more sense to provide an offset value or something, rather than the instruction itself - but the rationale might be obvious from better documentation/examples here.

codegen/masm/src/emit/mem.rs
```diff
@@ -9,6 +9,84 @@ use crate::{OperandStack, lower::NativePtr};
...
+    fn emit_counted_loop_header(&mut self, count_dup: masm::Instruction, span: SourceSpan) {
+        self.emit_push(0u32, span);
+        self.emit(count_dup, span);
+        self.emit_push(0u32, span);
+        self.emit(masm::Instruction::U32Gt, span);
+    }
+
+    /// Emit the loop back-edge condition for a counted `while.true` loop.
+    ///
+    /// The caller provides the `dup` instruction needed to bring `count` to the top of the stack
```
bitwalker: Same here

codegen/masm/src/emit/mem.rs
```diff
@@ -9,6 +9,84 @@ use crate::{OperandStack, lower::NativePtr};
...
+    }
+
+    /// Convert the byte pointer on top of the stack to a word-aligned element address.
+    ///
+    /// This traps unless the input byte address is aligned to a 16-byte Miden word boundary.
+    fn emit_word_aligned_element_addr_from_byte_ptr(&mut self, span: SourceSpan) {
+        self.emit_all(
+            [
+                masm::Instruction::U32DivModImm(16.into()),
+                masm::Instruction::Assertz,
```
bitwalker: We should probably consider using an instruction variant with an error message, so it is clear _why_ this fails when it fails

codegen/masm/src/emit/mem.rs
```diff
@@ -9,6 +9,84 @@ use crate::{OperandStack, lower::NativePtr};
...
+            ],
+            span,
+        );
+    }
+
+    /// Build a raw MASM block whose stack protocol is managed by the caller.
+    ///
+    /// This is used for branch bodies which operate on a known stack shape from the enclosing
+    /// emitter, but which do not need to synchronize typed operand-stack state back to it.
+    fn build_raw_block(
```
bitwalker: Is there a distinction between a "raw" MASM block and a non-"raw" MASM block? I think I might just call this `build_masm_block` or something

codegen/masm/src/emit/mem.rs
```diff
@@ -720,89 +818,68 @@ impl OpEmitter<'_> {
...
-                    value_ty.clone(),
-                    value_size,
-                    span,
-                );
+                let then_blk = self.build_raw_block(span, |then_emitter| {
+                    then_emitter.emit_all(
+                        [
+                            // Convert `src` to element address
+                            masm::Instruction::U32DivModImm(4.into()),
+                            masm::Instruction::Assertz,
```
bitwalker: We should probably try and make these asserts produce useful errors

codegen/masm/intrinsics/mem.masm
greenhat: Thanks! Done in c1802172cdb87c5c8266553baac4edf004315600

codegen/masm/src/emit/mem.rs
greenhat: Done in c1802172cdb87c5c8266553baac4edf004315600

codegen/masm/src/emit/mem.rs
greenhat: Done in c1802172cdb87c5c8266553baac4edf004315600

codegen/masm/src/emit/mem.rs
greenhat: Awesome! Thank you! Done in c1802172cdb87c5c8266553baac4edf004315600

codegen/masm/src/emit/mem.rs
greenhat: Done in c1802172cdb87c5c8266553baac4edf004315600

codegen/masm/src/emit/mem.rs
greenhat: Done in c1802172cdb87c5c8266553baac4edf004315600

greenhat: I also went and added error messages to all emitted asserts at 16a766932a633ee97a4351bf7340d2e39f402e58

PR #1005, opened 2026-03-12
fix: add symlink support in the new project template `cargo miden new`

### LINKED ISSUE #1002: Symlink support in the new project template in the `cargo miden new` command
`render_template` fails on symlink. Handle the symlink (re-create).

Restore the symlinks in the new project template removed in https://github.com/0xMiden/project-template/pull/22

greenhat:
Close #1002

### DIFF
--- a/tools/cargo-miden/src/template.rs
+++ b/tools/cargo-miden/src/template.rs
@@ -294,10 +294,123 @@ fn create_symlink(_source: &Path, target: &Path, destination: &Path) -> Result<(
     Ok(())
 }
 
-/// Creates a symlink at `destination` that points to `target`.
+/// Materializes the symlink target in the destination on Windows.
+#[cfg(windows)]
+fn create_symlink(source: &Path, target: &Path, destination: &Path) -> Result<()> {
+    materialize_symlink_target(source, target, destination)
+}
+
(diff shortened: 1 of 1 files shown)


tools/cargo-miden/src/template.rs
```diff
@@ -273,6 +278,28 @@ fn render_template(
...
+    std::os::unix::fs::symlink(target, destination).with_context(|| {
+        format!("Failed to create symlink '{}' -> '{}'", destination.display(), target.display())
+    })?;
+    Ok(())
+}
+
+/// Creates a symlink at `destination` that points to `target`.
+#[cfg(not(unix))]
+fn create_symlink(_source: &Path, _target: &Path, _destination: &Path) -> Result<()> {
+    bail!("Template symlink recreation is only supported on Unix platforms")
```
bitwalker: I would suggest we make a copy of the file on non-Unix targets, and maybe add a version for Windows that uses hardlinks - but we probably shouldn't support anything in templates that doesn't at least work on all targets where `cargo miden` may be run. We don't really properly support Windows at the moment (at least not intentionally), but we will need to eventually.

greenhat: Sure. Done in 34ffc86581801b34260398ecab96fbec6c84a6b3

Issue #1013, opened 2026-03-16
Enable benchmarking the size of compiled artifacts


mooori:
### Gooals

- [ ] Compare sizes resulting from different `--debug` options (`none, line, full`), which controls decorators.
- [ ] Analyze sizes of package sections, e.g. nodes, basic block data, strings etc.
  - [ ] Do that on different masm contracts, to compare e.g. handwritten masm to masm produced by the compiler 

### Implementation

Provide a cli that takes:

- a path to a project
- debug options to compare

and outputs the resulting artifact sizes.

### Potential follow-ups

- Integrate it into CI to measure how changes to the compiler affect sizes
- Option to strip decorators from existing artifacts
- Compare different midenc versions

PR #1015, opened 2026-03-18
feat(tools): add cli to measure decorator size in masp package

mooori:
Context: #1013

Currently `cargo miden build` will always include the debug info most relevant for this kind of analysis. So as first step this PR adds tooling to strip decorators from a `MastForest` and report the resulting size change.

Here's an example of how to use the tool and the output it produces:

```
# Assumes `examples/basic-wallet` was built in release mode
# Can be pointed at any `.masp` file
> cargo run -p objtool -- decorators examples/basic-wallet/target/miden/release/basic_wallet.masp

Input: examples/basic-wallet/target/miden/release/basic_wallet.masp
Package kind: library
Artifact: library

Metric                 KB   Delta  Delta %
original masp       36.65       -        -
original forest     36.13    0.00   +0.00%
without decorators  17.95  -18.19  -50.33%
compacted forest    13.10  -23.03  -63.74%
```

Users who wish to run it on their own masp packages may want to build the binary and add it to their path.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ clap = { version = "4.5", default-features = false, features = [
     "std",
     "env",
     "help",
+    "usage",
     "suggestions",
     "error-context",
 ] }


Cargo.toml
```diff
@@ -63,6 +63,7 @@ clap = { version = "4.5", default-features = false, features = [
     "std",
     "env",
     "help",
+    "usage",
```
mooori, line 66: Adding this to auto-generate text for the `USAGE` section in clap command help output. My understanding is:

- other CLIs in the repo that previously had empty `USAGE` now get it auto-populated by clap
- explicit overrides of `USAGE` are not affected by adding the `usage` feature

In the example below, there was `Usage:` followed by empty string before adding the `usage` feature:

```sh
cargo run -p objtool -- --help
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/objtool --help`
Inspect Miden compilation artifact sizes

Usage: objtool <COMMAND>

Commands:
  decorators  Compare serialized MAST forest sizes after stripping decorators
  help        Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

PR #1018, opened 2026-03-20
Fix `IndexSwitch` lowering when explicit cases start non-zero

### LINKED ISSUE #1017: `IndexSwitch` lowering miscompiles switches whose explicit cases do not start at `0`
Discovered in #985 

## Summary

The bug is that the MASM binary-search lowering for `scf::IndexSwitch` assumes the default arm only covers values **above** the largest explicit case. That assumption is invalid: after canonicalization, the default arm may also cover values **below** the smallest explicit case.

This means a legal switch shape such as explicit cases `[1, 2]` with the default arm covering `0` and all out-of-range values can be miscompiled.

## Rust code

```rust
#[inline(never)]
fn on_damage(out: &mut u32) {
    *out = 11;
}

#[inline(never)]
fn on_heal(out: &mut u32) {
    *out = 22;
}

#[inline(never)]
fn on_stat_mod(out: &mut u32) {
    *out = 33;
}

#[inline(never)]
fn classify(ability: AbilityType, out: &mut u32) {
    match ability {
        AbilityType::Damage => on_damage(out),
        AbilityType::Heal => on_heal(out),
        AbilityType::StatMod => on_stat_mod(out),
    }
}

let abilities = [AbilityType::Damage, AbilityType::Heal, AbilityType::StatMod];
let ability = abilities[which as usize];
let mut out = 0u32;
classify(ability, &mut out);
Felt::from_u32(out)
```

## What the current HIR shows

For the current reduced test, the relevant HIR is the `classify` helper:

```text
builtin.function private extern("C") @...classify(%14: i32, %15: i32) {
    ...
    %19 = hir.cast %18 <{ ty = #builtin.type<u32> }>;
    cf.switch %19 #builtin.u32<0> -> ^block12,
        #builtin.u32<1> -> ^block11,
        #builtin.u32<2> -> ^block10, ^block12 : (u32);
^block10:
    %22 = hir.load_local <{ local = #builtin.local_variable<1, i32> }>;
    hir.exec ::...::@on_stat_mod(%22) : extern("C") (i32) -> ();
    cf.br ^block9;
^block11:
    %21 = hir.load_local <{ local = #builtin.local_variable<1, i32> }>;
    hir.exec ::...::@on_heal(%21) : extern("C") (i32) -> ();
    builtin.ret;
^block12:
    %20 = hir.load_local <{ local = #builtin.local_variable<1, i32> }>;
    hir.exec ::...::@on_damage(%20) : extern("C") (i32) -> ();
    builtin.ret;
}
```

The key observation is that:

- `case 0` goes to `^block12`
- the fallback also goes to `^block12`
- `^block12` calls `on_damage`

So `case 0` and the fallback already overlap in the CFG form.

## What the WAT shows

The relevant frontend-generated Wasm for the same helper is:

```wat
(func $...classify (param i32 i32)
  block ;; label = @1
    block ;; label = @2
      block ;; label = @3
        local.get 0
        i32.const 255
        i32.and
        br_table 0 (;@3;) 1 (;@2;) 2 (;@1;) 0 (;@3;)
      end
      local.get 1
      call $...on_damage
      return
    end
    local.get 1
    call $...on_heal
    return
  end
  local.get 1
  call $...on_stat_mod
)
```

This `br_table` has the same overlap:

- selector `0` branches to `@3`, which calls `on_damage`
- selector `1` branches to `@2`, which calls `on_heal`
- selector `2` branches to `@1`, which calls `on_stat_mod`
- the default arm also branches to `@3`, which calls `on_damage`

So both the WAT and the translated HIR agree that selector `0` and the fallback share the same
destination before later canonicalization/lowering steps.

## Canonicalization removes `case 0`

There is an existing canonicalization that removes cases that are identical to the fallback:

https://github.com/0xMiden/compiler/blob/0ff3d972403a1590ec32571ffb6726212422f09f/dialects/cf/src/canonicalization/simplify_switch_fallback_overlap.rs?plain=1#L62-L70

This canonicalization is registered by default for `cf::Switch`.
So after canonicalization, it is completely valid for the switch to become:

- explicit cases `[1, 2]`
- default arm for `0` and all non-matching selectors

This is not malformed IR and should not require preserving an explicit `case 0`.

## How that shape reaches `scf::IndexSwitch`

`cf.switch` is converted to `scf.index_switch` by forwarding the surviving explicit cases and the
fallback region:

https://github.com/0xMiden/compiler/blob/0ff3d972403a1590ec32571ffb6726212422f09f/dialects/scf/src/transforms/cfg_to_scf.rs?plain=1#L200-L219

At this point, `scf.index_switch` legally contains:

- explicit cases `[1, 2]`
- a default region that also handles selector `0`

## Actual root cause

The bug is in the binary-search lowering in `codegen/masm/src/lower/utils.rs`.

For the two-case path, the helper uses logic of this form:

https://github.com/0xMiden/compiler/blob/0ff3d972403a1590ec32571ffb6726212422f09f/codegen/masm/src/lower/utils.rs?plain=1#L168-L170

That is only correct if the default arm handles values **greater than** the explicit range.
For explicit cases `[1, 2]`, it wrongly routes selector `0` into the explicit-case search because `0 <= 2` is true.

In the repro, that is exactly what happened:

- selector `0` should have taken the default arm and selected `on_damage`, which writes `11`
- instead it entered the explicit-case partition for `[1, 2]`
- that miscompiled the minimal dispatch helper and produced the `StatMod` result (`33`)

greenhat:
Close #1017 

This branch started from a case where removing the zero case, because it matched the default branch, exposed incorrect MASM `scf.index_switch` lowering. 
It now lowers sparse or small case sets with a linear search, keeps binary search for contiguous ranges with explicit interval guards, and uses shared branch helpers that preserve selector liveness, branch spans, and case-to-region pairing. 
It also expands switch coverage with new lowering snapshots and execution regressions for nonzero, unsorted, sparse, and `u32::MAX`-adjacent selectors, plus the minimized `resolve_turn` enum-dispatch reproduction.

### DIFF
--- a/codegen/masm/src/lower/utils.rs
+++ b/codegen/masm/src/lower/utils.rs
@@ -315,6 +315,10 @@ fn emit_default_block(
 }
 
 /// Emit a single out-of-range guard for `cases`, then enter the in-bounds binary search.
+///
+/// The interval bounds are inclusive. We only emit comparisons for bounds that can actually
+/// exclude values: selectors are unsigned, so there is no need to guard against values below `0`,
+/// and no selector can be greater than `u32::MAX`.
 fn emit_binary_search_with_interval_guard(
     op: &scf::IndexSwitch,
     emitter: &mut BlockEmitter<'_>,


tests/integration/src/rust_masm_tests/misc.rs
```diff
@@ -348,3 +348,76 @@ fn test_func_arg_order() {
     eval_package::<Felt, _, _>(&test.compile_package(), [], &args, &test.session, |trace| Ok(()))
         .unwrap();
 }
+
+/// Minimized regression test for the original `resolve_turn` reproduction.
+///
+/// This keeps only the enum dispatch shape that triggered the `IndexSwitch` bug.
+#[test]
+fn test_resolve_turn_index_switch_enum_dispatch() {
```
bitwalker: This feels like it should go in `codegen/operations/switch.rs`, since I'd really like to reorganize all the tests in `rust_masm_tests` (amongst others), into a cleaner structure, and that would be an appropriate place for this I think.

codegen/masm/src/lower/lowering.rs
```diff
@@ -325,16 +325,18 @@ stack on exit from 'after': {:#?}
...
-        // the selector against each non-default case to determine whether control should enter
-        // that block. The final else contains the default case.
-        let mut cases = self.cases().iter().copied().collect::<SmallVec<[_; 4]>>();
-        cases.sort();
+        // Lowering `hir.index_switch` is done with nested `if.true`/`else` regions that either
+        // compare the selector to each explicit case or partition a contiguous selector range.
+        let cases = utils::sorted_switch_cases(self);
+        let is_contiguous = cases
+            .windows(2)
+            .all(|pair| pair[0].selector().checked_add(1) == Some(pair[1].selector()));
```
bitwalker: If the cases are sorted, then it shouldn't ever be possible for this addition to fail, right? We don't allow duplicate selectors, so the only case where the `checked_add` returns `None` would be if the selector is `u32::MAX` - but that case will never be evaluated by this predicate, because if there is only a single case, then the iterator is empty; if there are two or more cases, then the `u32::MAX` value will always be `pair[1]`.

I only mention it, because when looking at this, the `checked_add` is suspicious due to always comparing against `Some(pair[1].selector())` - would it really be the case that the cases are guaranteed non-contiguous if the checked addition fails? No - but that can't happen anyway.

Anyway, my recommendation for clarity here would be to use `pair[0].selector() + 1` - if that ever fails, we've got a different problem.

codegen/masm/src/lower/utils.rs
```diff
@@ -87,360 +59,352 @@ stack on exit from 'else': {else_stack:#?}
...
+        .iter()
+        .copied()
+        .enumerate()
+        .map(|(index, selector)| SwitchCase::new(selector, op.get_case_region(index)))
+        .collect::<SmallVec<[SwitchCase; 4]>>();
+    cases.sort_by_key(SwitchCase::selector);
+    cases
+}
+
+/// The explicit selector interval spanned by a sorted `scf.index_switch` case slice.
```
bitwalker: I'd maybe document whether the interval bounds are inclusive or exclusive

codegen/masm/src/lower/utils.rs
```diff
@@ -87,360 +59,352 @@ stack on exit from 'else': {else_stack:#?}
...
+/// Emit a single out-of-range guard for `cases`, then enter the in-bounds binary search.
+fn emit_binary_search_with_interval_guard(
+    op: &scf::IndexSwitch,
+    emitter: &mut BlockEmitter<'_>,
+    cases: &[SwitchCase],
+    interval: SwitchCaseInterval,
+) -> Result<(), Report> {
+    let span = op.span();
+
+    match (interval.lower > 0, interval.upper < u32::MAX) {
```
bitwalker, line 330: It's not super clear from context why `0` and `u32::MAX` are excluded here - I'm assuming that's handled explicitly somewhere else, but when reading this function it isn't documented as an assumption - and my reading of the semantics seems to indicate that we'd want `>= 0` and `<= u32::MAX`

tests/integration/src/rust_masm_tests/misc.rs
greenhat: Yep. Thanks! Done in c8adee121ed7a3e15550cd3e42298971389d8742.

codegen/masm/src/lower/lowering.rs
greenhat: Good point! Thank you. Reworked in 5d699f56a15a894ed44314bb9f288b8dff300e07.

codegen/masm/src/lower/utils.rs
greenhat: Done in 95b6547a95df4a998d6fc03b7eca80b1a1f7c3af.

codegen/masm/src/lower/utils.rs
greenhat: When the explicit case interval already starts at 0 or ends at u32::MAX, there is nothing extra to guard on that side. I clarified the comment in 950e2c7e5c9c2747c7acbc4d5545d1aa7699d282

PR #1019, opened 2026-03-20
feat: add `I32Load8S` to `wasm` dialect

mooori:
Motivation:

- more pecialized representation of Wasm semantics
- enable optimizations across multiple Wasm instructions later on

In isolation, the generated masm is the same as before when lowering through the more general `load_sext`, see details below.

I'm sending a PR adding only one op first to figure out the approach to handling the `I(32|64)Load(8|16|32)(U|S)` family of instructions. Once the first of them is in the Wasm dialect, it should be easier to decide on how to add the remaining ones (generalizing `I32Load8S` vs adding more ops).

### Analysis of generated masm

It's based on compiling this Wat:

```wat
(module
  (memory (;0;) 16384)
  (global $MyGlobalVal (;0;) (mut i32) i32.const 42)
  (export "test_wrapper" (func $test_wrapper))
  (func $test_wrapper (;0;) (type 0) (param i32) (result i32)
    local.get 0
    i32.load8_s
  )
)
```

with the command below to get the following `masm`, which is identical when compiled off `next` and off this branch:

```sh
cargo run -p midenc -- load.wat --entrypoint=load::test_wrapper --emit=ir
```

```
@locals("1")
@callconv("C")
pub proc test_wrapper(i32) -> i32
    u32divmod.4
    swap.1
    swap.1
    dup.1
    mem_load
    swap.1
    push.8
    u32wrapping_mul
    u32shr
    swap.1
    drop
    push.255
    u32and
    dup.0
    push.128
    u32and
    eq.128
    push.0
    push.4294967040
    movup.2
    cdrop
    u32or
end
```

#### Why no optimizations

The masm does the following. Even when working with a fixed set of types, I didn't see an easy way to emit more efficient masm:

- Convert byte address → word address + offset
  - needed due to the difference in address schemes
- Load 32 bit word
  - cannot load less than 32-bits
- Extract the specific byte using shift and mask
  - only need 8 out of 32 loaded bits but can't assume anything about the unused 24 bits
- Sign-extend from 8 bits to 32 bits

There are [bit twiddling hacks](https://graphics.stanford.edu/~seander/bithacks.html) for `sext`, but I didn't go for it because:

- not sure if we want to do that at the moment
- assessing their performance on a non-standard architecture is not trivial

### DIFF
--- a/frontend/wasm/src/code_translator/mod.rs
+++ b/frontend/wasm/src/code_translator/mod.rs
@@ -257,7 +257,10 @@ pub fn translate_operator<B: ?Sized + Builder>(
             translate_load_zext(U16, U32, memarg, state, builder, span)?;
         }
         Operator::I32Load8S { memarg } => {
-            translate_load_sext(I8, I32, memarg, state, builder, span)?;
+            let addr_int = state.pop1();
+            let addr = prepare_addr(addr_int, &I8, Some(memarg), builder, span)?;
+            let val = builder.i32_load8_s(addr, span)?;
+            state.push1(val);
         }
         Operator::I32Load16S { memarg } => {
             translate_load_sext(I16, I32, memarg, state, builder, span)?;
--- a/tests/integration/src/codegen/wasm.rs
+++ b/tests/integration/src/codegen/wasm.rs
@@ -1,8 +1,11 @@
 use miden_debug::ToMidenRepr;
+use midenc_dialect_hir::HirOpBuilder;
 use midenc_dialect_wasm::WasmOpBuilder;
-use midenc_hir::{Felt, SourceSpan, Type, ValueRef, dialects::builtin::BuiltinOpBuilder};
+use midenc_hir::{
+    Felt, PointerType, SourceSpan, Type, ValueRef, dialects::builtin::BuiltinOpBuilder,
+};
 
-use crate::testing::{compile_test_module, eval_package};
+use crate::testing::{Initializer, compile_test_module, eval_package};
 
 #[test]
 fn test_i32_extend8_s() {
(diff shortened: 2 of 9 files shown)


frontend/wasm/src/code_translator/mod.rs
```diff
@@ -257,7 +257,10 @@ pub fn translate_operator<B: ?Sized + Builder>(
             translate_load_zext(U16, U32, memarg, state, builder, span)?;
         }
         Operator::I32Load8S { memarg } => {
-            translate_load_sext(I8, I32, memarg, state, builder, span)?;
+            let addr_int = state.pop1();
+            let addr = prepare_addr(addr_int, &I8, Some(memarg), builder, span)?;
```
mooori, line 261: Let's move `prepare_addr` to the `wasm` dialect? Calling it from inside the `I32Load8S` op would feel more natural and would allow to clean up the test in `codegen/wasm.rs`.

tests/integration/src/codegen/wasm.rs
```diff
@@ -374,3 +377,55 @@ fn test_i64_extend32_s() {
...
+    let mem_addr = 17 * 2u32.pow(16);
+
+    let (package, context) = compile_test_module([Type::I32], [Type::I32], |builder| {
+        let block = builder.current_block();
+        let addr_i32 = block.borrow().arguments()[0] as ValueRef;
+
+        let addr_u32 = builder.bitcast(addr_i32, Type::U32, span).unwrap();
+        let ptr = builder
+            .inttoptr(addr_u32, Type::from(PointerType::new(Type::I8)), span)
+            .unwrap();
```
mooori, line 393: This could be removed if `I32Load8S` could call `prepare_addr`, see comment above.

frontend/wasm/src/code_translator/mod.rs
bitwalker: That seems fine to me, unless @greenhat has any objections

mooori: As moving it affects a couple of call sites in `frontend`, I'd prefer to do it in a separate PR. So I've opened #1022 and go ahead with merging this PR as is.

greenhat: I'm fine with moving `prepare_addr` to the `wasm` dialect.

PR #1032, opened 2026-03-25
refactor: move `prepare_addr` from frontend to wasm dialect

### LINKED ISSUE #1022: refactor: move `fn prepare_addr` from frontend to wasm dialect
After adding the `wasm` dialect, it would be beneficial to call `prepare_addr` from ops in the `wasm` dialect. Testing ops would be easier and could be decoupled from frontend implementation details [[example](https://github.com/0xMiden/compiler/pull/1019/changes/BASE..d19da55b842c3157fde899e78749ec38635728dc#r2967403253)].

### TODO after moving it

- [x] Cleanup [`test_i32_load8_s`](https://github.com/0xMiden/compiler/blob/d19da55b842c3157fde899e78749ec38635728dc/tests/integration/src/codegen/wasm.rs#L382)

mooori:
Closes #1022

Previously the ops for `i32.load8_s` were built partly in the frontend and partly in the wasm dialect. This refactor allows building all of them in the wasm dialect. Ultimately all load ops constructed in the wasm frontend should go through the wasm dialect, which motivates moving `prepare_addr` there.

Needs to introduce something like `WasmMemOpBuilder` to allow the `builder` passed to `prepare_addr` calling `ArithOpBuilder, HirOpBuilder` methods. I feel like it's worth it for getting rid of the split mentioned above. So far all relevant `builders` already meet this requirement.

### DIFF
--- a/dialects/wasm/src/builders.rs
+++ b/dialects/wasm/src/builders.rs
@@ -3,7 +3,9 @@ use midenc_hir::{
     dialects::builtin::FunctionBuilder,
 };
 
-pub trait WasmOpBuilder<'f, B: ?Sized + Builder> {
+use crate::{WasmMemArg, mem::WasmMemOpBuilder, prepare_addr};
+
+pub trait WasmOpBuilder<'f, B: ?Sized + Builder>: WasmMemOpBuilder<'f, B> {
     fn sign_extend(
         &mut self,
         arg: ValueRef,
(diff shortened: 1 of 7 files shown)


dialects/wasm/src/mem.rs
```diff
@@ -0,0 +1,71 @@
...
+impl WasmMemArg {
+    pub const fn new(offset: u64, align: u8) -> Self {
+        Self { offset, align }
+    }
+}
+
+pub trait WasmMemOpBuilder<'a, B: ?Sized + Builder>:
+    ArithOpBuilder<'a, B> + HirOpBuilder<'a, B>
+{
+}
```
mooori, line 29: Motivation: Allow the `builder` in `prepare_address` to be used as `ArithOpBuilder` and `HirOpBuilder`.

dialects/wasm/src/builders.rs
```diff
@@ -3,21 +3,29 @@ use midenc_hir::{
     dialects::builtin::FunctionBuilder,
 };
 
-pub trait WasmOpBuilder<'f, B: ?Sized + Builder> {
+use crate::{WasmMemArg, mem::WasmMemOpBuilder, prepare_addr};
+
+pub trait WasmOpBuilder<'f, B: ?Sized + Builder>: WasmMemOpBuilder<'f, B> {
```
mooori, line 8: The `WasmMemOpBuilder` bound might also be moved to only those methods that need it.

dialects/wasm/src/mem.rs
```diff
@@ -0,0 +1,71 @@
...
+
+pub fn prepare_addr<'a, B: ?Sized + Builder>(
+    addr_int: ValueRef,
+    ptr_ty: &Type,
+    memarg: Option<WasmMemArg>,
+    builder: &mut (impl WasmMemOpBuilder<'a, B> + ?Sized),
+    span: SourceSpan,
+) -> Result<ValueRef, Report> {
+    let addr_int_ty = addr_int.borrow().ty().clone();
+    let addr_u32 = if addr_int_ty == Type::U32 {
```
bitwalker: I think we should require that the input type already be `I32` - the purpose of the `wasm` dialect is in large part to ensure that the Wasm type system semantics are maintained (which means that we would never expect to see `Type::U32` or `Type::Ptr` as input, unless we've got an operation that is not producing valid Wasm types).

To the extent that we still need such casts, we should push that into the frontend - but here in the Wasm dialect I think we should be stricter about this. That will also help simplify some of this code that may still be defending against things that the frontend never does anymore.

It will also guide us towards other ops that may be worth representing in the `wasm` dialect explicitly.

dialects/wasm/src/builders.rs
bitwalker: I think it's fine for now - if we run into friction due to the requirement, we'll revisit it then.

dialects/wasm/src/mem.rs
mooori: I've tightened the `addr_int` type requirement by panicking if it's type is not `I32`. Assuming the frontend only translates valid Wasm and therefore `addr_int` should never have another type.

For now I left the signature unchanged at `addr_int: ValueRef`. IIUC in this context there is no direct way to pop a value from the stack in the frontend (`let addr_int = state.pop1()`) and then pass `addr_int` to a function whose signature ensures that `addr_int` is `I32`.

We might add a type like `pub struct I32ValueRef(ValueRef)` which can only be constructed from `I32` values.

bitwalker: Yeah, this kind of issue would either be caught by a type check here in `prepare_addr`, or by the verifier (and I think the former is most appropriate here, so we can get a stack trace pointing to exactly what violated the contract).

PR #1040, opened 2026-03-27
feat: add remaining `load + sext` ops to wasm dialect

mooori:
#### Summary

- Factors out boilerplate of `wasm` op `I32Load8S` into traits, to make it reusable for the other `load + sext` ops
- Adds the remaining `i*.load*_s` ops to the `wasm` dialect

#### Generated code

- The `hir` changes, see the commited `.hir` expect files
- No changes in emitted `masm`, tested with the same approach is in #1019 

#### Review note

Can be reviewed per commit.

### DIFF
--- a/codegen/masm/src/lower/lowering.rs
+++ b/codegen/masm/src/lower/lowering.rs
@@ -1336,15 +1336,23 @@ impl HirLowering for wasm::SignExtend {
     }
 }
 
-impl HirLowering for wasm::I32Load8S {
-    fn emit(&self, emitter: &mut BlockEmitter<'_>) -> Result<(), Report> {
-        let result = self.result();
-        let pointer_ty = self.addr().ty();
-        let pointee_ty = pointer_ty.pointee().expect("pointer should have been verified").clone();
-        let mut inst_emitter = emitter.inst_emitter(self.as_operation());
-        inst_emitter.load(pointee_ty, self.span());
-        inst_emitter.sext(result.ty(), self.span());
-
-        Ok(())
-    }
+macro_rules! impl_hir_lowering_load_sext {
+    ($op:ty) => {
+        impl HirLowering for $op {
+            fn emit(&self, emitter: &mut BlockEmitter<'_>) -> Result<(), Report> {
+                let pointee_ty =
+                    self.addr().ty().pointee().expect("pointer should have been verified").clone();
+                let mut inst_emitter = emitter.inst_emitter(self.as_operation());
+                inst_emitter.load(pointee_ty, self.span());
+                inst_emitter.sext(self.result().ty(), self.span());
+                Ok(())
+            }
+        }
+    };
 }
+
+impl_hir_lowering_load_sext!(wasm::I32Load8S);
+impl_hir_lowering_load_sext!(wasm::I32Load16S);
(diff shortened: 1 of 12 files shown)


codegen/masm/src/lower/lowering.rs
```diff
@@ -1367,15 +1367,21 @@ impl HirLowering for wasm::SignExtend {
...
-impl HirLowering for wasm::I32Load8S {
+trait LowerWasmLoadSext: wasm::WasmLoadSext + Op {}
+
+impl LowerWasmLoadSext for wasm::I32Load8S {}
+impl LowerWasmLoadSext for wasm::I32Load16S {}
+impl LowerWasmLoadSext for wasm::I64Load8S {}
+impl LowerWasmLoadSext for wasm::I64Load16S {}
+impl LowerWasmLoadSext for wasm::I64Load32S {}
+
+impl<T: LowerWasmLoadSext> HirLowering for T {
```
bitwalker: This will eventually present an issue with trait coherence (i.e. you are defining a blanket implementation of `HirLowering` for all `T`, just bounded). It's not so much of an issue of the wrong impl being used, but rather that we can only use this trick _once_.

I would use `macro_rules!` for this, _or_ do the following for each op:

```rust
impl HirLowering for wasm::I32Load8 {
        lower_sign_extended_load(self, emitter)
    }
}
```

then define `lower_sign_extended_load` as a generic function.

dialects/wasm/src/ops.rs
```diff
@@ -166,7 +166,37 @@ impl Foldable for SignExtend {
     }
 }
 
-/// Load `result` from the heap at `addr` and sign-extend it to `i32`.
+/// Provides functions that enable blanket-like implementations of evaluation and HIR lowering.
+pub trait WasmLoadSext {
```
bitwalker: This should be defined as an operation trait IMO, otherwise you have no way to get the implementation from an `&Operation`/`OperationRef`

eval/src/eval.rs
```diff
@@ -2065,14 +2065,19 @@ impl Eval for wasm::SignExtend {
     }
 }
 
-impl Eval for wasm::I32Load8S {
+trait EvalWasmLoadSext: wasm::WasmLoadSext {
+    /// Sign-extends the value loaded from memory to the result type.
+    fn sext_immediate(&self, value: Value, evaluator: &HirEvaluator) -> Result<Value, Report>;
+}
+
+impl<T: EvalWasmLoadSext> Eval for T {
```
bitwalker: Same thing regarding trait coherence here. I'd solve this the same way I recommended you solve it for `HirLowering`

codegen/masm/src/lower/lowering.rs
mooori: Changed it `macro_rules!` based lowering.

eval/src/eval.rs
mooori: Changed it `macro_rules!` based eval.

dialects/wasm/src/ops.rs
mooori: With eval and lowering now based on `macro_rules!`, I didn't see an immediate use case for `WasmLoadSext`, so I've removed it for now.

PR #1041, opened 2026-03-30
Add `README` to `tools/objtool` and `objtool` install task

mooori:
Explains set-up and usage of `objtool` and the `decorators` subcommand.

### DIFF
--- a/tools/objtool/README.md
+++ b/tools/objtool/README.md
@@ -6,21 +6,13 @@ Provides the `objtool` CLI to analyze compilation artifacts.
 
 The compilation artifacts to be examined must have been produced by a `midenup` toolchain version compatible with the `compiler` version used to build `objtool`. Otherwise you may run into errors like `unsupported version`.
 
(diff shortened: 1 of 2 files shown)


tools/objtool/README.md
```diff
@@ -0,0 +1,62 @@
+# `objtool`
+
+Provides the `objtool` CLI to analyze compilation artifacts.
+
+## Compatibility
+
+The compilation artifacts to be examined must have been produced by a `midenup` toolchain version compatible with the `compiler` version used to build `objtool`. Otherwise you may run into errors like `unsupported version`.
+
+## Usage
```
bitwalker: I'd suggest we add a `Makefile.toml` task to build `objtool` just like our other tools, and then have the usage point out how to build the tool, and then run it directly, rather than via `cargo run` (feel free to mention that as another way to use it, but I'd suggest our user-facing documentation assume the tool is available in PATH)

PR #1049, opened 2026-04-02
fix: unsound IR handle conversions that were rebuilding `RawEntityRef`

### LINKED ISSUE #1047: Compiler built with `release` profile crashes on the simplest Rust program compilation
When the `cargo-miden` compiled with the `release` profile the new Rust program template compilation (`cargo miden build --release`) fails with an error:
```
thread 'main' (66579789) panicked at /Users/green/src/miden/compiler-tree/bump-new-project-template/dialects/arith/src/builders.rs:35:58:
called `Result::unwrap()` on an `Err` value:   x invalid operation arith.constant
```

Discovered at https://github.com/0xMiden/rust-templates/actions/runs/23844868871/job/69509545152?pr=40

When the `cargo-miden` is compiled with the `debug` profile the new Rust program template compilation (`cargo miden build --release`) **succeeds**.

Rust program source:
```rust
#![no_std]
#![feature(alloc_error_handler)]

#[cfg(not(test))]
#[panic_handler]
fn my_panic(_info: &core::panic::PanicInfo) -> ! {
    loop {}
}

#[cfg(not(test))]
#[alloc_error_handler]
fn my_alloc_error(_info: core::alloc::Layout) -> ! {
    loop {}
}

#[no_mangle]
pub fn entrypoint() -> i32 {
    // TODO: Implement your program logic here
    0
}
```

Wasm:
```
(module $my_program_proj.wasm
  (type (;0;) (func (result i32)))
  (memory (;0;) 16)
  (export "memory" (memory 0))
  (export "entrypoint" (func $entrypoint))
  (func $entrypoint (;0;) (type 0) (result i32)
    i32.const 0
  )
)
```

greenhat:
Close #1047

This PR fixes unsound IR handle conversions that were rebuilding `RawEntityRef` values from borrowed concrete and trait-object references. The compiler now preserves the original intrusive handle when casting between concrete ops, attributes, symbols, and erased or trait-object forms, which avoids corrupting metadata, borrow identity, and hidden property storage in optimized builds.

- add handle-preserving cast helpers for operation and attribute refs, and route parser, matcher, builder, SCF, and other IR call sites through them instead of reconstructing refs from borrowed values
- rebind property-backed symbol refs in place, including generated symbol setters, and unlink/relink symbol uses when a symbol-backed property is updated
- extend the symbol/callable path with handle-preserving `SymbolRef` trait casts so `SymbolTable::resolve`, `CallOpInterface::set_callee`, and related conversions stop rebuilding refs from borrowed trait objects
- add regression coverage for `ImmediateAttr` round-tripping, trait-based matcher paths, property-backed `Call::set_callee`, and `CallOpInterface::set_callee` via symbol resolution
- keep compiler crates under release-like test settings for the integration test package so the optimized-build failure mode stays covered

Adds a `test-release` Cargo profile, so local `cargo make test` keeps the faster debug-oriented workflow. To keep optimized-build coverage, it also adds `cargo make test-release` and a separate CI job that runs the Rust test suite with compiler crates rebuilt under the new profile.


Cargo.toml
```diff
@@ -218,12 +218,155 @@ opt-level = 3
...
 opt-level = 3
 
-# Compiler packages that might be used in tests
-[profile.test.package.midenc-frontend-wasm]
-opt-level = 3
+# Build the compiler with release-like settings only when explicitly requested
+# via the test-release profile. Cargo only honors these overrides in the
+# workspace root, even though the consuming crate lives under tests/integration.
+[profile.test-release]
+inherits = "test"
```
bitwalker, line 224: Do settings here not apply to packages? Seems like all the packages below use the same settings, which I believe can just be moved here

hir/src/ir/symbols.rs
```diff
@@ -89,6 +94,27 @@ impl AsSymbolRef for SymbolRef {
...
+    {
+        let ptr = {
+            let symbol = self.borrow();
+            symbol
+                .as_symbol_operation()
+                .name()
+                .upcast_raw::<Trait>(Self::into_raw(self).cast())?
+        };
+        let (_, metadata) = ptr.to_raw_parts();
+        Some(unsafe { self.cast_unsized_unchecked::<Trait>(metadata) })
```
bitwalker:
```suggestion
        let symbol = self.borrow();
        symbol
            .as_symbol_operation()
            .as_trait_ref::<Trait>()
```

No need to re-implement `as_trait_ref` here, since you have to downcast to `Operation` anyway (and fallible upcasting to `Trait` from `Operation` is sound)

hir/src/ir/operation.rs
```diff
@@ -783,9 +838,9 @@ impl Operation {
         while let Some(op) = parent.take() {
             parent =
                 op.parent().and_then(|block| block.parent()).and_then(|region| region.parent());
-            let op = op.borrow();
-            if let Some(t_ref) = op.downcast_ref::<T>() {
-                return Some(unsafe { UnsafeIntrusiveEntityRef::from_raw(t_ref) });
+            let is_match = { op.borrow().is::<T>() };
+            if is_match {
+                return op.try_downcast_op::<T>().ok();
```
bitwalker:
```suggestion
            if let Ok(op) = op.borrow().try_downcast_op::<T>() {
                return Some(op);
```

hir/src/ir/operation.rs
```diff
@@ -682,18 +726,29 @@ impl Operation {
...
+            {
+                let prop = self.get_property(attr_name);
+                let prop = prop.downcast_ref::<SymbolRefAttr>().unwrap();
+                self.unlink_symbol_use(prop);
+            }
 
-            // Track the usage of `symbol` by `self`
+            let path = symbol.borrow().path();
+            let symbol_ref = crate::dialects::builtin::attributes::SymbolRef::new(path, None);
+            let mut attr = context.create_attribute::<SymbolRefAttr, _>(symbol_ref);
```
bitwalker: This is broken - this attribute isn't being attached to anything (i.e. stored as the value of the property), and the old property will still be used after this point

hir/src/ir/operation.rs
```diff
@@ -682,18 +726,29 @@ impl Operation {
         let owner = self.as_operation_ref();
         let context = self.context_rc();
         let user = if self.has_property(attr_name) {
-            let mut prop = self.get_property_mut(attr_name);
-            let prop = prop.downcast_mut::<SymbolRefAttr>().unwrap();
-            let attr = unsafe { UnsafeIntrusiveEntityRef::from_raw(prop) };
+            {
```
bitwalker: I don't think this change is required, or rather it is doing more than necessary to fix what I presume is the issue here - we can mutate properties, we don't need to recreate them whole cloth. With `SymbolRefAttr` specifically, it is the `SymbolUse` which carries the link, not the `SymbolRefAttr` itself (the `SymbolUse` acts as an indirection to link the user and used symbol, while the `SymbolRefAttr` is what carries the path that is referenced (along with the `SymbolUse` reference) - the `SymbolRefAttr` doesn't get borrowed at all when reading/writing the `SymbolUse` it carries.

So when setting a symbol ref attribute, the only thing we have to do is unlink the previous `SymbolUse` (if linked) from the entity list it was linked into; and then link it into the new target entity list.

hir/src/ir/operation.rs
```diff
@@ -658,6 +679,29 @@ impl Operation {
 
 /// Symbol Attributes
 impl Operation {
+    /// Unlinks the current use stored in `attr`, if it is still attached to a symbol.
+    fn unlink_symbol_use(&self, attr: &SymbolRefAttr) {
```
bitwalker: This doesn't need to be a method on `OperationRef` (and indeed, it would be error-prone to do so, as nothing here guarantees that `attr` belongs to `self`).

More to the point, if a `SymbolRefAttr` is linked (via its `SymbolUse`), then it will have a `user`, and the resolution of the symbol path _must_ be relative to `user`, not some other operation (e.g. `self` in this case).

I would make this a method of `SymbolRefAttr` to address those issues.

hir/src/ir/operation.rs
```diff
@@ -345,6 +345,31 @@ impl OperationRef {
...
 
+    /// Returns this operation as a handle to an implemented trait object, if supported.
+    ///
+    /// The returned handle preserves the original intrusive allocation identity and swaps only the
+    /// pointee metadata for the requested trait object.
+    pub fn as_trait_ref<Trait>(self) -> Option<UnsafeIntrusiveEntityRef<Trait>>
+    where
+        Trait: ?Sized + Pointee<Metadata = DynMetadata<Trait>> + 'static,
+    {
+        let ptr = self.name().upcast_raw::<Trait>(OperationRef::into_raw(self).cast())?;
```
bitwalker, line 356: This is definitely wrong - you must use `self.container()` on the `Operation` reference to get the pointer to the concrete `Op` (from which you can then cast to `dyn Trait`).

To the extent this works, its because `&Operation` has the same address as `&T` where `T: Op` - but casting from `Operation` like this is not correct.

bitwalker (changes requested):

greenhat:
Thank you! I addressed all the notes. Please do another round.

hir/src/ir/entity.rs
```diff
@@ -1202,13 +1149,19 @@ impl<T: ?Sized, Metadata> RawEntityMetadata<T, Metadata> {
     #[track_caller]
     pub(crate) fn borrow(&self) -> EntityRef<'_, T> {
         let ptr = self as *const Self;
-        unsafe { (*core::ptr::addr_of!((*ptr).entity)).borrow() }
```
bitwalker, line 1205: Could you clarify this change for me? This appears to be entirely redundant (i.e. reconstructing the EntityRef/EntityMut that is already returned for no reason)

hir/src/dialects/builtin/attributes/symbol_ref.rs
```diff
@@ -101,6 +101,31 @@ impl SymbolRefAttr {
     pub fn set_user(&mut self, user: SymbolUseRef) {
         self.value.set_user(user);
     }
+
+    /// Resolves this symbol reference relative to the operation that owns the tracked use.
+    pub fn resolve(&self) -> Option<crate::SymbolRef> {
```
bitwalker, line 110: We probably should document the fact that this immutably borrows the user, so that the caller can ensure they aren't mutably borrowing it already, or they risk an aliasing violation

hir/src/ir/symbols/symbol_use.rs
```diff
@@ -19,17 +20,80 @@ pub struct SymbolUse {
...
+                .or_else(|| owner.implements::<dyn SymbolTable>().then_some(self.owner))
+        }?;
+        let symbol_table = symbol_table.borrow();
+        symbol_table.as_symbol_table()?.resolve(path)
+    }
+
+    /// Unlinks this use from its current symbol, if any.
+    pub fn unlink_from_symbol(
+        &mut self,
+        user: SymbolUseRef,
```
bitwalker: This isn't making sense to me - we are mutably borrowing `self`, but are passing in a SymbolUseRef that presumably points to `self`? I don't think that's a good idea. Why not just use `self: SymbolUseRef' instead? IIRC, we've got the features enabled to use that

hir/src/ir/symbols/symbol_use.rs
```diff
@@ -19,17 +20,80 @@ pub struct SymbolUse {
...
+        Some(symbol)
+    }
+
+    /// Links this use to `symbol`.
+    pub fn link_to_symbol(&mut self, user: SymbolUseRef, mut symbol: SymbolRef) {
+        debug_assert!(
+            !user.is_linked(),
+            "symbol use must be unlinked before it can be linked again"
+        );
+        symbol.borrow_mut().insert_use(user);
```
bitwalker: We should be using EntityList callbacks for this, not managing this manually, or this will be very fragile

hir/src/ir/symbols/symbol_use.rs
```diff
@@ -19,17 +20,80 @@ pub struct SymbolUse {
     pub owner: OperationRef,
     /// The symbol attribute of the op that stores the symbol
     pub attr: UnsafeIntrusiveEntityRef<SymbolRefAttr>,
+    /// The symbol currently linked to this use, if any.
+    pub used: Option<SymbolRef>,
```
bitwalker: I'm leery of adding this field, there was a reason I hadn't done it before. Critically, if one was to work backwards, i.e. go through the uses of a symbol from that symbols use list, you'd have to now also visit every SymbolUse and remove the stored `used` value, where previously you could just clear the list and the link would be unset automatically for all uses

hir/src/ir/entity.rs
greenhat: I agree. I delegated to `self.entity` in d6925c97b91719f43ad5bb2a4b862b0ff461beab.

hir/src/dialects/builtin/attributes/symbol_ref.rs
greenhat: Done in 177e41df247c213c8e9e8ad236bb77d21773cff8.

greenhat:
I addressed all the comments. The symbol-use-related notes are addressed in the last two commits: c778a9fee70c69f53978a6ca48cc29185cfb71b7 and 12472a6a0ae385c240c015b7fc5af88731284ed7. Please do another round.

hir/src/ir/operation.rs
```diff
@@ -118,6 +118,8 @@ pub struct Operation {
     /// If this operation represents control flow, this field stores the set of successors,
     /// and successor operands.
     pub successors: OpSuccessorStorage,
+    /// The set of symbol uses tracked for this operation when it acts as a symbol.
+    pub uses: SymbolUseList,
```
bitwalker: Unfortunately this means that every single operation now carries this extra data structure, and we really want to avoid that to keep memory usage reasonable. The reason it is stored on only the Symbol ops themselves is primarily for this reason.

hir/src/ir/operation.rs
```diff
@@ -256,6 +258,12 @@ impl EntityParent<Region> for Operation {
     }
 }
 
+impl EntityParent<SymbolUse> for Operation {
```
bitwalker: This would allow any op to be treated like a Symbol, which isn't semantically correct

bitwalker (changes requested):
We're super close, but not quite there - sorry for all iterations 😅. Unfortunately, the symbol use stuff needs a few further modifications, see the couple comments I left on that. I'd suggest keeping the changes there super narrow, because it is a very subtle area of the IR. On the one hand, this is a great way to learn the idiosyncrasies of this part of the code, but due to the importance of the other fixes here, it may make sense to do other types of refactoring after the unsoundness fixes are landed.

hir/src/ir/operation.rs
greenhat: Makes sense. I deleted the last commit with the callback refactor. This also addresses the comment below.

bitwalker (approved):

PR #1051, opened 2026-04-03
feature(sdk): make note script a library and generate `@note-script` and `@auth_script` MASM attribute

### LINKED ISSUE #926: Convert `RollupTarget::NoteScript` to produce a library
Ref https://github.com/0xMiden/miden-base/issues/2280

The note script entrypoint marked with `#[note_script]` attribute macros has to be marked with `@note_script` attribute in the generated MASM package.

### LINKED ISSUE #982: Use MASM attribute to mark authentication procedure in the authentication component (instead of "auth_" prefix)

greenhat:
Close #982 
Close #926 

Compile the note script as a library and mark the entrypoint(`#[note_script]`) with the `@note_script` MASM attribute.
Propagate the component method marked with the `#[auth_script]` attribute macro and note script attribute through a dedicated Wasm custom section so the frontend can attach without relying on hardcoded procedure names. Also reject multiple such methods per project.
Create the `midenc-frontend-wasm-metadata` crate so that SDK macros emit what methods were marked and the Wasm frontend parses it from the custom section and creates IR attributes.


frontend/wasm/src/component/lift_exports.rs
```diff
@@ -29,6 +29,8 @@ pub fn generate_export_lifting_function(
     export_func_name: &str,
     export_func_ty: FunctionType,
     core_export_func_path: SymbolPath,
+    is_auth_procedure: bool,
```
bitwalker: *Nit:* Since the set of flags/metadata might grow, and booleans are a bit opaque from the caller side - maybe it'd make sense to create a new struct that contains all of the flags/metadata needed for lifting/lowering, and pass that instead?

sdk/base-macros/src/wit_builder.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /// Finishes rendering and returns the final WIT source.
+    pub(crate) fn finish(self) -> String {
+        self.source
+    }
+}
+
+/// Writes indented lines inside a WIT interface or world block.
+pub(crate) struct WitBody<'a> {
```
bitwalker: *Nit:* Rather than reimplement this kind of formatting plumbing, it's probably a lot easier to just build off the primitives provided by `midenc_hir::formatter`, e.g. the `PrettyPrint` trait, `Document`, and the constructors like `indent(4, doc)`, etc.

The `Document` type implements `AddAssign`, so you can still use it imperatively like you would building a string, but it ensures the output is generally formatted nicely.

sdk/base-macros/src/wit_world.rs
```diff
@@ -0,0 +1,384 @@
...
+
+/// Parsed package metadata from the consuming crate's manifest.
+pub(crate) struct ManifestPackage {
+    manifest_dir: PathBuf,
+    package_table: Table,
+}
+
+impl ManifestPackage {
+    /// Loads the current crate's `[package]` table from `Cargo.toml`.
+    pub(crate) fn load(error_span: Span) -> Result<Self, syn::Error> {
```
bitwalker, line 24: We'll want to load this stuff from `miden-project.toml` instead once we support that in compiler/cargo-miden - but since we haven't done that yet, just noting it here

sdk/base-sys/src/bindings/types.rs
```diff
@@ -3,18 +3,20 @@ extern crate alloc;
 use miden_field_repr::FromFeltRepr;
 use miden_stdlib_sys::{Felt, Word, felt};
 
-/// Packs a scalar felt into the low limb of a protocol word.
-fn padded_word_from_felt(value: Felt) -> Word {
-    Word::new([felt!(0), felt!(0), felt!(0), value])
+/// Packs a scalar felt into the leading limb of a protocol word.
+#[inline(always)]
+pub fn padded_word_from_felt(value: Felt) -> Word {
```
bitwalker, line 7: *Note:* This is probably an example of something not worth decorating with `#[inline(always)]`, in order to avoid unnecessary bloating of the MAST.

sdk/base-sys/src/bindings/types.rs
```diff
@@ -3,18 +3,20 @@ extern crate alloc;
...
+    Word::new([value, felt!(0), felt!(0), felt!(0)])
 }
 
-/// Extracts a scalar felt from a protocol word with zero-padded high limbs.
-fn felt_from_padded_word(value: Word) -> Result<Felt, &'static str> {
-    if value[0] != felt!(0) || value[1] != felt!(0) || value[2] != felt!(0) {
-        return Err("expected zero padding in the upper three felts");
+/// Extracts a scalar felt from a protocol word with zero-padded trailing limbs.
+#[inline(always)]
+pub fn felt_from_padded_word(value: Word) -> Result<Felt, &'static str> {
```
bitwalker, line 12: Same here

tests/integration/src/rust_masm_tests/examples.rs
```diff
@@ -14,6 +14,37 @@ use proptest::prelude::*;
 
 use crate::{CompilerTest, CompilerTestBuilder, cargo_proj::project, testing::executor_with_std};
 
+/// Asserts that exactly one exported procedure carries `attribute` and that its leaf export name
+/// matches `expected_export_name`.
+fn assert_single_protocol_export(
```
bitwalker: *Nit:* I'd maybe change `single` here to `unique`, since a package might have multiple protocol exports, but what we're asserting here is that there is exactly one instance of an export associated with a specific attribute.

bitwalker (approved):
Looks good! I left a few comments with nits/notes, but am marking this approved as we don't need to do a second round of review if/when you address any of them.

greenhat:
Thank you! I implemented everything.

PR #1055, opened 2026-04-04
fix: update outdated doc paths in rust-toolchain.toml comment

CryptAm:
Update comment paths from docs/src/usage/ to docs/external/src/usage/ to reflect actual repository structure.


bitwalker:
This was recently fixed as part of updating our toolchain - thanks for the contribution!

PR #1063, opened 2026-04-06
test(codegen): extend memory load tests with random addresses

### LINKED ISSUE #469: Extend memory intrinsic tests to also take random addresses as input
This will provide complete coverage of the various ways these intrinsics can be called. Minimal refactoring of the existing tests added in #468 is needed to do this, but it was left to a subsequent PR.

giwaov:
Extends the aligned memory load tests to use randomly generated addresses via proptest, as described in #469.

## Changes

- Added \
andom_word_aligned_addr()\ helper that generates addresses in pages 17..256 with varying word offsets, ensuring they are above the Rust stack and properly aligned
- Updated \load_sw\, \load_dw\, \load_u8\, \load_u16\, and \load_bool\ tests to randomize both the value and the address in each proptest iteration
- Removed the fixed \write_to = 17 * 2u32.pow(16)\ from each updated test

The store tests and unaligned load/store tests still use hardcoded base addresses since the address is compiled into the IR and would require more significant refactoring to parameterize at runtime.

Closes #469

### DIFF
--- a/tests/integration/src/codegen/intrinsics/mem.rs
+++ b/tests/integration/src/codegen/intrinsics/mem.rs
@@ -8,20 +8,26 @@ use midenc_hir::{
     dialects::builtin::{BuiltinOpBuilder, attributes::Signature},
 };
 use proptest::{
-    prelude::any,
+    prelude::{any, Strategy},
     prop_assert_eq,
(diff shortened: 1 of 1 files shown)


bitwalker:
This has been merged manually, but your authorship was preserved

PR #1066, opened 2026-04-07
fix: propagate verification errors and add PassManager integrity test

### LINKED ISSUE #605: Add a test to check that the `PassManager` verifies integrity after modification
### What should be done?

There should be a test that checks whether the `PassManager` verifies the integrity of `Operation`s *after* they are modified. 

### How should it be done?

A test could be added that aims to break an invariant of an Operation, and then checks if the PassManager detects it.

### When is this task done?

When the test correctly detects that the `PassManager` broke an `Operation`'s invariant 

### Additional context

For the full context, see https://github.com/0xMiden/compiler/pull/591

giwaov:
## Summary

Fix a bug in `OpToOpPassAdaptor::run()` where verification errors were
silently dropped after a pass completed successfully. The original code
used `result.map_err(|_| verification_result)` which is a no-op when
`result` is `Ok(())`, meaning verification failures were never propagated
to the caller.

The fix changes this to `result = Err(verification_result)` so that
verification errors are always reported.

## Test

Add a test (`pass_manager_detects_broken_invariant`) that:
1. Creates a function with a `test::Add` operation (a BinaryOp requiring 2 operands)
2. Runs a `CorruptBinaryOp` pass that clears the operands of the first BinaryOp it finds
3. Asserts that the PassManager returns an error when verification is enabled

Without the bug fix, this test would pass silently (the verification error would be dropped).

Closes #605

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,25 +134,25 @@ wasmparser = { version = "0.227", default-features = false, features = [
 ] }
 
 # Workspace crates
-midenc-codegen-masm = { version = "0.8.0", path = "codegen/masm" }
-midenc-dialect-arith = { version = "0.8.0", path = "dialects/arith" }
-midenc-dialect-hir = { version = "0.8.0", path = "dialects/hir" }
-midenc-dialect-scf = { version = "0.8.0", path = "dialects/scf" }
-midenc-dialect-cf = { version = "0.8.0", path = "dialects/cf" }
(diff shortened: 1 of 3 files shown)


bitwalker (approved):

bitwalker:
I manually merged this PR without the fixup commits, authorship of the original commit was preserved

PR #1067, opened 2026-04-08
Fix VM v0.22.1 API breaking changes

greenhat:
The network integration test suite is disabled until #926 is implemented.

### DIFF
--- a/codegen/masm/src/artifact.rs
+++ b/codegen/masm/src/artifact.rs
@@ -233,7 +233,9 @@ impl MasmComponent {
     ) -> Result<AssemblyArtifact, Report> {
         if let Some(entrypoint) = self.entrypoint.as_ref() {
             self.assemble_program(entrypoint, link_libraries, link_packages, session)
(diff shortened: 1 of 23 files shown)


codegen/masm/src/artifact.rs
```diff
@@ -175,19 +175,70 @@ impl fmt::Display for MasmComponent {
     }
 }
 
+/// The library-backed MAST produced by assembling a [`MasmComponent`].
+#[derive(Clone)]
+pub struct AssemblyArtifact {
```
bitwalker, line 180: Note: The assembler can produce Package directly now, when assembling a project. So while this makes sense temporarily, we'll plan to remove it ASAP in favor of project assembly.

bitwalker (approved):

codegen/masm/src/artifact.rs
greenhat: Thanks! I made https://github.com/0xMiden/compiler/issues/1070 for tracking.

PR #1071, opened 2026-04-09
refactor: use project assembler to build a `Package`

### LINKED ISSUE #1070: Use project assembly to build a `Package`
From https://github.com/0xMiden/compiler/pull/1067#discussion_r3054441897

greenhat:
Close #1070 

- Switch MASM package assembly over to the VM project assembler and move the project-specific packaging logic into `codegen/masm::artifact::project_support`.
- Register linked libraries and linked packages through an in-memory package registry so assembly preserves exact dependency resolution, reuses versions already declared by linked packages for raw `-l` libraries, and rejects conflicting external version requirements.
- Normalize assembled package outputs after project assembly by restoring Wasm component-model export paths, carrying forward account component metadata and rodata advice-map entries, and using the filtered user module set to derive the library target namespace.
- Add regression coverage for dependency-version inference/conflicts and for keeping package manifest exports aligned with the assembled library exports.


codegen/masm/src/artifact/project_support.rs
```diff
@@ -0,0 +1,686 @@
...
+    store: &mut InMemoryPackageRegistry,
+    dependencies: &mut BTreeMap<Arc<str>, ProjectDependency>,
+    link_packages: &BTreeMap<Symbol, Arc<MastPackage>>,
+) -> Result<(), Report> {
+    let mut pending = BTreeMap::<PackageId, Arc<MastPackage>>::default();
+    for package in link_packages.values() {
+        pending.insert(package.name.clone(), package.clone());
+    }
+
+    while !pending.is_empty() {
```
greenhat, line 222: @bitwalker I'm not happy with this "brute force" approach, but the registry requires the dependencies to be registered in topological order.

bitwalker: We can provide our own registry implementation, which lets us just stick all the packages we know about into the store right out the gate - there isn't actually anything special about `InMemoryPackageRegistry`, it's basically a simple wrapper around a `BTreeMap`.

That said, this works for now, and it'll change when we handle projects in the frontend anyway, so I think we can just leave this as-is.

bitwalker (approved):
I think this works fine until we update the frontend to support projects - I'm working on some stuff there now, so I can handle that part while I'm in there.

I'll work on getting `miden-core-lib` and `miden-protocol` carrying packages rather than libraries, and we can strip out the need for handling `Library` at all. I need to think about the best way to handle package resolution moving forward - I think all the cases where we'd want to pass link libraries on the command line, we _wouldn't_ be compiling a Miden project, but something ad-hoc, in which case synthesizing a project manifest solves the resolution problem. We'd want to reject specifying link libraries when a Miden project manifest is present though, since all dependencies should go there, and there isn't really any benefit to passing them manually in those cases - but it's something we may need to revisit in the future.

codegen/masm/src/artifact/project_support.rs
greenhat: Funny enough, initially I did make my own registry. But then I noticed `InMemoryPackageRegistry` and in the spirit of DRY decided to go with it. 😄

Issue #1024, opened 2026-03-23
Implement `trace`-based printing infrastructure


bitwalker:
### What should be done?

The Miden VM does not directly provide any stdio functionality, but it does provide certain escape hatches for hosts to provide custom behavior for MASM programs being executed. The primary ones are `emit` (for events) and `trace` (for discardable instrumentation/events). The `Host` callbacks for these can access the processor state, and in the case of `emit`, also produce side effects via the advice provider.

The downside to `emit`, is that it is an actual operation in MAST, which means that when stripping debug info/instrumentation, `emit` would remain in the program. The `trace` "instruction" however, is actually a decorator, not an operation, and so it is stripped in those scenarios. This makes it particularly ideal for debug and instrumentation use cases where it is desirable to strip the instrumentation in release builds.

We want to implement the ability to print Rust strings during execution using `trace` events.

### How should it be done?

The implementation I've sketched out consists of implementing a new compiler intrinsic (and corresponding primop in the `hir` dialect) that prints a Rust string reference (i.e. `&str`, a fat pointer), by emitting `trace.PRINT_TRACE_ID` while the string reference is on top of the operand stack.

The intrinsic signature in Rust would be `unsafe extern "C" fn __println(ptr: *const u8, len: usize)`, which would translate to Wasm as `func $println (param i32 i32)`, where the first parameter is the pointer to the start of the string, the second is the length. This would be converted to the primop, i.e. `hir.println`, by the frontend.

We'd surface the low-level plumbing for this in the SDK via:

```rust
pub fn println(s:  &str) {
    unsafe {
        let bytes = s.as_bytes();
        __println(bytes.as_ptr(), bytes.len());
    }
}
```

A related task to this one is providing our own `println!` macro that formats and prints a string using the `println` intrinsic.

The host side of all this would live in `miden-debug`, by handling `PRINT_TRACE_ID` via the `on_trace` callback as follows:

1. Extract the string reference components from the top of the operand stack, accessing it via `ProcessState`
2. Validate that the string reference meets certain criteria (i.e. the memory address is in the addressable range, the size of the string is no larger than some soft limit)
3. Read the raw bytes from linear memory via `ProcessState`
4. Convert the bytes to a Rust string using `core::str::from_utf8`, and then print them to stdout

### When is this task done?

With the plumbing above implemented, there are two objectives here:

1. Calling `println` from Rust prints the string correctly to stdout when executed under the debug executor
[...]

PR #1075, opened 2026-04-10
feat: add lowering for `hir::ConstantPointer`

mooori:
In my implementation for #1024 there's a `hir::ConstantPointer` reaching masm codegen. This PR adds lowering for it.

Or is there no lowering on purpose and the issue is that `ConstantPointer` shouldn't reach masm codegen?


bitwalker (approved):
Looks good! We just weren't ever observing it prior to this, hence why it wasn't lowered yet

PR #1077, opened 2026-04-13
feat: enable trace based `println`

mooori:
Context: #1024 

Adds trace-based println support across the compiler pipeline:

- Introduces a new `hir` `PrintLn` op and a `MASM TRACE_PRINT_LN` event
- Adds lowering for `PrintLn` that manages `addr, size` of the string on the operand stack, such that evaluator and debug executor can pick it up.
- Adds WASM debug intrinsics and the SDK `println(&str)` helper to generate that op from Rust.
- Extends the evaluator/debug path to read bytes from memory, decode UTF-8, and collect printed
  lines.

Companion PR in `miden-debug`: [54](https://github.com/0xMiden/miden-debug/pull/54)

### TODO

- [ ] Remove patched `miden-debug` GH dependency once changes in `miden-debug` were released


dialects/hir/src/ops/primop.rs
```diff
@@ -76,3 +76,21 @@ pub struct MemCpy {
...
+#[derive(EffectOpInterface, OpPrinter, OpParser)]
+#[operation(
+    dialect = HirDialect,
+    implements(MemoryEffectOpInterface, OpPrinter)
+)]
+pub struct PrintLn {
+    // There is no write, but without `MemoryEffect::Write`, the `PrintLn` is not lowered to masm.
+    // With `Read` only, `PrintLn` probably gets eliminated since it is not producing a result.
+    #[operand]
+    #[effects(MemoryEffect(MemoryEffect::Read, MemoryEffect::Write))]
```
mooori, line 92: Is there another way to prevent `PrintLn` getting eliminated? Using a fake `Write` is not ideal

eval/src/eval.rs
```diff
@@ -732,6 +732,53 @@ impl Eval for hir::MemCpy {
     }
 }
 
+impl Eval for hir::PrintLn {
```
mooori, line 735: Looks like the evaluator is currently only usable as lib, not as cli. In that case I suppose it's sufficient to make printed lines available programmatically. So no logging/printing to stdout here.

dialects/hir/src/ops/primop.rs
greenhat: Create a new effect? I mean this operation produces a side effect that is handled in the debug executor (prints/logs/whatever).

dialects/hir/src/ops/primop.rs
```diff
@@ -76,3 +76,21 @@ pub struct MemCpy {
...
+
+/// Prints a string to the debug output.
+///
+/// The string bytes are read from memory at the given pointer address and length.
+#[derive(EffectOpInterface, OpPrinter, OpParser)]
+#[operation(
+    dialect = HirDialect,
+    implements(MemoryEffectOpInterface, OpPrinter)
+)]
+pub struct PrintLn {
```
greenhat, line 88: Would it be crazy to think that we should consider a new dialect for this operation?

greenhat (commented):
Looking good! Please see my notes.

dialects/hir/src/ops/primop.rs
bitwalker: I would avoid adding a new effect for just this, and instead remove the effect annotation entirely - this will force analysis to be conservative with respect to this op (i.e. it cannot be presumed to be side-ffect free, and thus removable - and ops with effects cannot be reordered past it).

The effect mechanism is specifically there to opt-in operations to optimization, by declaring that the specific effects those ops have, and thus allowing the optimizer to make decisions about scheduling/reordering/etc. based on the details of those effects. In cases where we explicitly do not want the optimizer to be optimizing a specific op (and especially removing it due to perceived lack of use), it is best to avoid annotating effects at all.

eval/src/eval.rs
bitwalker: Correct, it's intended use is primarily for verifying the behavior of the IR without lowering all the way to MASM, in cases where we can evaluate a program separately from Miden VM (for obvious reasons, we can't use the evaluator with things that require linking against MASM code, or that are dependent on details of the Miden VM that are not defined as part of the IR).

dialects/hir/src/ops/primop.rs
bitwalker: My thought was to place it in the debug dialect that @djolertrk created/was going to create - but since that isn't present in `next`, I think it is fine to put it in `hir` initially, and we can move it later.

codegen/masm/src/emit/primop.rs
```diff
@@ -265,6 +265,26 @@ impl OpEmitter<'_> {
         self.push(ty);
     }
 
+    /// Emit a trace marker for `hir.println`, leaving the operands intact until after the trace.
+    pub fn println(&mut self, span: SourceSpan) {
+        // Leave them on the stack so debug executor can read them.
```
bitwalker: I think this comment is misleading - when we emit ops, the input operands are semantically consumed by the operation we emit as MASM - but in cases like this, where there is no actual op that consumes the operands (in fact, `trace` is not an op at all, but a decorator), it is the `drop drop` that acts as the "real" operation consuming the operands.

So it isn't that we are leaving them on the stack - we aren't - we are consuming the operands via `drop drop` below. The reason why we don't `pop` the operands here like we do elsewhere, is because we want to let the call to `dropn(2)` handle it.

codegen/masm/src/events.rs
```diff
@@ -11,12 +11,21 @@ pub const TRACE_FRAME_START: u32 = 0xf0;
 /// The mnemonic here is F = frame, C = close
 pub const TRACE_FRAME_END: u32 = 0xfc;
 
+/// This event is emitted via `trace`, and indicates that a line should be printed.
+///
+/// The bytes representing the string are expected in memory. The executor reads the start address
+/// and length from the operand stack.
+///
+/// The mnemonic here is ASCII `PLN`.
+pub const TRACE_PRINT_LN: u32 = 0x50_4c_4e;
```
bitwalker: As mentioned in the `miden-debug` PR, I think we should use `PRNT` as the mnemonic

sdk/sdk/src/debug.rs
bitwalker, line 1: I would suggest we also define the `println!` macro, as follows:

```rust
#[macro_export]
macro_rules! println {
    ($message:literal) => {{
        $crate::debug::println($message);
    }};

    ($format:literal, $($arg:tt),+) => {{
        let message = ::alloc::format!($format, $($arg),*);
        $crate::debug::println(&message);
    }}
}
```

Then in downstream crates of the SDK, we can do stuff like:

```rust
use miden_sdk::println;

println!("hello world");

# or

println!("{greeting} {}", greeting = "hello", "world");
```

sdk/sdk/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 #![no_std]
 #![deny(warnings)]
 
+pub mod debug;
+
+pub use debug::println;
```
bitwalker: Let's not export `println` from the root like this - only the `println` macro I outlined above would be exported from the root

sdk/stdlib-sys/src/intrinsics/debug.rs
```diff
@@ -14,10 +17,26 @@ pub fn breakpoint() {
...
+
 /// Sets a breakpoint in the emitted Miden Assembly at the point this function is called.
 #[inline(always)]
 #[track_caller]
 #[cfg(not(all(target_family = "wasm", miden)))]
 pub fn breakpoint() {
     unimplemented!("debug intrinsics are only available when targeting the Miden VM")
 }
+
+/// Prints the string pointed to by `ptr` in the debug executor.
```
bitwalker, line 37: Functions with multiple cfg-dependent definitions should be defined next to each other when possible.

bitwalker (changes requested):
Looks good overall! I left a few notes on changes I'd like to make, but once this is ready, we'll merge it after `miden-debug` ships support for it.

sdk/sdk/src/debug.rs
mooori: Using `format!` is not yet possible as it hits `call_indirect`. I've opened #1102 to track this.

dialects/hir/src/ops/primop.rs
mooori: #1103

sdk/stdlib-sys/src/intrinsics/debug.rs
mooori: Could you clarify what to change? To me it looks like functions are grouped by cfgs.

sdk/sdk/src/debug.rs
bitwalker: That's fine, we can just add the literal-only case for now, and extend it with formatting facilities later

bitwalker: I would still add the format arguments case, but emit `compile_error!("unsupported use of println! with format arguments")` internally

sdk/stdlib-sys/src/intrinsics/debug.rs
bitwalker: The `breakpoint` function is between the two `println` definitions with different `cfg` criteria, I'd normally expect them to be right next to each other so that I can visually compare their behaviors more easily

mooori: Ah now I understand, thanks

codegen/masm/src/emit/primop.rs
mooori: Fixed by 6e92b04?

sdk/sdk/src/debug.rs
mooori: Done and I've also added the following arm:

```
    ($message:expr) => {{
        $crate::debug::println($message);
    }};
```

to enable using the macro as in `debug_executor_println_dynamic.rs`.

mooori:
I've addressed the review comments and rebased on latest `next` to fix build failures related to #1085. Old commits weren't modified, new changes start with commit b111edc.

With `println` now being emitted to the logger and the `trace.printed_lines` field removed, testing has become a bit tricky. It's documented in `tests/integration/tests/README.md`.

bitwalker (approved):

PR #1083, opened 2026-04-20
feat(objtool): add `masm2masp` command

mooori:
Enables generating `masp` from `masm`.

With this we can compare `masp` built from handwritten `masm` to `masp` generated by the compiler from Rust code.

### DIFF
--- a/tools/objtool/Cargo.toml
+++ b/tools/objtool/Cargo.toml
@@ -16,5 +16,12 @@ edition.workspace = true
 [dependencies]
 anyhow.workspace = true
 clap.workspace = true
+miden-assembly = { workspace = true, features = ["std"] }
 miden-core.workspace = true
 miden-mast-package.workspace = true
+miden-package-registry = { workspace = true, features = ["std", "resolver"] }
+miden-project = { workspace = true, features = ["std", "serde"] }
+miden-protocol.workspace = true
+
+[dev-dependencies]
+tempfile = "3"
--- a/tools/objtool/src/main.rs
+++ b/tools/objtool/src/main.rs
@@ -1,4 +1,5 @@
 mod decorators;
+mod masm2masp;
 
 use anyhow::Result;
 use clap::{Parser, Subcommand};
@@ -21,6 +22,8 @@ struct Cli {
 enum Commands {
     /// Compare serialized MAST forest sizes after stripping decorators.
     Decorators(decorators::DecoratorsCommand),
+    /// Convert a .masm file to a .masp package
+    Masm2masp(masm2masp::Masm2MaspCommand),
 }
 
 fn main() -> Result<()> {
@@ -28,5 +31,6 @@ fn main() -> Result<()> {
 
     match cli.command {
         Commands::Decorators(command) => decorators::run(command),
+        Commands::Masm2masp(command) => masm2masp::run(command),
     }
 }
--- a/tools/objtool/src/masm2masp.rs
+++ b/tools/objtool/src/masm2masp.rs
@@ -1,0 +1,266 @@
+use std::{fs, path::PathBuf, sync::Arc};
+
+use anyhow::{Context, Result};
+use clap::{ArgAction, Args};
+use miden_assembly::{Assembler, Path, ProjectTargetSelector};
+use miden_package_registry::InMemoryPackageRegistry;
+use miden_project::{Package, Target, TargetType};
+use miden_protocol::ProtocolLib;
+
+#[derive(Debug, Clone, Args)]
+pub struct Masm2MaspCommand {
+    /// Path to the input .masm file
+    pub masm_path: PathBuf,
+    /// Target type (e.g. account-component)
+    pub target_type: String,
+    /// Path of output directory to create, must *not* yet exist
+    // TODO since using `virtual` target, try without creating out_dir and manifest file
+    pub out_path: PathBuf,
+    /// Don't link against the default protocol library
+    #[arg(long = "no-protocol-lib", action = ArgAction::SetFalse, default_value_t = true)]
+    link_protocol_lib: bool,
+}
+
+pub fn run(command: Masm2MaspCommand) -> Result<()> {
+    let target_type: TargetType = command
+        .target_type
+        .parse()
+        .map_err(|e| anyhow::anyhow!("failed to parse target_type: {e}"))?;
+    if target_type != TargetType::AccountComponent {
+        // TODO add support for more `TargeType`s as needed
+        anyhow::bail!("only AccountComponent target type is currently supported");
+    }
+
+    if command.out_path.exists() {
+        anyhow::bail!("output directory '{}' already exists", command.out_path.display());
+    }
+
+    fs::create_dir_all(&command.out_path)
+        .with_context(|| format!("failed to create directory '{}'", command.out_path.display()))?;
+    // TODO cleanup the output directory if any of the below fails
+
+    let masm_filename = command
+        .masm_path
+        .file_stem()
+        .and_then(|s| s.to_str())
+        .with_context(|| format!("invalid masm file path '{}'", command.masm_path.display()))?;
+
+    let package_name = masm_filename.to_string();
+    let namespace_str = masm_filename.replace('-', "::");
+
+    let namespace_path: Arc<Path> = Path::new(&namespace_str).to_path_buf().into();
+    let target = Target::r#virtual(target_type, package_name.as_str(), namespace_path)
+        .with_path(format!("{masm_filename}.masm"));
+
+    let version = miden_project::semver::Version::new(0, 1, 0);
+    let package = Package::new(package_name.as_str(), target).with_version(version);
+
+    let manifest_toml = package
+        .to_toml()
+        .map_err(|e| anyhow::anyhow!("failed to generate manifest: {e}"))?;
+
+    let manifest_path = command.out_path.join("miden-project.toml");
+    fs::write(&manifest_path, &manifest_toml)
+        .with_context(|| format!("failed to write '{}'", manifest_path.display()))?;
+
+    let masm_dest = command.out_path.join(format!("{masm_filename}.masm"));
+    fs::copy(&command.masm_path, &masm_dest)
+        .with_context(|| format!("failed to copy masm file to '{}'", masm_dest.display()))?;
+
+    let mut store = InMemoryPackageRegistry::default();
+    let mut assembler = Assembler::default();
+    if command.link_protocol_lib {
+        // TODO record the protocol dependency in `miden-project.toml`
+        assembler = assembler
+            .with_dynamic_library(ProtocolLib::default())
+            .map_err(|e| anyhow::anyhow!("failed to link protocol library: {e}"))?;
+    }
+    let mut project_assembler = assembler
+        .for_project_at_path(&manifest_path, &mut store)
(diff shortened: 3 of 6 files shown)


mooori:
In the meantime I've learned that `cargo miden / midenc` will support `masm -> masp`. So `objtool masm2masp` is at most an ad-hoc tool to do some benchmarking of compilation artifact sizes. Hence there are some unresolved `TODO`s left in the diff.

@bitwalker could you take a look if this is correctly assembling `masp` from `masm`? That's handled in `fn run` in `masm2masp.rs`. Then we could report some numbers regarding the size difference of compiled vs handwritten programs.

I've tested `objtool masm2masp` on the handwritten [basic wallet](https://github.com/0xMiden/protocol/blob/6e78e3a6c31778adcb753b8bce83606b072be02b/crates/miden-standards/asm/standards/wallets/basic.masm).

tools/objtool/src/main.rs
```diff
@@ -21,12 +22,15 @@ struct Cli {
 enum Commands {
     /// Compare serialized MAST forest sizes after stripping decorators.
     Decorators(decorators::DecoratorsCommand),
+    /// Convert a .masm file to a .masp package
+    Masm2masp(masm2masp::Masm2MaspCommand),
```
bitwalker, line 26: Let's just name this subcommand `assemble`/`asm`, since that's basically what it is doing, i.e. assembling an ad-hoc MASM file or project

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+
+use anyhow::{Context, Result};
+use clap::{ArgAction, Args};
+use miden_assembly::{Assembler, Path, ProjectTargetSelector};
+use miden_package_registry::InMemoryPackageRegistry;
+use miden_project::{Package, Target, TargetType};
+use miden_protocol::ProtocolLib;
+
+#[derive(Debug, Clone, Args)]
+pub struct Masm2MaspCommand {
```
bitwalker, line 11: I think we should probably tweak the interface of this a bit, to allow any of the following:

1. `objtool asm path/to/foo.masm`
2. `objtool asm path/to/miden-project.toml --target <target_type>`

In the case of 1, we will require that the MASM source file correspond to a program module (i.e. contains a `begin` block). The name of the package will be derived from the basename of the `.masm` file, and the version set to `0.0.0`.

In the case of 2, what we do depends on what is specified in the `miden-project.toml` file for the requested target:

1. If the specified target specifies a `path`, then we know that the target is a MASM source project, and the assembler can handle all the details of assembling the target itself, we need only construct a `ProjectAssembler` and tell it to assemble the requested target.
2. If the specified target has no `path`, then we know that the target is written in a language other than MASM (i.e. Rust), and can punt to the compiler to handle compiling to a package via `cargo miden build --manifest-path path/to/Cargo.toml` (i.e. we presume that the directory containing the `miden-project.toml` also contains a `Cargo.toml`).

This is basically the functionality that we will provide via `miden build` itself, so once we've implemented assembly support there, we can basically replace all uses of `objtool asm` with `miden build` (or keep it around as a lower-level plumbing tool for our own uses, if we feel it has value for that).

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// Path to the input .masm file
+    pub masm_path: PathBuf,
+    /// Target type (e.g. account-component)
+    pub target_type: String,
```
bitwalker, line 15:
```suggestion
    #[arg(long, default_value_t = TargetType::Executable)]
    pub target_type: TargetType,
```

The `TargetType` enum implements `FromStr`, so you should be able to use it as shown here, though you may have to fiddle around with the `value_parser` option a bit if it doesn't just work as-is

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// Path of output directory to create, must *not* yet exist
```
bitwalker, line 16: Why?

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+    // TODO since using `virtual` target, try without creating out_dir and manifest file
+    pub out_path: PathBuf,
+    /// Don't link against the default protocol library
```
bitwalker, line 19: We can remove this entirely, and require use of a `miden-project.toml` file when dependencies are required

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+pub fn run(command: Masm2MaspCommand) -> Result<()> {
+    let target_type: TargetType = command
+        .target_type
+        .parse()
+        .map_err(|e| anyhow::anyhow!("failed to parse target_type: {e}"))?;
+    if target_type != TargetType::AccountComponent {
+        // TODO add support for more `TargeType`s as needed
+        anyhow::bail!("only AccountComponent target type is currently supported");
```
bitwalker, line 31: Is there a particular reason for this? There shouldn't be any reason why we can't support all target types out of the box when a `miden-project.toml` is provided (it is only when assembling a standalone `.masm` file where we must reject anything other than executable or library type.

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+
+#[derive(Debug, Clone, Args)]
+pub struct Masm2MaspCommand {
+    /// Path to the input .masm file
+    pub masm_path: PathBuf,
+    /// Target type (e.g. account-component)
+    pub target_type: String,
+    /// Path of output directory to create, must *not* yet exist
+    // TODO since using `virtual` target, try without creating out_dir and manifest file
+    pub out_path: PathBuf,
```
bitwalker, line 18: I would make `out_path` a file path, not a directory, and rename it to `output_file` instead, and make it optional (if not given, we can derive the `output_file` path from the input project/source file name, and place it in the current directory).

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+    let mut store = InMemoryPackageRegistry::default();
+    let mut assembler = Assembler::default();
+    if command.link_protocol_lib {
+        // TODO record the protocol dependency in `miden-project.toml`
+        assembler = assembler
+            .with_dynamic_library(ProtocolLib::default())
+            .map_err(|e| anyhow::anyhow!("failed to link protocol library: {e}"))?;
+    }
+    let mut project_assembler = assembler
+        .for_project_at_path(&manifest_path, &mut store)
```
bitwalker, line 79: You don't need to write a `miden-project.toml` to disk, you can just use `Assembler::for_project` and provide the `miden_project::Package` directly, since you've already constructed it in memory.

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    let masm_dest = command.out_path.join(format!("{masm_filename}.masm"));
+    fs::copy(&command.masm_path, &masm_dest)
+        .with_context(|| format!("failed to copy masm file to '{}'", masm_dest.display()))?;
+
+    let mut store = InMemoryPackageRegistry::default();
+    let mut assembler = Assembler::default();
+    if command.link_protocol_lib {
+        // TODO record the protocol dependency in `miden-project.toml`
+        assembler = assembler
```
bitwalker, line 74: This is not needed - the project assembler knows how to resolve dependencies, so long as they are available in the configured package store, and specified in the `miden-project.toml` manifest (i.e. `miden_project::Package`).

We should probably add flags to `objtool asm` to specify the path to packages on disk to seed the store with, since we do not want `objtool` to be depending on `miden-protocol` or `miden-core-lib` directly. Instead, we can just tell `objtool asm` to seed its package registry with a package using one or more instances of `-L path/to/lib.masp`. This will make `objtool asm` a lot more flexible as a tool IMO.

tools/objtool/src/masm2masp.rs
```diff
@@ -0,0 +1,266 @@
...
+            .with_dynamic_library(ProtocolLib::default())
+            .map_err(|e| anyhow::anyhow!("failed to link protocol library: {e}"))?;
+    }
+    let mut project_assembler = assembler
+        .for_project_at_path(&manifest_path, &mut store)
+        .map_err(|e| anyhow::anyhow!("failed to load project: {e}"))?;
+
+    let package = project_assembler
+        .assemble(ProjectTargetSelector::Library, "release")
```
bitwalker, line 83: We should derive `ProjectTargetSelector` using:

```rust
if target_type.is_executable() {
    ProjectTargetSelector::Executable
} else {
    ProjectTargetSelector::Library
}
```

tools/objtool/Cargo.toml
```diff
@@ -16,5 +16,12 @@ edition.workspace = true
 [dependencies]
 anyhow.workspace = true
 clap.workspace = true
+miden-assembly = { workspace = true, features = ["std"] }
 miden-core.workspace = true
 miden-mast-package.workspace = true
+miden-package-registry = { workspace = true, features = ["std", "resolver"] }
+miden-project = { workspace = true, features = ["std", "serde"] }
+miden-protocol.workspace = true
```
bitwalker, line 24: We don't want to depend on `miden-protocol` directly, nor `miden-core-lib`, which appears to be missing here, but would also be a required dependency for most anything non-trivial.

bitwalker (changes requested):
I think this is headed in the right direction, but I'd like to make a few changes (see commentary) that will put us in a better position once this lands:

* I'd like this to more closely align with the semantics we'd implement via `miden build`, but with special-case handling for assembling standalone `.masm` files as done here. This will make it more natural to use `objtool asm` in places where `miden build` would normally be used, but without the dependency on `midenup`. Ultimately, this behavior will be provided by `midenc`/`cargo miden`, so aligning the interface with how we eventually expect things to look will simplify things a bit.
* We don't have to actually write a project to disk to assemble a project - the manifest can be completely created in-memory and assembled from that without ever writing it to disk. That simplifies things a bit, since we don't need to deal with all the complexity that falls out of trying to create a "real" project.
* To be maximally useful, we will want to be able to assemble any target type supported by the project assembler, but since we can require that only executable/library are used in conjunction with a standalone `.masm` file, and all the complexity of other target types is handled via `miden-project.toml`, there isn't actually anything we need to do other than plumb the target type through.

tools/objtool/src/masm2masp.rs
bitwalker: As a follow-up note - we can also just simply reject assembly of projects that aren't MASM source projects for now (i.e. case 2 described above when a path to a `miden-project.toml` is given). That would overlap too much with what `cargo miden`/`midenc` are responsible for, and the primary use of this is to assemble MASM sources anyway.

lima-limon-inc:
As a sidenote, while working on migrating masm code to rust, I wrote a small CLI tool to display a `masp` file's metadata https://github.com/igamigo/miden-tools/tree/main/snowberry to ease debugging.

Just throwing it out there, maybe it helps!

bitwalker:
@lima-limon-inc If you'd like to port that functionality to `objtool`, I think it'd make a good addition!

mooori:
This is no longer needed as with #1135 `midenc` can compile MASM.

PR #1086, opened 2026-04-23
fix: route successor groups by explicit index to avoid group-0 collision

### LINKED ISSUE #1084: `index out of bounds` error in `EntityStorage::group`
## Bug Report

`cargo-miden 0.8.x` panics during `.masp` packaging when an account component has both `StorageValue` and `StorageMap` fields.

## Environment

- **cargo-miden**: 0.8.0 and 0.8.1 (both affected)
- **miden SDK crate**: 0.12.0
- **midenc-hir**: 0.8.1 (where the panic occurs)
- **Rust toolchain**: nightly-2025-12-10
- **Platform**: macOS (Darwin 24.6.0, aarch64)

### Minimal contract

```rust
#![no_std]
#![feature(alloc_error_handler)]

use miden::*;

#[component]
struct MyAccount {
    #[storage(description = "a value slot")]
    config: StorageValue<Word>,

    #[storage(description = "a map slot")]
    board: StorageMap<Word, Felt>,
}

#[component]
impl MyAccount {
    pub fn get_config(&self) -> Word {
        self.config.get()
    }

    pub fn get_cell(&self, key: Word) -> Felt {
        self.board.get(key)
    }
}
```

### Cargo.toml

```toml
[package]
name = "my-account"
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib"]

[dependencies]
miden = { version = "0.12" }

[package.metadata.component]
package = "miden:my-account"

[package.metadata.miden]
project-kind = "account"
supported-types = ["RegularAccountImmutableCode"]
```

### Build command

```
cargo miden build --release
```

### Output

```
   Compiling my-account v0.1.0
    Finished `release` profile [optimized] target(s)
Creating Miden package .../target/miden/release/my_account.masp

thread 'main' panicked at /.../.cargo/registry/src/.../midenc-hir-0.8.1/src/ir/entity/storage.rs:180:31:
index out of bounds: the len is 1 but the index is 1
```

The `.masp` file is **not** created.

## Analysis

The panic is at `midenc-hir-0.8.1/src/ir/entity/storage.rs:180`:

```rust
pub fn group(&self, group: usize) -> EntityRange<'_, T> {
    EntityRange {
        range: self.groups[group].as_range(),  // <-- panics here
        items: self.items.as_slice(),
    }
}
```

The `GroupedEntityStore` is initialized with 1 group (default), but the code tries to access `self.groups[1]` (index 1) when processing the `StorageMap` slot. It appears the second group (for map-type slots) is never created during metadata construction.

## Observations

- **Components with only `StorageValue` fields**: build and package fine.
- **Components with only `StorageMap` fields**: not tested in isolation.
- **Components with both `StorageValue` AND `StorageMap`**: always panic during packaging.
- **Note scripts that depend on affected components**: also fail because `cargo-miden` rebuilds the dependency component first and panics there.
- **Note scripts with no component dependencies** (e.g., a no-op result carrier): build fine.
- Reordering storage fields (map first vs. last) does not help.

## Impact

This blocks building any non-trivial account component that uses both value slots and storage maps — a very common pattern (e.g., game state + board data, token metadata + balances, etc.). It prevents migration from `miden` SDK 0.10 to 0.12 for projects that use `StorageMap`.

## Workaround

None found. Both `cargo-miden` 0.8.0 and 0.8.1 use the same `midenc-hir 0.8.1` and exhibit the same panic.

greenhat:
Close #1084 

`with_successors`/`with_keyed_successors` picked their target successor
group by checking `op.successors.is_empty()`, which only inspects items.
A switch produced with no cases (e.g. when `SimplifySwitchFallbackOverlap`
collapses every case into the fallback) left the default empty group 0
in place, so the follow-up fallback call saw `is_empty` still true and
was written into group 0 instead of group 1 — later panicking in
`Switch::fallback` when the accessor indexed the missing group 1.

Pass the group index from the macro (matching `with_operands_in_group`)
and have `with_successors_in_group`/`with_keyed_successors_in_group`
use `extend_group(group_index, …)` unconditionally, so the builder and
the macro-generated accessors share the same static group numbering by
construction.


hir/src/ir/operation/builder.rs
```diff
@@ -172,7 +172,13 @@ where
         });
     }
 
-    pub fn with_successors<I>(&mut self, succs: I)
+    /// Record a successor group at `group_index`, creating intervening groups as needed.
+    ///
+    /// The explicit `group_index` mirrors [`Self::with_operands_in_group`] and ensures the group
+    /// this call commits matches the index the macro-generated accessors use. Relying on the
+    /// successor storage's `is_empty` state to pick the target group is unsound — an earlier
+    /// empty group is indistinguishable from the pristine default state.
```
bitwalker:
```suggestion
    /// this call commits matches the index the macro-generated accessors use.
```

I'd remove the part starting with "Relying on", since that's not relevant to the caller in this case.

bitwalker (approved):
LGTM - just a nit about the one doc comment, but otherwise 👍

PR #1087, opened 2026-04-24
Add `midenc-fuzza` differential fuzzing harness

greenhat:
It's a PoC of a test suite where an agent is generating Rust programs for differential testing using the compiler's code coverage as feedback


tests/integration/src/end_to_end/differential/harness.rs
```diff
@@ -0,0 +1,142 @@
+//! Differential fuzzing harness for the Miden compiler.
+//!
+//! Each test case, checked in under `src/cases/`, is just the body of a
+//! `#[unsafe(no_mangle)] pub extern "C" fn entrypoint(u32, u32) -> u32`
+//! plus any helpers it needs. The harness prepends a fixed header
+//! (`#![no_std]` + `#[panic_handler]`) before writing the case as `src/lib.rs`
+//! of a generated cargo project, builds it twice — natively as a host `cdylib`
+//! and via `cargo-miden` to a MASM package — and compares outputs across
```
bitwalker, line 8: Isn't this redundant with our `proptest`-based tests (which have the added benefit of shrinking to find minimal repros)? Or is it meant more as a proof-of-concept, and the intended use would be for programs where `proptest` isn't as well suited vs a traditional fuzzer?

tests/integration/src/end_to_end/differential/harness.rs
```diff
@@ -0,0 +1,142 @@
...
+        .build();
+    let dylib_path = build_host_cdylib(&native_proj.root(), &pkg_name);
+
+    let lib = unsafe { libloading::Library::new(&dylib_path) }
+        .unwrap_or_else(|e| panic!("failed to load {}: {e}", dylib_path.display()));
+    type EntryFn = unsafe extern "C" fn(u32, u32) -> u32;
+    let entry: libloading::Symbol<EntryFn> = unsafe { lib.get(b"entrypoint\0") }
+        .unwrap_or_else(|e| panic!("missing `entrypoint` in {}: {e}", dylib_path.display()));
+
+    // Proptest: 16 cases, shrinking disabled — the whole case file IS the
```
bitwalker, line 57: I'm not sure that's actually true? The primary purpose of shrinking is to find the simplest input which triggers an issue, _not_ to find the smallest program that reproduces the issue. As far as I can tell here, the case file is just the program, not the minimal inputs.

Granted, if your only inputs to a program are just plain `u32` integer values, shrinking provides less value - but it can still reduce the failing input to something like `(u32::MAX, 0, 0)`, instead of a random set of values like `(u32::MAX, 123439, 8631234)`, which can often highlight what input is the problem (or at least remove some randomness from the inputs that obscure things unnecessarily).

tests/integration/src/end_to_end/differential/harness.rs
```diff
@@ -0,0 +1,142 @@
...
+
+[lib]
+crate-type = ["cdylib"]
+
+[profile.release]
+opt-level = 3
+panic = "abort"
+
+[profile.dev]
+panic = "abort"
```
bitwalker, line 115: If you compile a `cdylib` with `panic = "abort"` and then dynamically-load it into the current process, and execution hits a panic, you'll crash the whole process without having a way to catch the panic (because it will abort, rather than unwind as expected).

tests/integration/src/end_to_end/differential/harness.rs
greenhat: It's a PoC of using an agent to generate Rust programs using the compiler's code coverage as feedback. I missed a PR description.

tests/integration/src/end_to_end/differential/harness.rs
greenhat: The comment is a bit off. My reason is two-fold:
1. The shrinking generates a lot of noise that messes up the feedback for the agent.
2. I want to capture the exact inputs that triggered the miscompilation. Shrunk inputs might trigger another code path (another miscompilation?).

tests/integration/src/end_to_end/differential/harness.rs
greenhat: Thank you! I hadn't thought about that.

greenhat:
I'm pretty happy with how it turned out. During the tuning runs, it discovered #1093, #1094 and #1095.
The instructions on how to launch an agent are in the `tests/fuzza/README.md`.

bitwalker (approved):
Looks good! My only request is that we move this under `tools` (I want `tests` to contain only actual test sources, not tooling used by the test suite). We still have a pending task to clean up the organization of `tests` in general, but I mostly want to avoid adding new stuff under `tests` unless that is the only place that makes sense for it.

I'm marking this approved contingent on that move, so I don't hold things up while I'm offline

bitwalker (commented):
As an aside - can we have the test cases emitted to the `tests` directory under some appropriate test crate? Those don't feel like they belong under the `miden-fuzza` tool itself.

Ideally, `miden-fuzza` can be used to generate the cases, and then they would be executed by our standard test suite, without the need to use `miden-fuzza` for that. In other words, `miden-fuzza` solves the problem of identifying interesting test cases, generating them, validating whether they capture a real regression or improve test coverage, and then those generated test cases get merged into our standard test suite.

AIUI, that totally fits within the intended usage of this new tool, but let me know if I'm missing anything

greenhat:
Good point! I'll split it.

Issue #532, opened 2025-05-27
Explore the alternatives(removing) to explicitly passing low-level Miden SDK WIT files


greenhat:
Related to #531 

### Motivation

There should be a better way to make provide the imports for Miden SDK extern functions to `wit-component` than writing and maintaining WIT interface files.

### Potential options

1. Make a core Wasm module that exports functions (either stubs or doing the ABI transformation) and try to pass it to `wasm-ld`. The benefit of this approach is to eliminate the component layer when calling Miden stdlib and tx kernel functions.

2. Adapters in `wasm-ld` at https://github.com/bytecodealliance/wasm-component-ld/blob/main/src/lib.rs

3. Take a look an `wasi-sdk` and how it's linked.

greenhat:
Also, stub libraries - https://lld.llvm.org/WebAssembly.html

greenhat:
[component.wat.txt](https://github.com/user-attachments/files/21750998/component.wat.txt)

@bitwalker I did some digging, and here is what I found out.

### Wasm Component Model Adapter

In the `wit-component` encoder, the code that classifies the core Wasm module import [code](https://github.com/bytecodealliance/wasm-tools/blob/a2d966c7ba79497953375006a2075a6ed52e9cfc/crates/wit-component/src/validation.rs?plain=1#L487-L748) is pretty straightforward.

The concept of adapter caught my eye. It is a core Wasm module, whose exports are used to satisfy the imports of the main core Wasm module. Unfortunately, the adapter module is instantiated indirectly using shims since it has imports that are satisfied by the main core Wasm module exports (see [code](https://github.com/bytecodealliance/wasm-tools/blob/a2d966c7ba79497953375006a2075a6ed52e9cfc/crates/wit-component/src/encoding.rs?plain=1#L1580-L1589)). 
I ended up testing it using the following [example](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasi-preview1-component-adapter/README.md) and confirmed that the produced Wasm component will be a lot harder for us to process than the current synthetic WIT interface approach. I could not create the gist with the resulting Wasm component, so I attached it to this comment.

### wasm-component-ld (wasm-ld)

I've tested the stub libraries and played with other options (`--unresolved-symbols`, etc), and nothing useful came out of it. The core Wasm module produced has imports, and the `wit-component` encoder doesn't like it.

greenhat:
https://github.com/0xMiden/compiler/issues/531#issuecomment-3196819965

[...]

PR #1090, opened 2026-04-28
LTO-safe linker stubs

### LINKED ISSUE #1089: Fix the generated code when `lto=true`
From #1065

> 6. `lto=true` is super interesting. It gives the most significant (50%-90%) decrease in size but the compiled code no longer works (47 tests failed). I want to dig into it more. I've checked the basic-walled contract code and besides very aggressive inlining where [compiler/sdk/base-sys/src/bindings/native_account.rs](https://github.com/0xMiden/compiler/blob/ff6e2256bfef27905b0fc9650150fa71ca97388e/sdk/base-sys/src/bindings/native_account.rs?plain=1#L93) was inlined with most of the glue code optimized away.

greenhat:
Close #1089 

This PR:
- Enables the fat LTO (`lto = true`).
- Revamps the stubs to use [`black_box`](https://doc.rust-lang.org/beta/core/hint/fn.black_box.html) to survive LTO.

## Mast size changes (stripped debug info)

| Example | `next` | PR | Change |
| --- | ---: | ---: | ---: |
| `basic-wallet` | 29016 | 15764 | -45.67% |
| `basic-wallet-tx-script` | 45884 | 17045 | -62.85% |
| `p2id-note` | 38430 | 19745 | -48.62% |
| `p2ide-note` | 43003 | 23039 | -46.42% |

## Basic wallet cycle changes

| Measurement | `next` | PR | Change |
| --- | ---: | ---: | ---: |
| `p2id` mint note | 20211 | 7327 | -63.75% |
| `p2id` tx script processing | 25232 | 6799 | -73.05% |
| `p2id` note | 20211 | 7327 | -63.75% |
| `p2ide` reclaim note | 21699 | 8340 | -61.57% |

TODO:
- rebase
- try what was blocked by --allow-undefined
- fork wasm-component-ld


sdk/alloc/stubs/heap_base.rs
```diff
@@ -3,7 +3,16 @@
...
+fn stub<T: Default>() -> T {
+    core::hint::black_box(core::panic::Location::caller());
+    core::hint::black_box(T::default())
+}
+
 #[unsafe(export_name = "intrinsics::mem::heap_base")]
+#[inline(never)]
 pub extern "C" fn __intrinsics_mem_heap_base_stub() -> *mut u8 {
-    unsafe { core::hint::unreachable_unchecked() }
+    stub()
```
bitwalker, line 17: There are two things we could try rather than these stubs:

1. Declare extern functions as having `extern_weak` linkage. According to the [wasm-ld docs](https://lld.llvm.org/WebAssembly.html), the linker will emit stubs that contain an `unreachable` in such cases. AIUI, LTO would not apply in these cases, but testing would be required:

```rust
#[unsafe(export_name = "...")]
#[unsafe(linkage = "extern_weak")]
extern "C" fn foo() -> result;
```

2. Use [stub libraries](https://lld.llvm.org/WebAssembly.html#stub-libraries). This is used directly with the linker, rather than relying on rustc/LLVM to do things correctly. This would require writing the stubs file to disk, and passing custom linker arguments.

I'm not sure if there are any `lld` options that control specific optimizations that apply here, but digging deeper there might be worth it as well, rather than relying on implementation details of rustc/LLVM to obscure the code enough to disable link-time optimizations (e.g. `black_box`). We'd definitely need really good regression tests to catch if something slips through that subtly breaks things if we do end up relying on that.

greenhat: 1. It ends up as an `env` imports. The link command contains `--allow-undefined`. That means the unresolved weak externs become imports rather than local `unreachable` trap stubs. I also tried adding `--unresolved-symbols=ignore-all` explicitly through `cargo-miden`, but that did not override the import behavior.

2. I tried it when I did https://github.com/0xMiden/compiler/issues/532#issuecomment-3182983595. I tried it again and the result is the same - we end up with `env` imports in the Wasm module that stop the Wasm CM linking.

As long as there is `--allow-undefined` as I explained in https://github.com/0xMiden/compiler/issues/532#issuecomment-3219393197 these approaches will not bring the desired outcome.

greenhat: > I'm not sure if there are any `lld` options that control specific optimizations that apply here, but digging deeper there might be worth it as well, rather than relying on implementation details of rustc/LLVM to obscure the code enough to disable link-time optimizations (e.g. `black_box`). 

I think I covered it when I worked on #532.

> We'd definitely need really good regression tests to catch if something slips through that subtly breaks things if we do end up relying on that.

The intrinsics and stdlib should be covered by the semantic tests. The protocol bindings are tested only for compilation, but we can add `litcheck_filecheck::filecheck!` checks for the linked MASM protocol functions.

**EDIT**: Added in 59dc389c521752320ac77023374723bdcc23fa8b

bitwalker: I believe the latest Rust toolchain removes `--allow-undefined` (there was a blog post about the change being potentially breaking recently). Maybe we should switch to a recent nightly and check how things work now.

greenhat: I tried in your `bitwalker/nightly-2026-04-29` branch and the linker is still invoked with `--allow-undefined`:

```

 INFO rustc_codegen_ssa::back::link LC_ALL="C"
 PATH="/Users/green/.rustup/toolchains/nightly-2025-12-10-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin:/Users/green/.rustup/toolchains/nightly-2025-12-10-aarch64-apple-darwin/lib/rustlib/aarch64-apple-darwin/bin/self-contained:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Library/Apple/usr/bin:/Users/green/s:/Users/green/go/bin:/Applications/Visual
 Studio Code.app/Contents/Resources/app/bin:/Users/green/.cargo/bin:/usr/local/sbin:/Users/green/bin:/Users/green/.local/bin:/Applications/Ghostty.app/Contents/MacOS"
 VSLANG="1033" "wasm-component-ld" "-flavor" "wasm" "--export" "cabi_realloc"
 "--export" "miden:basic-wallet/basic-wallet@0.1.0#move-asset-to-note"
 "--export" "miden:basic-wallet/basic-wallet@0.1.0#receive-asset" "-z"
 "stack-size=1048576" "--stack-first" "--allow-undefined" "--no-demangle"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/basic_wallet.basic_wallet.56b8d497f1b93d29-cgu.0.rcgu.o"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/basic_wallet.coa0wum726y21wwrxtf86rp2t.rcgu.o"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libmiden-42e5d7d24e7c8bf7.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libwit_bindgen-5f8118f25035a17c.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libmiden_sdk_alloc-444e74dee0b069b8.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libmiden_base-877e8b817cbd86ce.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libmiden_base_sys-27783d9b72c6f525.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libmiden_field_repr-3e9599b4a68e206c.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libmiden_stdlib_sys-d3fa3524d2787be6.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libmiden_field-f6efeda34afe9ace.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libthiserror-de5d7c64c8c5150d.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/liballoc-7fdb75bc4385100d.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libcore-c245813a48747169.rlib"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/libcompiler_builtins-0f99848336997ae1.rlib"
 "-l" "miden_base_sys_stubs" "-L"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/build/miden-base-sys-777b7a54c3f199c2/out"
 "-L"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/build/miden-stdlib-sys-5a4e6d72c097a3c0/out"
 "-L"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/build/miden-sdk-alloc-81923d26d362fa92/out"
 "-L"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/build/wit-bindgen-5a0885edbd6f1c59/out"
 "-L"
 "/Users/green/.rustup/toolchains/nightly-2025-12-10-aarch64-apple-darwin/lib/rustlib/wasm32-wasip2/lib/self-contained"
 "-o"
 "/Users/green/src/miden/compiler-tree/bitwalker-nightly-2026-04-29/examples/basic-wallet/target/wasm32-wasip2/release/deps/basic_wallet.wasm"
 "--gc-sections" "--no-entry" "-O2" "--strip-debug" "--fatal-warnings"
```

To reproduce:
``` 
RUSTC_LOG=rustc_codegen_ssa::back::link=info \
  cargo test -p miden-integration-tests basic_wallet_and_p2id -- --nocapture
```

greenhat: The blog post: https://blog.rust-lang.org/2026/04/04/changes-to-webassembly-targets-and-handling-undefined-symbols/

bitwalker: I'm still looking to see why this isn't in effect in the latest nightly (it seems it may be applied to `wasm32-unknown-unknown` already, but not the WASI targets).

However, something that occurred to me when going down the rabbit hole again (and the blog post reminded me) is that we can use `#[link(wasm_import_module = "whatever")]` on the `extern_weak` symbols (solution 1 I mentioned earlier), which will at least ensure that the symbol ends up as an import from a module we can recognize. I can't recall why imports from `env` were specifically an issue, but if we can control what module these are imported from, maybe it becomes less of an issue for us.

I'll do a bit more poking around here and follow up

bitwalker: Ok, so here's what I found.

Given the following:

```rust
# sdk/alloc/src/lib.rs
#![feature(linkage)]
extern "C" {
    #[linkage = "extern_weak"]
    #[link_name = "intrinsics::mem::heap_base"]
    fn heap_base() -> *mut u8;
}

# sdk/alloc/stubs/heap_base.rs
#![feature(optimize_attribute)]

#[unsafe(export_name = "intrinsics::mem::heap_base")]
#[optimize(none)]
#[inline(never)]
pub extern "C" fn __intrinsics_mem_heap_base_stub() -> *mut u8 {
    unsafe { core::hint::unreachable_unchecked() }
}
```

Running `cargo miden build` on the latest toolchain in `examples/counter-contract` (just an arbitrary choice here to keep things simple) produces a package successfully, and the corresponding Wasm binary is drastically smaller (69K vs 3.3M), but the size of the `.masp` file is only slightly smaller, here's the `objtool` output:

```
❯ bin/miden-objtool decorators examples/counter-contract/counter_contract_no_lto.masp
Input: examples/counter-contract/counter_contract_no_lto.masp
Package kind: library
Artifact: library

Metric                  KB    Delta  Delta %
original masp       217.63        -        -
original forest     215.58     0.00   +0.00%
without decorators   84.67  -130.91  -60.73%
compacted forest     57.69  -157.89  -73.24%

❯ bin/miden-objtool decorators examples/counter-contract/counter_contract_lto.masp
Input: examples/counter-contract/counter_contract_lto.masp
Package kind: library
Artifact: library

Metric                  KB    Delta  Delta %
original masp       226.51        -        -
original forest     224.46     0.00   +0.00%
without decorators   88.02  -136.44  -60.79%
compacted forest     56.13  -168.33  -74.99%
```

Curiously, the initial size of the `.masp` _with_ LTO is actually larger - not really sure why that is tbh, but ultimately it ends up smaller after stripping/compaction.

Anyhow, the important thing is that this method seemingly does _not_ perform interprocedural optimizations during LTO, so the unreachable stubs remain in the module, and the callers do not optimize out the calls to them.

bitwalker: ~Note I did find a bug that will be fixed in `miden-vm` v0.22.3 (see [here](https://github.com/0xMiden/miden-vm/pull/3087) for details), that will be released today, so shouldn't be a blocker, but just FYI~

Updated to v0.22.3 on `next`

greenhat: Thank you! Implemented in #1107.

greenhat:
Closing this in favor of #1107

Issue #1091, opened 2026-04-28
Overflowing and checked arith for small signed ints are broken


mooori:
Overflowing arith for small signed ints (`i8`, `i16`) was broken until recently. This had been fixed by changes in the debugger and by lowering some instructions through the new wasm dialect. After these changes landed the `test_overflowing_*_(i8|i16)` tests had been unignored in #992 and #996.

They have been ignored again in [this commit](https://github.com/0xMiden/compiler/commit/35f0c5e9de707eb11b7352ff2190edb63e7f0965) which migrated `miden-vm 0.20 -> 0.21.1` and `miden-base 0.13 -> 0.14`. So that might be a starting point for the investigation.

Can be replicated by running the ignored tests:

```sh
cargo nextest run -p miden-integration-tests --no-fail-fast --run-ignored ignored-only \
  -- rust_masm_tests::instructions::test_overflowing
```

Tests `test_checked_*_(i8|i16)` are failing too, probably due to the same root cause.

bigeez:
Hi, I’d like to investigate this if you’re happy to assign it to me.

I’ll start by reproducing the ignored i8 / i16 overflowing and checked arithmetic tests, then trace whether the regression came from the lowering path or from the miden-vm / miden-base upgrade mentioned in the issue.

Please assign me if that sounds useful.

mooori:
Thanks @bigeez for your interest. Though I’m planning to take this on myself since I’ve worked on it before.

Issue #977, opened 2026-02-23
test: integer `checked wide` and `overflowing` arith ops


mooori:
### What should be done?

Continuation of #959

Once the issues collected in #971 are resolved, add the following tests:

- [ ] `wide-overflowing`
- checked:
  - [x] `normal` #1092 
  - [ ] `wide`

### How should it be done?

Add tests in [rust_masm_tests/instructions.rs](https://github.com/0xMiden/compiler/blob/c002d1a263f78b441e48a56c4fea0d9a6e870ade/tests/integration/src/rust_masm_tests/instructions.rs)

### When is this task done?

Tests are added

bitwalker:
Done in #1354

PR #1092, opened 2026-04-28
test: add proptests for `checked` arith

mooori:
Addresses the `checked normal` part of #977

This mainly mirrors the setup for `overflowing` arith tests.

More tests than expected are failing and need to be ignored due to #1091


greenhat (approved):
Looking good!

PR #820, opened 2025-12-15
Track variable locations

djolertrk:
Track variable locations at HIR and MASM levels.
Add DebugVar decorator and new debug_info section in the final .masp.

Also, add `miden-debugdump` tool to dump .debug_info section, so we can explore the custom section from the package.

vm part: https://github.com/0xMiden/miden-vm/pull/2471

PR #1097, opened 2026-04-30
feat: implement debug info dialect and variable tracking

### LINKED ISSUE #714: Add means of debugging variables in Rust
### Feature description

Currently, debugging Miden Rust programs is a bit difficult. The simplest way currently to inspect a variable is via the `assert_eq!` macro:

```rs
let mut some_variable = 5;
// some_variable is updated to 6 somewhere
assert_eq!(some_variable, 6u32);
```

While this works, it interrupts program execution and is not ideal for inspecting multiple variables or complex state.

It would be very useful to have a Miden Rust-specific debug macro or function, similar to `println!`, that can emit the value of a variable at a particular point in execution without halting the program. For example:
```rs
debug_var!(some_variable); // hypothetically prints the current value
```

This would make debugging compiled Miden Rust code far more straightforward and efficient.

Benefits:
* Inspect variables without breaking execution.
* Simplifies debugging of complex programs.
* Provides a familiar interface for Rust developers (println!-like).

### Why is this feature needed?

This would make debugging Miden Rust programs much more straightforward.

### LINKED ISSUE #745: Introduce variable location representation and tracking
### Feature description

Currently, the compiler infrastructure (from generated DWARF) parses source locations only (line:column/.debug_line section). The idea is to parse variables locations (`DW_AT_location`) as well.

DWARF standard was extended with `DW_OP_WASM_location` that can represent locals, globals and stack operands.

HIR can be extended with a new `builtin`/debug instruction - `builtin.debug_value`, that contains several operands, like pointer to an attribute representing variable (`DIlocalVariableAttr`), function (`DISubprogramAttr`) and expression (`DIExpression`) that represents the value.

### Why is this feature needed?

_No response_

bitwalker:
This is @djolertrk's work from #820, but rebased on top of #1096, with various refactorings/fixes made to it in preparation for merging to `next`.

The three tip commits on this branch contain all the changes I made, so that I could preserve @djolertrk's original commits. See the commit message for the `HEAD` commit, as that has details on all of the more subtle changes/refactorings, and is the one that deserves the most review.

The other two are more obvious, so I'll summarize them here:

1. I merged the separate `debugdump` tool into `miden-objtool` as a subcommand, specifically `miden-objtool dump debug-info`. The implementation is largely identical, with some superficial refactorings around how the config is constructed, etc.
2. I removed the temporary workaround code that was added to `midenc-compile` to emit the custom debug info sections which are now produced by the assembler natively when assembling projects. That means that _after_ this is merged, there will be a short period where we do not have the debug info sections available, until we merge #1071 (unless that gets merged first, which may very well be the case, as that is next on my TODO list to review).

Some issues I found while reviewing the changes in #820, and which I have not attempted any fix here:

1. The debug info tests added in `tests/integration/src/rust_masm_tests/debug.rs` do not seem to result in any debug ops being emitted in all but one case, and in that one case, the location information is wrong (i.e. `"unknown"` source file). Presumably this will be addressed by improvements to the frontend, but it is worth noting.
2. The `di.declare` and `di.kill` ops are not currently emitted anywhere, and I'd expect them to be (even though we don't really handle them during codegen, except for `di.declare` which I added the lowering for). The `di.kill` op will likely require a new decorator in MAST unfortunately.
3. The frontend does not handle most DWARF expressions, only a small subset, which means that variable tracking is likely to be inaccurate. I plan to discuss how to handle this with @djolertrk ASAP - but after doing some investigation, I believe we're likely best off emitting DWARF bytecode directly, and then using `gimli` to decode and evaluate it in the debugger. `gimli` provides an evaluator that delegates all the machine-specific details to us, which works perfectly for our needs.


djolertrk (changes requested):
Since we marked the `di.*` ops as `Transparent`, they get deleted in passes like `DCE`, so we have zero variables (locations) at the end in debug info sections. We need to teach passes not to delete them.

djolertrk:
> Since we marked the `di.*` ops as `Transparent`, they get deleted in passes like `DCE`, so we have zero variables (locations) at the end in debug info sections. We need to teach passes not to delete them.

Latest commits address this, and also adds testing of `di.debug_value` ops.

djolertrk:
> Since we marked the `di.*` ops as `Transparent`, they get deleted in passes like `DCE`, so we have zero variables (locations) at the end in debug info sections. We need to teach passes not to delete them.

Hmm. As we discussed, we should actually keep `DCE` perform on these as well, since we do not need to keep dead values in HIR anyway (even though we could salvage some values, like the one depending on constants, or if a variables value can be described in terms of other SSA value that is still alive - but lets not worry about this scenario at this point).

I am working on a fix. For one use case that I am hitting issues with, we could be using `di.debug_declare` - e.g. if we have a parameter that is placed at an offset from frame register, we won't have any SSA value to connect that with at HIR level, so, in those cases, we can just describe it with a DWARF expression.

PR #1098, opened 2026-05-01
chore: update toolchain to nightly-2026-04-29

bitwalker:
Updates our Rust toolchain to 1.97/nightly-2026-04-29, and updates all of our dependencies as much as possible (though there are still a couple cases where we keep an older version) - I also didn't update any of the `wit-bindgen`-related deps.

I'm hoping this might make #1090 a little cleaner now that Rust has removed `--allow-undefined` from its default `wasm-ld` flags, but we'll see.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,36 +259,7 @@ jobs:
       - name: Test
         # Skip running the some test since they are a disc space hog.
         run: |
-          cargo make test -E 'package(cargo-miden) - test(new_project_integration_tests_pass) - test(test_all_templates_and_examples)'
-
-  cargo_miden_tests_mpt:
-    name: cargo-miden test Miden project template
-    runs-on: ubuntu-latest
-    # We only want to run the integration tests if the unit tests pass, and that has the added
-    # benefit that we can re-use the cache from the unit test job for all integration tests
-    needs: [unit_tests]
-    steps:
-      - uses: actions/checkout@v5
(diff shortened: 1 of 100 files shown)


bitwalker:
@greenhat I think we should remove the `cargo-miden` integration tests (specifically the ones that generate the `integration` workspace from the `project-template` repo) from this repo, and move them to the `project-template` repo (which already does in fact run the same tests AIUI). Those two tests fail every time there is some kind of weird divergence between nightlies (e.g. in this case, an older version of `hashbrown` was specializing on `Copy` and that isn't supported now, so it fails to compile). Since those tests aren't actually running the right version of `cargo-miden` anyway (because the version the generated `integration` workspace uses is pulled from crates.io), there is basically no point in running them in our CI.

greenhat:
> @greenhat I think we should remove the `cargo-miden` integration tests (specifically the ones that generate the `integration` workspace from the `project-template` repo) from this repo, and move them to the `project-template` repo (which already does in fact run the same tests AIUI). Those two tests fail every time there is some kind of weird divergence between nightlies (e.g. in this case, an older version of `hashbrown` was specializing on `Copy` and that isn't supported now, so it fails to compile). Since those tests aren't actually running the right version of `cargo-miden` anyway (because the version the generated `integration` workspace uses is pulled from crates.io), there is basically no point in running them in our CI.

I agree. Make them ignored and I'll take care of moving them to the `project-template` repo.

PR #1104, opened 2026-05-03
refactor: clean up integration test organization

### LINKED ISSUE #629: Rename integration test crates and extract shared code into a separate crate
Following https://github.com/0xMiden/compiler/pull/555#discussion_r2248146332 and https://github.com/0xMiden/compiler/pull/555#discussion_r2248158572

bitwalker:
This PR closes #629  by restructuring our integration test crates, making their naming consistent, separating test support code from the tests themselves, and extracting fixtures to a common directory.

Tests are now structured along the following axes:

### Rust-based vs CLI tests

* Integration tests that do not require the protocol/MockChain are maintained in `tests/integration`
* Integration tests that do are maintained in `tests/integration-network`
* Integration tests that invoke CLI interfaces and assert something about the output use lit/FileCheck, and are maintained in `tests/lit`

### Unit vs Integration

* Tests which do not require exclusive access to some global resource are defined as normal Rust unit tests
* Tests which do are defined as Rust integration tests. Currently this consists of just the `println` tests, as exercising them contends on the global logger

### Test area and subject (what we're testing)

* Tests are broken up within the integration test crates based on what area of compilation is being tested (i.e. frontend, middle-tier, codegen, or the full Rust pipeline)
* Tests are then further broken up by the specific testing topic (e.g. diagnostics, specific lowerings, semantic tests, etc.)

### Test type (how we're testing it)

* Finally, tests are broken up by the testing method (i.e. property-based tests, functional tests, etc.)

**NOTE:** Not everything is necessarily perfectly organized yet - if you spot anything off, let me know. If you think you have a better organization than what was thrown together here, feel free to suggest changes - we're not committed to anything specific yet.


greenhat (approved):
Looking great! 
The only thing that sticks out to me is the `expect/functional/property` level in the `tests/integration/src/end_to_end`. I think it's artificial and redundant. 
The `expect` module tests are mainly to test successful compilation and the `expect!` calls are there to check/show some extra stuff. 
The `functional` vs `property` divide also seems artificial to me. I see `property` tests as "the same code but tested with more than one set of inputs" compared to the `functional` tests.

mooori (approved):
Nice organization!

PR #1113, opened 2026-05-06
[1/3] FPI

### LINKED ISSUE #697: Foreign Procedure Invocation (FPI) support
Implement the counter contract copier contract from the tutorial at https://0xmiden.github.io/miden-docs/imported/miden-tutorials/src/rust-client/foreign_procedure_invocation_tutorial.html and implement missing parts.

>The only tx-kernel function I found with the signature that cannot be represented in Rust is execute_foreign_procedure - https://github.com/0xMiden/miden-base/blob/5dfa4310abf68c294357edf742b6dcfc2d5bbc3e/crates/miden-lib/asm/miden/tx.masm?plain=1#L202-L217 The reason is foreign procedure inputs and outputs are dynamic and since there is no variadic arguments in Rust the only option to handle them is to write an intrinsic but that'd throw a wrench in the automated bindings generation. So the proper way to handle it seems to be to change the MASM signature

>For execute_foreign_procedure, I think it probably doesn't need to be invoked "manually" and so no need to expose it to the user. I'm imagining that in the future, we'll generate calls to this procedure automatically somehow - this would happen when one account invokes a method on another account, and I think we probably still need to flash out both the ergonomics of how this looks like in Rust and how this gets translated into WIT/MASM on the back end.

greenhat:
Ref #697 

Adds end-to-end Foreign Procedure Invocation support for SDK account components and note scripts. The new `#[foreign_account(package_name1, package_name2)]` SDK macro now discover Miden account dependencies from package metadata, synthesize caller-only FPI imports, and generate methods for the struct that let callers invoke dependency methods through `execute_foreign_procedure` with normal Rust signatures.

- During binding generation, the macro loads the selected WIT world and injects a synthetic `fpi-<method>` import for each freestanding method exported by a Miden dependency interface. Each synthetic import prepends the foreign account id prefix, account id suffix, and procedure root word to the original method parameters while preserving the original result type. `wit-bindgen` then renders these imports as Rust `fpi_<method>` free functions.
- After `wit-bindgen` emits bindings, the macro scans dependency import modules for generated `fpi_*` functions, loads the dependency `.masp` package, resolves exported procedure roots, and appends methods to the defined struct. The public wrapper methods remove the internal prefix/suffix/root parameters from the user-facing signature and call the generated `fpi_*` function with `self.0.prefix`, `self.0.suffix`, and the resolved root.
- The Wasm frontend recognizes imports whose WIT function name starts with `fpi-` and routes them through dedicated FPI lowering instead of generic component import lowering. For direct canonical ABI calls, it rewrites the inner import call into an HIR `exec` of `miden::protocol::tx::execute_foreign_procedure`, reordering the account id to the protocol ABI order, expanding the procedure root word, forwarding flattened procedure inputs, validating the 16-input/16-result limits, and storing multi-felt results through the canonical output pointer when needed.
For FPI calls whose flattened parameters exceed the direct canonical ABI limit, the canonical ABI passes a single argument tuple pointer, and the generated wrapper function reloads every flattened value from that tuple in HIR (sign-extending narrow signed integers), so the indirect shape never leaves the frontend. From there, every FPI call — direct, indirect, and raw — lowers to the same dedicated `hir.exec_fpi` op: its operands are the up-to-16 flattened procedure input felts, while the 6-felt executor prefix (account id suffix and prefix, already reordered, plus the procedure root word) is stored into felt function locals referenced by the op. MASM codegen then pads the scheduled inputs with zeros to the fixed 16-felt executor input width, loads the prefix locals on top of the stack, and invokes the real `execute_foreign_procedure` — no compiler-internal marker callee, no layout attributes, and no stack shuffling beyond the addressable 16-element window.

The note to account example:
```rust
...
#[foreign_account(mixed_scalar_struct_account)]
struct CounterContract;
...
#[note]
impl CounterCaller {
    /// Checks that a mixed scalar record crosses the FPI boundary in both directions.
    #[note_script]
    pub fn run(self, _arg: Word) {
        let count_acc = CounterContract::from_account(self.counter_account_id);
        let result = count_acc.echo_mixed_scalar_record(MixedScalarRecord {
            first_u64: FIRST_U64,
            second_u64: SECOND_U64,
            felt_value: felt!(77),
            u32_value: U32_VALUE,
            u8_value: U8_VALUE,
        });
```
See tests/integration-network/src/mockchain/fpi/note/mixed_scalar_struct.rs for the complete example.

The account to account example:
```rust
#[foreign_account(account_to_account_callee_account)]
struct CalleeAccount;

/// Account component which forwards reads to another account through FPI.
#[component]
struct CallerAccount;

#[component]
impl CallerAccount {
    /// Reads a counter value from the provided foreign account.
    pub fn read_foreign_count(&self, callee_account_id: AccountId) -> Felt {
        let callee = CalleeAccount::from_account(callee_account_id);
        let key = Word::new([felt!(13), felt!(21), felt!(34), felt!(55)]);
        callee.get_count(key)
    }
}
```
See tests/integration-network/src/mockchain/fpi/account/account_to_account.rs for the complete example.

The note to account example with multiple packages:
```rust
#[foreign_account(multiple_packages_first_account, multiple_packages_second_account)]
struct ForeignCounters;

/// Note script input containing the foreign account id.
#[note]
struct CounterCaller {
    /// Account id with both counter components deployed.
    foreign_account_id: AccountId,
}

#[note]
impl CounterCaller {
    /// Checks that a multi-package foreign account binding exposes both component methods.
    #[note_script]
    pub fn run(self, _arg: Word) {
        let counters = ForeignCounters::from_account(self.foreign_account_id);

        let first = counters.get_first_count();
        let second = counters.get_second_count();
...
```

This PR also adds raw bindings for the kernel `execute_foreign_procedure` procedure so that the MASM account methods can be called via FPI:
```rust
...
let outputs = tx::execute_foreign_procedure(
    self.foreign_account_id,
    procedure_root,
    inputs,
);
...
```


greenhat:
@bitwalker @bobbinth I finished the FPI implementation with the `#[foreign_account(package_name1, package_name2)]` attribute macro we discussed in https://github.com/0xMiden/compiler/issues/697#issuecomment-4433777471. Check out the PR description and tests for examples.

EDIT: We support account-to-account, note-to-account, and direct `execute_foreign_procedure` calls (for MASM accounts).

greenhat:
Following our discussion to go with the 2.1 `#[account(...)]` macro for both current and foreign accounts, I renamed the `#[foreign_account(...)]` macro to `#[account(...)]` in 9966e55341ce946ec8f40e239304b6c311dd2046

codegen/masm/src/lower/fpi.rs
bitwalker, line 1: Could the module doc here elaborate  a bit more on how FPI is handled during lowering? In particular, the aspects related to ABI handling, but I would also document any assumptions/expectations on code that triggers the FPI handling.

bitwalker (commented):
I'm still working my way through this, but I think some additional documentation in the codegen backend would be helpful here, as there are a few aspects that aren't super clear to me that I suspect documentation would fix.

codegen/masm/src/lower/fpi.rs
greenhat: Done in d9c45876e42a70af58bbc1dc7f23d81e3070094a

codegen/masm/src/lower/fpi.rs
```diff
@@ -0,0 +1,638 @@
...
+//! (`AccountId`, procedure root, and FPI context), and the remaining 16 operands are the flattened
+//! procedure inputs. The executor returns one felt for each possible procedure input slot.
+//!
+//! Lowering handles two call shapes produced by earlier frontend/component ABI lowering:
+//!
+//! - Direct FPI calls target `execute_foreign_procedure` itself and carry the already-flattened
+//!   operands directly on the `hir.exec`. The frontend omits unused trailing procedure input slots
+//!   so the generic operand scheduler only solves for the real operands. After scheduling those
+//!   operands, this module pads the stack with zero felts until it reaches the 22-felt executor
+//!   width. The 16-operand boundary needs one backend scratch local because MASM stack shuffling can
```
bitwalker: Maybe it would be simpler if we arranged to store the first 6 operands (account id, procedure root, FPI context) in locals, and then load them on to the operand stack manually in the lowering of an `hir.exec` that targets `execute_foreign_procedure` after padding the stack out to 16 elements, thereby getting all 22 operands on the operand stack with minimal fuss.

With only a single scratch local, it takes a lot more shuffling to get things into place.

This also seems to me to be a good fit for a special instruction specifically for FPI, that largely looks like `hir.exec`, but rather than the first 6 operands (3 values) being SSA values, they would be references to 3 local variables that are stored to prior to the FPI instruction. Lowering for the new instruction would then be able to guarantee that up to 16 operands would be on the operand stack on entry to the lowering of that op, and the lowering could manually load the values of the 3 local variables on top of the operand stack in the correct order. The only special handling that would be needed in the lowering code would be the padding aspect.

What do you think? It's a bit of an adjustment, but I think it would make lowering more straightforward, more efficient, and a bit easier to reason about (not to mention it would avoid complicating the lowering for `hir.exec`.

greenhat: Great idea! Thank you! Implemented in the last two commits (6119c12a8a4e347439dbd73c5be0557ee6a3e06b and 39dafb825820effffad292b559dd3aeb449e3850)

dialects/hir/src/ops/invoke.rs
```diff
@@ -126,6 +131,90 @@ impl OpParser for Exec {
...
+/// operand stack: it pads them with zeroes to [`ExecFpi::MAX_INPUT_FELTS`], then loads the six
+/// locals on top to form the full [`ExecFpi::EXECUTOR_INPUT_FELTS`]-felt executor ABI without any
+/// stack shuffling beyond the addressable 16-element window.
+#[derive(EffectOpInterface, OpPrinter, OpParser)]
+#[operation(
+    dialect = HirDialect,
+    implements(InferTypeOpInterface, MemoryEffectOpInterface, OpPrinter)
+)]
+#[effects(MemoryEffect(MemoryEffect::Read, MemoryEffect::Write))]
+pub struct ExecFpi {
```
bitwalker, line 152: I can't help but feel it would be better if we could somehow use the "real" signature of the foreign procedure somehow, probably by storing the type signature in an attribute and making the operands/results `AnyType` - but I see that more as a future improvement - this will do for now I think.

bitwalker (approved):

dialects/hir/src/ops/invoke.rs
greenhat: I agree. I made #1176.

PR #1114, opened 2026-05-06
ci: add shear, workspace inheritance, and zizmor checks, and move heavy GitHub Actions jobs to WarpBuild

huitseeker:
1. The dependency check now uses [`cargo-shear`](https://github.com/Boshen/cargo-shear), replacing `cargo-machete`. `cargo shear` found unused dependencies that were still present in
  this workspace, and it can also catch dependencies placed in the wrong section.
  2. [`zizmor`](https://github.com/zizmorcore/zizmor) scans GitHub Actions files for risky patterns, like loose action pins or unsafe defaults.
  3. The heavier CI jobs now run on larger [WarpBuild](https://www.warpbuild.com/) runners.
  4. The dependency policy check now also uses `cargo-workspace-inheritance-check`. It makes sure crates shared across the workspace use one workspace-level dependency entry instead of
  repeating versions in many crate manifests.

  The PR also removes the unused dependencies that `shear` found, pins action versions more tightly, and promotes shared dependencies to the workspace table.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,25 +23,47 @@ env:
   # Set the new value when the cache grows too much and we hit the runner's disk space limit
   # via https://github.com/rust-lang/docs.rs/pull/2433
   RUST_CACHE_KEY: rust-cache-20251212
+  # Reduce cache size by not storing debug info in dev-profile artifacts.
+  CARGO_PROFILE_DEV_DEBUG: 0
 
 jobs:
-  lint:
-    name: lint
+  workspace_inheritance:
+    name: Workspace inheritance
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
+      # Pin install-action to a release and request the tool explicitly; shortcut
+      # tags can trip zizmor's impostor-commit audit when they are later retagged.
+      - name: Install cargo-binstall
+        uses: taiki-e/install-action@055f5df8c3f65ea01cd41e9dc855becd88953486 # v2.75.18
+        with:
+          tool: cargo-binstall
+      - name: Install workspace inheritance checker
+        run: cargo binstall --no-confirm cargo-workspace-inheritance-check@1.2.0
+      - name: Check workspace inheritance
+        run: cargo workspace-inheritance-check --promotion-threshold 2 --promotion-failure
+
+  lint:
+    name: lint
+    runs-on: warp-ubuntu-latest-x64-8x
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
       - uses: ./.github/actions/cleanup-runner
       - name: Install Rust
         run: |
           rustup update --no-self-update
           rustc --version
       - name: Cache Cargo
-        uses: Swatinem/rust-cache@v2
+        uses: WarpBuilds/rust-cache@9d0cc3090d9c87de74ea67617b246e978735b1a1 # v2.9.1
         with:
           # Use a common cache for the basic compilation/formatter/clippy checks
           shared-key: ${{ github.workflow }}-shared
           prefix-key: ${{ env.RUST_CACHE_KEY }}
-          save-if: ${{ github.ref == 'refs/heads/next' }}
+          save-if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/next' }}
       - name: Install cargo-make
         run: |
           if ! cargo make --version 2>/dev/null; then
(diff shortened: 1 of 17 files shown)


bitwalker (changes requested):
Looks like we both had similar notions at the same time 😅 - I've already implemented the switch to `shear` in next, but if you rebase the other changes, we can get this merged

tools/hir-opt/Cargo.toml
```diff
@@ -20,11 +20,6 @@ path = "src/main.rs"
 [dependencies]
 midenc-compile.workspace = true
 midenc-hir.workspace = true
-midenc-dialect-arith.workspace = true
```
bitwalker, line 23: Note that these are false positives - the dependencies are actually used, but it is no surprise that `cargo-shear` can't detect this, because the dependency is indirect (we need these crates linked into the executable so that their dialects are available, and dialect registration relies on some linker tricks via `inventory`).

Cargo.toml
```diff
@@ -162,6 +159,15 @@ cargo-miden = { version = "0.8.1", path = "tools/cargo-miden" }
...
 miden-field = { version = "^0.25" }
+miden-base-sys = { version = "0.12.0", path = "sdk/base-sys" }
+miden-field-repr = { version = "0.12.0", path = "sdk/field-repr/repr" }
+miden-stdlib-sys = { version = "0.12.0", path = "sdk/stdlib-sys" }
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+
+[workspace.metadata.cargo-shear]
+# `litcheck-filecheck::filecheck!` expands to paths in `litcheck`, which
+# cargo-shear cannot see from the source before macro expansion.
+ignored = ["litcheck"]
```
bitwalker: This can be removed, the fix is to import `litcheck` using `use litcheck_filecheck::litcheck`. The `filecheck` macro is going to be fixed in a patch release soon so that this isn't necessary, but I haven't gotten around to it yet

Cargo.toml
```diff
@@ -162,6 +159,15 @@ cargo-miden = { version = "0.8.1", path = "tools/cargo-miden" }
 midenc-integration-test-support = { path = "tests/support" }
 midenc-expect-test = { path = "tools/expect-test" }
 miden-field = { version = "^0.25" }
+miden-base-sys = { version = "0.12.0", path = "sdk/base-sys" }
+miden-field-repr = { version = "0.12.0", path = "sdk/field-repr/repr" }
+miden-stdlib-sys = { version = "0.12.0", path = "sdk/stdlib-sys" }
+serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
```
bitwalker: We separate our workspace crates from all other deps - I'd move `serde_json` up to the section with all the third-party deps

.github/workflows/ci.yml
```diff
@@ -23,25 +23,61 @@ env:
   # Set the new value when the cache grows too much and we hit the runner's disk space limit
   # via https://github.com/rust-lang/docs.rs/pull/2433
   RUST_CACHE_KEY: rust-cache-20251212
+  # Reduce cache size by not storing debug info in dev-profile artifacts.
+  CARGO_PROFILE_DEV_DEBUG: 0
```
bitwalker: This is likely to break integration tests, since we rely on debug info for various compiler tests - I guess if CI is passing, then no worries, but that surprises me (it may be that `cargo make` is changing it back though, who knows)

.github/workflows/ci.yml
```diff
@@ -85,52 +109,37 @@ jobs:
...
-        run: |
-          if ! cargo make --version 2>/dev/null; then
-            cargo install cargo-make --force
-          fi
-      - name: Clippy
-        run: |
-          cargo make unused
+          persist-credentials: false
+      - name: Install cargo-shear
+        uses: taiki-e/install-action@055f5df8c3f65ea01cd41e9dc855becd88953486 # v2.75.18
```
bitwalker, line 124: I'm not stoked about the idea of having CI diverge from how we run stuff locally during development. If possible, I'd prefer to use our `cargo make unused` task.

`cargo make unused` will use an existing `cargo shear` if it is already installed, otherwise it installs it on demand from crates.io

.github/workflows/ci.yml
```diff
@@ -85,52 +109,37 @@ jobs:
   check_unused:
     name: check for unused items/dependencies
     runs-on: ubuntu-latest
-    # Run this job after the linter, so the cache is hot
-    needs: [lint]
-    # But run this check even if the lint check failed
-    if: ${{ always() }}
```
bitwalker, line 91: This is what I was asking about with the previous "why was this removed" comment

.github/workflows/ci.yml
bitwalker: It occurs to me that we won't know if this breaks CI until this PR gets merged and the next one picks up the changes - I'll test this out locally and see what happens

huitseeker: CI runs on this PR do pick up the CI options set in this PR.

.github/workflows/ci.yml
huitseeker: using `make unused`, thereby making this change obsolete

bitwalker (approved):

PR #1119, opened 2026-05-08
fix(frontend/wasm): handle `movement_count >= value_width`

mooori:
Closes #1110
Closes #1095

Makes the Wasm frontend emit instructions to calculate `movement_count % value_width` before doing a shift or rotation. Wasm allows `movement_count > value_width` because it implicitly calculates `movement_count % value_width` before shifting or rotating. In contrast Miden VM traps if `movement_count > value_width`. So in the Wasm frontend we need to mask the count before the shift/rotate. See these comments for more details and references to the wasm spec: [1](https://github.com/0xMiden/compiler/issues/1110#issuecomment-4400360586), [2](https://github.com/0xMiden/compiler/issues/1095#issuecomment-4476359457).

Here are compiler explorer showcases for [shift](https://godbolt.org/z/1WPzb85Ed) and [rotate](https://godbolt.org/z/nsEWj3qfe). Note how `x & 31` is optimized away because that's guaranteed by Wasm semantics.

### Performance impact

As measured by `expect` tests it's negligible.

- Some slight increases as now shift/rotate are preceded by `band` to calculate `%`. Probably that can be optimized away in many cases.
- Some slight decreases as `cast` was replaced by `trunc` when converting a 64 bit movement count to 32 bits.

### Test status and new issues

- These changes allow unignoring most of the wrapping shift tests
- Fuzza's `bitops` test passes
- Issues popping up after this change: #1120, #1121, #1122. Ignore reasons for tests have been updated.
- Note that #1120 requires ignoring a test that passed previously: `rust_sdk_invalid_stack_offset_movup_16_issue_831` for a reason other than what it was originally reproducing.

PR #1129, opened 2026-05-13
CSE non-determinism fix

greenhat:
Discovered in #1113 and #1119. 

### What
It manifested only when the compiler crates are compiled with the `release` profile and only on Linux. The manifestation was that when compiling the same code multiple times, the compiled MAST was different in some iterations while the Wasm binary stayed the same. This resulted in either a `procedure with ... root not found` error when a note script compiled the account as a dependency and got a different MAST versus the deployed account (compiled separately). Or a different cycle count due to the note compiling to a different MAST on CI vs locally.
Here is the report gathered when the issue was located: https://gist.github.com/greenhat/f3e7f59599bffcb9d6d464ad34fa6d2a

### Fix
Fix CSE in graph regions by making candidate lookup resilient to operand rewrites and by tracking visited operations by identity. The pass now uses hash-backed candidates for SSA-dominance regions, rescans candidates where graph regions can mutate previously seen operations, keeps operand payloads in sync during partial use replacement, and adds regression coverage for a use-before-def graph-region case that previously exposed stale equivalence state.


hir-transform/src/cse.rs
```diff
@@ -505,6 +614,323 @@ builtin.function public extern("C") @simple_constant() -> (i32, i32) {
...
+// CHECK-NEXT: [[CANONICAL_USER:%\d+]] = arith.sub [[CANONICAL]], [[RHS]]
+// CHECK-NEXT: [[VISITED:%\d+]] = arith.sub [[DUP:%\d+]], [[RHS]]
+// CHECK-NEXT: [[DUP]] = arith.constant 1 : i32;
+// CHECK-NEXT: hir.assert_eq [[CANONICAL_USER]], [[CANONICAL_USER]]
+// CHECK-NEXT: hir.assert_eq [[VISITED]], [[VISITED]]
+            "#
+        );
+    }
+
+    #[test]
```
bitwalker: I think we should start writing these tests under the `lit` directory, and use `hir-opt` to run specific passes like this. It'll be a lot easier to review the input IR and what is expected to change, compared to trying to map the Rust APIs to HIR text in your head when looking at the expected output.

If there are specific obstacles to doing so, we should extend `hir-opt` as needed to support whatever functionality we require, since it's primary purpose for existing is these sorts of tests.

bitwalker (approved):
Marking approved, but I think it would be worth testing out whether the tests here can be written as pure `lit` tests using `hir-opt`, and if something is missing in order for it to work, opening issues so we can add those features when we get to it. I'm not going to block merging on that though, but I would really like to get away from verbose IR tests written in Rust - they're just super brittle and hard to follow.

hir-transform/src/cse.rs
greenhat: Thank you! Done in c32409a09bc082fd0745427b181f154088527057.

greenhat:
I don't want to merge it because it might mess with #1135.

PR #1131, opened 2026-05-15
feat: add `load_qw` intrinsic

### LINKED ISSUE #1115: Implement `load_qw` intrinsic
After implementing `store_qw`, the test `overflowing_rem_i128` hits `load_qw`.

Context: #1111

mooori:
Closes #1115 by implementing the `load_qw` intrinsic.

Tests that needed `load_qw` now fail for another reason, so `#[ignore = "..."]` have been updated.


tests/integration/src/codegen/memory/load_qw.rs
```diff
@@ -0,0 +1,158 @@
...
+            &[Felt::new(offs as u64)],
+            context.session(),
+            |_trace| Ok(()),
+        )
+        .unwrap();
+
+        assert_eq!(output, expected, "offset {offs}");
+    };
+
+    for offs in 0..=3 {
```
greenhat: Consider expanding `initial_bytes` to at least 32 bytes and test offsets `0..=15`, which covers every byte alignment for a 16-byte load.

tests/integration/src/codegen/memory/load_qw.rs
```diff
@@ -0,0 +1,158 @@
...
+fn load_qw_with_offset() {
+    setup::enable_compiler_instrumentation();
+
+    // Use the start of the 17th page (1 page after the 16 pages reserved for the Rust stack).
+    let read_from = 17 * 2u32.pow(16);
+
+    // Generate a `test` module with `main` function that invokes `load_qw` when lowered to MASM.
+    // The parameter is the offset to be applied to the base address (`read_from`).
+    // Compile once outside the test loop.
+    let (package, context) = compile_test_module([Type::U32], [Type::I128], |builder| {
```
greenhat, line 14: Should we also test `Type::U128` or is the lowering path sign-agnostic?

greenhat (changes requested):
Looking good! Please check my notes.

tests/integration/src/codegen/memory/load_qw.rs
mooori: Good catch! Done

tests/integration/src/codegen/memory/load_qw.rs
mooori: The `proc load_qw` intrinsic itself is sign-agnostic, but the lowering path for the test module isn't. Load/store `Type::U128` is not yet supported. I've opened #1137 with notes to test `U128` as well in `load_qw` and `store_qw` tests.

mooori:
Force-pushed after rebasing on latest `next`. Only [75394da](https://github.com/0xMiden/compiler/pull/1131/commits/75394da42400664f413867a8b917209ea9895dc7) is new, previous commits are unmodified.

codegen/masm/intrinsics/mem.masm
```diff
@@ -310,6 +310,39 @@ pub proc realign_dw
     u32or # [x_lo, x_hi]
 end
 
+# Realign an unaligned quad-word.
+#
+# See `realign_dw` for a detailed description of the approach.
+#
+# Stack transition is [e4, e3, e2, e1, e0, offset] -> [x0, x1, x2, x3]
+pub proc realign_qw
```
bitwalker: Let's add a type signature here, i.e.:

```
pub proc realign_qw(qw: u128, offset: u8) -> u128
```

You can also do the following if you want to preserve the naming of each limb:

```
pub proc realign_qw(e4: u32, e3: u32, e2: u32, e1: u32, e0: u32, offset: u8) -> (x0: u32, x1: u32, x2: u32, x3: u32)
```

Note that in the both cases we're being clear about the assumption that the input/output is comprised of 4 `u32` limbs (as that is the defined representation of `u128`), and the procedure body here also makes that assumption.

codegen/masm/intrinsics/mem.masm
```diff
@@ -359,6 +392,40 @@ pub proc load_dw # [addr, offset]
     end
 end
 
+# Load four 32-bit words to the operand stack
+pub proc load_qw # [addr, offset]
```
bitwalker: Same here with type signature:

```
pub proc load_qw(addr: ptr<u32, addrspace(felt)>, offset: u8) -> u128
```

bitwalker (approved):
I only skimmed the implementation for now, but overall looks good. I left a couple of nits about adding procedure signatures for the intrinsics, but this can be merged once that's addressed.

greenhat (approved):

PR #1132, opened 2026-05-16
test: add regression test for production of invalid bitcast

mooori:
Adds a test to reproduce #1133 which can be run with

```
cargo make midenc
MIDENC_BIN_DIR="$PWD/bin" \
    litcheck lit run -v tests/lit/regression/width_changing_bitcast.wat
```

Could be added as-is and then updated once the issue was fixed. Or be merged only after the fix. Any preferences?

### DIFF
--- a/examples/auth-component-no-auth/Cargo.lock
+++ b/examples/auth-component-no-auth/Cargo.lock
@@ -1039,7 +1039,6 @@ version = "0.12.0"
 dependencies = [
  "heck",
  "miden-formatting",
- "miden-mast-package",
  "miden-protocol",
  "midenc-frontend-wasm-metadata",
  "proc-macro2",
--- a/tests/lit/regression/width_changing_bitcast.wat
+++ b/tests/lit/regression/width_changing_bitcast.wat
@@ -1,0 +1,25 @@
+;; In this sequence, `i32.wrap_i64` produced an invalid bitcast from `i64` to `i32`.
+;;
+;; TODO stop producing the invalid bitcast + make this test assert valid ops are generated
+;;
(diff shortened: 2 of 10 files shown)


tests/lit/regression/width_changing_bitcast.wat
```diff
@@ -0,0 +1,25 @@
+;; In this sequence, `i32.wrap_i64` produced an invalid bitcast from `i64` to `i32`.
+;;
+;; TODO stop producing the invalid bitcast + make this test assert valid ops are generated
+;;
+;; RUN: /bin/sh -c "TMPDIR=\$(mktemp -d) && midenc %s --entrypoint=width_changing_bitcast::wrap_i64 --emit=hir=\"\$TMPDIR\" --emit=masm=\"\$TMPDIR\" -o \"\$TMPDIR/width_changing_bitcast.masp\" && cat \"\$TMPDIR/root.hir\"" | filecheck %s
```
bitwalker, line 5:
```suggestion
;; RUN: midenc %s --emit=hir=- | filecheck %s
```

You can drastically simplify the `RUN` line here since you are only checking the HIR output, and `midenc` can output the HIR directly to stdout for you

examples/auth-component-no-auth/Cargo.lock
bitwalker, line 1: I would avoid checking in these files unless your changes directly resulted in them changing, otherwise you have to deal with merge conflicts (ideally we wouldn't check in the `Cargo.lock` for these at all, but they're still needed for now I think).

bitwalker (commented):
I would suggest we _not_ merge this, since this is not a regression that is likely to resurface again, as the issue is a very straightforward bug in the frontend translation from Wasm, rather than the result of rewrites/optimizations, or something in codegen.

That said, if we do merge it, that's fine too, so I've noted the only real change that I'd want to make first.

PR #1135, opened 2026-05-18
feat: major refactor of midenc/cargo-miden to support Miden projects and MASM compilation

bitwalker:
This branch started as a spike to develop a disassembler from MASM to HIR, and then integrate that into `midenc` to use in linting/analyzing MASM. This effectively required integrating the new `miden-project.toml` support into `midenc`, which in turn meant that the way we were previously compiling Rust/Cargo projects also needed to be reworked. Supporting `miden-project.toml` also means reworking how the SDK macros integrate with projects, and how metadata is extracted for use. All of this also necessarily impacts how we resolve dependencies and compile projects in workspaces.

Long story short, this branch ended up being a massive chunk of work over the past couple of weeks, across almost 100 commits. I'm opening this as a PR primarily so that the diff is available for posterity, but I don't expect it to actually be reviewed before merging - there is simply too much change involved to accomplish the necessary refactorings. Instead, I would recommend looking at the parts of the code you are familiar with, for anything that you are concerned about, in case anything looks out of place, or that you think requires follow-up, and we can discuss in more detail in new issues or in our compiler sync.

Highlights of what is implemented here:

* `midenc` can now be used directly to compile Rust or MASM source files, Cargo projects, and Miden projects (whether Cargo-based or MASM-based). The `cargo-miden` extension now largely just delegates to `midenc` via `midenc-compile`
* The compiler pipeline is now broken up into multiple sub-pipelines, and some stages broken up into smaller/more precise stages so as to allow greater re-use while simplifying each stage. Stages for Rust, Cargo, and MASM inputs have been added
* A new analysis stage was added, which will run a new unconstrained advice taint analysis over an HIR world, and identify various interesting cases where advice data is used without asserting/validating its shape/range. This is primarily a proof-of-concept, but provides a foundation we can expand on, or be used as a reference for how to build such an analysis using the compiler's dataflow analysis framework.
* A new `midenc-frontend-masm` crate which is used to lift/disassemble MASM into HIR. This is done best-effort, but can handle both untyped and typed MASM procedures, the former by inferring the stack effects and types based on the procedure body. Liberal use of `builtin.unrealized_conversion_cast` is used to handle type mismatches for now.
* Compiler configuration and session setup is now centralized in `midenc-session`, in order to reduce drift between various entrypoints to compilation, and types of inputs/targets.
* The use of `Cargo.toml` for most Miden/component configuration has been moved to `miden-project.toml`. The latter is now required in all Rust projects.

There are also a handful of other small improvements/bug fixes (such as a fix to `Operation::is_a_proper_ancestor_of_b` which was not visiting more than the immediate parent of an operation due to a mismanaged cursor).

I plan to merge this sometime Monday, after our compiler sync, since I would prefer we base all of our ongoing work on top of these changes.


tools/cargo-miden/src/template.rs
```diff
@@ -86,6 +86,23 @@ pub fn generate(args: GenerateArgs) -> Result<PathBuf> {
 
     render_template(&source_root, &project_dir, &parser, &variables, &config)?;
 
+    // Generate miden-project.toml directly temporarily, if not rendered by template.
```
greenhat, line 89: I think we should add a `miden-project.toml` in our templates rather than generate it.

examples/p2id-note/miden-project.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+
+[lib]
+kind = "note"
+namespace = "miden:p2id/miden-p2id@0.1.0"
+
+[dependencies]
+miden-core = "*"
+miden-protocol = "*"
+basic-wallet = { path = "../basic-wallet" }
```
greenhat, line 12: We should remove the fields "handled" by `miden-project.toml` from the `Cargo.toml` in our examples (`kind`, `namespace`, dependencies, etc.). I think we should also remove support for those fields from the `Cargo.toml` and only support them in `miden-project.toml` to avoid confusion.

greenhat (approved):
Super cool! I've gone through the SDK, cargo-miden and Wasm frontend. Please check my notes.

examples/p2id-note/miden-project.toml
bitwalker: Agreed, I'll follow up on this PR with another to warn if the old Cargo.toml-based config is detected alongside `miden-project.toml` (and improve the handling of missing `miden-project.toml` in general

PR #1136, opened 2026-05-18
feat: improve debug loc tracking

djolertrk:
Improve variable tracking.

This goes on top of  `Add synthetic source locations` https://github.com/0xMiden/compiler/pull/812.

### DIFF
--- a/frontend/wasm/src/module/function_builder_ext.rs
+++ b/frontend/wasm/src/module/function_builder_ext.rs
@@ -196,10 +196,10 @@ impl<B: ?Sized + Builder> FunctionBuilderExt<'_, B> {
 
         if let Some((file_symbol, _directory, line, column)) = self.span_to_location(span) {
             attr.file = file_symbol;
-            if line != 0 {
+            if should_fill_debug_attr_location(attr.line, attr.column, line, column) {
                 attr.line = line;
+                attr.column = column;
             }
-            attr.column = column;
         }
 
         // If DWARF didn't provide a location expression, synthesize one from the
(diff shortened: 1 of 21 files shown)


frontend/wasm/src/module/function_builder_ext.rs
```diff
@@ -712,9 +750,18 @@ impl<B: ?Sized + Builder> FunctionBuilderExt<'_, B> {
         let context = self.inner.builder().context();
         let session = context.session();
         let source_file = session.source_manager.get(span.source_id()).ok()?;
-        let uri = source_file.uri().as_str();
-        let path = Path::new(uri);
-        let file_symbol = Symbol::intern(uri);
+        let uri = source_file.uri();
```
bitwalker, line 753: This will need to respect `--remap-path-prefix` once rebased on `next`

hir/src/patterns/driver.rs
```diff
@@ -486,6 +486,10 @@ impl GreedyPatternRewriteDriver {
 
         let op = op_ref.borrow();
 
+        if op.implements::<dyn Transparent>() {
```
bitwalker: I don't think this is a good idea - this will make it impossible to write rewrite patterns which are anchored to a `Transparent` op, and we may want to do so.

tests/lit/debuginfo/inline-stubs.wat
```diff
@@ -1,6 +1,7 @@
 ;; RUN: midenc %s --entrypoint=test_felt_add --emit=hir=- 2>&1 | filecheck %s --check-prefix=HIR-ADD
 ;; RUN: midenc %s --entrypoint=test_felt_assert_eq --emit=hir=- 2>&1 | filecheck %s --check-prefix=HIR-ASSERT
 ;; RUN: midenc %s --entrypoint=test_felt_arithmetic --emit=hir=- 2>&1 | filecheck %s --check-prefix=HIR-ARITH
+;; RUN: midenc %s --entrypoint=inline_stubs_test::test_felt_add -Z print-hir-source-locations --emit=hir=- 2>&1 | filecheck %s --check-prefix=HIR-LOC
```
bitwalker:
```suggestion
;; RUN: midenc %s -Zprint-hir-source-locations --emit=hir=- -Canalyze-only 2>&1 | filecheck %s --check-prefix=HIR-LOC
```

No need to run the full compilation pipeline for this test, and you can just compile it as a library, rather than specifying an entrypoint

tools/cargo-miden/src/commands/build.rs
```diff
@@ -211,13 +211,13 @@ fn build_cargo_args(cargo_opts: &CargoOptions) -> Vec<String> {
     // Configure profile settings
     let cfg_pairs: Vec<(&str, &str)> = vec![
         ("profile.dev.panic", "\"abort\""),
-        ("profile.dev.opt-level", "1"),
+        // Keep dev builds unoptimized so Rust preserves source variable locations for debugger
```
bitwalker: I think we should probably derive this value from `Options::opt_level`, so that we can control it a bit better from the CLI

bitwalker: I'll also note that `miden-project.toml` supports build profiles, and we'll be deriving certain configuration options from there as well, so in general we'll want to make more of these options configurable (or at least derived from user-provided configuration)

bitwalker (changes requested):
Looks good! I left a few comments, but this is largely ready to merge, once its rebased on `next`

tools/cargo-miden/src/commands/build.rs
djolertrk: sure, I am addressing this

djolertrk:
NOTE: This is not ready yet, as I am chasing a bug in debug value instruction handling in some transformations.

djolertrk:
> NOTE: This is not ready yet, as I am chasing a bug in debug value instruction handling in some transformations.

But it seems it is not a debug info related, but rather a transformation/codegen bug. Will open PR soon.

djolertrk:
PR: https://github.com/0xMiden/compiler/pull/1141

djolertrk:
hey @bitwalker, is this ready to be merged? :)

bitwalker (approved):

PR #1139, opened 2026-05-20
fix(frontend/wasm): make some `i64` unary instructions return `i64`

### LINKED ISSUE #1121: native/Masm divergence for `u128::overflowing_div`
### Packages versions

#1119 on top of `next`

### Bug description

native/Masm divergence for `u128::overflowing_div`

### How can this be reproduced?

```sh
cargo nextest run -p midenc-integration-tests \
    --run-ignored all \
    -E 'test(/overflowing_div_u128/)'
```

Requires #1119

### Relevant log output

```shell
thread 'end_to_end::arithmetic::overflowing_div_u128' (86208) panicked at tests/integration/src/end_to_end/arithmetic.rs:924:13:
    Found minimal(shrinked) failing case: (250781882541897244803871015896621252608, 17)
    Failure: Reason("assertion failed: `(left == right)` \n  left: `[240, 240, 240, 240, 240, 240, 240, 240, 240, 240, 16, 15, 157, 27, 25, 11, 0]`, \n right: `[0, 208, 220, 2, 80, 56, 36, 251, 240, 240, 16, 15, 157, 27, 25, 11, 0]`: VM output mismatch at tests/integration/src/end_to_end/arithmetic.rs:917")
```

### LINKED ISSUE #1122: native/Masm divergence for `u128::overflowing_rem`
### Packages versions

#1119 on top of `next`

### Bug description

native/Masm divergence for `u128::overflowing_rem`

### How can this be reproduced?

```sh
cargo nextest run -p midenc-integration-tests \
    --run-ignored all \
    -E 'test(/overflowing_rem_u128/)'
```

### Relevant log output

```shell
thread 'end_to_end::arithmetic::overflowing_rem_u128' (92751) panicked at tests/integration/src/end_to_end/arithmetic.rs:924:13:
    Found minimal(shrinked) failing case: (231051897244068755437035091323978252288, 5)
    Failure: Reason("assertion failed: `(left == right)` \n  left: `[3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`, \n right: `[95, 28, 53, 108, 40, 94, 245, 111, 255, 255, 255, 255, 255, 255, 255, 255, 0]`: VM output mismatch at tests/integration/src/end_to_end/arithmetic.rs:917")
```

### LINKED ISSUE #1133: Invalid bitcast generation `i64 -> i32` (different widths)
Ops `div`, `rem` with 128 bit-wide operands lead to HIR bitcast from `i64` to `i32`, which is invalid:

https://github.com/0xMiden/compiler/blob/fec54ae8a69bc146135e75185e69ab2489d310f1/dialects/hir/src/builders.rs#L482-L483

## Reproduction

See the test added in #1132

## Proposed fix

- Add guards to avoid emitting invalid bitcasts and report an error instead
- Produce `trunc/zext/sext/cast` instead of `bitcast` when the width changes

mooori:
Closes #1133
Closes #1121
Closes #1122
Allows unignoring some tests in `arithmetic.rs`

Change Wasm translation such that `i64.clz, i64.ctz, i64.popcnt` return `i64` instead of `i32`.

According to the [Wasm spec](https://webassembly.github.io/profiles/core/syntax/instructions.html#numeric-instructions) they return `i64` because:

> Unary Operations: consume one operand and produce one result of the respective type.

Also the [instruction overview](https://webassembly.github.io/profiles/core/appendix/index-instructions.html) lists them with `[i64] -> [i64]`.

Here's a [compiler explorer showcase](https://godbolt.org/z/1EGGPTjvr): Rust's `leading_zeros()` returns `u32`, so there's a `i32.wrap_i64` to convert `i64.clz` output.

### Performance impact

None in expect tests. Might be that none of the instructions is hit in tests that assert cycle counts. Which aligns with the bug being hidden so far. Recently added overflowing arith tests for 128 bit ints surfaced it.

For some programs the cycle count might slightly increase as now there's an extra `zext`.

### `arith` dialect return types

Currently `arith` ops `Clz, Ctz, Popcnt` all return `UInt32`. Could/should that be changed to `Or<UInt32, UInt64>`? Might avoid the `zext`.

### DIFF
--- a/frontend/wasm/src/code_translator/expected/i64_clz.hir
+++ b/frontend/wasm/src/code_translator/expected/i64_clz.hir
@@ -1,7 +1,8 @@
 builtin.function public extern("C") @test_wrapper() {
     %0 = arith.constant 1 : i64;
     %1 = arith.clz %0;
-    %2 = hir.bitcast %1 <{ ty = #builtin.type<i32> }>;
+    %2 = arith.zext %1 <{ ty = #builtin.type<u64> }>;
+    %3 = hir.bitcast %2 <{ ty = #builtin.type<i64> }>;
     cf.br ^block5;
 ^block5:
     builtin.ret;
--- a/frontend/wasm/src/code_translator/expected/i64_ctz.hir
+++ b/frontend/wasm/src/code_translator/expected/i64_ctz.hir
@@ -1,7 +1,8 @@
 builtin.function public extern("C") @test_wrapper() {
     %0 = arith.constant 1 : i64;
     %1 = arith.ctz %0;
-    %2 = hir.bitcast %1 <{ ty = #builtin.type<i32> }>;
+    %2 = arith.zext %1 <{ ty = #builtin.type<u64> }>;
+    %3 = hir.bitcast %2 <{ ty = #builtin.type<i64> }>;
     cf.br ^block5;
 ^block5:
     builtin.ret;
--- a/frontend/wasm/src/code_translator/expected/i64_popcnt.hir
+++ b/frontend/wasm/src/code_translator/expected/i64_popcnt.hir
@@ -1,0 +1,9 @@
+builtin.function public extern("C") @test_wrapper() {
+    %0 = arith.constant 1 : i64;
+    %1 = arith.popcnt %0;
+    %2 = arith.zext %1 <{ ty = #builtin.type<u64> }>;
+    %3 = hir.bitcast %2 <{ ty = #builtin.type<i64> }>;
+    cf.br ^block5;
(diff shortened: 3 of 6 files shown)


bitwalker:
> Currently arith ops Clz, Ctz, Popcnt all return UInt32. Could/should that be changed to Or<UInt32, UInt64>? Might avoid the zext.

Lets keep it as `u32`, the underlying instructions return 32-bit values, so for Wasm the extension will always be required

bitwalker (approved):

PR #1140, opened 2026-05-21
fix: edge case handling in `i32::overflowing_mul` intrinsic

### LINKED ISSUE #1093: native/MASM divergence in the wrapping `mul` and `add`
Discovered by fuzza (#1087)

See `mulladd` test.

mooori:
Closes #1093 and allows unignoring arithmetic test `overflowing_mul_i128`.

### `u32widening_mul` returns `[hi, lo]`

In the compiler intrinsic `i32::overflowing_mul` the result of `u32widening_mul` was misinterpreted:

```
[overflowed, result] // previously
[hi, lo]             // actually
```

Addressed in this PR by converting hi to overflowed.

### `a * b = i32::Min` is not overflow

Currently this is incorrectly interpreted as overflow because it overflows the positive `i32` range:

> [...] the supposedly unsigned result has its sign bit set, which could only happen if we overflowed the positive i32 range

Fixed by handling that edge case.

### Testing compiler intrinsics

Hitting a particular intrinsic in tests can be difficult, which made me think about a way to test intrinsics directly. See #1142.

### DIFF
--- a/codegen/masm/intrinsics/i32.masm
+++ b/codegen/masm/intrinsics/i32.masm
@@ -174,14 +174,19 @@ pub proc overflowing_mul # [b, a]
         movup.2 cdrop                     # [-a or a, is_b_signed, b, negate_result]
         swap.2 dup.0 exec.unchecked_neg   # [-b, b, is_b_signed, -a or a, negate_result]
         movup.2 cdrop                     # [-b or b, -a or a, negate_result]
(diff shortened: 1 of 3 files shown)


greenhat (approved):
Looking good!

mooori:
(force-pushed after rebase on latest `next`, no changes)

PR #1141, opened 2026-05-22
fix(scf): remove unused index_switch results

djolertrk:
Fixes `scf.index_switch` canonicalization so unused switch results are removed, along with the corresponding operands from each case/default `scf.yield`.

Previously, dead switch results could keep dead yielded values threaded through control flow, which later confused MASM operand scheduling. This mirrors the existing unused-result cleanup already implemented for scf.if and adds a focused HIR test covering the rewrite.


bitwalker (approved):

bitwalker:
@djolertrk Just need to sign your commit

djolertrk:
addressed the comment and signed the commit. Thanks @bitwalker!

PR #1148, opened 2026-05-28
feat(hir): implement dialect conversion infrastructure

bitwalker:
This introduces dialect conversion infrastructure similar to MLIR, so that higher-level dialects can be introduced and then converted via legalization to dialects that the codegen backend understands.

Today, we rely on the compiler pipeline to only produce ops which implement `HirLowering` when passing IR to the backend. If an illegal/unlowerable op reaches the backend, we currently just panic with an error explaining that we expected the op to implement `HirLowering`.

With this change, we now define what dialects (and in some cases, a subset of ops in a dialect) are legal for codegen, and predicate this on those ops implementing `HirLowering`. If a new dialect or op is added, we now get a legalization diagnostic if there is no path to legalizing that dialect (or op). Not only will this catch ops that we haven't explicitly legalized, it will also catch ops we've stated are legal but which don't implement `HirLowering`. If we add a new dialect `A`, and define how to convert it to dialect `B`, and there is a conversion path from `B` to `C`, where `C` is legal for codegen - then `A` can be implicitly legalized by conversion to `C` through `B` - this is handled by the dialect conversion framework.

My main motivation for implementing this, aside from it being in the backlog for a long time, is that it allows my new experimental programming language to use a new HIR dialect as its intermediate representation, and then lower through our existing dialects using this dialect conversion infrastructure.

### DIFF
--- a/codegen/masm/src/legalization.rs
+++ b/codegen/masm/src/legalization.rs
@@ -1,0 +1,210 @@
+use alloc::rc::Rc;
+
+use midenc_dialect_arith as arith;
+use midenc_dialect_cf as cf;
+use midenc_dialect_hir as hir;
+use midenc_dialect_scf as scf;
+use midenc_dialect_ub as ub;
+use midenc_dialect_wasm as wasm;
+use midenc_hir::{
+    Context, EntityMut, Operation, OperationName, Report,
+    conversion::{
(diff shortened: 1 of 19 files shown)


greenhat (approved):
Looking good!

PR #1149, opened 2026-05-29
Wasm CM `variant` (Rust `enum`), `option` and `result` support

### LINKED ISSUE #111: Full Wasm CM types support

greenhat:
Ref #111 

Adds component-model support for Rust enums and generic wrappers across the SDK macros and the Wasm frontend, so variant-shaped values can round-trip through the canonical ABI.

- **WIT variants for Rust enums**: the SDK component macros encode exported enums as WIT variants, and the frontend translates component variants/enums into HIR `EnumType` values. Canonical ABI load/store follows the active case's payload layout while joining flat payload slots position by position, preserving the flattened ABI contract across imports and exports.
- **`Option<T>` and `Result<T, E>` mapping**: macro type resolution now maps these to WIT `option`/`result` with recursive dependency tracking, so nested SDK core types and exported records keep their imports and validation.


greenhat:
I rebased and cleaned up the code after the FPI merge. I extracted the `variant` in FPI support into #1177 since this PR is already quite big.

greenhat:
Rebased.

frontend/wasm/src/code_translator/mod.rs
```diff
@@ -414,6 +417,10 @@ pub fn translate_operator<B: ?Sized + Builder>(
             let val = state.pop1_bitcasted(Felt, builder, span);
             state.push1(val);
         }
+        Operator::I32ReinterpretF32 => {
```
bitwalker, line 420: This probably needs to use a safer cast, since converting felts to `u32`/`i32` isn't guaranteed to succeed. I would double check what the semantics of this operation are, and either use `cast` or `trunc`, depending on whether this conversion is fallible or not.

frontend/wasm/src/component/lift_exports.rs
```diff
@@ -164,6 +183,22 @@ fn generate_lifting_with_transformation(
          have a pointer in the result",
     );
 
+    // The lowered core function implementing a transformed export takes the flattened
```
bitwalker, line 185: Does the core function _always_ return a pointer, or only in the cases where the result is not an immediate type, e.g. `i32`? I'd naturally assume that pointer indirection is only required when the result type requires heap allocation (or cannot be returned by-value as a supported immediate type in Core Wasm, e.g. `i32`/`i64`/`f32`).

frontend/wasm/src/component/flat.rs
```diff
@@ -39,40 +41,51 @@ pub enum CanonicalTypeError {
...
-/// Byte layout for one value produced by canonical ABI flattening.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct CanonicalFlatLayoutEntry {
-    /// Byte offset of this flattened value relative to the containing tuple.
-    pub offset: u32,
-    /// Source type to load from memory for this flattened value.
-    pub ty: Type,
+/// Identifies the wrapper transformation required by canonical ABI flattening.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum CanonicalAbiTransformation {
```
bitwalker: Since this isn't a transformation per se, but really an implementation detail of our internal parameter/result passing ABI, let's rename this enum to `CanonicalAbiIndirection` and the variants to:

* `None` - no change
* `In` - replaces `ParamTuple`, indicates there is pointer indirection used for passing inputs/parameters
* `Out` - replaces `ResultOutPtr`, indicates there is pointer indirection used for returning outputs/results
* `InOut` - replaces `Both`, indicates there is pointer indirection used for both inputs/outputs

This better reflects the ABI detail that it represents, and makes the names a bit more intuitive (i.e. it both makes it immediately clear whether there is indirection or not, and if there is, what the dataflow of the indirection is). It is unlikely we would add multiple methods of input/output indirection, so naming these variants after the method used I think obfuscates the meaning a bit.

frontend/wasm/src/component/canon_abi_utils.rs
```diff
@@ -73,64 +79,610 @@ pub fn load<B: ?Sized + Builder>(
...
+    span: SourceSpan,
+) -> WasmResult<()> {
+    if payload_flat_types.is_empty() {
+        validate_variant_discriminant(fb, discriminant, cases.len(), span)?;
+        return Ok(());
+    }
+
+    let payload_addr = offset_addr(fb, ptr, payload_offset32, span)?;
+    let join_block = fb.create_block_with_params(payload_flat_types.iter().cloned(), span);
+    let case_blocks = switch_variant_cases(fb, discriminant, cases.len(), span)?;
```
bitwalker: Just a note - but this is an example of a place where the frontend could just emit ops from the `scf` dialect directly, rather than lowering to the `cf` dialect and relying on the SCF transform. We know that the control flow here is guaranteed to be structured (i.e. will contain no branches), and that will produce more efficient IR than if we rely on the SCF transform (while the transform should be able to recover equivalent representation to what we'd emit by hand, it's not guaranteed).

frontend/wasm/src/component/types/mod.rs
```diff
@@ -1347,6 +1361,494 @@ impl VariantInfo {
...
+    pub ir: FunctionType,
+    /// Canonical ABI type trees for the function parameters.
+    pub params: Box<[CanonicalAbiType]>,
+    /// Canonical ABI type trees for the function results.
+    pub results: Box<[CanonicalAbiType]>,
+}
+
+/// Canonical ABI lowering metadata for one component-model type.
+#[derive(Clone, Debug)]
+pub struct CanonicalAbiType {
```
bitwalker: Do we really need a parallel type hierarchy just for the Canonical ABI? As best I can tell, the only thing it provides is memoized access to the flattened representation/size - but are we accessing it enough where that pays off? I would imagine that in the vast majority of cases, computing the flattened representation would be very cheap, so this adds quite a lot of code for what appears to be not a lot of benefit.

What's your take? Does it pull its weight in your opinion?

bitwalker (changes requested):
Looks good! I had a few nits/notes/questions, but nothing major

frontend/wasm/src/code_translator/mod.rs
greenhat: We discussed it on one of our calls. It's emitted by rustc (CanonABI) when joining the `enum` variants before the function call. For a brief moment, the felt is cast to `i32` before passing into the function. On the callee side it is cast back to `f32` (felt). For example, in the
`result_with_felt_ok_and_u64_error_payloads` test in the `pub fn
roundtrip(&self, value: Result<Felt, u64>) -> ...` when value is
`Ok([SOME_FELT])` the joined enum is encoded as `i64` with the following Wasm:

```wat
i32.reinterpret_f32
i64.extend_i32_u
```

And then on the callee side back to `felt` with:

```wat
i32.wrap_i64
f32.reinterpret_i32
```

I used the maximum `Felt` values to check that felt survives untouched in the tests. It's ugly, but I haven't found another way.

frontend/wasm/src/component/lift_exports.rs
greenhat: No, not always. Your assumption is correct; the core function only returns a pointer when the result is not an immediate type. But this (`generate_lifting_with_transformation`) function handles the indirect case only. The direct case is handled in the `generate_direct_lifting` function.

frontend/wasm/src/component/types/mod.rs
greenhat: Absolutely, it is used very extensively. It ties the type with its shape and ABI info so they are always traveling together. Plus, there is transformation from the Wasm CM types and a handful of helpers.

frontend/wasm/src/component/flat.rs
greenhat: Done in af526e5c63abf85b4f2cf9c332df9046e070f9f5

greenhat:
Thank you! I addressed the notes. Please do another round.

frontend/wasm/src/code_translator/mod.rs
bitwalker: Let's document this in the code here, since I believe this is the only reason we would want to allow this op to appear in Wasm inputs, and it'll prevent us from regressing this accidentally in the future. Unfortunately, it doesn't seem like there will be a way for us to catch if this is used for something other than the specific case you outlined - but since we document that `f32` is unsupported, if it causes unintended behavior elsewhere, hopefully it is due to user error, not some change in how our use of `f32` for felts is emitted by rustc/LLVM

frontend/wasm/src/component/types/mod.rs
bitwalker: I guess what I'm saying is that aside from memoizing the flattened representation, this appears to be almost exactly 1:1 lined up with the type system in `midenc-hir-types`. Transforming Wasm CM types to the types in `midenc-hir-types` would look basically identical to the conversion to this parallel set of types. The types in `midenc-hir-types` also carry their shape and ABI details - the only thing they lack is the conversion to the "flattened" representation which is defined by the Canonical ABI (and they only lack that because it doesn't make much sense to compute that up front, versus on-demand).

I'd like to avoid duplicating details like this if possible - but if this set of types is truly different, and has its place, I can accept that; I'm just isn't apparent to me that it really is. Assume for a moment we had defined conversion helpers in `midenc-hir-type` to obtain the flattened representation for the Canonical ABI, what else is different enough to warrant a separate set of types? That might help clarify why these need to exist.

greenhat: I think you’re right that the extra hierarchy is more than we need. One caveat: we cannot use HIR’s existing field/enum offsets directly because HIR layout differs from Canonical ABI layout, e.g., 64-bit scalar alignment (HIR offset is 4 vs. CanonABI is 8).

I took this direction and removed the extra `CanonicalAbiType` hierarchy in 1b5b90b9408ef061c54da7c312d23f7a5ac74b53.

The HIR `Type` tree already carries the record/enum shape we need, so the remaining Canonical ABI-specific pieces are now query/layout helpers over HIR types.

bitwalker: Ah interesting - I guess we should consider whether we should adjust the alignment for 64-bit scalars in `midenc-hir-types` to 8. IIRC though, it is defined as 4 somewhere (I can't recall if it is in the LLVM data layout for `wasm32` or somewhere else) - but we could confirm what it is elsewhere in case we've chosen too small an alignment for some reason.

bitwalker (approved):

PR #1151, opened 2026-05-29
feat: preserve ABI types

djolertrk:
Typed ABI/debug consumers need to distinguish the lowered execution ABI from the source-level/component ABI. Without preserving the semantic signature, tools only see flattened `felt` parameters/results and cannot recover types like `account-id`, `word`, or `asset`.

### DIFF
--- a/codegen/masm/src/artifact.rs
+++ b/codegen/masm/src/artifact.rs
@@ -214,10 +214,12 @@ impl MasmComponent {
 
         let mut exe = Box::new(masm::Module::new_executable());
         let span = SourceSpan::default();
+        let mut invoked = Vec::new();
         let body = {
             let mut block = masm::Block::new(span, Vec::with_capacity(64));
             // Invoke component initializer, if present
             if let Some(init) = self.init.as_ref() {
+                invoked.push(masm::Invoke::new(masm::InvokeKind::Exec, init.clone()));
                 block.push(Op::Inst(Span::new(span, Inst::Exec(init.clone()))));
             }
 
@@ -231,6 +233,7 @@ impl MasmComponent {
                 span,
                 Inst::Trace(TraceEvent::FrameStart.as_u32().into()),
             )));
+            invoked.push(masm::Invoke::new(masm::InvokeKind::Exec, entrypoint.clone()));
             block.push(Op::Inst(Span::new(span, Inst::Exec(entrypoint.clone()))));
             block
                 .push(Op::Inst(Span::new(span, Inst::Trace(TraceEvent::FrameEnd.as_u32().into()))));
@@ -242,16 +245,18 @@ impl MasmComponent {
                 let qualified = masm::QualifiedProcedureName::new(module.as_path(), name);
                 InvocationTarget::Path(Span::new(span, qualified.into_inner()))
             };
+            invoked.push(masm::Invoke::new(masm::InvokeKind::Exec, truncate_stack.clone()));
             block.push(Op::Inst(Span::new(span, Inst::Exec(truncate_stack))));
             block
         };
-        let start = masm::Procedure::new(
+        let mut start = masm::Procedure::new(
             span,
             masm::Visibility::Public,
             masm::ProcedureName::main(),
             0,
             body,
         );
+        start.extend_invoked(invoked);
         exe.define_procedure(start, source_manager)
             .into_diagnostic()
             .wrap_err("failed to define executable `main` procedure")?;
--- a/codegen/masm/src/lower/component.rs
+++ b/codegen/masm/src/lower/component.rs
@@ -739,6 +739,7 @@ impl MasmFunctionBuilder {
             };
             let span = SourceSpan::default();
             // Add init call to the emitter's target before emitting the function body
+            emitter.invoked.insert(masm::Invoke::new(masm::InvokeKind::Exec, init.clone()));
             emitter
                 .target
                 .push(masm::Op::Inst(Span::new(span, masm::Instruction::Exec(init))));
@@ -761,6 +762,7 @@ impl MasmFunctionBuilder {
                 InvocationTarget::Path(Span::new(SourceSpan::default(), qualified.into_inner()))
             };
             let span = SourceSpan::default();
+            invoked.insert(masm::Invoke::new(masm::InvokeKind::Exec, truncate_stack.clone()));
             body.push(masm::Op::Inst(Span::new(span, masm::Instruction::Exec(truncate_stack))));
         }
         let Self {
@@ -862,9 +864,10 @@ fn masm_component_type_expr_from_hir(ty: &Type) -> masm::TypeExpr {
                     .with_span(SourceSpan::UNKNOWN),
             )
         }
-        Type::Ptr(ptr) => masm::TypeExpr::Ptr(masm::PointerType::new(
-            masm_component_type_expr_from_hir(ptr.pointee()),
-        )),
+        Type::Ptr(ptr) => masm::TypeExpr::Ptr(
+            masm::PointerType::new(masm_component_type_expr_from_hir(ptr.pointee()))
+                .with_address_space(ptr.addrspace()),
+        ),
         Type::Function(_) => masm::TypeExpr::Ptr(masm::PointerType::new(
             masm::TypeExpr::Primitive(Span::unknown(Type::Felt)),
         )),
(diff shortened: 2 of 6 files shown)


djolertrk:
NOTE: For this to work end-to-end, we need a PR in VM, but since compiler is not migrated yet here is branch based on the latest supported version: https://github.com/walnuthq/miden-vm/tree/pr/debug-type-names-v0.22.3 (without this patch, we would see anonymous structs).

Basically, there we salvage ABI type names from user level into the `.debug_types`, instead of remembering raw felts only.

```bash
$ miden-objtool dump debug-info
...
  [  10] STRUCT miden:base/core-types@1.0.0/account-id (size: 8 bytes)
    +   0: prefix : struct miden:base/core-types@1.0.0/felt
    +   4: suffix : struct miden:base/core-types@1.0.0/felt
```

codegen/masm/src/lower/component.rs
```diff
@@ -671,14 +673,8 @@ impl MasmFunctionBuilder {
...
 
-        let sig = function.signature();
-        let args = sig.params.iter().map(|param| masm::TypeExpr::from(param.ty.clone())).collect();
-        let results = sig
-            .results
-            .iter()
-            .map(|result| masm::TypeExpr::from(result.ty.clone()))
-            .collect();
-        let signature = masm::FunctionType::new(sig.cc, args, results);
+        let signature =
```
bitwalker, line 676: We need to make sure that this signature is the same as the Component ABI signature when the calling convention is `component-model`, or this signature will not be correct.

We at the very least need an integration test with a mixture of types that asserts that the ABI expressed by the "semantic" signature here, and the one actually expressed at the Component ABI level are the same, so that we can catch drift between them.

I'm expecting that in general we'll be relying on WIT for this anyway once I finish up the binding generation work - but I'm fine with the signature we encode in MASM/package metadata to match the Component ABI types, since we must always derive the flattened representation for codegen anyway.

bitwalker (changes requested):
I _think_ this is ok, but we really need integration test coverage here to make sure we don't end up encoding inaccurate signatures that we then later rely on for binding generation, etc.

codegen/masm/src/lower/component.rs
djolertrk: Thanks! It is addressed!

codegen/masm/src/lower/component.rs
```diff
@@ -816,6 +812,115 @@ impl MasmFunctionBuilder {
...
+                    ty: masm_component_type_expr_from_hir(&field.ty),
+                }
+            });
+            masm::TypeExpr::Struct(
+                masm::StructType::new(name, fields)
+                    .with_repr(Span::unknown(struct_ty.repr()))
+                    .with_span(SourceSpan::UNKNOWN),
+            )
+        }
+        Type::Ptr(ptr) => masm::TypeExpr::Ptr(masm::PointerType::new(
```
greenhat: `PointerType::new` defaults to the element address space. Should we preserve `ptr`'s address space?
Also, we have a similar (to `masm_component_type_expr_from_hir`) conversion already defined in `miden-assembly-syntax`:
```rust

impl From<Type> for TypeExpr {
    fn from(ty: Type) -> Self {
        match ty {
            Type::Array(t) => Self::Array(ArrayType::new(t.element_type().clone().into(), t.len())),
            Type::Struct(t) => Self::Struct(StructType::new(
                None,
                t.fields().iter().enumerate().map(|(i, ft)| {
                    let name = Ident::new(format!("field{i}")).unwrap();
                    StructField {
                        span: SourceSpan::UNKNOWN,
                        name,
                        ty: ft.ty.clone().into(),
                    }
                }),
            )),
            Type::Ptr(t) => Self::Ptr((*t).clone().into()),
            Type::Function(_) => {
                Self::Ptr(PointerType::new(TypeExpr::Primitive(Span::unknown(Type::Felt))))
            },
            Type::List(t) => Self::Ptr(
                PointerType::new((*t).clone().into()).with_address_space(AddressSpace::Byte),
            ),
            Type::I128 | Type::U128 => Self::Array(ArrayType::new(Type::U32.into(), 4)),
            Type::I64 | Type::U64 => Self::Array(ArrayType::new(Type::U32.into(), 2)),
            Type::Unknown | Type::Never | Type::F64 => panic!("unrepresentable type value: {ty}"),
            ty => Self::Primitive(Span::unknown(ty)),
        }
    }
}
```

codegen/masm/src/artifact.rs
```diff
@@ -240,16 +243,18 @@ impl MasmComponent {
...
         };
-        let start = masm::Procedure::new(
+        let mut start = masm::Procedure::new(
             span,
             masm::Visibility::Public,
             masm::ProcedureName::main(),
             0,
             body,
         );
+        start.extend_invoked(invoked);
```
greenhat, line 259: The component export wrappers inject `init` and `truncate_stack` with raw `Exec` ops. We should track them as well.
https://github.com/0xMiden/compiler/blob/3c03b01fe7f92c870fff227238a71efa339fa700/codegen/masm/src/lower/component.rs?plain=1#L764
https://github.com/0xMiden/compiler/blob/3c03b01fe7f92c870fff227238a71efa339fa700/codegen/masm/src/lower/component.rs?plain=1#L744

greenhat (changes requested):
Looking good overall! Please see my comments.

codegen/masm/src/lower/component.rs
djolertrk: Yes, nice catch! Thanks!!

codegen/masm/src/lower/component.rs
```diff
@@ -816,6 +814,117 @@ impl MasmFunctionBuilder {
...
+    let Type::Function(ty) = subprogram.ty.as_ref()? else {
+        return None;
+    };
+
+    let args = ty.params().iter().map(masm_component_type_expr_from_hir).collect();
+    let results = ty.results().iter().map(masm_component_type_expr_from_hir).collect();
+    Some(masm::FunctionType::new(ty.calling_convention(), args, results))
+}
+
+fn masm_component_type_expr_from_hir(ty: &Type) -> masm::TypeExpr {
```
greenhat: I checked again and I did not find any difference between this conversion and the one I posted above from the `miden-assembly-syntax` crate. Is there any reason for not using the one from `miden-assembly-syntax`?

codegen/masm/src/lower/component.rs
```diff
@@ -816,6 +814,117 @@ impl MasmFunctionBuilder {
...
+        Type::List(element_ty) => masm::TypeExpr::Ptr(
+            masm::PointerType::new(masm_component_type_expr_from_hir(element_ty))
+                .with_address_space(masm::types::AddressSpace::Byte),
+        ),
+        Type::Unknown | Type::Never | Type::F64 => panic!("unrepresentable type value: {ty}"),
+        ty => masm::TypeExpr::Primitive(Span::unknown(ty.clone())),
+    }
+}
+
+fn masm_type_expr_from_hir(ty: &Type) -> masm::TypeExpr {
```
greenhat: This conversion differs from the `masm_component_type_expr_from_hir` only by handling `I64` and `I128` types.

greenhat (changes requested):
Please see my note regarding the conversions.

codegen/masm/src/lower/component.rs
djolertrk: the thing is that there is a difference, see, like

`TypeExpr::from(Type::U64)` becomes `[u32; 2]`,

and the function I am adding (now renamed) `component_abi_type_expr_from_hir(Type::U64)` stays `u64`

codegen/masm/src/lower/component.rs
djolertrk: yeah, it is a duplicate, leftover -- I locally handled it

greenhat (approved):
Thank you!

djolertrk:
can we merge this one? :)

codegen/masm/src/lower/component.rs
bitwalker: That seems like a bug in `TypeExpr::from` - we should be preserving it as a primitive type in the conversion. Can you open a PR against `miden-vm` for that and add a comment to this conversion helper like:

```rust
// TODO(pauls): Remove once miden-vm#XXXX is merged and ships in the next stable release, expected to be v0.24
```

tests/integration/src/sdk/mod.rs
```diff
@@ -44,6 +46,135 @@ fn assert_manifest_exports_match_library(package: &miden_mast_package::Package)
...
+            .filter_map(|export| match export {
+                PackageExport::Procedure(export) => Some(export.path.as_ref().as_str().to_string()),
+                PackageExport::Constant(_) | PackageExport::Type(_) => None,
+            })
+            .collect::<Vec<_>>(),
+    );
+    matches[0]
+}
+
+fn type_brief(ty: &Type) -> String {
```
bitwalker: We should probably just use the existing `Display` impl for `Type` here

bitwalker (commented):
I think this is probably ready, just a couple of small comments before merging

codegen/masm/src/lower/component.rs
djolertrk: Sure. I will add this TODO, and fix in VM.

djolertrk:
is this ready now?

codegen/masm/src/lower/component.rs
djolertrk: The fix in VM is: https://github.com/0xMiden/miden-vm/pull/3253

bitwalker (approved):
Just needs a rebase and this can be merged

bitwalker:
Looks like the `rust_sdk_cross_ctx_account_and_note_word` test is failing after the rebase

PR #1159, opened 2026-06-04
[2/4] `#[account(...)]` macro generates the native account wrapper (instead of hardcoded `Account` type)

### LINKED ISSUE #1157: `#[account(...)]` macro should generate API for note's active account (instead of `Account` type)
As discussed in https://github.com/0xMiden/compiler/issues/697#issuecomment-4541740522 we adopted the 2.1 option `#[account(...)]` macro for both current and foreign accounts. 

Currently, we generate an `Account` type and pass it as a value in the note script entrypoint. With the `#[account(...)]` macro, the note script code should look like this:

```rust
use crate::bindings::*;
use miden::AccountId;

#[account(package_x, package_y)]
pub struct MyAccount;

#[account(package_z)]
pub struct Oracle;

#[note]
struct MyNote;

#[note]
impl MyNote {
    #[note_script]
    pub fn run(self, _arg: Word, account: &mut MyAccount) {
        let oracle = Oracle::new(AccountId::new(..));
        if oracle.package_z_component_method() {
             account.package_x_component_method();
        }
        ..snip..
    }
}
```

Where `MyAccount` should have the API that we're currently generating for the `Account` type.

Since the `#[account(...)]` macro generates both active and foreign account APIs, its implementation should look like this:

```rust
pub struct MyAccount {
    foreign_account_id: Option<AccountId>
}

impl MyAccount {
    fn some_account_method(&self, args: ...) -> ... {
        if let Some(foreign_account_id) = self.foreign_account_id {
            execute_foreign_procedure(foreign_account_id, proc_root, args)
        } else {
            package_[x|y]::some_account_method(args)
        }
    }

    ..snip..
}

### PR DESCRIPTION
**Pull request by greenhat**
Close #1157 

**This PR is stacked on #1113 and should be merged after it.**

Unifies the `#[account(...)]` wrapper so a single user-declared type represents both the transaction's native (active) account and a foreign (FPI) account, replacing the auto-generated `crate::bindings::Account` struct.

- `#[account(...)]` wrapper methods now dispatch at runtime: a wrapper created with `new(account_id)` (renamed from `from_account`) routes calls through `execute_foreign_procedure`, while the wrapper passed to a `#[note]`/`#[tx_script]` entrypoint invokes the active account's native imports directly.
- **BREAKING:** the auto-generated `bindings::Account` struct is removed; entrypoint account parameters must reference an `#[account(...)]` type. This is enforced via the `AccountWrapper` marker trait rather than by type name, so a wrapper may itself be named `Account`, and non-wrapper types fail with a guiding trait-bound diagnostic.
- Dependency methods that would shadow `ActiveAccount` built-ins (e.g. `get_id`) are rejected at macro expansion to prevent silent method-resolution changes.

### CODE REVIEW

**Comment by greenhat** [comment]
@bitwalker Rebased and ready.

File: sdk/base-macros/src/fpi.rs [hunk 1]
```diff
@@ -36,7 +36,32 @@ pub(crate) const WIT_FUNCTION_PREFIX: &str = "fpi-";
 /// Rust identifier prefix generated by wit-bindgen for FPI WIT functions.
 pub(crate) const RUST_FUNCTION_PREFIX: &str = "fpi_";
 
-const FROM_ACCOUNT_METHOD: &str = "from_account";
+const NEW_METHOD: &str = "new";
+
+/// Method names of the `ActiveAccount` trait (`sdk/base-sys/src/bindings/active_account.rs`).
+///
+/// Dependency functions mapping to these names are rejected: the generated inherent method would
```
bitwalker, line 43: Is this actually true? I believe Rust requires disambiguating with turbofish syntax in such cases (since it cannot know which trait method you are actually intending to call when both are in scope).

You do mention inherent method though, but IIRC the plan was to have component methods defined in the trait implementation, so it's a bit unclear to me why we're generating inherent methods at all vs trait implementations. Account methods like `get_id` I'd expect to be provided by implementing some kind of `Account` trait, rather than as inherent methods of the account struct itself. That would also avoid the issue with method resolution.

None of that prevents someone from adding something like `get_id` or whatever to their account/component struct, but if all of our stuff is trait-implemented, then they'd still be required to disambiguate with turbofish syntax when the conflicting trait is in scope (which would basically always be the case I think).

sdk/base-macros/src/fpi.rs
```diff
@@ -490,50 +541,94 @@ fn build_wrapper_method(
...
+/// A `None` id selects the transaction's native (active) account; a `Some` id selects a foreign
+/// account reached through FPI.
 fn foreign_account_struct(account_struct: &ItemStruct) -> syn::Result<ItemStruct> {
     let attrs = &account_struct.attrs;
     let vis = &account_struct.vis;
     let ident = &account_struct.ident;
+    // User attributes are re-emitted as-is, so drop macro-added derives the user already
+    // requested (matched by the trait's final path segment) to avoid duplicate impls.
+    let user_derives = user_derived_names(attrs);
+    let added_derives = ["Clone", "Copy", "Debug", "Default"]
```
bitwalker: Maybe we should just never emit any of these, and just leave it up to the user - that simplifies things here, and only requires adding one line to account component declarations that need any of these trait impls (which we're seemingly anticipating that they'd be adding anyway).

sdk/base-sys/src/bindings/active_account.rs
```diff
@@ -273,60 +273,82 @@ pub fn has_procedure(proc_root: Word) -> bool {
...
 pub trait ActiveAccount {
+    /// Guard hook invoked by every active-account operation before it runs.
+    ///
+    /// The default implementation is a no-op. Types that can also represent a *foreign* account
+    /// (for example the struct generated by the `#[account(...)]` macro) override this to reject
+    /// calls made on a foreign binding, because the active-account operations always target the
+    /// transaction's native account rather than the foreign one.
+    #[doc(hidden)]
+    #[inline]
+    fn __assert_active_account(&self) {}
```
bitwalker, line 288: I would make this method required, rather than providing a default no-op implementation, since presumably we'll have to emit `impl ActiveAccount` for all `#[account(...)]`  structs anyway.

sdk/base-sys/src/bindings/active_account.rs
```diff
@@ -338,48 +360,74 @@ pub trait ActiveAccount {
...
+///
+/// The `#[note]` and `#[tx_script]` macros instantiate their entrypoint account parameter
+/// through this trait, so that parameter must be a type implementing it. The `#[account(...)]`
+/// macro implements it automatically; it is not a sealed capability boundary, so manual
+/// implementations are possible but normally unnecessary.
+#[diagnostic::on_unimplemented(
+    message = "`{Self}` is not an account wrapper generated by `#[account(...)]`",
+    note = "define a struct with `#[account(...)]` and use it as the entrypoint account parameter"
+)]
+pub trait AccountWrapper: Default {
```
bitwalker: Doesn't this also need `ActiveAccount` as a supertrait as well? Otherwise you could define an impl of `AccountWrapper` for something like `i32`, which we probably don't want to treat as valid.

sdk/base-sys/src/bindings/active_account.rs
```diff
@@ -338,48 +360,74 @@ pub trait ActiveAccount {
...
+    /// Creates a binding to the transaction's native (active) account.
+    #[inline(always)]
+    fn native() -> Self {
```
bitwalker: I'm a bit confused about the mixture of native vs active terminology - the latter makes sense to me, the former does not (an account being "native" really conveys nothing meaningful to me at all - "active"/"current" is way more intuitive IMO).

bitwalker (changes requested):
Looks good! I have a few notes/questions, but nothing major I think

sdk/base-macros/src/fpi.rs
greenhat: > You do mention inherent method though, but IIRC the plan was to have component methods defined in the trait implementation, so it's a bit unclear to me why we're generating inherent methods at all vs trait implementations. Account methods like `get_id` I'd expect to be provided by implementing some kind of `Account` trait, rather than as inherent methods of the account struct itself. That would also avoid the issue with method resolution.

The inherent method is defined to route inter-component vs FPI calls on the `struct` under the `#[account(...)]` macro on the callee side. It depends on the optional account id (for FPI) field defined in this `struct`. I don't think we planned to implement it as a trait. Here is an example of such a wrapper for the `BasicWallet::receive_asset`:
```rust
    pub fn receive_asset(&self, asset: ::miden::Asset) -> () {
        match self.foreign_account_id {
            ::core::option::Option::Some(__miden_foreign_account_id) => {
                __miden_foreign_account_wallet::miden::basic_wallet::miden_basic_wallet::fpi_receive_asset(
                    __miden_foreign_account_id.prefix,
                    __miden_foreign_account_id.suffix,
                    ::miden::Word::new([
                        {
                            const VALUE: u64 = 14498277730569837599u64 as u64;
                            ::miden_stdlib_sys::Felt::new(VALUE).unwrap()
                        },
                        {
                            const VALUE: u64 = 11350922803923988613u64 as u64;
                            ::miden_stdlib_sys::Felt::new(VALUE).unwrap()
                        },
                        {
                            const VALUE: u64 = 3366787023335181781u64 as u64;
                            ::miden_stdlib_sys::Felt::new(VALUE).unwrap()
                        },
                        {
                            const VALUE: u64 = 15705799660361837834u64 as u64;
                            ::miden_stdlib_sys::Felt::new(VALUE).unwrap()
                        },
                    ]),
                    asset,
                )
            }
            ::core::option::Option::None => {
                __miden_foreign_account_wallet::miden::basic_wallet::miden_basic_wallet::receive_asset(
                    asset,
                )
            }
        }
    }
```

sdk/base-macros/src/fpi.rs
greenhat: We need `Default` for `AccountWrapper::native()` to work since `AccountWrapper: Default` and `native()` is just `Self::default()`. I removed the rest in ac7af62a9b67c7be74303259b4f680eb0b784f0b

sdk/base-sys/src/bindings/active_account.rs
greenhat: There's exactly one macro producing dual native/foreign types (#[account]), and it always emits the override (`panic!` if the foreign account is set). If we make it required, it'd move the no-op implementation into boilerplate, and we'd have to `impl ActiveAccount for T { fn __assert_active_account(&self) {} }` on each `impl`.

sdk/base-sys/src/bindings/active_account.rs
greenhat: Good catch! Thank you! Done in 5d297a28e99de8f8b964ae260cd53936146323ad

sdk/base-sys/src/bindings/active_account.rs
greenhat: Renamed to `active()` in f905a16f9c9760c6c72c7bb4f4c5fddbd93c7791

greenhat:
Thank you! I addressed all the review notes. Please do another round.

sdk/base-macros/src/fpi.rs
bitwalker: Could you dump the full expanded code for that crate? I think that'll clarify things a bit for me. 

My intuition would be that `BasicWallet` would be a component trait, implemented for a struct annotated with `#[account]`. The only difference AFAIK would be in whether the implementation is generated or hand-written: In cases where you are declaring an `#[account]` that implements `BasicWallet` (where `BasicWallet` is defined in a dependency package), a `BasicWallet` trait is generated, and an implementation for the `#[account]` structure is generated, which simply wraps calls to the "real" `wit-bindgen`-generated `Guest` trait methods. In cases where the component is hand-written, then the trait and implementation are hand-written (and decorated with `#[component]`).

sdk/base-sys/src/bindings/active_account.rs
bitwalker: Are there places where the default impl actually gets used? AIUI we're always emitting an explicit impl

bitwalker (commented):
Sorry for the turn-around time on this review @greenhat - just want to make sure I understand how all the pieces fit together as we merge them. Don't worry about the release - we'll get to it when we get to it.

I think we're just about ready to merge this, just have a few more follow-up questions.

sdk/base-macros/src/fpi.rs
greenhat: > Could you dump the full expanded code for that crate? I think that'll clarify things a bit for me.

Sure. Here is the [expanded p2id note](https://gist.github.com/greenhat/8c08d8bc2a685ee1c841081462d196b8) for the following [user code](https://github.com/0xMiden/compiler/blob/c800022cf38657b243667468ee3abae9521bf2c6/examples/p2id-note/src/lib.rs?plain=1)

 
> My intuition would be that `BasicWallet` would be a component trait, implemented for a struct annotated with `#[account]`. The only difference AFAIK would be in whether the implementation is generated or hand-written: In cases where you are declaring an `#[account]` that implements `BasicWallet` (where `BasicWallet` is defined in a dependency package), a `BasicWallet` trait is generated, and an implementation for the `#[account]` structure is generated, which simply wraps calls to the "real" `wit-bindgen`-generated `Guest` trait methods. In cases where the component is hand-written, then the trait and implementation are hand-written (and decorated with `#[component]`).

Just to be sure, we're talking about the callee side (e.g. p2id note). I don't understand why `BasicWallet` would be a component trait at all. Every method of `BasicWallet` struct routes either to `fpi_`-prefixed wrapper for `execute_foreign_procedure` (if the account id is set) or to the free function generated by `wit-bindgen`.

sdk/base-sys/src/bindings/active_account.rs
greenhat: It is a guard hook and it is invoked by every active-account method before it calls the external tx kernel function).

Types that can also represent a *foreign* account (for example the struct generated by the `#[account(...)]` macro) override this to reject calls made on a foreign binding, because the active-account operations always target the transaction's active account rather than the foreign one.

sdk/base-macros/src/fpi.rs
bitwalker: > I don't understand why `BasicWallet` would be a component trait at all

Specifically to deal with the ambiguity that arises if multiple components are in scope for a given account struct that have overlapping/conflicting method names. They can't all be implemented as inherent methods of the account struct - the only way to support such overlap/conflict is via traits, so that you can disambiguate with turbofish syntax by specifying which trait/method you actually want.

For example, in the example you've provided, if `Wallet` had another `foo` component implemented for it, in addition to `basic_wallet`, and both components defined a `receive_asset` method - you'd end up with a compilation error with no way to resolve it, because we're emitting the component methods as inherent methods of the `Wallet` struct, rather than generating a `BasicWallet` trait, and a `Foo` trait, and implementations of both for `Wallet` (in which case the call to `account.receive_asset(asset)` in the body of `script` would need to be rewritten as `<Wallet as BasicWallet>::receive_asset(account, asset)` to disambiguate from `Foo::receive_asset`).

This obviously doesn't come up in the p2id example currently, because the `Wallet` struct only derives a single component (the `BasicWallet` component) - but we need to have a solution for multi-component scenarios, or we'll have no way for users to deal with conflicting methods between components - generating traits and trait impls solves this neatly (and in a way that is largely invisible to the user, except in cases where they need to disambiguate).

sdk/base-sys/src/bindings/active_account.rs
bitwalker: The purpose of the hook is clear - I think the part that is unclear to me is in what scenarios we have a struct that _can't_ represent a foreign account. I was under the impression we basically always emit a struct that has an optional account id field, but if that's not the case, that would explain the confusion on my part.

greenhat: The account component struct. We implement the `ActiveAccount` trait for it with a one-liner that uses the default guard implementation.

sdk/base-macros/src/fpi.rs
greenhat: Ah, I see. Yeah, it makes sense. Thank you for a thorough explanation. 
Since the implementation would have a lot of common code with `#[component(...)]`, which is implemented in the next PR (#1171), I would prefer to implement this in a new PR. Preferably, after we merge this (2/4, 3/4, 4/4) PR train.

greenhat: I made #1208 where I put a code example and how it'd be expanded. I'd appreciate it if you took a look.

bitwalker: I'll sign off on this then and we can deal with it in a new PR on top of all the others

bitwalker (approved):

PR #1160, opened 2026-06-04
test: add tests for `i32` arithmetic intrinsics

mooori:
Sets up testing of intrinsics by building an executable from MASM that calls the intrinsic to be tested. This is possible now that the toolchain can handle MASM. We then assert that the return values of the intrinsic correspond to the expected result computed natively via Rust.

#### Motivation

- There is no direct and stable mapping from Rust to MASM intrinsics. For example we proptest Rust's `i32::overflowing_mul` but the generated MASM does not call the `i32::overflowing_mul` intrinsic. So #1161 was discovered only after directly testing `proc i32::overflowing_mul` here.
- When implementing new intrinsics it can be helpful to test them directly in a simple way.

#### Implementation details

- One simple test helper is sufficient for all `i32.masm` intrinsics because
  - All Miden VM inputs/outputs are `Felts`
  - All Rust inputs are `i32` and all outputs are either `i32` or can be easily mapped to `i32` (e.g. `bool` becomes either `0i32` or `1i32`)
- `NumericStrategy` has been moved such that it can be used in the existing tests in `arithmetic.rs` and in the new tests in `intrinsics/i32_arithmetic.rs`.
  - Existing code was moved without modifications and then a some new `Strategy` values and `NumericStrategy` methods were added.
- Tests cases in which the VM is expected to succeed asserting the expected outcome
- Testws cases in which the VM is expected to trap asserting the trap occurs with the expected reason.

#### Issues discovered by these tests

#1161, #1162, #1163


greenhat (approved):
Looking good!

PR #1167, opened 2026-06-09
fix: edge case handling in intrinsic `i32::overflowing_mul`

### LINKED ISSUE #1161: Native/MASM divergence for intrinsic `i32::overflowing_mul`
Reproduction on top of #1160:

```sh
cargo nextest run -p midenc-integration-tests i32_overflowing_mul --run-ignored all

thread 'end_to_end::intrinsics::i32_arithmetic::i32_overflowing_mul' (16108) panicked at tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs:95:13:
Found minimal failing case: (-2147483648, 0)
assertion failed: `(left == right)`
  left: `[1, 0]`,
 right: `[0, 0]` at tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs:89
```

There's seems to be a logic error in the `either operand are MIN` branch.

mooori:
Stacked on top of #1160 to have intrinsic testing available.

### `i32::MIN * 0` does not overflow

Closes #1161

Previously: Overflow was reported for `i32::MIN * 0`, though the result is 0.
Now: Don't report overflow

While working on this I found another issue:

### `i32::MIN * odd_number` equals `(true, i32::MIN)`

Previously: This would return `(true, 0)` because of an incorrect assumption:

https://github.com/0xMiden/compiler/blob/30544eb610c8315550aafbf07944b225f5399c76/codegen/masm/intrinsics/i32.masm#L145

Now: Returns `(true, i32::MIN)`, for example 

```rust
i32::MIN.overflowing_mul(3) // == i32::MIN
```

### DIFF
--- a/tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
+++ b/tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
@@ -484,7 +484,6 @@ fn i32_checked_sub() {
 }
 
 #[test]
-#[ignore = "https://github.com/0xMiden/compiler/issues/1161"]
 fn i32_checked_mul() {
     let proc_body = r#"
     # Stack: [b, a]
--- a/tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
+++ b/tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
@@ -211,7 +211,6 @@ fn i32_overflowing_sub() {
 }
 
 #[test]
-#[ignore = "https://github.com/0xMiden/compiler/issues/1161"]
 fn i32_overflowing_mul() {
     let proc_body = r#"
     # Stack: [b, a]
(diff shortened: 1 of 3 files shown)


tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
```diff
@@ -211,7 +211,6 @@ fn i32_overflowing_sub() {
 }
 
 #[test]
-#[ignore = "https://github.com/0xMiden/compiler/issues/1161"]
```
greenhat, line 214: Given that `checked_mul` is just `overflowing_mul` followed by `assertz`, the `i32_checked_mul` test should also work now.

greenhat (approved):
Looking good!

tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
mooori: Oh yeah, forgot to remove the `ignore`. Done now.

PR #1168, opened 2026-06-09
fix: trap on overflow in `i32::checked_div`

### LINKED ISSUE #1162: Incorrect edge case handling in intrinsic `i32::checked_div`
`proc i32::checked_div` is supposed to trap on overflow, however it does not trap on inputs `i32::MIN, -1`.

### Reproduction

Requires #1160 

```sh
cargo test -p midenc-integration-tests i32_checked_div_min_by_neg1 -- --nocapture
```

mooori:
Closes #1162 

Stacked on top of #1160 to have access to intrinsic testing.

---

The intrinsic `i32::checked_div` should trap on `i32::MIN/-1` because the result overflows `i32`. It did not trap because it didn't account for this edge case.

### DIFF
--- a/tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
+++ b/tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
@@ -519,11 +519,11 @@ fn i32_checked_div() {
         NumericStrategy::<i32>::div_signed_checked(),
         binary_i32op_inputs_to_stack,
         |(a, b): &(i32, i32)| {
-            // TODO(#1162) verify MIN/-1 traps VM
             if *b == 0 {
                 Err(TrapExpectation::DivideByZero)
+            } else if *a == i32::MIN && *b == -1 {
+                Err(TrapExpectation::FailedAssertionOverflow)
             } else {
-                // TODO(#1162) use checked_div
                 Ok(vec![a.wrapping_div(*b)])
             }
         },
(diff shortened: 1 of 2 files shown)


tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
```diff
@@ -550,43 +550,3 @@ fn i32_checked_shr() {
         },
     );
 }
-
-// TODO(#1162): remove once #1162 is fixed and handle MIN/-1 in `i32_checked_div` above
-/// This test reproduces #1162
-#[test]
-fn i32_checked_div_min_by_neg1_i1162_reproducer() {
```
mooori, line 557: Can be removed as the test `i32_checked_div` now asserts that `(i32::MIN, -1)` causes a `FailedAssertionOverflow` trap.

greenhat: It is one weighted branch in a randomized `prop_oneof!`, so a proptest run can pass without executing the exact `i32::MIN / -1` case.

greenhat (changes requested):
Looking good! Please see my comment.

tests/integration/src/end_to_end/intrinsics/i32_arithmetic.rs
mooori: I assumed weights are chosen such that the probability of each case being executed is "high enough". Though your comment made me take another look as this is relevant for all tests that use `NumericStrategy`. I've opened #1169 to address this. WDYT?

bitwalker (approved):

greenhat (approved):

PR #1169, opened 2026-06-10
test: ensure arith proptests execute edge cases

mooori:
### Problem

Arithmetic proptests using `NumericStrategy`s currently might not run on all edge cases, as cases are drawn randomly.

### Ideal solution

1. Deterministically run edge cases exactly once
2. Use proptest to generate random cases

AFAIK proptest does not support having a Strategy for which some cases are executed exactly once. So 1. would require running the edge cases in a loop and then using a proptest runner for 2. That would be more verbose / require more complicated test setup.

### Solution proposed in this PR

1. Run 512 instead of 256 cases for `NumericStrategy`s. The cost of that should be bearable, for example on my machine I'm getting the following times for tests in `i32_arithmetic.rs`:

```
256 cases: Summary [   0.244s] 18 tests run: 18 passed, 436 skipped
512 cases: Summary [   0.490s] 18 tests run: 18 passed, 436 skipped
```

2. Lower the weight of random cases to *basically* get `all edge cases and some random cases`.


greenhat (approved):
Looking good!

Issue #1166, opened 2026-06-09
Account component sibling calls


greenhat:
See the 1.3 trait-based design at https://github.com/0xMiden/compiler/issues/697#issuecomment-4541740522

PR #1171, opened 2026-06-11
[3/4] Redesign `#[component]` around a trait API and a storage struct

greenhat:
Ref #1166 

This PR implements the first part of the #1166 - `#[component]` redesign around traits. The sibling component traits and calls will be implemented in the next PR(s).

**This PR is stacked on #1159 and should be merged after it.**

An account component is no longer a struct with an inherent impl; it is a `#[component_storage]`
struct (storage fields + metadata + runtime boilerplate), a `#[component]` trait (the API — its
kebab-cased name becomes the WIT interface name), and a `#[component] impl Trait for Storage`
(the behavior). The macro architecture, the component's identity model, and one MASM-lowering
detail change with it; everything else (examples, fixtures, templates, ~50 binding tests) is
migration churn.

## The core architectural decision: two stateless expansions

The redesign's central problem is that proc macros cannot see the trait and the impl in one
invocation, so the two expansions must agree on the WIT interface without talking to each other.
Two designs were on the table:

- **A process-global registry** (trait expansion records the API; impl expansion looks it up) —
  the same pattern as the existing `#[export_type]` registry. Rejected: it adds a hard
  trait-before-impl expansion-order dependency (module order can violate it for valid Rust),
  collides on bare trait names across modules, serves stale data in rust-analyzer's persistent
  proc-macro server, and needs ~150 lines of string-round-tripped mirror types.
- **Stateless re-derivation** (what this PR does): *all* generation — WIT, `miden::generate!`,
  guest bindings, `export!` — happens at the **impl** expansion, derived from the impl block's
  own signatures. rustc's trait checking is the consistency mechanism: if the impl doesn't match
  the trait, the user gets a normal Rust error. The trait expansion is a pure validator plus one
  artifact (see "auth" below).

Consequences worth knowing as a reviewer:

- The trait may live in any module (e.g. `mod api`) since `self::bindings` is created where it's
  used. There is no coordination state anywhere.
- Forgotten annotations cannot fail silently. Two language-level "handshakes" exist: the trait
  expansion injects a hidden associated const and the storage expansion a hidden inherent const;
  the impl expansion references both, so a missing `#[component]`/`#[component_storage]` becomes
  a missing-item error naming the constant (documented in the `#[component]` rustdoc). These
  travel with the item through imports/aliases, which a registry key wouldn't.
- Validation is deliberately *duplicated* where stateless-ness requires it: the namespace check
  runs at the trait site (best span) **and** the impl site, because the impl may spell the trait
  through an alias and the WIT is named after the impl-side spelling — without the impl-side
  check, `use api::Foo as Alias; impl Alias for ...` would silently generate interface `alias`
  and resurrect the undefined-`::init` linker failure.
- One deliberate asymmetry: `#[auth_script]` lives on the **trait** method, and the auth frontend
  link section is emitted by the trait expansion. Rationale (recorded in code): auth-ness is
  contract, and given the stateless design, trait placement *forces* trait-site emission — the
  governing rule is "API-derived artifacts at the API site; behavior-derived artifacts (WIT,
  bindings, exports) at the impl site". Misplacement on impl methods is a hard error (the outer
  `#[component]` sees raw attribute tokens before the standalone macro runs, so it must reject
  rather than strip).

## The identity model and its trade-offs

`[lib].namespace` in `miden-project.toml` stays the source of truth for the component's library
identity (it is registry-exposed and the project assembler overrides the component root module
path with it). The WIT interface name is inferred from the trait name. Since both must exist,
they can drift — the chosen reconciliation is **hard validation**: `[lib].namespace` must equal
the full `miden:<package>/<interface>@<version>` derived from the manifest and trait (full
equality, not just the interface segment — a wrong package/version would let the MASM identity
drift from the generated WIT paths). The error message prints the exact expected value.
Rejected alternatives: deleting the namespace from the toml (it's a future-registry field) and an
explicit `#[component(name = ...)]` override (identity should be inferred from Rust).

**Storage slot names** now derive from the namespace interface segment instead of the storage
struct's Rust name. Trade-off: renaming the *private* struct can no longer re-key deployed
storage (the old failure mode — slot names feed `StorageSlotId`), but renaming the *trait* (and
updating the namespace per the error's advice) now does. That hazard is explicitly warned about
in the mismatch error itself and in the CHANGELOG. Net: slot identity is tied to the public,
validated, deliberate name instead of an implementation detail. The examples' slot names revert
to their original pre-branch values, which is a good sanity signal.

The validation/diagnostic surface is symmetric across all three expansion sites by construction
(a shared `reject_generics` helper; manifest-missing guards on trait, impl, and fielded storage):
the early review rounds showed "hardening one site and forgetting the siblings" was the recurring
defect, so symmetry is structural now.

## The one codegen change: symbolic `init`

The rodata/globals `init` prologue in component export wrappers used to `exec` a fully-qualified
path derived from the component id — which must match the user-editable `[lib].namespace` to
link, the root cause of the historical cryptic failures. It is now an unqualified
`InvocationTarget::Symbol("init")`, which resolves module-locally. This is sound because the
export wrappers carrying the prologue and the generated `init` procedure are both emitted into
the root component module (verified against actual emitted MASM; the inner lifted functions in
child modules carry no prologue). That co-location invariant is documented at the emission site,
including the escape hatch if wrappers ever move (thread a qualified target). A user-exported
WIT method named `init` collides at definition time with a loud "symbol conflict" error — it
cannot silently shadow the target. Note: every Rust component has globals (shadow stack
pointer), so *every* component e2e test in the suite exercises this path.

## Accepted constraints (intentional, not gaps)

- Custom `#[export_type]` types in component signatures must be nameable from the crate root
  (documented). Module-aware resolution isn't achievable (proc macros can't know their module
  path); a violation currently fails inside generated bindings with poor spans — a spanned probe
  was considered and deferred.
- One component per crate (a second `generate!` collides on `bindings`); cross-crate component
  traits are out of scope (the export-type registry is per-crate anyway).
- `cargo miden new` needs a rust-templates release with trait-named components.
  `test_all_templates` is `#[ignore]`d meanwhile.
- The macro parses `[lib].namespace` with string surgery; moving that into
  `miden-project::Target` (e.g. `interface_name()`) is noted follow-up work, not blocking.

## Review focus suggestions

- `sdk/base-macros/src/component_macro/mod.rs` is the heart (~everything above lives there);
  `storage.rs` for slot naming; `codegen/masm/src/lower/component.rs` for the init change.
- The diagnostics matrix is the highest-value review target: each contract violation (missing
  annotation on either side, alias spelling, generic trait/impl/storage, default bodies, missing
  manifest, wrong namespace package/version/interface, misplaced `#[auth_script]`) has a
  dedicated cargo-check regression test in `tests/integration/src/sdk/macros.rs` asserting the
  specific message — reading those tests is the fastest way to audit the user experience.
- Examples/fixtures/binding-test changes are mechanical migration; the binding-test scaffolding
  itself was de-stringified (syn-parsed fixtures, derived storage names) with zero expect-file
  churn, which confirms the refactor didn't change compiled output.

## Breaking changes

- `#[component]` no longer applies to structs or inherent impl blocks; components are written as
  `#[component_storage]` struct + `#[component]` trait + `#[component]` trait impl.
- The WIT interface name is derived from the component trait name and must match the interface
  segment of `[lib].namespace` in `miden-project.toml` (validated with an actionable error).
- Storage slot names change for components whose storage struct name does not match the
  namespace interface segment (the example slot names revert to their original values, e.g.
  `counter_contract::counter_contract::count_map`).


greenhat:
@bitwalker Rebased and ready. I stumbled upon `Linear` stack scheduling producing non-deterministic results. I made a fix and explained what surfaced it in 1acfaf751a3c78e19d2e9964cb4f9f4b7d7d5550.

bitwalker (commented):
I'm finding the AI-generated PR description here pretty obtuse - it's doing more to obfuscate decisions/rationale than clarify them IMO.

One thing that sticks out to me already - `#[auth_script]` serves to define an auth component, which are different from normal account components (i.e. they have unique constraints, are defined by having a single entrypoint - the auth procedure, the signature of which is also fixed AIUI). I would suggest that we use a separate `#[auth_component]` attribute for defining auth components. It would work somewhat similarly to `#[component]`, in that it decorates a trait - but it can be a lot stricter about shape/assumptions (i.e. it can require that the trait define only a single method, and that method must adhere to the signature requirements of auth procedures). It isn't clear to me why we'd use `#[auth_script]` to decorate component trait methods, since that implies that you could decorate more than one trait method with `#[auth_script]`, or that multiple traits could specify `#[auth_script]` methods and be used together, when in fact neither of those are the case. By modeling auth components as distinct entities, we regain some control over how they are used (or can at the very least prevent/catch more incorrect usages).

The `#[component_storage]` attribute feels off to me somehow. I know why we need it for technical reasons, but the naming is definitely awkward at best. The struct serves two purposes:

1. Define the storage schema for a component
2. Act as an anchor for the component trait impl

If I was to look at this as an outsider, this would indicate to me that the struct is more than just storage, it's defining a concrete instance of a component - which largely is about storage, yes, but a component with no storage still requires a struct. I don't have a great answer off the top of my head - but I do wonder if maybe we can make the attribute name better reflect this duality of purpose. It's not a blocking issue - but I do think it presents a pedagogical one.

I still need to finish going through the PR - but wanted to leave some initial comments until I'm done doing that

greenhat:
> I'm finding the AI-generated PR description here pretty obtuse - it's doing more to obfuscate decisions/rationale than clarify them IMO.
> 
> One thing that sticks out to me already - `#[auth_script]` serves to define an auth component, which are different from normal account components (i.e. they have unique constraints, are defined by having a single entrypoint - the auth procedure, the signature of which is also fixed AIUI). I would suggest that we use a separate `#[auth_component]` attribute for defining auth components. It would work somewhat similarly to `#[component]`, in that it decorates a trait - but it can be a lot stricter about shape/assumptions (i.e. it can require that the trait define only a single method, and that method must adhere to the signature requirements of auth procedures). It isn't clear to me why we'd use `#[auth_script]` to decorate component trait methods, since that implies that you could decorate more than one trait method with `#[auth_script]`, or that multiple traits could specify `#[auth_script]` methods and be used together, when in fact neither of those are the case. By modeling auth components as distinct entities, we regain some control over how they are used (or can at the very least prevent/catch more incorrect usages).

I like the controlling aspect. I made https://github.com/0xMiden/compiler/issues/1201 to discuss it further. 

> 
> The `#[component_storage]` attribute feels off to me somehow. I know why we need it for technical reasons, but the naming is definitely awkward at best. The struct serves two purposes:
> 
> 1. Define the storage schema for a component
> 2. Act as an anchor for the component trait impl
> 
> If I was to look at this as an outsider, this would indicate to me that the struct is more than just storage, it's defining a concrete instance of a component - which largely is about storage, yes, but a component with no storage still requires a struct. I don't have a great answer off the top of my head - but I do wonder if maybe we can make the attribute name better reflect this duality of purpose. It's not a blocking issue - but I do think it presents a pedagogical one.
> 

I'm not super happy with the name, but I cannot come up with a better one. 
As for the "no storage" case, I embraced the idea that an empty struct will highlight that the account storage is "empty". Meaning that everyone has to define it and acknowledge that this account component doesn't store anything.

greenhat:
> I'm finding the AI-generated PR description here pretty obtuse - it's doing more to obfuscate decisions/rationale than clarify them IMO.

Got it. I'll scale it down. I wanted to experiment with a high-level overview of the code changes.

greenhat:
Fixed https://github.com/0xMiden/compiler/issues/1202 in 9a9502c44cbca5cdbaf7c6d79982f4922d9369e1

hir-analysis/src/analyses/dce.rs
```diff
@@ -222,6 +222,7 @@ impl PredecessorState {
 
     pub fn join(&mut self, predecessor: OperationRef) -> ChangeResult {
         if self.known_predecessors.insert(predecessor) {
+            self.known_predecessors.sort_by(stable_operation_cmp);
```
bitwalker, line 225: This can be a very expensive operation, and joins can be computed many times during analysis - it isn't clear to me why this needs a stable sort though, could you clarify?

bitwalker (commented):
Looks good - I just have one question before we merge

hir-analysis/src/analyses/dce.rs
greenhat: I mentioned the issue in https://github.com/0xMiden/compiler/pull/1171#issuecomment-4766828138 above, which manifests as different MASM is generated for the same HIR. My take is that linux's ASLR + release profile affects the observation/insertion order in `PredecessorState::known_predecessors`, which is later consumed as an ordered list.

bitwalker: I guess we'll have to eat the overhead from this for now, but ideally there is a cheaper way to sort the elements of the predecessor set deterministically - it is what it is for now though. This is almost certainly not the only source of non-determinism in the compiler pipeline either, but we can cross that bridge when we get to it.

bitwalker (approved):

PR #1175, opened 2026-06-12
Fix the silently failing lit suite and make printed HIR round-trip through the parser

greenhat:
The lit/filecheck suite has been failing on `next` without anyone noticing: litcheck 0.4.3 always
exits 0 regardless of test results (despite its `--help` saying otherwise), so the CI job stayed
green while `FAIL:` lines scrolled by (example: <https://github.com/0xMiden/compiler/actions/runs/27338221879/job/80768422463>).
Three broken lit tests and two unparseable `XFAIL:` directives accumulated behind that. This PR
fixes the harness so it cannot happen again, fixes everything that had piled up, and pins the
parser behavior with tests so regressions surface as unit failures rather than lit-only ones.

- **Harness guard**: the `test-lit` task now derives its exit code from the litcheck summary and
  fails the build on failed/unresolved tests (workaround until litcheck propagates failures via
  its exit code). To be removed after the `litcheck` is released with https://github.com/bitwalker/litcheck/pull/2 fix.
- **Printed HIR now round-trips through the parser.** The printer and parser had diverged on the
  format in many places; reading back the printed form of any non-trivial function was impossible.
  Fixed at the machinery level: regions of operand-bearing ops parse with the printed
  `^block(...)` headers declaring their entry arguments, trailing variadic operand groups, fixed
  operand-group sizing, multi-name result bindings, labeled block definitions (incl. backward
  branch references), `scf.while` result types, `scf.index_switch` region order, and
  `builtin.ret_imm` immediates.
- **Successor argument lists round-trip too.** Branch targets printed as `^block:(%v)`, a form
  the parser has no grammar for; they now print as `^block(%uses, types)`, the form it accepts.
  This also unblocked `cf.switch`: keyed case parsing, integer attribute parsing, and successor
  group placement (parsed successors previously all landed in one flat group, breaking
  `cases()`/`fallback()` accessors).
- **Tests**: the `structured_control_flow.hir` lit test is rewritten as a canonical print/parse
  fixpoint (it previously had no CHECK lines at all), and new unit tests in `hir`, `scf`, and
  `cf` pin each parser behavior via a `testing::parse_function_fixpoint` helper.
- **Smaller fixes**: frontmatter Rust builds honor an inherited `RUSTUP_TOOLCHAIN` instead of
  forcing the ambient `nightly`; the two debugdump tests use the `XFAIL: *` spelling litcheck
  understands; the `#[component]` no-arguments error names the attribute, matching its committed
  lit expectation.

**Reviewer note**: the printed format changes in two ways — `scf.while` now prints its result
types (functional signature like `scf.if`), and successor arguments print as `^block(%uses,
types)` — so the `.hir` expect files across `dialects/{scf,hir}`, `frontend/wasm`, and
`hir-analysis` were regenerated; their diffs are mechanical.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -435,27 +435,16 @@ dependencies = ["cargo-miden"]
 [tasks.test-lit]
 category = "Test"
 description = "Runs the lit/filecheck test suite"
+command = "litcheck"
 env = { MIDENC_BIN_DIR = "${MIDENC_BIN_DIR}" }
-script = ['''
-#!/usr/bin/env bash
-set -euo pipefail
-# litcheck (up to 0.4.3) always exits 0, even when tests fail or cannot be run, so a meaningful
-# exit code must be derived from the printed summary; otherwise lit failures never fail CI.
-log_file="$(mktemp)"
-trap 'rm -f "${log_file}"' EXIT
-litcheck lit run --verbose \
-    --path "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/bin" \
-    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/tests/lit" 2>&1 | tee "${log_file}"
-if ! grep -q "Total Discovered Tests" "${log_file}"; then
-    echo "error: litcheck did not produce a test summary" >&2
-    exit 1
-fi
-# Note: "Failed As Expected" (XFAIL) sections intentionally do not match this pattern.
-if grep -qE "^\*+ (Failed|Unresolved) Tests" "${log_file}"; then
-    echo "error: the lit suite reported failed or unresolved tests" >&2
-    exit 1
-fi
-''']
+args = [
+    "lit",
+    "run",
+    "--verbose",
+    "--path",
+    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/bin",
+    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/tests/lit",
+]
 dependencies = ["litcheck", "midenc", "cargo-miden", "hir-opt", "miden-objtool"]
 
 [tasks.lit]
@@ -473,7 +462,8 @@ dependencies = ["litcheck", "midenc", "cargo-miden", "hir-opt", "miden-objtool"]
 [tasks.litcheck]
 category = "Test"
 description = "Set up the litcheck utility"
-install_crate = { crate_name = "litcheck", test_arg = "--help" }
+# 0.4.4 is the first release that propagates lit failures through the process exit code.
+install_crate = { crate_name = "litcheck", min_version = "0.4.4", test_arg = "--help" }
 
 [tasks.docs]
 category = "Documentation"
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -435,16 +435,27 @@ dependencies = ["cargo-miden"]
 [tasks.test-lit]
 category = "Test"
 description = "Runs the lit/filecheck test suite"
-command = "litcheck"
 env = { MIDENC_BIN_DIR = "${MIDENC_BIN_DIR}" }
-args = [
-    "lit",
-    "run",
-    "--verbose",
-    "--path",
-    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/bin",
-    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/tests/lit",
-]
+script = ['''
+#!/usr/bin/env bash
+set -euo pipefail
+# litcheck (up to 0.4.3) always exits 0, even when tests fail or cannot be run, so a meaningful
+# exit code must be derived from the printed summary; otherwise lit failures never fail CI.
+log_file="$(mktemp)"
+trap 'rm -f "${log_file}"' EXIT
+litcheck lit run --verbose \
+    --path "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/bin" \
+    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/tests/lit" 2>&1 | tee "${log_file}"
+if ! grep -q "Total Discovered Tests" "${log_file}"; then
+    echo "error: litcheck did not produce a test summary" >&2
+    exit 1
+fi
+# Note: "Failed As Expected" (XFAIL) sections intentionally do not match this pattern.
+if grep -qE "^\*+ (Failed|Unresolved) Tests" "${log_file}"; then
+    echo "error: the lit suite reported failed or unresolved tests" >&2
+    exit 1
+fi
+''']
 dependencies = ["litcheck", "midenc", "cargo-miden", "hir-opt", "miden-objtool"]
 
 [tasks.lit]
--- a/hir/src/ir/parse/operation.rs
+++ b/hir/src/ir/parse/operation.rs
@@ -733,9 +733,12 @@ where
                 );
             }
 
-            // Add definitions for each of the parsed result names, mapping them onto the
-            // operation's results in order. Names do not correspond 1:1 to result groups: a
-            // single variadic result group may be bound by multiple names (e.g. `scf.if`).
+            // Bind each parsed result name to a result of the operation, in order. The printer
+            // emits one SSA name per result (e.g. `%a, %b, %c = scf.if`) - the canonical form for
+            // an op with a single result group - so the names map directly onto the flattened
+            // result list; the `%name:count` spelling instead binds `count` consecutive results to
+            // `%name#0..#count`. The previous code indexed `results().group(i)` by the i-th *name*,
+            // which panicked once a single variadic group was bound by more than one name.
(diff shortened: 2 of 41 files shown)


hir/src/ir/print/asm_printer.rs
```diff
@@ -201,6 +201,27 @@ impl<'a> AsmPrinter<'a> {
         self.document += doc;
     }
 
+    /// Prints a successor argument list: the operand uses followed by their types, e.g.
+    /// `(%0, %1, u32, u32)`, matching the grammar accepted by `parse_successor_and_use_list`.
```
bitwalker: This is incorrect - the grammar rule for `ssa-use-and-type-list` is `ssa-use-list ':' type-list-no-parens`. So this should be expecting `(%0, %1 : u32, u32)` NOT `(%0, %1, u32, u32)`

hir/src/ir/parse/operation.rs
```diff
@@ -385,6 +385,11 @@ where
         }
 
         self.parse_comma_separated_list(Delimiter::None, Some("SSA use list"), |parser| {
+            // Stop at the first non-value token: a use list may be followed by further
```
bitwalker: This is incorrect - an undelimited comma-separated list can never be followed by other other items. Such grammar would be ambiguous, and we do not support it.

SSA use lists will always either be delimited, or followed by a `:` (when the types of the SSA operands are being specified).

hir/src/ir/parse/operation.rs
```diff
@@ -407,9 +412,10 @@ where
             })?
             .into_parts();
 
-        // If we have an attribute ID, it is a result number.
+        // If we have an all-digit attribute ID, it is a result number; any other hash
```
bitwalker: I'm having trouble understanding where we can ever have an SSA use be followed by an attribute - that isn't supported anywhere in the grammar that I'm aware of.

hir/src/ir/parse/operation.rs
```diff
@@ -528,7 +534,8 @@ where
         }
 
         let mut types = SmallVec::<[Type; 4]>::new_const();
-        self.parser.token_stream_mut().expect(Token::Comma)?;
+        // The comma separating the uses from their types was consumed by the use list loop
```
bitwalker: The bug here was that `Comma` was supposed to be `Colon` - I'm assuming this is the source of the incorrect assumptions around parsing of SSA uses that I've flagged elsewhere

hir/src/ir/parse/operation.rs
```diff
@@ -733,9 +740,12 @@ where
                 );
             }
 
-            // Add definitions for each of the result groups.
-            for (result_group, result_record) in result_ids.iter().enumerate() {
-                let group = op.results().group(result_group);
+            // Add definitions for each of the parsed result names, mapping them onto the
```
bitwalker: Can you provide an example of this - I don't believe this is true. There are two ways results can be printed:

1. For operations with a single result group, you may either name each result individually, or use the second form below for the whole group
2. All result groups are collapsed into %index_or_name#N` form, e.g.` %0#1`, `%pair#2`.

This is what allows the parser to recognize how many results there are, and what groups they belong to, particularly for ops with variadic results.

If we're currently violating this rule somewhere, we should fix that, rather than attempting to allow arbitrary mixture of the two forms.

hir/src/ir/parse/parser.rs
```diff
@@ -1457,18 +1457,6 @@ pub trait ParserExt<'input>: Parser<'input> {
         T: AttributeRegistration,
     {
         let name = self.context().get_registered_attribute_name::<T>();
-        let default_dialect = *self.state().default_dialect_stack.last().unwrap();
```
bitwalker, line 1460: Why was this removed?

hir/src/ir/parse/operation.rs
```diff
@@ -1298,20 +1312,25 @@ impl<'input, P> OperationParser<P>
 where
     P: Parser<'input>,
 {
-    /// Parse a new block into 'block'.
+    /// Parse a new block into 'block', returning the block that was actually defined.
+    ///
+    /// Note that the returned block may differ from `block`: a label that was previously
```
bitwalker: I don't believe the returned block should ever differ from `block`, because the caller is expecting the block they passed to be the one that gets filled by the parser. If a block was explicitly provided, but that block was also forward-declared, then I would panic, so that we can ensure the caller is fixed to properly handle the possibility of forward-declared blocks, or to explicitly pass `None` and just let `parse_block` give them back a block (as you did above on line 1296).

midenc-compile/src/stages/parse/rust.rs
```diff
@@ -243,8 +243,11 @@ trim-paths = [\"diagnostics\", \"object\"]
     let cargo_path = cargo_env.as_deref().unwrap_or_else(|| Path::new("cargo"));
 
     let mut cargo = std::process::Command::new(cargo_path);
-    // Ensure we specify the nightly toolchain if a specific cargo wasn't set
-    if cargo_env.is_none() {
+    // Ensure we specify the nightly toolchain if a specific cargo wasn't set. An inherited
```
bitwalker: I would just call the `rustup_toolchain` helper that is defined in the `midenc_compile::rust` module (and is used for this purpose in the `cargo` stage as well), as it is much more comprehensive.

Makefile.toml
```diff
@@ -435,16 +435,27 @@ dependencies = ["cargo-miden"]
...
 env = { MIDENC_BIN_DIR = "${MIDENC_BIN_DIR}" }
-args = [
-    "lit",
-    "run",
-    "--verbose",
-    "--path",
-    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/bin",
-    "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/tests/lit",
-]
+script = ['''
```
bitwalker: This can be removed by upgrading to `litcheck v0.4.4`

bitwalker (changes requested):
Thanks for tackling this! I have a few questions, and flagged some issues, but overall nice work!

I've published v0.4.4 of `litcheck` with the fix for the exit code issue

hir/src/ir/print/asm_printer.rs
greenhat: Fixed in 91aa0022965716870ee34e7cc8df181d42569994 — `print_successor_arguments` now emits `(%0, %1 : u32, u32)`, and the
doc/grammar reference is corrected.

hir/src/ir/parse/operation.rs
greenhat: Agreed — reverted in 91aa0022965716870ee34e7cc8df181d42569994; `parse_optional_ssa_use_list` is back to its original form.

hir/src/ir/parse/operation.rs
greenhat: Right, nowhere. This only came up because `cf.switch` printed its keyed cases bare
(`%sel #key -> ^bb, ^fallback`), which put the selector use directly next to the first case key. The
cases are now bracket-delimited (`%sel [#key -> ^bb], ^fallback`), so the use is followed by `[`, and
this workaround is reverted in 91aa0022965716870ee34e7cc8df181d42569994.

hir/src/ir/parse/operation.rs
greenhat: Fixed in 91aa0022965716870ee34e7cc8df181d42569994.

hir/src/ir/parse/operation.rs
greenhat: Example: `%63, %64, %65, %66 = scf.while ...`. `scf.while` has a single variadic `#[results]` group, and the printer names each result individually (your form 1.). The original code indexed `results().group(i)` by the i-th **name**, so any name past the first addressed a non-existent group and panicked. Reworded in 832f5064ba82df6fe8b938ff20036fb0ee08f4fa.

hir/src/ir/parse/parser.rs
greenhat: Name and dialect parsing are delegated entirely to `parse_extended_attribute`, which resolves the dialect and errors on an unknown attribute. The `is::<T>()` check below then validates the concrete type. I added a comment with this clarification in 91aa0022965716870ee34e7cc8df181d42569994.

hir/src/ir/parse/operation.rs
greenhat: Done in 832f5064ba82df6fe8b938ff20036fb0ee08f4fa

Makefile.toml
greenhat: Done in f850c4099c2f4caef99556566b8ccacbd8f02aa5.

midenc-compile/src/stages/parse/rust.rs
greenhat: Done in 82c499eb4685ef7d4827f21473c41afdeca7bbf8.

greenhat:
Thank you! I addressed all the notes and rebased the branch. Please do another round.

bitwalker (approved):

PR #1178, opened 2026-06-15
test: add tests for `i64` intrinsics

### LINKED ISSUE #1142: Test compiler intrinsics against natively executed Rust
Compiler intrinsics perform non-trivial logic, in many cases directly mapping to Rust functions. For example there's `proc overflowing_mul` in `i32.masm` which maps to `i32::overflowing_mul` in Rust. In cases without direct mapping, it still might be easy to replicate the intrinsic's behavior in Rust.

For a given intrinsic, the test flow would be:

1. Construct Masm calling that intrinsic, execute it in Miden VM
2. Construct Rust with equivalent behavior, execute it natively
3. Generate inputs, for instance using proptest strategies
4. Assert that 3. and 4. yield the same result

The aim of constructing Masm and calling the intrinsic directly is decoupling intrinsic testing from the compilation pipeline. Otherwise it can be difficult target a specific intrinsic, see #1140.

The setup and tests in `end_to_end/arithmetic.rs` can serve as a reference.

mooori:
- Adds tests for all `i64` intrinsics
- Moves some code such that it can be used in tests for both `i32` and `i64` intrinsics
- Extends some `NumericStrategy`s
- Issues uncovered:
  - #1179
  - #1181 
  - #1185 
  - #1187
 
Tests for `i32` intrinsics have already been merged, so this PR will close #1142


tests/integration/src/end_to_end/intrinsics/i64_arithmetic.rs
```diff
@@ -0,0 +1,511 @@
...
+#[test]
+fn i64_pow2() {
+    let proc_body = r#"
+    # Stack: [n_lo, n_hi]
+    exec.::intrinsics::i64::pow2
+    # Stack: [r_lo, r_hi]
+"#;
+    test_i64_intrinsic(
+        proc_body,
+        NumericStrategy::<i64>::pow2(),
```
greenhat: `NumericStrategy::pow2()` generates only `0..=bit_width - 2`, while the MASM implementations assert `i32: n < 31` and `i64: n_hi == 0 && n < 63`. Consider adding explicit trap cases for `31`/`63` and invalid negative or nonzero-high-limb encodings so the tests cover the checked boundary, not just the happy path.

greenhat (approved):
Looking good! Just a minor note.

PR #1180, opened 2026-06-15
[4/4] Account component sibling calls

### LINKED ISSUE #697: Foreign Procedure Invocation (FPI) support

### LINKED ISSUE #1166: Account component sibling calls

greenhat:
Close #1166 
Close #697 

**This PR is stacked on #1171 and should be merged after it.**

Closes the `#697` series.

## What this adds

An account can be deployed with two or more components. Until now there was no Rust-level way for one component to call another. This PR adds **sibling component calls**: a component reaches the other components of the same account through the Wasm component-model boundary — wit-bindgen lift/lower plus a Miden cross-context `call` — which is the *same* mechanism a note script already uses to call an account.

```rust
#[component_storage]
struct MyComponentStorage { /* ... */ }

// `#[account(pkg::Interface)]` binds a *foreign* account reached via FPI — the
// inter-account counterpart to a sibling call. This is the `#[account]` machinery
// the series builds on; the interface segment is now required (see breaking change).
#[account(counter_contract::CounterContract)]
struct OtherAccount;

// `#[component(pkg::Interface, ...)]` on the trait declares the sibling components.
// Each reference generates a `pub trait` (Pausable, CounterContract) whose default
// methods call the sibling across the component boundary.
#[component(pausable::Pausable, counter_contract::CounterContract)]
trait MyComponent: Pausable + CounterContract {
    fn receive_asset(&mut self, asset: Asset);
    fn move_asset(&mut self, asset: Asset, other_id: AccountId);
}

#[component] // impl-site attribute stays bare
impl MyComponent for MyComponentStorage {
    fn receive_asset(&mut self, asset: Asset) {
        assert!(!self.is_paused());  // intra-account sibling call -> cross-context `call`
        self.increment_count();      // intra-account sibling call
        self.add_asset(asset);       // NativeAccount built-in (unchanged)
    }
    fn move_asset(&mut self, asset: Asset, other_id: AccountId) {
        assert!(!self.is_paused());                          // sibling call (this account)
        let count = OtherAccount::new(other_id).get_count(); // inter-account FPI call
        assert!(count < 100);
        self.remove_asset(asset);                            // NativeAccount built-in
    }
}
```

The same `CounterContract` interface is used two ways: `self.increment_count()` is a sibling call into the counter component deployed on *this* account, while `OtherAccount::new(id).get_count()` is an FPI call into a *different* account that also carries the counter component — the `sibling_and_fpi` test exercises exactly this mix in one crate.

The sibling traits attach to the storage struct automatically, so `self.is_paused()` / `self.increment_count()` type-check with no user-written glue, and they can be named as supertraits to make the dependency part of the component's API. Component traits may now declare supertraits (previously rejected).

## The load-bearing result: zero compiler-pipeline changes

The frontend classifies component-model imports purely by the `fpi-` name prefix (no caller-kind check), and the `#[account]` native-dispatch branch already compiles direct cross-context calls *inside* an account component. So an account component importing a sibling interface takes the exact same `hir.call` → MASM `call` path that note→account already exercises. Nothing in `frontend/wasm`, `codegen/masm`, or the assemble stage needed to change.

The one thing that had never actually been *executed* was a same-account component→component call at runtime — does the tx kernel accept it, and does a storage write made by the sibling-called component commit to the account? The mock-chain tests answer both: the single-sibling test seeds a sibling counter at 314 and asserts 315 after the transaction, proving kernel acceptance and storage-delta commitment.

## How the macro generates the sibling traits

All sibling generation happens at the **trait-site** `#[component]` expansion (`component_macro/sibling.rs`); the impl-site attribute stays bare. Per invocation:

1. **Synthesize a throwaway import-only WIT world.** A `package miden:sibling-bindings@1.0.0; world sibling-bindings { import miden:<pkg>/<iface>@<ver>; ... }` is rendered inline (one import per reference). wit-bindgen generates against a *world*, and the component's own world is export-only, so this scratch world is the vehicle that makes wit-bindgen emit the import stubs. It is never written to disk or recorded in the package — it exists only during expansion. (`#[account]`/FPI uses the identical pattern with a distinct `foreign-account-bindings` world, so both coexist in one crate; see the mixed test.)

2. **Run wit-bindgen and harvest the imports.** The generated free functions are wrapped in a hidden module `__miden_sibling_bindings_<trait_snake>`, then walked to build the trait. The harvesting reuses the FPI binding-walker, parameterized by a filter so the sibling path keeps the plain (non-`fpi_`) imports.

3. **Emit one `pub trait <Interface>` per reference** whose methods are **default methods** forwarding to those hidden-module imports (`&self` receivers, argument-forwarding, unit-return handling).

4. **Attach via an empty blanket impl** `impl<T: ::miden::native_account::NativeAccount> <Interface> for T {}`. `NativeAccount` is implemented exactly by `#[component_storage]` structs, so the storage struct picks the methods up with no glue, while `#[account]` FPI wrappers (which implement `ActiveAccount`/`AccountWrapper`, not `NativeAccount`) are deliberately excluded — only code acting as the transaction's native account can make intra-account sibling calls. The method bodies live in the trait as defaults, so the blanket impl stays empty.

**Resolution is at link time, not macro time.** The harvested imports lower to direct cross-context `call`s that the assembler resolves against the dependency package by interface path. Unlike FPI — which reads the dependency `.masp` during expansion to extract procedure roots — siblings read **no** `.masp` at macro time; only the dependency's generated WIT is needed.

**Owned types cross correctly.** A sibling interface that owns a `record`/`variant` (not just core `felt`/`word`) remaps that type to the caller's `crate::bindings::<module>::<Type>` via the shared `dependency_type_with_entries`. That resolves because the component's *own* impl-side `miden::generate!` imports the sibling (it is a declared dependency) and materializes the type there; the sibling scratch world borrows it through a wit-bindgen `with` redirection. This is the same proven remap FPI already relies on.

## Breaking change: `#[account(pkg::Interface)]`

`#[account(...)]` now requires the dependency's exported WIT interface segment — `#[account(counter_contract::CounterContract)]` instead of the old bare `#[account(counter_contract)]`. The interface is kebab-cased and validated against the interfaces the dependency's generated WIT actually exports; a bare reference gets a migration-message error.

A new `dependency_ref.rs` parses the `package::Interface` reference and is shared by **both** `#[account]` and `#[component]` (the package selects the manifest dependency; the interface names the generated Rust trait for `#[component]`, or selects the WIT interface for `#[account]`). To support selection by name, `wit_world.rs` dependency parsing now collects **all** exported interfaces of a dependency (was first-only) with find-by-name lookup. All existing `#[account(pkg)]` call sites (examples, FPI fixtures, docs) are migrated in this PR.

A sibling dependency must be declared in `[dependencies]` and have its generated WIT reachable through `[package.metadata.miden.dependencies].<name>.wit` in `miden-project.toml` — the same entry FPI dependencies already use. Macro expansion fails with a targeted diagnostic (naming the dependency and the exact TOML entry to add) if that entry is missing, instead of a raw wit-parser "package not found".


greenhat:
@bitwalker Rebased and ready.

bitwalker (approved):
LGTM! I didn't dig too deep into the frontend, but skimming over how the pieces fit together and how it all looks reads like what I was expecting to see, so I'm signing off under the assumption that any minor issues that may be present will be caught as we continue to iterate on things

greenhat:
Rebased and ready.

PR #1186, opened 2026-06-16
fix(wasm): correct memory op and int32 narrowing semantics

### LINKED ISSUE #1182: int32_to_uint checks dup.1 instead of the value being narrowed
## Summary

`codegen/masm/src/emit/int32.rs::int32_to_uint` and `try_int32_to_uint` duplicate stack slot 1 before applying the high-bit mask. The value being narrowed is on stack slot 0, so checked unsigned casts can validate an unrelated live word.

Affected commit checked: `17eb665e52f3c2a519f8aa58c7dff1335ba63887` on `next`.

## Steps to reproduce

Public repro:

https://github.com/Kuhai9801/miden-int32-to-uint-ci/actions/runs/27597467577

The repro models the current emitted sequence:

```text
dup.1
push.<mask>
u32and
assertz
```

and compares it with checking the top value.

## Expected behavior

A checked `u32/i32 -> u8/u16/i1` narrowing should reject when the value being narrowed has any bit set above the target width, and should not depend on the word below it.

## Actual behavior

The check depends on stack slot 1:

- `256 -> u8` can pass when the lower live word is `0`.
- `255 -> u8` can fail when the lower live word has high bits.

## Notes

The fix appears to be using the stack slot containing the value being narrowed, not the next lower live word, in both `int32_to_uint` and `try_int32_to_uint`.

### LINKED ISSUE #1183: Wasm memory.copy overlap lowers to forward-copy semantics
## Summary

Wasm `memory.copy` has memmove-style overlap semantics, but the current lowering routes it to `builder.memcpy`, whose fallback loop copies forward. For overlapping ranges with `dst > src`, the generated behavior can alias its own writes and produce different memory contents than the Wasm source program.

Affected commit checked: `17eb665e52f3c2a519f8aa58c7dff1335ba63887` on `next`.

## Steps to reproduce

Public repro:

https://github.com/Kuhai9801/miden-int32-to-uint-ci/actions/runs/27597467549

The repro compares Wasm overlap behavior against the current forward fallback loop for:

```text
memory = [1, 2, 3, 4, 5, 6]
memory.copy(dst = 1, src = 0, len = 4)
```

## Expected behavior

The result should preserve the original source range:

```text
[1, 1, 2, 3, 4, 6]
```

## Actual behavior

The forward loop re-reads bytes it just wrote:

```text
[1, 1, 1, 1, 1, 6]
```

## Notes

`frontend/wasm/src/code_translator/mod.rs` lowers `Operator::MemoryCopy` directly to `builder.memcpy`. The fallback path should either copy backward when ranges overlap in this direction, use a temporary buffer, or otherwise implement Wasm `memory.copy` semantics.

### LINKED ISSUE #1184: Wasm memory.size/memory.grow use Miden heap state instead of the module memory declaration
## Summary

The Wasm translator maps `memory.size` and `memory.grow` to Miden heap intrinsics. This can make a Wasm module observe heap page state rather than the module's declared linear memory size and max.

Affected commit checked: `17eb665e52f3c2a519f8aa58c7dff1335ba63887` on `next`.

## Steps to reproduce

Public repro:

https://github.com/Kuhai9801/miden-int32-to-uint-ci/actions/runs/27597467547

The repro models a Wasm memory declared with one page and compares that with the current zero-initialized heap-backed size.

## Expected behavior

For a module with one declared memory page, `memory.size` should report `1`. `memory.grow` should respect the Wasm memory maximum.

## Actual behavior

The heap-backed path reports `0` for the same initial state in the repro, and growth is checked against heap capacity rather than the Wasm memory declaration.

## Notes

`frontend/wasm/src/code_translator/mod.rs` maps `Operator::MemorySize` and `Operator::MemoryGrow` to `builder.mem_size` / `builder.mem_grow`. The parsed Wasm memory declaration appears not to be the source of truth for these operations.

1sgtpepper:
Fixes #1182
Fixes #1183
Fixes #1184

## Summary

This fixes the int32 unsigned narrowing stack duplication to check the value being narrowed.

For Wasm memory translation, `memory.size` now returns the module-declared initial memory size instead of the Miden heap size. `memory.grow` and single-memory `memory.copy` now fail explicitly because the current lowering does not preserve Wasm linear-memory growth or overlap-copy semantics.

## Testing

- `git diff --check`
- Targeted source review for the changed int32 narrowing and Wasm memory translation paths

No local compile or test run; verification should run in CI.

### DIFF
--- a/codegen/masm/src/emit/int32.rs
+++ b/codegen/masm/src/emit/int32.rs
@@ -300,7 +300,7 @@ impl OpEmitter<'_> {
         let reserved = 32 - n;
         let mask = (2u32.pow(reserved) - 1) << n;
         // Copy the input
-        self.emit(masm::Instruction::Dup1, span);
+        self.emit(masm::Instruction::Dup0, span);
         // Apply the mask
         self.emit_push(mask, span);
         self.emit(masm::Instruction::U32And, span);
@@ -323,7 +323,7 @@ impl OpEmitter<'_> {
         let reserved = 32 - n;
         let mask = (2u32.pow(reserved) - 1) << n;
         // Copy the input
-        self.emit(masm::Instruction::Dup1, span);
+        self.emit(masm::Instruction::Dup0, span);
         // Apply the mask
         self.emit_push(mask, span);
         self.emit(masm::Instruction::U32And, span);
--- a/codegen/masm/src/emit/int32.rs
+++ b/codegen/masm/src/emit/int32.rs
@@ -1080,31 +1080,73 @@ mod tests {
     use super::*;
     use crate::{OperandStack, masm::Op};
 
-    #[test]
-    fn int32_to_uint_masks_top_stack_word() {
-        let span = SourceSpan::default();
-        let context = Rc::new(Context::default());
-        let mut stack = OperandStack::new(context);
-        let mut invoked = BTreeSet::default();
-        let mut block = Vec::new();
-        let mut emitter = OpEmitter::new(&mut invoked, &mut block, &mut stack);
+    fn push_u32_op(value: u32, span: SourceSpan) -> Op {
+        Op::Inst(masm::Span::new(
+            span,
+            masm::Instruction::Push(masm::Immediate::Value(masm::Span::new(span, value.into()))),
+        ))
+    }
 
-        emitter.int32_to_uint(8, span);
+    fn new_emitter<'a>(
+        invoked: &'a mut BTreeSet<masm::Invoke>,
+        block: &'a mut Vec<Op>,
+        stack: &'a mut OperandStack,
+    ) -> OpEmitter<'a> {
+        OpEmitter::new(invoked, block, stack)
+    }
 
-        assert_eq!(&block[0], &Op::Inst(masm::Span::new(span, masm::Instruction::Dup0)));
+    fn assert_int32_to_uint_stack_contract(block: &[Op], span: SourceSpan, n: u32) {
+        let reserved = 32 - n;
+        let mask = (2u32.pow(reserved) - 1) << n;
+
+        assert_eq!(
+            &block[0],
+            &Op::Inst(masm::Span::new(span, masm::Instruction::Dup0)),
+            "must duplicate the top stack word, not the word underneath"
+        );
+        assert_eq!(&block[1], &push_u32_op(mask, span));
+        assert_eq!(&block[2], &Op::Inst(masm::Span::new(span, masm::Instruction::U32And)));
     }
 
     #[test]
-    fn try_int32_to_uint_masks_top_stack_word() {
+    fn int32_to_uint_masks_top_stack_word_and_preserves_input() {
         let span = SourceSpan::default();
         let context = Rc::new(Context::default());
         let mut stack = OperandStack::new(context);
         let mut invoked = BTreeSet::default();
         let mut block = Vec::new();
-        let mut emitter = OpEmitter::new(&mut invoked, &mut block, &mut stack);
+        let mut emitter = new_emitter(&mut invoked, &mut block, &mut stack);
+
+        emitter.int32_to_uint(8, span);
+
+        assert_int32_to_uint_stack_contract(&block, span, 8);
+        assert_eq!(
+            &block[3],
+            &Op::Inst(masm::Span::new(
+                span,
+                OpEmitter::assertz_with_message_inst(
+                    "value does not fit in unsigned 8-bit range",
+                    span
+                )
+            ))
+        );
+    }
+
+    #[test]
+    fn try_int32_to_uint_masks_top_stack_word_and_preserves_input() {
+        let span = SourceSpan::default();
+        let context = Rc::new(Context::default());
+        let mut stack = OperandStack::new(context);
+        let mut invoked = BTreeSet::default();
+        let mut block = Vec::new();
+        let mut emitter = new_emitter(&mut invoked, &mut block, &mut stack);
 
         emitter.try_int32_to_uint(8, span);
 
-        assert_eq!(&block[0], &Op::Inst(masm::Span::new(span, masm::Instruction::Dup0)));
+        assert_int32_to_uint_stack_contract(&block, span, 8);
+        assert_eq!(
+            &block[3],
+            &Op::Inst(masm::Span::new(span, masm::Instruction::EqImm(Felt::ZERO.into())))
+        );
     }
 }
--- a/codegen/masm/src/emit/int32.rs
+++ b/codegen/masm/src/emit/int32.rs
@@ -1070,3 +1070,41 @@ impl OpEmitter<'_> {
         self.raw_exec("::intrinsics::i32::max", span);
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use alloc::{collections::BTreeSet, rc::Rc};
+
+    use midenc_hir::Context;
+
+    use super::*;
+    use crate::{OperandStack, masm::Op};
+
+    #[test]
+    fn int32_to_uint_masks_top_stack_word() {
+        let span = SourceSpan::default();
+        let context = Rc::new(Context::default());
+        let mut stack = OperandStack::new(context);
+        let mut invoked = BTreeSet::default();
+        let mut block = Vec::new();
+        let mut emitter = OpEmitter::new(&mut invoked, &mut block, &mut stack);
+
+        emitter.int32_to_uint(8, span);
+
+        assert_eq!(&block[0], &Op::Inst(masm::Span::new(span, masm::Instruction::Dup0)));
+    }
+
+    #[test]
+    fn try_int32_to_uint_masks_top_stack_word() {
+        let span = SourceSpan::default();
+        let context = Rc::new(Context::default());
+        let mut stack = OperandStack::new(context);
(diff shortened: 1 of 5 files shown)


bitwalker:
Thanks for the contribution! However, in the future, do not address multiple unrelated issues with a single PR like this - not only is it more burdensome to review, it prevents merging any fixes until the PR in its entirety is ready to merge. Additionally, our contribution guidelines require that contributors be assigned to an issue before opening PRs, due to the large influx of AI-generated PRs.

I'm letting both of those slide in this case, because I feel that the issues are (in all but one case) important to address - but I just want to be clear that in general my default inclination is to just close PRs like this without any review.

codegen/masm/src/emit/int32.rs
```diff
@@ -5,6 +5,10 @@ use super::{OpEmitter, dup_from_offset, felt, masm, movup_from_offset};
 
 pub const SIGN_BIT: u32 = 1 << 31;
 
+fn unsigned_int32_reserved_mask(n: u32) -> u32 {
```
bitwalker, line 8: Just inline the body of this helper at the callsites - it is too simple and too vague to be useful as a helper function IMO, and makes it harder to follow the code where it is used.

codegen/masm/src/emit/int32.rs
```diff
@@ -1070,3 +1072,149 @@ impl OpEmitter<'_> {
...
+    use crate::{OperandStack, masm::Op};
+
+    fn push_u32_op(value: u32, span: SourceSpan) -> Op {
+        Op::Inst(masm::Span::new(
+            span,
+            masm::Instruction::Push(masm::Immediate::Value(masm::Span::new(span, value.into()))),
+        ))
+    }
+
+    fn new_emitter<'a>(
```
bitwalker, line 1092: This function is redundant, I would just replace its uses with `OpEmitter::new` directly

1sgtpepper:
My apologies, I had grouped these because the Wasm memory issues seemed closely related, and I thought the int32 narrowing fix was small enough to address alongside them. I understand the preference for keeping separate issues in separate PRs. What would be your preferred next step here?

codegen/masm/src/emit/int32.rs
```diff
@@ -1070,3 +1072,149 @@ impl OpEmitter<'_> {
...
+        invoked: &'a mut BTreeSet<masm::Invoke>,
+        block: &'a mut Vec<Op>,
+        stack: &'a mut OperandStack,
+    ) -> OpEmitter<'a> {
+        OpEmitter::new(invoked, block, stack)
+    }
+
+    fn assert_int32_to_uint_stack_contract(block: &[Op], span: SourceSpan, n: u32) {
```
bitwalker, line 1100: This is just a tautological assertion, and isn't actually useful IMO. We don't care about the specific instructions that get emitted, we care that the behavior is correct. I would remove this entirely.

codegen/masm/src/emit/int32.rs
```diff
@@ -1070,3 +1072,149 @@ impl OpEmitter<'_> {
...
+        assert_eq!(
+            &block[0],
+            &Op::Inst(masm::Span::new(span, masm::Instruction::Dup0)),
+            "must duplicate the top stack word, not the word underneath"
+        );
+        assert_eq!(&block[1], &push_u32_op(mask, span));
+        assert_eq!(&block[2], &Op::Inst(masm::Span::new(span, masm::Instruction::U32And)));
+    }
+
+    fn emitted_uint_check_succeeds(block: &[Op], initial_stack: &[u32]) -> bool {
```
bitwalker, line 1112: This should be replaced with a real behavioral test, i.e. construct an emitter for a block, call the emit helper being tested, build a MASM function from that block, assemble that to a program, and then execute the program using the processor. We recently added some tests in the integration test suite that assemble and execute MASM directly, in order to test the MASM-based compiler intrinsics - that functionality should be available to use here.

Doing the above will also make it easier to use property-based testing to randomize both the inputs and the target bit size of the output value.

codegen/masm/src/emit/int32.rs
```diff
@@ -1070,3 +1072,149 @@ impl OpEmitter<'_> {
...
+                    stack.insert(0, lhs & rhs);
+                }
+                inst => panic!("unexpected instruction in uint check prefix: {inst:?}"),
+            }
+        }
+
+        stack[0] == 0
+    }
+
+    #[test]
```
bitwalker, line 1145: I would remove this test - it's too brittle and doesn't really provide any meaningful regression protection.

codegen/masm/src/emit/int32.rs
```diff
@@ -1070,3 +1072,149 @@ impl OpEmitter<'_> {
...
+                OpEmitter::assertz_with_message_inst(
+                    "value does not fit in unsigned 8-bit range",
+                    span
+                )
+            ))
+        );
+    }
+
+    #[test]
+    fn try_int32_to_uint_masks_top_stack_word_and_preserves_input() {
```
bitwalker, line 1170: I would remove this test - like the one above, it's too brittle and is essentially tautological

codegen/masm/src/emit/int32.rs
```diff
@@ -1070,3 +1072,149 @@ impl OpEmitter<'_> {
...
+        emitter.try_int32_to_uint(8, span);
+
+        assert_int32_to_uint_stack_contract(&block, span, 8);
+        assert_eq!(
+            &block[3],
+            &Op::Inst(masm::Span::new(span, masm::Instruction::EqImm(Felt::ZERO.into())))
+        );
+    }
+
+    #[test]
```
bitwalker, line 1187: I think this case could be covered by the behavioral test I outlined in my other comment

codegen/masm/src/emit/int32.rs
```diff
@@ -1070,3 +1072,149 @@ impl OpEmitter<'_> {
...
+        let mut block = Vec::new();
+        let mut emitter = new_emitter(&mut invoked, &mut block, &mut stack);
+
+        emitter.int32_to_uint(32, span);
+
+        assert_int32_to_uint_stack_contract(&block, span, 32);
+        assert_eq!(&block[1], &push_u32_op(0, span));
+    }
+
+    #[test]
```
bitwalker, line 1202: I would remove this test as well - it is essentially testing that we aren't making the same mistake as the original bug, rather than testing that the behavior of the function is correct, which would not only exclude the bug, but also cover more cases.

frontend/wasm/src/code_translator/mod.rs
```diff
@@ -217,28 +217,39 @@ pub fn translate_operator<B: ?Sized + Builder>(
         }
         /******************************* Memory management *********************************/
         Operator::MemoryGrow { .. } => {
-            let arg = state.pop1_bitcasted(U32, builder, span);
-            let result = builder.mem_grow(arg, span)?;
-            // WASM memory.grow returns i32, so bitcast from U32 to I32
-            state.push1(builder.bitcast(result, I32, span)?);
+            unsupported_diag!(
```
bitwalker, line 220: This is just flat out incorrect - the previous lowering for this should be preserved

frontend/wasm/src/code_translator/mod.rs
```diff
@@ -217,14 +217,28 @@ pub fn translate_operator<B: ?Sized + Builder>(
...
+                    .with_message("MemorySize: module has no linear memory")
+                    .into_report()
+            })?;
+            if memory.imported {
+                unsupported_diag!(
+                    diagnostics,
+                    "MemorySize: imported linear memories are not supported yet"
+                );
+            }
+            let result = builder.u32(u32::try_from(memory.minimum).into_diagnostic()?, span);
```
bitwalker, line 241: This is also incorrect - not only are the semantics wrong (it doesn't respect `mem_grow`), but the minimum size of the memory doesn't dictate its size - the `maximum` does.

Regardless, this isn't even the right way to handle this. If we're going to bother to respect memory size limits then it needs to be done by integrating into the existing `mem_grow`/`mem_size` intrinsics (and I'm not convinced we have any reason to respect the limits declared for Wasm - there is little benefit to imposing the declared heap size limit, as it has almost no bearing on how much memory is ultimately consumed by the resulting MASM program).

Regardless - #1184 (from which this change is derived) should be handled as a separate PR entirely, as it is a bigger change than what you've implemented here.

1sgtpepper:
Quick update: I restored the existing single-memory `memory.copy` lowering because rejecting it caused the current CI failures across existing Rust/template/integration test paths. With that change, this PR no longer claims to fix #1183 and is now scoped to #1182 and #1184.

bitwalker (changes requested):
There are a number of issues with the changes made here, and I do not see any fix for #1183 here, which you said this PR closes.

I would suggest closing this PR, and opening new ones for each issue individually, taking into account the feedback I've left here - it is more likely that at least some of your fixes will get merged that way, and will avoid the PR getting too messy.

1sgtpepper:
Thanks for the review. I agree this PR is over-scoped and should be split. I’ll close it and stop treating #1183/#1184 as part of this PR.

Given your #1184 clarification, my next plan is a focused #1182 PR only: the int32 unsigned narrowing fix, including the n = 32 mask case, with behavioral MASM/VM regression coverage instead of instruction-shape tests.

Could you assign #1182 to me if that scope sounds right?

PR #1194, opened 2026-06-17
fix: determine sign by looking at hi limb in `i64::is_signed`

### LINKED ISSUE #1187: Native/Intrinsic divergence for `i64::is_signed`
### Reproduction

Requires #1178 

```sh
cargo nextest run -p midenc-integration-tests -E 'test(/i64_is_signed/)' --run-ignored all

   thread 'end_to_end::intrinsics::i64_arithmetic::i64_is_signed' (45234) panicked at tests/integration/src/end_to_end/intrinsics/i64_arithmetic.rs:78:13:
    Found minimal failing case: -878442593682194433
    assertion failed: `(left == right)`
      left: `[0]`,
     right: `[1]` at tests/integration/src/end_to_end/intrinsics/i64_arithmetic.rs:55
```

### Cause

The intrinsic drops `a_hi` instead of `a_lo` and then does sign check on the wrong limb.

mooori:
Closes #1187 

Previously the sign was determined by looking at the `lo` limb instead of the `hi` limb.

### DIFF
--- a/codegen/masm/intrinsics/i64.masm
+++ b/codegen/masm/intrinsics/i64.masm
@@ -20,7 +20,7 @@ end
 #
 # This function consumes `a`.
 pub proc is_signed # [a_lo, a_hi]
-    swap.1 drop
+    drop
     push.SIGN_BIT u32and push.SIGN_BIT eq
 end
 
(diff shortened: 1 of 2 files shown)


bitwalker (approved):
Ah, this is from back before the little-endian switch, I knew we must still have some lingering details of this around 😅 - good find

PR #1195, opened 2026-06-18
fix(codegen): check narrowed int32 operand

### LINKED ISSUE #1182: int32_to_uint checks dup.1 instead of the value being narrowed

1sgtpepper:
Closes #1182

`int32_to_uint` and `try_int32_to_uint` were duplicating the word below the value being narrowed before applying the unsigned range mask. With another live value below the operand, checked narrowing could reject a valid value or accept an out-of-range value based on the wrong stack word.

This changes both helpers to duplicate the top stack value before masking. The full-width `n == 32` case uses a zero mask so the mask construction does not shift by 32.

Regression coverage compiles HIR snippets and executes them through the package evaluator:

- checked casts from `u32` and `i32` to `i1`, `u8`, and `u16`;
- overflowing `u8` addition, which covers the `try_int32_to_uint` path;
- live guard values below the checked operand, so the old `dup.1` behavior fails on both valid and invalid inputs.

### DIFF
--- a/codegen/masm/src/emit/int32.rs
+++ b/codegen/masm/src/emit/int32.rs
@@ -296,15 +296,12 @@ impl OpEmitter<'_> {
     /// Execution traps if the value cannot fit in the unsigned N-bit range.
     pub fn int32_to_uint(&mut self, n: u32, span: SourceSpan) {
         assert_valid_integer_size!(n, 1, 32);
-        // Mask the value and ensure that the unused bits above the N-bit range are 0
-        let reserved = 32 - n;
-        let mask = (2u32.pow(reserved) - 1) << n;
-        // Copy the input
-        self.emit(masm::Instruction::Dup1, span);
-        // Apply the mask
+        // If the target bit width is 32, then use an empty mask.
+        let mask = if n == 32 { 0 } else { u32::MAX << n };
+        // The range check must inspect the top value, not another live value below it.
+        self.emit(masm::Instruction::Dup0, span);
         self.emit_push(mask, span);
         self.emit(masm::Instruction::U32And, span);
-        // Assert the masked value is all 0s
         self.emit(
             Self::assertz_with_message_inst(
                 format!("value does not fit in unsigned {n}-bit range"),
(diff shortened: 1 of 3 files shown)


codegen/masm/src/emit/mod.rs
```diff
@@ -1906,6 +1906,53 @@ mod tests {
         assert_eq!(emitter.stack()[0], Type::I32);
     }
 
+    #[test]
```
bitwalker: This test is essentially tautological - IMO it isn't useful, so I'd remove it. Tests should be behavioral (i.e. test that the output is correct for specific inputs), and when based on specific prior failures, we can specifically exercise those as a form of regression testing.

tests/integration/src/codegen/int32.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    compile_test_module([Type::U32, Type::U32, Type::U8, Type::U8], [Type::I1], |builder| {
+        let block = builder.current_block();
+        let expected_guard = block.borrow().arguments()[0] as ValueRef;
+        let live_guard = block.borrow().arguments()[1] as ValueRef;
+        let lhs = block.borrow().arguments()[2] as ValueRef;
+        let rhs = block.borrow().arguments()[3] as ValueRef;
+
+        let (overflowed, _sum) = builder.add_overflowing(lhs, rhs, span).unwrap();
+        // Keep both guards live below the sum so overflowing arithmetic must validate the sum,
```
bitwalker: The wording of this is pretty unclear, because it is referring to a specific failure mode of the original code that is fixed. I would remove this comment, and instead provide a comment that explains the purpose of the guard values (which is essentially to have stuff on the operand stack that would cause an error if values on the operand stack were consumed incorrectly).

codegen/masm/src/emit/int32.rs
```diff
@@ -319,15 +316,12 @@ impl OpEmitter<'_> {
     /// Places a boolean on top of the stack indicating if the conversion was successful
     pub fn try_int32_to_uint(&mut self, n: u32, span: SourceSpan) {
         assert_valid_integer_size!(n, 1, 32);
-        // Mask the value and ensure that the unused bits above the N-bit range are 0
-        let reserved = 32 - n;
-        let mask = (2u32.pow(reserved) - 1) << n;
-        // Copy the input
-        self.emit(masm::Instruction::Dup1, span);
-        // Apply the mask
+        // A 32-bit target has no unused high bits, so use a zero mask without shifting by 32.
```
bitwalker:
```suggestion
        // If the target bit width is 32, then use an empty mask
```

bitwalker (changes requested):

1sgtpepper:
@bitwalker Could you approve the CI run? Thanks!

bitwalker:
Fixed as part of #1356

PR #1174, opened 2026-06-12
feat: support `i32::checked_mod`

mooori:
- Adds intrinsic `i32::checked_mod`
- Adds codegen for `checked_mod_i32`

Closes #1152 and allows unignoring tests `checked_rem_i(8|16|32)`.

### `checked_mod` semantics

Assumes `checked_mod` corresponds to Rust's [`checked_rem`](https://doc.rust-lang.org/std/primitive.i32.html#method.checked_rem) because

- `u32mod` is based on the same `fn handle_division` as `u32div` [ref](https://github.com/0xMiden/miden-vm/blob/d1c853cd25497d17806745e64a9f0919e42f2455/crates/assembly/src/instruction/u32_ops.rs#L202)
- `u32mod` calculates the [remainder](https://github.com/0xMiden/miden-vm/blob/d1c853cd25497d17806745e64a9f0919e42f2455/crates/assembly/src/instruction/u32_ops.rs#L203) which corresponds to `a % b` in Rust [ref](https://github.com/0xMiden/miden-vm/blob/d1c853cd25497d17806745e64a9f0919e42f2455/processor/src/execution/operations/u32_ops/tests.rs#L230)

PR #1196, opened 2026-06-18
feat: support `i64::checked_mod`

### LINKED ISSUE #1000: Implement `checked_mod` for `i64`
Start at `OpEmitter::checked_mod` and add `OpEmmitter::checked_mod_i64` that calls the new intrinsic (add it) at `codegen/masm/intrinsics/i64.masm`.

mooori:
- Adds intrinsic `i64::checked_mod`
  - semantics correspond to Rust's [`checked_rem`](https://doc.rust-lang.org/std/primitive.i32.html#method.checked_rem), see #1174 for the reasoning
- Adds codegen for `checked_mod_i64`

Closes #1000 and unlocks `i64::overflowing_rem`, `i64::checked_rem`.

### DIFF
--- a/codegen/masm/intrinsics/i64.masm
+++ b/codegen/masm/intrinsics/i64.masm
@@ -253,6 +253,40 @@ pub proc checked_div # [b_lo, b_hi, a_lo, a_hi]
     end
 end
 
+# Computes `a % b`, asserting that both inputs are valid i64 values (u32 limbs).
+#
+# Execution traps on division by zero, and on `i64::MIN % -1` (the underlying division overflows).
+pub proc checked_mod # [b_lo, b_hi, a_lo, a_hi]
+    u32assertw
+
+    # Trap on i64::MIN % -1, since the underlying division overflows.
+    dup.1 dup.1                                                    # [b_lo, b_hi, b_lo, b_hi, a_lo, a_hi]
+    push.NEG1_HI push.NEG1_LO exec.::miden::core::math::u64::eq    # [b == -1, b_lo, b_hi, a_lo, a_hi]
+    dup.4 dup.4                                                    # [a_lo, a_hi, b == -1, b_lo, b_hi, a_lo, a_hi]
+    push.MIN_HI push.MIN_LO exec.::miden::core::math::u64::eq      # [a == MIN, b == -1, b_lo, b_hi, a_lo, a_hi]
+    and assertz                                                    # [b_lo, b_hi, a_lo, a_hi]
+
+    # The remainder takes the sign of the dividend `a`, so `is_signed_a` will be needed later.
+    dup.3 push.SIGN_BIT u32and push.SIGN_BIT eq                    # [is_a_signed, b_lo, b_hi, a_lo, a_hi]
+    movdn.4                                                        # [b_lo, b_hi, a_lo, a_hi, is_a_signed]
+
+    # abs(a)
+    movup.3 movup.3                                                # [a_lo, a_hi, b_lo, b_hi, is_a_signed]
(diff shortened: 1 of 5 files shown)


mooori:
I think the `lit` CI failure is unrelated to the changes in this PR:

- The `lit` suite has been failing silently for a while and the fixes from #1175 are not yet merged.
- The `litcheck` exit code issue has already been [fixed and published](https://github.com/0xMiden/compiler/pull/1175#pullrequestreview-4519193550)

So I suppose one of the pre-existing issues is now making the `lit` CI job red.

tests/integration/src/end_to_end/intrinsics/i64_arithmetic.rs
```diff
@@ -363,6 +363,30 @@ fn i64_checked_div() {
...
+    # Stack: [r_lo, r_hi]
+"#;
+    test_i64_intrinsic(
+        proc_body,
+        NumericStrategy::<i64>::rem_signed_checked(),
+        binary_i64op_inputs_to_stack,
+        |(a, b): &(i64, i64)| {
+            if *b == 0 {
+                Err(TrapExpectation::DivideByZero)
+            } else if *a == i64::MIN && *b == -1 {
```
greenhat, line 380: WebAssembly signed remainder is only undefined for a zero divisor; `i64.rem_s(i64::MIN, -1)` should return `0`, not trap ([WebAssembly numerics, `irem_s`](https://webassembly.github.io/spec/core/exec/numerics.html#op-irem)).

greenhat (changes requested):
Looking good overall! Please check my note.
Also, IIRC we have `arith.smod`. Should we use it here?

codegen/masm/intrinsics/i64.masm
```diff
@@ -253,6 +253,40 @@ pub proc checked_div # [b_lo, b_hi, a_lo, a_hi]
     end
 end
 
+# Computes `a % b`, asserting that both inputs are valid i64 values (u32 limbs).
+#
+# Execution traps on division by zero, and on `i64::MIN % -1` (the underlying division overflows).
+pub proc checked_mod # [b_lo, b_hi, a_lo, a_hi]
```
bitwalker, line 259: We should add type signatures to all new intrinsics

bitwalker (commented):
Looks good! Aside from @greenhat's notes, I just had one nit, otherwise 👍

tests/integration/src/end_to_end/intrinsics/i64_arithmetic.rs
mooori: > i64.rem_s(i64::MIN, -1) should return 0, not trap

I think this means we have a bug in the lowering of Wasm's `i64.rem_s` which tests do not catch:

- Wasm `i64.rem_s` is probably compiled to `i64::checked_mod` intrinsic
- For `(i64::MIN, -1)` Wasm returns 0 but the intrinsic traps

The arith tests `i64::checked_rem` and `i64::overflowing_rem` do hit the intrinsic `i64::checked_mod`. Though when compiling `i64::checked_rem`, `i64::overflowing_rem` from Rust to Wasm, the generated Wasm handles edge cases explicitly, thus hiding the bug.

My plan is:

- Add test(s) to check whether the `i64.rem_s` lowering is fine or not
- Fix it if needed. Maybe by adding intrinsic `i64::wrapping_mod` and lowering `i64.rem_s` to that.

The same issue should then also exist for `i32.rem_s` and the `i32::checked_mod` intrinsic.

greenhat: Sounds good!

mooori:
Reverted to draft to work on #1206 first.

PR #1199, opened 2026-06-18
fix(debuginfo): preserve exported type names

### LINKED ISSUE #1198: Struct are anon in debug dump info
In the case in contract we have custom struct `Point`, and some functions that take `Point`, `AccountId`, or `Word`:

```rust
pub struct Point {
    pub x: Felt,
    pub y: Felt,
}

pub fn add_points(&self, a: Point, b: Point) -> Point { ... }
pub fn take_word(&self, w: Word) -> Word { ... }
pub fn take_account_id(&self, id: AccountId) -> AccountId { ... }
```

The debug info keeps the real type, but in the debug dump(`miden-objtool dump debug-info`) each one is a `STRUCT`. `size` is in bytes,  with each field with struct: <anon>.

Here is a `Word` (4 felts, 16 bytes):

```
[  47] STRUCT str[1] (size: 16 bytes)
    +   0: str[0] : struct <anon>
    +   4: str[2] : struct <anon>
    +   8: str[3] : struct <anon>
    +  12: str[4] : struct <anon>
```

And a `Point` (2 felts, 8 bytes):

```
[ 238] STRUCT str[1] (size: 8 bytes)
    +   0: str[0] : struct <anon>
    +   4: str[2] : struct <anon>
```

The dump shows the real name, for example next to `STRUCT` or jsut the real name, and the real type next to each
field. So instead of `<anon>` you see `Point` and `Felt`.

`Point` (2 felts, 8 bytes):

```
[ 238] STRUCT Point (size: 8 bytes)
    +   0: x : Felt
    +   4: y : Felt
```

`Word` (4 felts, 16 bytes):

```
[  47] STRUCT Word (size: 16 bytes)
    +   0: Felt
    +   4: Felt
    +   8: Felt
    +  12: Felt
```

Dump example:
[raw_debug_dump.txt](https://github.com/user-attachments/files/29100135/raw_debug_dump.txt)

djolertrk:
Preserve the type names and pass it to VM.
The VM fix for it is https://github.com/0xMiden/miden-vm/pull/3269.

Resolves https://github.com/0xMiden/compiler/issues/1198.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,6 +168,7 @@ miden-stdlib-sys = { version = "0.12.0", path = "sdk/stdlib-sys" }
 
 [patch.crates-io]
 #miden-assembly = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
+#miden-assembly = { path = "../miden-vm/assembly" }
 # miden-debug = { git = "https://github.com/0xMiden/miden-debug", branch = "main" }
 #miden-protocol = { git = "https://github.com/0xMiden/protocol", rev = "a53bbe2209f506df87876c8b9c9a1730214f456b" }
 #miden-standards = { git = "https://github.com/0xMiden/protocol", rev = "a53bbe2209f506df87876c8b9c9a1730214f456b" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -169,7 +169,15 @@ miden-stdlib-sys = { version = "0.12.0", path = "sdk/stdlib-sys" }
 [patch.crates-io]
 #miden-assembly = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
 #miden-assembly = { path = "../miden-vm/assembly" }
+#miden-assembly-syntax = { path = "../miden-vm/assembly-syntax" }
+#miden-core = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
+#miden-core = { path = "../miden-vm/core" }
 # miden-debug = { git = "https://github.com/0xMiden/miden-debug", branch = "main" }
+#miden-debug-types = { path = "../miden-vm/crates/debug/types" }
+#miden-processor = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
+#miden-processor = { path = "../miden-vm/processor" }
+#miden-mast-package = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
+#miden-mast-package = { path = "../miden-vm/package" }
 #miden-protocol = { git = "https://github.com/0xMiden/protocol", rev = "a53bbe2209f506df87876c8b9c9a1730214f456b" }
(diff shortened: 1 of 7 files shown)


djolertrk:
Marking this as draft until this https://github.com/0xMiden/miden-vm/pull/3269 is merged

djolertrk:
> Marking this as draft until this [0xMiden/miden-vm#3269](https://github.com/0xMiden/miden-vm/pull/3269) is merged

This is merged, but we need to wait for the new patch release to be published to crates.io

tests/lit/hir/parsing/structured_control_flow.hir
```diff
@@ -1,72 +1,22 @@
 // RUN: hir-opt %s | filecheck %s
 
-// XFAIL: *
```
greenhat: This test is fixed in https://github.com/0xMiden/compiler/pull/1175

greenhat (changes requested):
Looking good! Just a minor note.

djolertrk:
> Looking good! Just a minor note.

Thanks! Will address that along with VM version bump once the new patch release is ready on crates.io.

greenhat (approved):
Looking good!

greenhat:
@djolertrk Please sign the commits.

djolertrk:
> @djolertrk Please sign the commits.

done, thanks!

Cargo.toml
```diff
@@ -167,17 +167,7 @@ miden-field-repr = { version = "0.12.0", path = "sdk/field-repr/repr" }
 miden-stdlib-sys = { version = "0.12.0", path = "sdk/stdlib-sys" }
 
 [patch.crates-io]
-#miden-assembly = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
```
bitwalker, line 170: Please leave these lines in this file - they're used sometimes during local development when we need to patch these crates; commit hashes and such in them don't have to be kept up to date, but they should remain here for convenience

djolertrk: I agree, done.

djolertrk:
This is ready I guess :)

Cargo.toml
```diff
@@ -168,16 +168,7 @@ miden-stdlib-sys = { version = "0.12.0", path = "sdk/stdlib-sys" }
 
 [patch.crates-io]
 #miden-assembly = { git = "https://github.com/0xMiden/miden-vm", rev = "614cd7f9b52f45238b0ab59c71ebb49325051e5d" }
-#miden-assembly = { path = "../miden-vm/assembly" }
```
greenhat, line 171: Seems like a regression after https://github.com/0xMiden/compiler/pull/1199#discussion_r3479207384

greenhat (changes requested):

Cargo.toml
djolertrk: I think that this was my comment since that is the way I organised directories :)

djolertrk: But I can return it back if you think it is useful?

greenhat: I think Paul asked not to remove them (commented deps), but I see them removed.

djolertrk: Yeah, you are right, both lines should be there. I just returned it back.

bitwalker: @djolertrk Looks like these lines are still removed - I'd just revert the changes to this file entirely in your branch. Ping me once you're ready and I'll get this merged

djolertrk: Ouch, done

bitwalker (approved):

PR #1203, opened 2026-06-22
Don't export internal procedures in the MAST/.masp

### LINKED ISSUE #1197: Don't export internal procedures in the compiled package
Discovered in #1192 

The package exports and the MAST exports contain internal procedures:
```
[tests/integration/src/end_to_end/examples/counter_note.rs:35:5] package_exports = [
    "\"miden:counter-contract\"::\"miden-counter-contract@0.1.0\"::\"get-count\" -> 0x42e6ad875cc27716b62c2653a94a84c86b42ebfaef9b1ceae58769fd8db54a35",
    "\"miden:counter-contract\"::\"miden-counter-contract@0.1.0\"::\"increment-count\" -> 0x028811d3f83f498dc3aebd90f388c11654d80a764f901428a64c986edb7d30d4",
    "::\"miden:counter-contract/miden-counter-contract@0.1.0\"::\"get-count\" -> 0xfbf5d9582443e18037624698c8a930eb0f71b211caa60a49ba3de0031acdd9aa",
    "::\"miden:counter-contract/miden-counter-contract@0.1.0\"::\"increment-count\" -> 0x3fb4f19a6dec50b8f0eb80e05b3218c0a8210c878047403f4584d72512720c0e",
    "::\"miden:counter-contract/miden-counter-contract@0.1.0\"::counter_contract::cabi_realloc -> 0xb66c0ef9551334e8922c0dcc297ea411e569a9c75f4e4509d005c55912a49fa1",
    "::\"miden:counter-contract/miden-counter-contract@0.1.0\"::init -> 0x5ed60c1dcdccfe9a25e144b12a93f6f6803746bd3dabf4707bac4a024b253911",
]
```

greenhat:
Close #1197 

Compiled Miden packages now export only the lifted Component Model (CM) wrapper procedures. Internal procedures — lowered core Wasm functions, the component `init`, `cabi_*`, and intrinsics — are kept off the package export surface. Previously they leaked in as public exports.

**Frontend (`frontend/wasm`)**
- Core Wasm exports inside a component are translated as `Private`; the specific core export wrapped by a lifted function is promoted to `Internal` so the wrapper can still resolve it. Standalone (non-component) module exports stay `Public`.

**Codegen (`codegen/masm`)**
- At assembly, non-root support modules are statically linked into the assembler instead of being surfaced as package source, so internal procedures never reach `Library.exports()` and now-unreachable ones are pruned by DCE.
- The component `init` procedure is now a private, root-local procedure, invoked via a same-module symbol from the lifted wrappers and the generated entrypoint.
- Executable `main`/test-harness generation moved from assembly-time into lowering.

**Session (`midenc-session`)**
- Centralized executable-target detection in `Session::is_executable_target()`.

### DIFF
--- a/codegen/masm/CHANGELOG.md
+++ b/codegen/masm/CHANGELOG.md
@@ -8,7 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- keep internal component procedures out of compiled MASM package exports
+- keep internal component procedures out of compiled MASM package exports so library packages expose
+  only lifted Component Model wrappers
+- preserve generated component initializer callees in MASM procedure metadata
 
 ## [0.5.1](https://github.com/0xMiden/compiler/compare/midenc-codegen-masm-v0.5.0...midenc-codegen-masm-v0.5.1) - 2025-11-13
 
--- a/codegen/masm/CHANGELOG.md
+++ b/codegen/masm/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- keep internal component procedures out of compiled MASM package exports
+
 ## [0.5.1](https://github.com/0xMiden/compiler/compare/midenc-codegen-masm-v0.5.0...midenc-codegen-masm-v0.5.1) - 2025-11-13
 
 ### Other
--- a/codegen/masm/src/artifact.rs
+++ b/codegen/masm/src/artifact.rs
@@ -17,11 +17,13 @@ pub struct MasmComponent {
     ///
     /// All components must have a canonical root module, even if empty
     pub root: Arc<Path>,
-    /// The symbol name of the component initializer function
+    /// Whether this component requires an initializer procedure.
     ///
-    /// This function is responsible for initializing global variables and writing data segments
+    /// The initializer is responsible for initializing global variables and writing data segments
     /// into memory at program startup, and at cross-context call boundaries (in callee prologue).
-    pub init: Option<masm::InvocationTarget>,
+    /// When set, a private root-local `init` procedure is generated and invoked by the lifted
+    /// export wrappers (and the generated executable entrypoint) via a same-module symbol.
+    pub requires_init: bool,
     /// The symbol name of the program entrypoint, if this component is executable.
     ///
     /// If unset, it indicates that the component is a library, even if it could be made executable.
--- a/codegen/masm/src/artifact.rs
+++ b/codegen/masm/src/artifact.rs
@@ -34,6 +34,11 @@ pub struct MasmComponent {
     pub heap_base: u32,
     /// The address of the `__stack_pointer` global, if such a global has been defined
     pub stack_pointer: Option<u32>,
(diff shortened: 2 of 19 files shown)


bitwalker (changes requested):
The way the module hierarchy is defined (and thus the public surface of the package that gets assembled) has changed in the upcoming VM release. In particular, the public surface of the package is determined based on the public symbols of the root module, and any of its public submodules; rather than any public symbol in either the root or its support modules. I think we have to keep the current way we assemble the target namespace (which will expose more symbols than will be exposed after we migrate to the new VM) - but we should ensure that intrinsics (and other library code that isn't part of the target namespace) aren't linked as support modules, and are instead statically-linked separately.

> At assembly, non-root support modules are statically linked into the assembler instead of being surfaced as package source

This is fine for things that don't belong to the target namespace (e.g. the compiler intrinsics, statically-linked libraries, etc.). For modules that belong to the target namespace, we must continue to provide all of them as the public surface of the package - when we migrate to the next VM release, the public surface will automatically be pruned based on what is actually visible according to the module hierarchy.

> The component init procedure is now a private, root-local procedure, invoked via a same-module symbol from the lifted wrappers and the generated entrypoint.

I think the component init procedure probably needs to remain public, since it must be called by more than just the `component-model` exports themselves, but before any exported procedure from the Miden component can be used (since those exports will assume that linear memory, etc., has been initialized). Codegen can't assume that the Miden component is a Wasm component (and therefore only has `component-model` exports).

In practice, currently, the component level init procedure is unlikely to be used outside the `component-model` exports, or the generated main procedure - but it is technically a requirement for using a component, so it should be exported IMO.

greenhat:
> The way the module hierarchy is defined (and thus the public surface of the package that gets assembled) has changed in the upcoming VM release. In particular, the public surface of the package is determined based on the public symbols of the root module, and any of its public submodules; rather than any public symbol in either the root or its support modules. I think we have to keep the current way we assemble the target namespace (which will expose more symbols than will be exposed after we migrate to the new VM) - but we should ensure that intrinsics (and other library code that isn't part of the target namespace) aren't linked as support modules, and are instead statically-linked separately.
> 
> > At assembly, non-root support modules are statically linked into the assembler instead of being surfaced as package source
> 
> This is fine for things that don't belong to the target namespace (e.g. the compiler intrinsics, statically-linked libraries, etc.). For modules that belong to the target namespace, we must continue to provide all of them as the public surface of the package - when we migrate to the next VM release, the public surface will automatically be pruned based on what is actually visible according to the module hierarchy.

Got it.

> 
> > The component init procedure is now a private, root-local procedure, invoked via a same-module symbol from the lifted wrappers and the generated entrypoint.
> 
> I think the component init procedure probably needs to remain public, since it must be called by more than just the `component-model` exports themselves, but before any exported procedure from the Miden component can be used (since those exports will assume that linear memory, etc., has been initialized). Codegen can't assume that the Miden component is a Wasm component (and therefore only has `component-model` exports).
> 
> In practice, currently, the component level init procedure is unlikely to be used outside the `component-model` exports, or the generated main procedure - but it is technically a requirement for using a component, so it should be exported IMO.

You're right. I forgot that the end goal for the `init` procedure to be called by the VM before the first invocation of any component method only once (per component instance).

Given all of the above, I think we can safely close this PR.

bitwalker:
Closing as discussed on our call today

Issue #1206, opened 2026-06-24
Miscompilation of Wasm's `rem_s` instruction


mooori:
### Problem

Wasm's `(i32|i64).rem_s` is compiled to the `(i32|i64)::checked_mod` intrinsic. However they behave differently for `(MIN, -1)`:

- Wasm's `rem_s` does not trap and returns 0
- The `checked_mod` intrinsic traps because the underlying division overflows

### Not (yet) surfaced by tests

Existing tests have not surfaced this. Rust's `(i32|i64)::(overflowing|checked)_rem` hit this path, however rustc emits wasm that explicitly checks edge cases, including `(MIN, -1)`. So the bug is hidden. See #1196 for context.

#1207 adds coverage.

### Reproduction

Run the ignored test from #1207

### Proposed solution

Add intrinsic `(i32|i64)::wrapping_mod` and compile `(i32|i64).rem_s` to that.

### TODO

- [x] `i32` #1233
- [ ] `i64`

PR #1207, opened 2026-06-24
test: compiled Wasm result vs interpreted for some  `i32` arith

mooori:
Adds test coverage for the `i32` variant of the bug described in #1206. AFAIK there is not yet a test setup to compare the result of:

- Wasm instruction executed/interpreted by Wasm runtime/interpreter
- vs compiling that Wasm instruction and executing it on Miden VM

If you like this approach, I would suggest that eventually we add such tests for most/all Wasm instructions. For now this just covers:

- `i32.add`, `i32.div_s` to show the approach works
- `i32.rem_s` to reproduce #1206 

### Wasm interpreter

Using `wasmi` as it's lightweight and easy to use.

### DIFF
--- a/tests/integration/Cargo.toml
+++ b/tests/integration/Cargo.toml
@@ -47,3 +47,5 @@ blake3 = "1.5"
 cargo_metadata = "0.19"
 concat-idents = "1.1"
 libloading = "0.8"
+wasmi = "1.1.0"
+wat.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3743,6 +3743,8 @@ dependencies = [
  "quote",
  "sha2 0.11.0",
  "syn 2.0.117",
+ "wasmi",
+ "wat",
 ]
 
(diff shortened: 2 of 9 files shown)


tests/integration/Cargo.toml
```diff
@@ -45,3 +45,5 @@ blake3 = "1.5"
 cargo_metadata = "0.19"
 concat-idents = "1.1"
 libloading = "0.8"
+wasmi = "1.1.0"
+wat.workspace = true
```
mooori, line 51: I suppose these new dependencies satisfy the comment, but not sure if my interpretation is correct:

> NOTE: Use local paths for dev-only dependency to avoid relying on crates.io during packaging

Comment not applicable since the new dependencies are from `crates.io`.

greenhat (approved):
I love the idea!

mooori:
(force-pushed after rebasing on `next`, no changes)

PR #1217, opened 2026-06-26
feat: add i32 intrinsics `overflowing_mod`, `wrapping_mod`

mooori:
- Adds i32 intrinsics `overflowing_mod`, `wrapping_mod`
- Changes `checked_mod` to re-use `wrapping_mod`

### Follow up

Lower Wasm's `i32.rem_s` through `wrapping_mod` instead of `checked_mod` to match Wasm semantics. See #1206 for context. Potentially by adding op `I32RemS` to the Wasm dialect.


greenhat (approved):
Looking good!

bitwalker (approved):

PR #1218, opened 2026-06-29
feat(sdk)!: generate per-component traits for #[account] wrappers

### LINKED ISSUE #1208: `#[account(...)]` macro should generate traits
> I don't understand why `BasicWallet` would be a component trait at all

Specifically to deal with the ambiguity that arises if multiple components are in scope for a given account struct that have overlapping/conflicting method names. They can't all be implemented as inherent methods of the account struct - the only way to support such overlap/conflict is via traits, so that you can disambiguate with turbofish syntax by specifying which trait/method you actually want.

For example, in the example you've provided, if `Wallet` had another `foo` component implemented for it, in addition to `basic_wallet`, and both components defined a `receive_asset` method - you'd end up with a compilation error with no way to resolve it, because we're emitting the component methods as inherent methods of the `Wallet` struct, rather than generating a `BasicWallet` trait, and a `Foo` trait, and implementations of both for `Wallet` (in which case the call to `account.receive_asset(asset)` in the body of `script` would need to be rewritten as `<Wallet as BasicWallet>::receive_asset(account, asset)` to disambiguate from `Foo::receive_asset`).

This obviously doesn't come up in the p2id example currently, because the `Wallet` struct only derives a single component (the `BasicWallet` component) - but we need to have a solution for multi-component scenarios, or we'll have no way for users to deal with conflicting methods between components - generating traits and trait impls solves this neatly (and in a way that is largely invisible to the user, except in cases where they need to disambiguate).

_Originally posted by @bitwalker in https://github.com/0xMiden/compiler/pull/1159#discussion_r3469787044_

I'm thinking about implementing it so that expanded code look like the following:
Source:

```rust
#[account(foo::Foo, bar::Bar)]
struct Wallet;

/// New SDK trait — the "expose the account id" trait.
pub trait AccountBinding {
  fn foreign_account_id(&self) -> Option<AccountId>;
}
```

`#[account(...)]` macro expansion:

```rust
// (1) hidden wit-bindgen bindings: native + `fpi_` import per function (UNCHANGED)
#[doc(hidden)]
#[allow(dead_code)]
pub mod bindings {
  pub mod foo { pub mod foo {
      pub fn receive_asset(asset: Asset) { /* native import */ }
      pub fn fpi_receive_asset(p: Felt, s: Felt, root: Word, asset: Asset) { /* FPI import */ }
  }}
  pub mod bar { pub mod bar {
      pub fn receive_asset(asset: Asset) { /* … */ }
      pub fn fpi_receive_asset(p: Felt, s: Felt, root: Word, asset: Asset) { /* … */ }
      pub fn get_balance() -> Felt { /* … */ }
      pub fn fpi_get_balance(p: Felt, s: Felt, root: Word) -> Felt { /* … */ }
  }}
}

#[derive(Default)]
struct Wallet {
  foreign_account_id: Option<::miden::AccountId>,
}

impl ::miden::account::AccountBinding for Wallet {
  #[inline(always)]
  fn foreign_account_id(&self) -> Option<::miden::AccountId> {
      self.foreign_account_id
  }
}

// ONE TRAIT PER COMPONENT — the dispatch lives in the default methods,
// reached through the AccountBinding supertrait (so it works without field access)
trait Foo: ::miden::account::AccountBinding {
  #[inline(always)]
  fn receive_asset(&self, asset: ::miden::Asset) {
      match self.foreign_account_id() {
          Some(id) => bindings::foo::foo::fpi_receive_asset(
              id.prefix, id.suffix,
              ::miden::Word::new([/* foo::receive_asset MAST root, baked at expansion */]),
              asset,
          ),
          None => bindings::foo::foo::receive_asset(asset),
      }
  }
}
impl Foo for Wallet {}

trait Bar: ::miden::account::AccountBinding {
  #[inline(always)]
  fn receive_asset(&self, asset: ::miden::Asset) {
      match self.foreign_account_id() {
          Some(id) => bindings::bar::bar::fpi_receive_asset(id.prefix, id.suffix, /*root*/, asset),
          None     => bindings::bar::bar::receive_asset(asset),
      }
  }
  #[inline(always)]
  fn get_balance(&self) -> ::miden::Felt {
      match self.foreign_account_id() {

```

greenhat:
Follow-up to #1157 addressing PR #1159 review feedback. Closes #1208.

## Overview

Reworks the `#[account(...)]` macro to expose each referenced account component as its **own generated trait** — named after the component's interface, with the wrapper's visibility, implemented for the wrapper struct — instead of flattening every component's methods into inherent methods on the wrapper. This lifts the previous restriction that no two components on one account could share a method name, and makes call sites explicit when they do.

## What changes

- **One trait per component.** `#[account(pkg::Interface)]` generates `trait Interface` and implements it for the wrapper. Single-component accounts with no name overlap still call `account.method(..)` unchanged (with the trait in scope).
- **Overlapping method names now allowed.** Two components exporting the same method coexist on one wrapper; disambiguate with UFCS `<Wallet as Interface>::method(account, ..)` — previously a hard compile error.
- **`as Alias` renaming.** `#[account(pkg::Interface as MyTrait)]` renames the generated trait, for when an interface name would clash with the wrapper struct, another reference, or a same-interface sibling `#[component(...)]`.
- Component methods that shadow an `ActiveAccount` built-in (e.g. `get_id`) or are named `new` are now permitted (resolved via UFCS / the inherent constructor).

## Breaking / migration

The wrapper struct must be named differently from every generated trait, and overlapping / `ActiveAccount`-shadowing calls need UFCS. Full guide in `sdk/sdk/MIGRATION.md` (`0.13.0 -> 0.14.0`); ships in the next SDK release.


greenhat:
Rebased and ready.

bitwalker (approved):

PR #1231, opened 2026-07-01
test(cargo-miden): remove test_all_templates and its exclusive infra

### LINKED ISSUE #1205: Move the contract templates testing (`tests_all_templates` test) to the template's repo
Keeping them in the compiler repo creates too much friction when making breaking changes. 
Update the release process docs so that the first item is to make sure both new project templates and contract templates are updated to work with the current version of the compiler.

greenhat:
Close #1205 

The test_all_templates integration test was moved to the rust-templates repository, where the project templates now live, so it is no longer needed here.

Remove the test and everything used only by it:

- Delete tests/build.rs (the test plus its private new_project_args and build_new_project_from_template helpers) and its tests/mod.rs entry.
- Drop the [dev-dependencies] miden-mast-package entry; build.rs was the only test referencing miden_mast_package. It remains a normal dependency used by the crate itself.
- Remove the dedicated cargo_miden_tests_templates_and_examples CI job, whose only purpose was running test_all_templates, and drop the now-stale exclusion from the cargo-miden integration-tests filter.


greenhat:
Rebased and ready.

bitwalker (approved):

PR #1232, opened 2026-07-01
Fix invalid trait object slice example

CryptAm:
The Rust example uses `&[dyn Foo]`, which does not compile because trait objects (`dyn Foo`) are unsized and cannot be used directly as slice elements.

Updated the function signature to use references to trait objects (`&[&dyn Foo]`), keeping the example behavior while making the snippet valid Rust.


bitwalker:
No longer relevant

PR #1233, opened 2026-07-01
feat: add `I32RemS` op to `wasm` dialect

mooori:
- Adds op `I32RemS` to the `wasm` dialect
- Translates Wasm's `i32.rem_s` to `I32RemS`, which is lowered to intrinsic `i32::wrapping_mod` instead of `i32::checked_mod`

Fixes the `i32.rem_s` miscompilation (see #1206) and makes the `i32_rem_s` test pass.

### DIFF
--- a/eval/src/eval.rs
+++ b/eval/src/eval.rs
@@ -931,6 +931,47 @@ macro_rules! binop_overflowing {
     }};
 }
 
+macro_rules! binop_wrapping_nonzero_rhs {
+    ($op:ident, $evaluator:ident, $operator:ident) => {{
+        let lhs = $op.lhs();
+        let lhs_value = $evaluator.use_value(&lhs.as_value_ref())?;
+        let rhs = $op.rhs();
+        let rhs_value = $evaluator.use_value(&rhs.as_value_ref())?;
+
+        let lhs_ty = lhs.ty();
+        let rhs_ty = lhs.ty();
+        if lhs_ty != rhs_ty {
+            return Err($evaluator.report(
+                "evaluation failed",
+                $op.span(),
+                format!("operand types do not match: {lhs_ty} vs {rhs_ty}"),
+            ));
+        }
+
+        if rhs_value == 0 {
+            return Err($evaluator.report(
+                "evaluation failed",
+                $op.span(),
+                format!("rhs must not be zero"),
+            ));
+        }
+
+        match (lhs_value, rhs_value) {
+            (Immediate::I8(x), Immediate::I8(y)) => Immediate::I8(x.$operator(y)),
+            (Immediate::U8(x), Immediate::U8(y)) => Immediate::U8(x.$operator(y)),
+            (Immediate::I16(x), Immediate::I16(y)) => Immediate::I16(x.$operator(y)),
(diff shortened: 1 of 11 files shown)


eval/src/eval.rs
```diff
@@ -931,6 +931,47 @@ macro_rules! binop_overflowing {
...
+macro_rules! binop_wrapping_nonzero_rhs {
+    ($op:ident, $evaluator:ident, $operator:ident) => {{
+        let lhs = $op.lhs();
+        let lhs_value = $evaluator.use_value(&lhs.as_value_ref())?;
+        let rhs = $op.rhs();
+        let rhs_value = $evaluator.use_value(&rhs.as_value_ref())?;
+
+        let lhs_ty = lhs.ty();
+        let rhs_ty = lhs.ty();
+        if lhs_ty != rhs_ty {
```
greenhat, line 943: They are both `lhs.ty()` (see two lines above).

greenhat (approved):
Looking good! Please see my comment.

greenhat:
@bitwalker I almost forgot that we have an IR evaluator. We are not using it in tests (except the `is_prime` test). Do we need it?

mooori:
> I almost forgot that we have an IR evaluator. We are not using it in tests (except the is_prime test). Do we need it?

I think removing it would be a separate issue/PR, so I'll go ahead and merge this PR.

---

Force pushed after rebase: initial commit unchanged, last commit is new.

eval/src/eval.rs
mooori: Addressed in 577e292 for the macro added in this PR.

There are other macros with `rhs_ty = lhs.ty()` and other issues in type checks. Will send a separate PR.

bitwalker:
@greenhat The purpose of the IR evaluator is for testing semantics - if an IR program fails under the evaluator, then that indicates that either the IR semantics for one or more ops are broken; a validator is missing or improperly implemented; or some other IR-level issue is present (e.g. a bad rewrite, analysis, or other bug of that sort).

We don't use it in tests currently, because we largely do end-to-end testing, which necessarily catches both IR-level issues as well as frontend and codegen issues - but that broad scope is also problematic when trying to triage the root cause in some cases. The evaluator is a useful diagnostic tool when we have isolated a problem to a relatively narrow scope, where executing an IR reproducer with the evaluator allows us to verify whether the IR we emitted from the frontend is actually semantically valid. If so, that tells us we have an issue with codegen, or at the VM level; if it fails in the evaluator, then we know we have a problem in the frontend or the IR.

It does have some constraints though, that limit its broader usefulness:

* It does not have any way to execute foreign code at all, so reproducers must be self-contained, and pure IR
* Coverage of all dialect operations is incomplete, so not all programs can be executed. This is fixable, but has been relatively low priority/motivated by need.
* Its correctness depends on correctly modeling the semantics of operations it supports - so it does not provide a guarantee of correctness. This isn't really a problem in practice, because the IR only provides very low-level primitives, so there isn't a lot of room to get it wrong - but it is a potential pitfall.

In any case, maintaining it is low-effort, so I would prefer to keep it around for the time being.

PR #1236, opened 2026-07-02
feat: improve debug info dialect and debug info generation

djolertrk:


djolertrk:
Resolves https://github.com/0xMiden/miden-debug/pull/95

frontend/wasm/src/module/debug_info.rs
```diff
@@ -175,8 +175,18 @@ fn build_function_debug_info(
 ) -> Option<FunctionDebugInfo> {
     let func_name = module.func_name(func_index);
 
-    let (file_symbol, directory_symbol) = determine_file_symbols(parsed_module, addr2line, body);
-    let (line, column) = determine_location(addr2line, body.body_offset);
+    // Translate the function body's file offset into the coordinate space used by the DWARF
+    // line info, mirroring the per-instruction lookups in `parse_function_body`: for standalone
+    // modules DWARF addresses are relative to the code section start, while for modules embedded
+    // in components they are offset by the module's base offset within the component.
+    let dwarf_offset = if parsed_module.wasm_file.module_base_offset > 0 {
```
bitwalker: Component DWARF location schedules still use the wrong address space. This new conversion correctly uses component-relative DWARF coordinates for the subprogram lookup, but `parse_function_body` still consumes `location_schedule` with module-relative `code_offset`, and `low_pc_map` is built in that relative space as well. Raw DIE/location-list ranges remain component-relative, so declarations and kills can be delayed or never emitted. Please centralize this conversion and use the DWARF-space offset consistently.

frontend/wasm/src/module/debug_info.rs
```diff
@@ -396,21 +407,30 @@ fn build_location_schedule(locals: &[Option<LocalDebugInfo>]) -> Vec<LocationSch
...
-                schedule.push(LocationScheduleEntry {
-                    offset: end,
-                    var_index,
-                    storage: None,
-                });
+                // Only schedule a kill when the lifetime actually has a gap here: when another
+                // location for this variable begins exactly at this offset, the variable is
+                // merely changing location, and the kill would only produce a spurious
+                // dead-then-redeclared flicker in the record stream.
+                let contiguous = supported.iter().any(|other| other.start == end);
```
bitwalker: A syntactically supported next range is not guaranteed to emit a backend location. `WasmGlobal` becomes an operandless `debug_declare`, and `WasmStack` does too when the frontend stack lookup fails; MASM lowering emits no record for either case. Suppressing the kill here therefore leaves the previous location incorrectly active. Preserve the kill unless replacement emission is guaranteed, or emit an explicit kill when the replacement cannot be represented.

frontend/wasm/src/module/function_builder_ext.rs
```diff
@@ -197,12 +195,17 @@ impl<B: ?Sized + Builder> FunctionBuilderExt<'_, B> {
             return;
         }
 
-        if let Some((file_symbol, _directory, line, column)) = self.span_to_location(span) {
+        // Only fill in the variable's declaration location when DWARF did not provide one; in
+        // particular the file must never be overwritten with the current instruction's file, or
+        // records emitted while translating inlined code would corrupt the variable's identity
+        // (e.g. a kill for `a` declared in the user's source being attributed to
+        // `core/src/iter/range.rs`).
+        if let Some((file_symbol, _directory, line, column)) = self.span_to_location(span)
```
bitwalker: This fills the declaration location only on a cloned `Variable` attribute. The stored `info.locals[idx].attr` remains unresolved, so later assignments or range events clone it again and may choose a different file/line from their current instruction, especially across inline frames. Persist the first fallback into the stored local metadata, or populate unresolved locals when the subprogram fallback is established.

dialects/hir/src/transforms/local2reg.rs
```diff
@@ -1,20 +1,78 @@
...
+            if matches!(
+                expr.as_value().operations.as_slice(),
+                [ExpressionOp::WasmLocal(index)] if *index == local_index
+            ) {
+                declares.push(op);
+            }
+        }
+    });
+
+    let [declare_op] = declares.as_slice() else {
```
bitwalker: Returning here is not conservative because the callers still erase the local stores. Multiple matching declarations, or a declaration with a transformed expression such as `[WasmLocal(index), DW_OP_deref]`, remain pointed at an unwritten/stale slot. Please skip promotion/store deletion when every affected declaration cannot be safely converted, or explicitly rewrite/kill all of them.

hir-transform/src/dce.rs
```diff
@@ -73,7 +73,14 @@ impl DeadCodeElimination {
...
         let mut changed = PostPassStatus::Unchanged;
         op.raw_postwalk_all::<Backward, _>(|op: OperationRef| {
-            let is_dead = op.borrow().is_trivially_dead();
+            // Transparent ops (e.g. debug info) have no results and no memory effects, so they
+            // always appear trivially dead; erasing them here would silently strip all debug
+            // info. They are only removed when their operand's producer is dead, which is the
+            // job of the region simplification DCE.
+            let is_dead = {
+                let op = op.borrow();
+                !op.implements::<dyn midenc_hir::traits::Transparent>() && op.is_trivially_dead()
```
bitwalker: Preserving the transparent op makes an otherwise-dead producer appear used, so standalone DCE now retains runtime computations whose only use is `di.debug_value` (the new lit test explicitly retains such an `arith.sub`). Debug metadata should not affect generated code. Use transparent-aware liveness and salvage/kill the debug value while still eliminating the dead producer.

sdk/base-macros/src/fpi.rs
```diff
@@ -946,14 +946,24 @@ fn find_dependency_package_in_dir(
...
                 .and_then(|value| value.to_str())
-                .is_some_and(|file_stem| file_stem == stem)
-        }) {
-            return Ok(Some(package.clone()));
-        }
+                .is_some_and(|file_stem| package_stems.iter().any(|stem| stem == file_stem))
+        })
+        .collect::<Vec<_>>();
+    candidates
+        .sort_by_key(|path| std::cmp::Reverse(fs::metadata(path).and_then(|m| m.modified()).ok()));
```
bitwalker: Modification time is not package identity. Accepted dependency aliases and the canonical artifact can coexist; in the existing `component-macros-note` fixture this can select stale `component_macros.masp` over canonical `component_macros_account.masp`, including on equal timestamps due tie ordering. Resolve `[package].name` from `miden-project.toml` as the deterministic primary filename and use legacy aliases only as fallback.

midenc-compile/src/cargo.rs
```diff
@@ -277,6 +277,55 @@ fn cargo_build(
...
+            output_path.display()
+        ))
+    })?;
+
+    // Temporary files are created with restrictive permissions; restore the
+    // conventional mode for a build artifact
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        let _ = std::fs::set_permissions(&output_path, std::fs::Permissions::from_mode(0o644));
```
bitwalker: Forcing mode `0644` after publication overrides the caller's Unix umask. For example, an artifact that previously became `0600` under `umask 077` is made world-readable here. Configure the tempfile with mode `0666` before creation so the OS applies the umask, and avoid silently discarding permission errors.

hir/src/dialects/debuginfo/attributes/expression.rs
```diff
@@ -268,17 +268,42 @@ impl ExpressionOp {
...
+                    magnitude as i128
+                };
+                *byte_offset = i64::try_from(signed).map_err(|_| {
+                    crate::parse::ParserError::InvalidIntegerLiteral {
+                        span: offset_span,
+                        reason: format!("byte offset '{signed}' is out of range for i64"),
+                    }
+                })?;
+                *global_index = if is_local {
+                    encode_frame_base_local_index(index).unwrap_or(index)
```
bitwalker: Indices with bit 31 set collide with `FRAME_BASE_LOCAL_MARKER`: accepted textual HIR such as `global, 2147483648+0` reprints as `local, 0+0`, and the local fallback here has the same collision. Practical Wasm limits are lower, but the textual grammar accepts full `u32`; reject indices outside the taggable range so parsing is not silently lossy.

bitwalker (changes requested):
Overall this looks good, but there were some issues identified during automated review that I'd like to have you address before we merge this

frontend/wasm/src/module/debug_info.rs
djolertrk: hmm, yeah. I centralized the address-space conversion in `WasmFileInfo::dwarf_offset()` and this should be covered now

frontend/wasm/src/module/debug_info.rs
djolertrk: fixed

frontend/wasm/src/module/function_builder_ext.rs
djolertrk: fixed

dialects/hir/src/transforms/local2reg.rs
djolertrk: fixed, it does make sense, the promotion/store deletion is now skipped whenever all affected debug declarations cannot be **safely** converted

hir-transform/src/dce.rs
djolertrk: yes, fixed

sdk/base-macros/src/fpi.rs
djolertrk: hmm, yeah, addressed

djolertrk:
> Overall this looks good, but there were some issues identified during automated review that I'd like to have you address before we merge this

Thanks!

djolertrk:
Hi @bitwalker. Please let me know if this is ready now. Thanks a lot!

hir/src/ir/parse.rs
```diff
@@ -742,3 +742,31 @@ impl core::fmt::Debug for UnresolvedOperand {
 
 /// A type alias for a vector of [NamedAttribute]
 pub type ParsedAttrs = SmallVec<[NamedAttribute; 1]>;
+
+/// Parse an attribute of type `A` from `source`, for use in unit tests.
+///
+/// This drives the same parser machinery as full IR parsing, positioned directly at the
+/// attribute body, so attribute print/parse round-trips can be tested in isolation — attributes
+/// like `di.subprogram` are only ever attached to op attribute dictionaries that the custom op
```
bitwalker: If we have custom op printers that aren't printing their attribute dictionaries - that's a bug. It's required that all custom printers print that information - it is left up to the custom printer to decide whether certain attributes get special treatment in the printed output, but it is a contract between the printer and parser machinery that all of the elements must be printed.

frontend/wasm/src/module/module_env.rs
```diff
@@ -206,6 +206,20 @@ pub struct WasmFileInfo {
...
 }
 
+impl WasmFileInfo {
+    /// Convert a module-relative file offset into the address space used by this module's DWARF.
+    ///
+    /// Standalone modules use code-section-relative addresses, while DWARF attached to a component
+    /// uses offsets relative to the component containing the embedded module.
+    pub fn dwarf_offset(&self, module_offset: u64) -> u64 {
+        if self.module_base_offset > 0 {
+            self.module_base_offset.saturating_add(module_offset)
```
bitwalker: [P1] `module_offset` is already component-absolute for embedded modules: wasmparser constructs the nested `Parser` with the component range start, so function-body and operator positions already include `module_base_offset`. Adding the base again here double-counts it, causing addr2line lookups and location-schedule events to miss. Please treat the parser offsets as authoritative, and replace the synthetic relative-offset test with a real nested-component case.

dialects/hir/src/transforms/local2reg.rs
```diff
@@ -1,20 +1,102 @@
...
+    rewriter: &mut R,
+    function_op: OperationRef,
+    local: &LocalVariable,
+    stores: &[(OperationRef, ValueRef)],
+) -> bool {
+    let local_index = local.as_usize() as u32;
+    let mut declares = SmallVec::<[OperationRef; 2]>::new();
+    function_op.raw_prewalk_all::<Forward, _>(|op: OperationRef| {
+        let operation = op.borrow();
+        if let Some(declare) = operation.downcast_ref::<DebugDeclare>() {
```
bitwalker: [P1] This scan only handles `DebugDeclare`, but the Wasm frontend also emits `DebugValue` with a `WasmLocal(idx)` expression for fallback locals. Promotion then erases the store while leaving that expression pointed at the removed local slot, so MASM lowering reports a location that is never written. Please rewrite matching debug-value references to the promoted SSA value, or preserve the store when they cannot be rewritten safely.

hir-transform/src/dce.rs
```diff
@@ -73,9 +73,30 @@ impl DeadCodeElimination {
...
+                            .iter()
+                            .map(|result| result.borrow().as_value_ref())
+                            .collect::<Vec<_>>(),
+                    )
+                }
+            };
+            if let Some(results) = dead_results {
                 changed = PostPassStatus::Changed;
+                for result in results {
+                    midenc_hir::dialects::debuginfo::transform::erase_debug_info(&result);
```
bitwalker, line 98: [P1] This salvage path only fixes the standalone DCE pass. The production backend currently runs canonicalizer/region simplification while standalone DCE is disabled, and the region DCE implementation directly erases the dead producer and its debug-value user without emitting a kill. That leaves the variable's earlier location active. Please route the production DCE path through equivalent salvage/kill handling.

codegen/masm/src/lower/lowering.rs
```diff
@@ -1554,41 +1554,68 @@ fn debug_var_location_from_expression(
...
-                .map(|pos| emitter.stack.effective_index(pos) as u8)
-                .map(DebugVarLocation::Stack),
+            // Constants only lower to a Const location when they are canonical field elements;
+            // otherwise the felt would silently wrap modulo the field prime and the debugger
+            // would display a different value than the program's. Preserve such constants (and
+            // all negative ones) in serialized form instead.
+            ExpressionOp::ConstU64(val) => Felt::new(*val)
+                .ok()
+                .map(DebugVarLocation::Const)
+                .or_else(|| Some(DebugVarLocation::Expression(expr.to_bytes()))),
```
bitwalker: [P1] Serializing non-canonical or negative constants is not safe with the locked `miden-debug-engine` decoder: it reconstructs these expression operands with `Felt::new(...).expect(...)` (casting signed values to `u64` first), so a valid expression such as `DW_OP_consts(-1)` will panic the debugger. Please avoid emitting this serialized form until the decoder can represent it without panicking, or add a safe end-to-end representation.

hir/src/dialects/debuginfo/attributes/expression.rs
```diff
@@ -268,17 +268,51 @@ impl ExpressionOp {
...
+                        _ => None,
                     })?
                     .into_inner();
                 parser.parse_comma()?;
-                let index = parser.parse_decimal_integer::<u32>()?.into_inner();
-                parser.parse_comma()?;
-                *byte_offset = parser.parse_decimal_integer::<i64>()?.into_inner();
-                *global_index = encode_frame_base_local_index(index).unwrap_or(index);
+                let (index_span, index) = parser.parse_decimal_integer::<u32>()?.into_parts();
+                if index >= FRAME_BASE_LOCAL_MARKER {
```
bitwalker: [P2] Rejecting high-bit indices only avoids one textual round-trip symptom; the underlying representation still uses bit 31 as a local marker, so a valid resolved global/stack-pointer address with that bit set is decoded as an FMP-local offset. This also unnecessarily rejects valid global indices. Please use an explicitly tagged local/global representation instead of overloading a `u32` value bit.

frontend/wasm/src/module/debug_info.rs
```diff
@@ -913,8 +930,49 @@ fn resolve_decl_file<R: gimli::Reader<Offset = usize>>(
...
     let raw = dwarf.attr_string(unit, file.path_name()).ok()?;
-    let path = raw.to_string_lossy().ok()?;
-    Some(Symbol::intern(path.as_ref()))
+    let file_name = raw.to_string_lossy().ok()?;
+    let comp_dir = unit
+        .comp_dir
+        .as_ref()
+        .and_then(|raw| raw.to_string_lossy().ok())
+        .map(|path| path.into_owned());
+    let directory = if file.directory_index() == 0 {
```
bitwalker: [P2] Directory index zero is the implicit compilation directory only for DWARF versions 2–4. In DWARF v5 it indexes the first entry in the directory table, so this branch drops a valid directory and builds the wrong source path. Please branch on the line-program version or defer to gimli's version-aware `file.directory(header)` handling.

examples/fibonacci/.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
...
+  "configurations": [
+    {
+      "type": "miden",
+      "request": "launch",
+      "name": "Debug Fibonacci with miden-debug",
+      "runtime": "debugger",
+      "programPath": "${workspaceFolder}/target/miden/debug/fibonacci.masp",
+      "inputsPath": "${workspaceFolder}/inputs.toml",
+      "linkLibraries": [
+        "${workspaceFolder}/../../target/debug/build/miden-core-lib-06b2e87178feb654/out/assets/miden-core.masp"
```
bitwalker: [P2] This embeds a concrete Cargo build-script fingerprint, which changes with the host, toolchain, features, and dependency rebuilds, so a fresh checkout will usually point at a nonexistent sysroot. Please discover the generated asset path dynamically or expose it through a stable project-local path. The same hard-coded fingerprint appears in the Zed launch configurations.

bitwalker (changes requested):
This is just about ready, but automated review found some issues that I was able to validate - see the comments for details

examples/fibonacci/.vscode/launch.json
djolertrk: will remove those files for IDEs for now

frontend/wasm/src/module/debug_info.rs
djolertrk: Wow, nice catch! :D

djolertrk:
@bitwalker done, thank you!

codegen/masm/src/lower/component.rs
```diff
@@ -1435,3 +1415,53 @@ fn patch_debug_var_locals_in_block(
...
+                    base: FrameBase::Local(local_index),
+                    byte_offset,
+                },
+            ] = expression.operations.as_slice()
+            else {
+                return None;
+            };
+            let local_index = i16::try_from(*local_index).ok()?;
+            let local_offset = local_index - aligned_num_locals as i16;
+            let expression = Expression::with_ops(vec![ExpressionOp::ResolvedFrameBase {
```
bitwalker: **[P1] Use a debugger-supported frame-base encoding**

This serializes the new `ResolvedFrameBase` opcode for ordinary frame-local variables, but the current `miden-debug` decoder—including the locked 0.10.1 release—does not recognize its tag. A direct consumer-level reproduction resolves these locations to `None`, making common Rust locals unavailable. Coordinate a decoder-compatible encoding or update the debugger first, and add an end-to-end compatibility test.

codegen/masm/src/lower/lowering.rs
```diff
@@ -1552,43 +1552,68 @@ fn debug_var_location_from_expression(
...
+            ExpressionOp::Deref
+            | ExpressionOp::PlusUConst(_)
+            | ExpressionOp::Minus
+            | ExpressionOp::Plus
+            | ExpressionOp::StackValue
+            | ExpressionOp::Piece(_)
+            | ExpressionOp::BitPiece { .. }
+            | ExpressionOp::Unsupported(_) => None,
         },
         _ => Some(DebugVarLocation::Expression(expr.to_bytes())),
```
bitwalker: **[P1] Validate constants in compound expressions**

The single-operation cases reject negative or non-canonical constants, but compound expressions bypass those checks and are serialized wholesale. For example, `ConstS64(-1)` followed by `PlusUConst(1)` is accepted by the frontend and still panics `miden-debug` during field conversion. Validate every operand before serialization or make the consumer evaluate these expressions safely.

.zed/debug.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+    "label": "Debug Fibonacci with miden-debug",
+    "adapter": "miden",
+    "config": {
+      "request": "launch",
+      "runtime": "debugger",
+      "programPath": "target/miden/debug/fibonacci.masp",
+      "midenDebugPath": "../../../miden-debug/target/debug/miden-debug",
+      "sysroot": "../../target/debug/build/miden-core-lib-06b2e87178feb654/out/assets",
```
bitwalker: **[P2] Remove the remaining Cargo fingerprint**

The IDE cleanup removed the example launch files but left this root Zed configuration with a build-specific `miden-core-lib` fingerprint. The Fibonacci README now also claims checked-in VS Code and Zed configurations use a stable project-local package, although the VS Code configuration was deleted and this remaining configuration does not. Remove or stabilize this file and update the README accordingly.

bitwalker (changes requested):
A few more issues remain before this can be merged, see the comments for details. Can you also rebase on `next`?

djolertrk:
> A few more issues remain before this can be merged, see the comments for details. Can you also rebase on `next`?

Yes, thanks @bitwalker!

bitwalker (approved):

PR #1238, opened 2026-07-03
feat(cargo-miden): print the compiled artifact path on build

### LINKED ISSUE #1237: `cargo miden build` should print the compiled artifact path
To help the user find the .masp file.

greenhat:
Closes #1237

A successful `cargo miden build` gave no indication of where the compiled package was written, leaving users to hunt for the .masp file under `target/miden/<profile>/` on their own.

### DIFF
--- a/tools/cargo-miden/src/main.rs
+++ b/tools/cargo-miden/src/main.rs
@@ -1,4 +1,4 @@
-use anyhow::Ok;
+use cargo_miden::CommandOutput;
 
 fn main() -> anyhow::Result<()> {
     // Initialize logger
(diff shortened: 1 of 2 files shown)


mooori (approved):

tools/cargo-miden/src/main.rs
```diff
@@ -7,9 +7,17 @@ fn main() -> anyhow::Result<()> {
     builder.format_timestamp(None);
     builder.init();
 
-    if let Err(e) = cargo_miden::run(std::env::args()) {
-        eprintln!("{e:?}");
-        std::process::exit(1);
+    match cargo_miden::run(std::env::args()) {
+        Ok(Some(CommandOutput::BuildCommandOutput { output })) => {
+            for artifact_path in output {
+                println!("Compiled {}", artifact_path.display());
```
bitwalker, line 13: Maybe we should print this in the output of `midenc`? It would be useful there as well, but this patch only shows the output path when invoking the compiler via `cargo miden build`

greenhat: Yep. I created #1274 for tracking.

PR #1239, opened 2026-07-03
fix: temporarily suppress UntrustedMastForest error logs from miden-core

### LINKED ISSUE #1211: Errors shown on `cargo miden build`, but they don't break the build
```
[ERROR miden_core::mast::serialization] UntrustedMastForest expected HASHLESS input; supplied artifact includes wire node hashes, and validation will recompute them and require them to match
[ERROR miden_core::mast::serialization] UntrustedMastForest expected STRIPPED input; supplied artifact includes DebugInfo and other optional payloads over the wire
[ERROR miden_core::mast::serialization] UntrustedMastForest expected HASHLESS input; supplied artifact includes wire node hashes, and validation will recompute them and require them to match
[ERROR miden_core::mast::serialization] UntrustedMastForest expected STRIPPED input; supplied artifact includes DebugInfo and other optional payloads over the wire
```

greenhat:
The proper fix to #1211 is to use trusted API
`Package::read_from_bytes_trusted` in the next VM v0.24.

`cargo miden build` and `midenc` print `UntrustedMastForest expected HASHLESS/STRIPPED input` error records emitted by `miden-core` when deserializing Miden packages whose MAST includes node hashes and debug info, as all compiler-produced packages do. The errors are harmless (validation recomputes and checks the hashes) but make every build look broken.

Add a `log::Log` adapter to `midenc-log` that drops exactly these two records (matched by level, target, and full message text) and forwards everything else, and install it in both binaries. `midenc-driver` already takes the logger as `Box<dyn Log>`, so the wrapping happens at the binary entry points only. The suppression is temporary and should be removed once the compiler migrates to the Miden VM v0.24 release and trusted API.

### DIFF
--- a/midenc-log/src/lib.rs
+++ b/midenc-log/src/lib.rs
@@ -315,6 +315,7 @@
 
 pub mod filter;
 mod logger;
+mod suppress;
 mod writer;
 
 pub mod fmt;
@@ -322,6 +323,7 @@ pub mod fmt;
 pub use self::{
     fmt::{Target, TimestampPrecision, WriteStyle},
     logger::*,
+    suppress::SuppressKnownDependencyErrors,
 };
 
 #[doc = include_str!("../README.md")]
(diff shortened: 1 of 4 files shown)


mooori (approved):

bitwalker (approved):
Not a super big fan of putting this in midenc-log itself, but since it's temporary, it's fine

PR #1241, opened 2026-07-05
[3/3] feat(fpi): support payload-carrying variants in typed FPI signatures

### LINKED ISSUE #1177: Support payload-carrying variants (WIT `variant`/`option`/`result`) in FPI signatures
The `variant`/`option`/`result` support was implemented before the FPI.
The #1176 might be relevant here.

## Context

Canonical ABI lowering supports payload-carrying variants end to end: Rust enums,
`Option<T>`, and `Result<T, E>` exported through the SDK component macros become WIT
variants, lower into HIR `EnumType` values, and flow through component imports and
exports with flat discriminant validation and switch-based payload load/store
(see #111).

FPI (foreign procedure invocation) deliberately supports a narrower, felt-only type
set: scalars, felt, records, and C-like enums. Payload-carrying variants are rejected,
so a foreign procedure cannot take or return an `option`, `result`, or payload enum,
even though the same signature works on a regular component boundary.

## Current behavior

Variants over FPI are rejected in two independent places, which must be lifted
together if support is added:

1. **Typed-signature validation** — `validate_fpi_value_type`
   (`frontend/wasm/src/component/lower_imports.rs`) rejects payload-carrying enums in
   FPI parameters and results:
   `typed FPI only supports enums without payload values`.
2. **Indirect-argument load layout** — `CanonicalAbiType::push_flat_layout`
   (`frontend/wasm/src/component/types/mod.rs`) returns
   `CanonicalTypeError::NonCLikeEnum` for payload-carrying variants, because they have
   no static flat load layout: which payload lanes are live depends on the runtime
   discriminant.

Gate 1 fires first for FPI imports; gate 2 is the structural backstop on the indirect
path and also documents *why* a static layout cannot exist.

## What support would require

Each sub-problem already has a solved counterpart in the canonical ABI lowering that
can be reused or mirrored:

- **Direct arguments:** flattened variant lanes (discriminant + joined payload slots,
  all `i32`/`i64`/`felt`) are felt-convertible, so `push_fpi_arg_felts` and the felt
  budget arithmetic largely work as-is. The missing piece is discriminant range
  validation before the value crosses the FPI boundary — mirror
  `canon_abi_utils::validate_flat_variants`, which the direct component wrappers
  already use.
- **Indirect arguments:** loading a variant from the argument tuple requires a
  discriminant switch over the case payload layouts, not a static offset table —
  mirror `canon_abi_utils::load`'s variant handling instead of extending
  `flat_layout`.
- **Results:** `lower_fpi_result_felts` / `push_fpi_result_value` would need variant
  reconstruction from result felts, plus discriminant validation *before* storing
  through the caller's output pointer (the canonical transformed-import path validates
  pre-store for the same reason).
- **SDK macros:** the `#[account(...)]`/foreign-account binding generation must accept
  `variant`/`option`/`result` types in foreign procedure WIT signatures.

## Acceptance criteria

- An FPI procedure can take and return `option<felt>`, a `result`, and a
  payload-carrying variant; invalid discriminants trap via the generated
  unreachable path on both the argument and result directions.
- Both rejection gates above are removed (or reduced to one shared policy point).
- Integration coverage in the mockchain FPI suite for direct and indirect variant
  arguments and variant results, including mixed-width payload lanes (`u64` + `felt`).

## Non-goals

- Lists/strings over FPI (separate pointer-like-type policy).
- Changing the count-based indirect threshold (`has_arg_ptr` must keep mirroring
  wit-bindgen's count rule).

greenhat:
Close #1177 

**This PR is stacked on #1250 and should be merged after it.**

Canonical ABI lowering has supported WIT variant/option/result end to end since #111, but typed FPI rejected payload-carrying enums in two independent places, so a foreign procedure could not take or return an option, result, or payload variant even though the same signature worked on a regular component boundary.


greenhat:
Rebased(ported) and ready.

bitwalker (approved):

Issue #1158, opened 2026-06-04
Implement batch kernel in Rust


greenhat:
The architecture is discussed at https://github.com/0xMiden/protocol/issues/1122#issuecomment-4611369385

The implementation follows the MASM implementation at https://github.com/0xMiden/protocol/pull/2905

Let's see how much we can squeeze it down.

greenhat:
After the first optimization run, the size is down to 125 KB.

For the following scenario (happy path): https://github.com/0xMiden/compiler/blob/f6b4007f093cccd4e5f17c50ca4f7497c4891311/tests/integration/src/end_to_end/batch_kernel.rs?plain=1#L300-L303

The cycle count is at 41095.

/cc @bitwalker @bobbinth

PR #1245, opened 2026-07-05
fix(wasm): bitcast the br_table selector instead of range-checking it

### LINKED ISSUE #1243: `br_table` selector rejected at runtime — `value does not fit in i32`
Discovered in #1242 

Compiled Rust guest programs trap in the Miden VM with

```
assertion failed with error message: value does not fit in i32
```

whenever control flow reaches a Wasm `br_table` whose selector, interpreted as u32, is `> 0x8000_0000`. LLVM routinely produces such selectors on purpose: it normalizes a `match`/switch by subtracting the smallest case value with **wrapping** arithmetic and relies on `br_table`'s unsigned out-of-range → default rule to catch everything else. A selector of `0 - 1 = 0xFFFFFFFF` is therefore a perfectly legal "go to the default arm" value, and the compiler-inserted assert is wrong.

greenhat:
Close #1243

Discovered in #1158 

Wasm defines the `br_table` selector as unsigned, with any out-of-range value branching to the default arm. LLVM relies on this when it normalizes a `switch` by subtracting the smallest case value with wrapping arithmetic, so selectors with the high bit set (e.g. `0 - 1`) are legitimate take-the-default values. The frontend re-typed the selector with a checked I32 -> U32 cast, which codegen lowers to an assertion rejecting exactly those values, trapping with "value does not fit in i32". Ordinary guest code such as a `match` on `core::cmp::Ordering` hit this at runtime.

### DIFF
--- a/tests/integration/src/end_to_end/regressions/mod.rs
+++ b/tests/integration/src/end_to_end/regressions/mod.rs
@@ -2,4 +2,5 @@ mod func_arg_order;
 mod func_arg_same;
 mod invalid_stack_index_16_issue_872;
 mod invalid_stack_index_4_word_1_felt_args;
+mod issue1243;
 mod vec_realloc_copies_data_issue_811;
--- a/frontend/wasm/src/code_translator/mod.rs
+++ b/frontend/wasm/src/code_translator/mod.rs
@@ -1029,12 +1029,11 @@ fn translate_br_table<B: ?Sized + Builder>(
         frame.br_destination()
     };
 
-    let selector = state.pop1();
-    let selector = if selector.borrow().ty().clone() != U32 {
-        builder.cast(selector, U32, span)?
-    } else {
-        selector
-    };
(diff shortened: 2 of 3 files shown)


tests/integration/src/end_to_end/regressions/mod.rs
```diff
@@ -1,3 +1,4 @@
+mod br_table_default_issue_1243;
```
bitwalker: Can we make the convention for these naming the module `issueXXX` with the test given a descriptive name - that erp keep these regressions ordered together in the order they were reported, and will make it easier to filter/select IMO

bitwalker (changes requested):
LGTM - just one nit about the way we organize these regression tests

tests/integration/src/end_to_end/regressions/mod.rs
greenhat: Done.

greenhat:
Thank you! Done. Please do another round.

bitwalker (approved):

PR #1247, opened 2026-07-05
[1/3] fix(hir): make Region::find_common_ancestor return the common region

### LINKED ISSUE #1244: `Region::find_common_ancestor` returns `None` or hangs for multi-operation queries
`Region::find_common_ancestor` in `hir/src/ir/region.rs` is broken for any query with two or more operations:

- returns `None` when all operations lie in the first operation's innermost region, instead of returning that region;
- never terminates when any operation lies outside that region (e.g. operations in sibling `scf.if` regions).

greenhat:
Close #1244 

For queries with two or more operations the function could never return `Some`: the candidate loop `take()`s the region binding before the containment check and `break`s on success, returning the leftover `None`, and it never advances past an operation that is not contained in the current candidate, looping forever whenever an operation lies outside the first operation's innermost region. As a result the greedy pattern-rewrite driver, the sole caller, silently ran multi-operation rewrites unscoped instead of confining them to the common ancestor region.

### DIFF
--- a/hir/src/ir/region.rs
+++ b/hir/src/ir/region.rs
@@ -3,6 +3,8 @@ mod interfaces;
 mod invocation_bounds;
 mod kind;
 mod successor;
+#[cfg(test)]
+mod tests;
 mod transforms;
 
 use alloc::rc::Rc;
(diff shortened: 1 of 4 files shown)


hir/src/ir/region.rs
```diff
@@ -584,31 +586,25 @@ impl Region {
...
-                            unsafe {
-                                remaining_ops.set_unchecked(index, false);
-                            }
-                        }
-                    }
-                    if remaining_ops.not_any() {
-                        break;
+                while let Some(r) = region {
+                    // Is every other op contained in `region`?
+                    if rest.iter().all(|op| r.borrow().find_ancestor_op(*op).is_some()) {
```
bitwalker: This reborrows on every op, rather than borrowing once for the search, which will add a fair amount of unnecessary overhead

bitwalker (changes requested):
LGTM - just one nit

hir/src/ir/region.rs
greenhat: Done.

greenhat:
Thank you! Done. Please do another round.

bitwalker (approved):

PR #1248, opened 2026-07-05
[2/2] feat: embed component WIT in the compiled Miden package

### LINKED ISSUE #345: Use account's Miden package WIT file in a note script project
### Including the WIT binary form in the Miden package

After #365

Assuming the WIT binary form is encoded in the custom section of the Wasm CM file, we need to pass it through the pipeline up to the assembly stage to be included in the Miden package.

I see two options:
1. Pass the WIT binary form as a metadata in the `Session`;
2. Pass the WIT binary form as an `hir::Component/masm::Library`'s attribute;

I lean towards the first option because I don't see a need for the WIT info anywhere else in the pipeline, and we already have the high-level type info from it.

### Using the packaged binary WIT in the dependent project
In a dependent project (note script, etc.) we need to build the WIT textual form from the WIT binary found in the Miden package, write WIT file to disk and feed the path to the `cargo-component` code the Wasm component generation.

### LINKED ISSUE #1298: Explore using `build.rs` in the contract to compile dependencies
Should be done after #1302.

So that macros that load dependency packages find them. Otherwise, they error out, making DX suboptimal (LSP in the note script showing an error because the dependency account package is not compiled, see #1215).

**EDIT**:

> My understanding was that for non-`midenc` orchestrated builds (e.g. LSP-driven), we'd use `build.rs` to trigger a nested build with `midenc`, with an environment variable set that ensures that the `midenc`-driven build of the crate doesn't recurse infinitely. We'd set this environment variable from `midenc`, not the build script. This would ensure that _all_ builds of the crate have `MIDENC_PACKAGE_CACHE` set at proc-macro expansion time, allowing us to completely sidestep the issue of reimplementing dependency management in the proc-macro crate (which we absolutely should not do, but will end up doing if we keep going down this road).

After the implementation, we should clean the package discovery in the SDK macros to always get it from the `MIDENC_PACKAGE_CACHE` and trust the found package file.

greenhat:
Close #345
Close #1298 

**This PR is stacked on #1306 and should be merged after it.**

## Summary

The component WIT generated by `#[component]` is now embedded in a Wasm custom section (`rodata,miden_wit`), carried through the compiler pipeline, and stored as a `wit` section of the compiled `.masp`. The `#[account(...)]`, sibling `#[component(pkg::Iface)]`, `#[note]`, `#[tx_script]`, and `generate!` macros read dependency WIT from the dependency's compiled package instead of from `target/generated-wit/`, so the `wit = "..."` path metadata in `miden-project.toml` is gone and prebuilt `.masp` file dependencies are self-contained.

Previously every consuming project carried `wit` entries pointing into another crate's target directory, needed only because WIT selection and binding resolution read WIT from different places — the trap behind the sibling "package not found" failures (the `augment_missing_sibling_wit` workaround is deleted). Now both read the same single source.

## Details

- **Pipeline**: both out-of-band payloads (account component metadata, WIT) travel in one `PackageSections` carrier struct from the frontend to assembly, where they are attached as package sections — a future payload needs no per-stage threading.
- **Manual authoring**: a bare `miden::generate!()` over a local `wit/` directory embeds the WIT too, provided it is a single self-contained file; anything else is skipped, and consumers get the accurate "does not embed component WIT" error.


greenhat:
Rebased and ready.

bitwalker (commented):
I think we need to retain the `wit` field in `miden-project.toml` - otherwise there will be no way to manually provide WIT for a package that does not contain it. It is an edge case of course, but unfortunately a necessary one to cover at this stage I think.

This only caught my eye because I just implemented WIT embedding in the branch containing the language I'm building, and there I have it so that the embedded WIT is the default, but the `wit` key in `miden-project.toml` is checked as a fallback/escape hatch.

I'm hoping to open the PR that introduces the language this weekend/early next week - just depends on how much time I have, but since linking between Rust-compiled packages and those compiled from this new language is an explicit goal of mine, I'm keeping a close eye on anything that affects how we link/bind packages.

greenhat:
> I think we need to retain the `wit` field in `miden-project.toml` - otherwise there will be no way to manually provide WIT for a package that does not contain it. It is an edge case of course, but unfortunately a necessary one to cover at this stage I think.
> 

Done in 2d7507aa03bd15d2347175744d89e95c640ced04. In the case where the WIT is both present in the Miden package and in the toml file, I raised an error.

greenhat:
Rebased and ready.

greenhat:
Rebased (ported) and ready.

bitwalker:
@greenhat I've followed up on #1300 with how I think we have to proceed with regards to dependency management, perhaps we should make those changes as part of this PR, or make them in a separate PR and rebase this PR on that one? Dependency management is so core to the way the SDK macros work, that I'd rather solve it _first_, than to merge more hacks that sidestep the fundamental problem.

greenhat:
> @greenhat I've followed up on #1300 with how I think we have to proceed with regards to dependency management, perhaps we should make those changes as part of this PR, or make them in a separate PR and rebase this PR on that one? Dependency management is so core to the way the SDK macros work, that I'd rather solve it _first_, than to merge more hacks that sidestep the fundamental problem.

I agree. 
This PR is already quite big. I'll do it in separate PR(s) and rebase this one afterwards.

greenhat:
Closing in favor of #1328.

PR #1250, opened 2026-07-05
[2/3] fix: prune spills by reachability, not dominance

### LINKED ISSUE #1249: Unsound dominance-based spill pruning in `TransformSpills`
Discovered in #1241 

## Summary

`rewrite_spill_pseudo_instructions` (`hir-transform/src/spill.rs`) erases a spill unless it
**dominates** a live reload of its value. But the spills analysis intentionally places spills
**per path** (e.g. one along each arm of a join), so a reload after the join is covered by a set
of spills, none of which individually dominates it. The dominance predicate erases all of them
while the reload survives.

## Expected behavior

A spill may be elided only if it provably **cannot reach** any live reload of its value —
reachability, not dominance, is the criterion (per-arm spills reach the join reload and must be
kept; a dead edge-spill whose value's only reload lies in the sibling arm reaches nothing and
can still be erased).

greenhat:
Close #1249 

The spill transform erased any spill that did not **dominate** a live reload of its value. The spills analysis places spills **per path**, so a reload after a join is covered by a set of per-arm spills, none of which dominates it: all were erased while the reload survived, panicking reload materialization (or, with partial coverage, reading a stale local — a silent miscompile).

Prune by **reachability**: a spill is elided only if no control-flow path from it can reach a live reload of its value, after normalizing nested operations to their ancestors in the innermost common region; unanalyzable cases conservatively keep the spill. Dead edge-spills are still erased.


hir-transform/src/spill.rs
```diff
@@ -935,3 +938,158 @@ fn rewrite_spill_pseudo_instructions(
...
+///
+/// Reachability is evaluated within a single execution of the innermost isolated-from-above
+/// ancestor (e.g. one function invocation). Queries that cross an isolation boundary, or whose
+/// positions cannot otherwise be related (graph regions, no common ancestor region), are
+/// conservatively reachable.
+///
+/// This query is exported from the crate root chiefly so the behavioral tests in
+/// `midenc-dialect-hir` (which have the parser and control-flow dialects needed to build
+/// fixtures) can exercise it directly; the in-crate consumer is spill pruning.
+pub fn op_reaches(from: OperationRef, to: OperationRef) -> bool {
```
bitwalker: Let's make this a static function hanging off `Operation`, since it is a very general query that could be useful in other contexts.

hir-transform/src/spill.rs
```diff
@@ -935,3 +938,158 @@ fn rewrite_spill_pseudo_instructions(
 
     Ok(())
 }
+
+/// Returns true if some control-flow path from `from` may reach `to`.
+///
+/// The result may be conservatively `true` when reachability cannot be reasoned about precisely;
```
bitwalker: The semantics of this query are incorrect:

* The name of this function is `op_reaches`, but that is not in fact the question it answers. Instead, it is more like `op_might_reach`. A more interesting question to ask (and answer), at least for the spills elision optimization, is something more like `op_cannot_reach`, i.e. you want to know whether you can _safely_ remove a spill because none of its corresponding reloads are reachable.
* Reachability cannot be statically proven as claimed by this function. The only way that property can be usefully proven in general is with a real reachability analysis (which this is not), performed in conjunction with the SCCP/DCE analyses (so you can determine what control flow paths are actually reachable at runtime). The cases that _are_ statically provable with this query as written, are the least interesting ones (i.e. the two operations have no conditional control flow between them).
* This function should return a more precise answer, so that the action taken when the query can't prove reachability is left up to the caller. Something like:

```rust
enum Reachability {
    /// Provably unreachable, i.e. no control flow path exists between `a` and `b`
    Impossible,
    /// Provably reachable, i.e. there is at least one control flow path guaranteed to reach from `a` to `b`
    Guaranteed,
    /// Reachability is not proven, but there is at least one control flow path that reaches from `a` to `b`, but full reachability analysis is required to prove whether the path(s) are truly executable
    Maybe,
    /// Cannot be determined without global reachability analysis, because the two ops are in different functions
    MaybeInterprocedurally,
    /// Cannot be determined because control flow between the two ops is not well-defined (i.e. the both belong to a graph-like region, or their common ancestor region is graph-like)
    Indeterminate,
}
```

The spills transform would treat both `MaybeInterprocedurally` and `Indeterminate` as errors, because they represent invalid states for the IR to be in (a reload of a spill must necessarily belong to the same function; and spills/reloads aren't valid in graph-like regions at all). Other users of this function might act on those results differently.

hir-transform/src/spill.rs
```diff
@@ -935,3 +938,158 @@ fn rewrite_spill_pseudo_instructions(
...
+/// conservatively reachable.
+///
+/// This query is exported from the crate root chiefly so the behavioral tests in
+/// `midenc-dialect-hir` (which have the parser and control-flow dialects needed to build
+/// fixtures) can exercise it directly; the in-crate consumer is spill pruning.
+pub fn op_reaches(from: OperationRef, to: OperationRef) -> bool {
+    // Operations in different isolation scopes (e.g. two functions) share no control flow to
+    // walk, so the query cannot be answered; report the conservative `true`.
+    if isolation_scope(from) != isolation_scope(to) {
+        return true;
```
bitwalker: In the case of spills/reloads, this is definitely a hard error kind of situation, or "provably unreachable" - because it is never the case that a spill and a reload in different isolation scopes are related to each other in any way (by definition, a reload can only reference a spill in the same isolation scope).

hir-transform/src/spill.rs
```diff
@@ -935,3 +938,158 @@ fn rewrite_spill_pseudo_instructions(
...
+    }
+
+    // Normalize both operations to the innermost region containing them both: an operation
+    // nested in a sub-region (e.g. structured control flow) is represented by its ancestor
+    // operation in the common region.
+    let Some(common_region) = Region::find_common_ancestor(&[from, to]) else {
+        // Operations without a common ancestor region cannot be reasoned about.
```
bitwalker: IMO, we absolutely can return `false` here. A lack of a common ancestor (at least in the context of spills) effectively makes it impossible for control flow to reach `b` from `a`, or `a` to `b` (since if either was possible, there would necessarily be a common ancestor, i.e. a region/block that is an ancestor of both).

Let me know if I'm missing an obvious case here - but it isn't at all clear to me why we'd be conservative in this case.

I should also note, that at least for spills, it is just a straight up error for a reload/spill pair to lack a common ancestor - all reloads of a value must be _strictly dominated_ by spills of that value. An individual spill might not dominate a given reload of that spilled value (e.g. a reload at a program point where control flow joins may have likely spilled different values along the paths taken to the join), but there cannot be a path to the reload that doesn't go through at least one spill of the reloaded value.

hir-transform/src/spill.rs
```diff
@@ -935,3 +938,158 @@ fn rewrite_spill_pseudo_instructions(
...
+
+    // Both operations normalize to the same ancestor op: either one op encloses the other (and
+    // trivially reaches it), or they sit in different sub-regions of that op, where transfer
+    // between the regions depends on the op's semantics (e.g. it can happen across loop
+    // iterations), so conservatively assume it does.
+    if from_ancestor == to_ancestor {
+        return true;
+    }
+
+    // Ancestors that are themselves isolated from above (e.g. two functions in one module) are
```
bitwalker: This comment is incorrect. The `IsolatedFromAbove` trait means that regions of that op cannot reference SSA values defined "above" that op (including the scope that the op itself appears in). It has nothing to do whatsoever with symbols, and does not imply anything about the region that contains it (i.e. an `IsolatedFromAbove` op can appear in either SSA CFG regions or graph-like regions).

So if the ancestor ops are isolated from above, then by definition spills/reloads cannot be shared between them (because no SSA value that was spilled above them could have been referenced within them). This function doesn't deal with spills/reloads directly, only control flow reachability though.

In any case, since `IsolatedFromAbove` implies nothing that would affect reachability from one op to another, I don't understand the purpose of this (and I suspect it is actually problematic to have this here)

hir-transform/src/spill.rs
```diff
@@ -935,3 +938,158 @@ fn rewrite_spill_pseudo_instructions(
...
+            return true;
+        }
+        current = owner_op.parent_region();
+    }
+    false
+}
+
+/// Returns the nearest proper ancestor of `op` that is isolated from above, i.e. the operation
+/// whose single execution scopes a reachability query involving `op`.
+fn isolation_scope(op: OperationRef) -> Option<OperationRef> {
```
bitwalker: This doesn't have any bearing on reachability between ops - at best you could search upwards until you find the nearest ancestor op whose parent region is a graph-like region (since reasoning about reachability beyond that point is impossible/not meaningful).

For spills, we'd necessarily be looking for the nearest `Function` - but since `Function` is always defined in a graph-like region, it trivially meets the above criteria.

bitwalker (changes requested):
This will need to be rebased unfortunately, since #1247 was merged and this seemingly was branched off that PR

There are a couple of issues with the way reachability is implemented here. Ultimately, what we really want is a proper reachability analysis that works in concert with SCCP/DCE, so we can elide spills even when there are reloads that are reachable from the view of the CFG, but only via paths that aren't actually executable (e.g. `spill(); if false { let v = reload(); foo(v); } else { bar(); }` where the reload can't possibly ever be executed).

That said, a simpler form that we can implement as a heuristic could still be useful, at least as a short-term workaround for our lack of reachability analysis - but I think we need to adjust the implementation here a bit, as noted in my comments.

greenhat:
Thank you for a thorough review! I implemented everything in the last commit. Please do another round.

bitwalker:
@greenhat Can you rebase this on `next`?

hir/src/ir/reachability.rs
```diff
@@ -0,0 +1,213 @@
...
+            // Unreachable per find_common_ancestor's postcondition (the returned region
+            // contains every queried op); kept as a defensive fallback.
+            return Reachability::Maybe;
+        };
+
+        // Both operations normalize to the same ancestor op: either one op encloses the other
+        // (and control entering it can reach the nested position), or they sit in different
+        // sub-regions of that op, where transfer between the regions depends on the op's
+        // semantics (e.g. it can happen across loop iterations).
+        if from_ancestor == to_ancestor {
```
bitwalker, line 144: **[P2] Consult the sibling-region graph**

Operations in mutually exclusive `scf.if` arms normalize to the same owner, but no child-to-sibling path exists unless an enclosing construct can re-execute the `if`. Returning `Maybe` here prevents spill pruning from removing provably dead stores. Consult `RegionBranchOpInterface` and outer re-entry before deciding.

hir/src/ir/reachability.rs
```diff
@@ -0,0 +1,213 @@
...
+        let (Some(from_block), Some(to_block)) =
+            (from_ancestor.borrow().parent(), to_ancestor.borrow().parent())
+        else {
+            return Reachability::Maybe;
+        };
+
+        // Within one block an earlier operation always flows into a later one; this is only a
+        // guarantee when neither position was normalized, since entering a sub-region of an
+        // ancestor op is generally conditional on that op's semantics.
+        if from_block == to_block && from_ancestor.borrow().is_before_in_block(&to_ancestor) {
```
bitwalker, line 158: **[P2] Do not guarantee passage through intervening ops**

Earlier placement in the same block does not ensure control reaches the later operation: an intervening region operation may loop indefinitely, return from the function, or abort. Return `Maybe` unless every intervening operation is proven to return to its parent on all paths.

hir/src/ir/reachability.rs
```diff
@@ -0,0 +1,213 @@
...
+    /// [Reachability::MaybeInterprocedurally] and [Reachability::Indeterminate] respectively,
+    /// leaving the interpretation of such queries to the caller.
+    pub fn reachability(from: OperationRef, to: OperationRef) -> Reachability {
+        // One operation enclosing the other is an intra-procedural relationship no matter where
+        // the encloser resides: control entering the enclosing op can reach the nested
+        // position, and control leaving the nested position flows back through the enclosing
+        // op. This must precede the scope comparison, which would otherwise misclassify
+        // enclosure by an op residing in a graph-like region (e.g. a function op and an op in
+        // its body) as interprocedural.
+        if from.borrow().is_proper_ancestor_of(&to.borrow())
```
bitwalker: **[P2] Preserve graph-region indeterminacy**

This ancestry shortcut runs before region-kind checks, so a module and a function nested in its graph-like body are reported as `Maybe` even though that region defines no control-flow order. Inspect the crossed region chain and return `Indeterminate` when enclosure crosses a graph boundary.

hir-transform/src/spill.rs
```diff
@@ -838,64 +852,75 @@ fn rewrite_inserted_phi_uses(
...
-                (used, dom_ok)
+                rl.reloaded().borrow().is_used()
             };
-            if reload_used && dom_ok {
-                is_used = true;
-                break;
+            if !reload_used {
+                continue;
+            }
+            match Operation::reachability(operation, reload_op) {
```
bitwalker: **[P2] Avoid repeated whole-CFG walks**

Every spill scans every reload, and each matching live pair starts a fresh structural reachability traversal here. This is `O(S × R × (B + E))` and can become cubic. Index live reloads by `ValueRef` and cache or reverse-compute block reachability per value.

bitwalker (changes requested):
This is just about ready I think, but Codex found some issues that escaped my first pass over the changes, and I agree with the conclusions. See the comments for details. **NOTE:** I did the review against a local merge of the latest `next` branch with your PR on top - so bear that in mind, but I don't think it changes anything as far as this review goes

greenhat:
@bitwalker Thank you! Done in 76c18ec7e65b1fbda2e774d0614577627d5e77eb and f8571dfb5e59875078e9b55750313a3086d21160. Please do another round.

bitwalker (approved):

PR #1251, opened 2026-07-06
[1/2] `call_indirect` implementation

### LINKED ISSUE #133: Wasm table support and `call_indirect` op translation
```[tasklist]
### Tasks
- [ ] Translate Wasm module tables;
- [ ] Translate module table imports/exports and table initialization;
- [ ] Translate Wasm `call_indirect` op;
```
See the semantic details in https://github.com/0xPolygonMiden/compiler/issues/32#issuecomment-1970655897

greenhat:
Close #133 

## How it works

- Each *dispatched-through* `funcref` table is lowered to a new module-level `builtin.function_table` op holding one `builtin.function_table_entry` per statically-initialized slot (from the table's active element segments).
- The linker allocates a word-aligned region for each table — one word (16 bytes) per slot — in the page following the globals table, and moves the dynamic heap base past it.
- The component `init` procedure fills each initialized slot with the callee's MAST root using `procref`;
- A call site lowers through the new `hir.exec_indirect` op to a bounds check, slot address computation, and a `dynexec`;

## Design notes

- **Tables are lowered lazily**, on first use by a `call_indirect`: wit-bindgen and rustc routinely emit `funcref` tables (empty, or holding only artifacts like `cabi_realloc`) in modules that never make an indirect call, and lowering those would burden every compiled program with a useless table and its initialization code. With lazy lowering, programs without indirect calls compile byte-identically to before this change.
- **The Wasm-mandated runtime type-signature check is intentionally omitted** (see the known-limitations appendix): safe Rust cannot produce a signature-mismatched call — reaching one requires code that is already undefined behavior — so the check would only add runtime cost to every indirect call. `dynexec` resolves the slot's word as a MAST root digest, so a corrupted or mismatched slot either matches a procedure already compiled into the program or fails execution; code injection is impossible.
- Table-referenced functions are promoted from private to internal visibility, since the assembler rejects cross-module `procref` of private procedures.

## Memory-layout hardening

Reviewing the safety of the new table region surfaced a pre-existing miscompile, fixed in the second commit: zero-initialized statics (`.bss`) occupy linear memory after the data segments without appearing in the wasm module, and the linker allowed only a single page of headroom for them — a larger `.bss` region silently overlapped the compiler-managed memory laid out above it (globals, function tables, heap).

The wasm module's declared minimum memory (previously parsed and discarded) is now recorded as a language-agnostic `reserved_memory` attribute on `builtin.Module`, expressed in bytes so no wasm units leak below the frontend, and the linker uses it as an additional floor for all compiler-managed regions. The one-page heuristic and the 17-page default reservation remain as fallbacks for HIR that carries no attribute (e.g. not produced by the Wasm frontend).


docs/external/src/appendix/known-limitations.md
```diff
@@ -43,79 +43,42 @@ find a better/more natural representation for `Felt` in WebAssembly.
...
+ignored) to a word-aligned region of linear memory holding one word (the MAST root digest of the
+referenced function) per table slot. The
+region is populated at program startup by the component initialization procedure using `procref`,
+and `call_indirect` dispatches through it with `dynexec` (i.e. in the caller's memory context),
+after a bounds check on the table index which traps deterministically with a
+"indirect call: function table index out of bounds" assertion failure.
+
+The following limitations remain:
+
+- The Wasm-mandated runtime type-signature check is intentionally omitted, with no plans to add
```
bitwalker: I disagree with the premise here - there are serious correctness issues with executing a call with a signature mismatch:

* We can't lower the call correctly unless we know the signature/ABI
* Not trapping will allow execution to proceed in potentially very dangerous ways - it would be a gaping huge security hole that anyone could leverage in an attack if the control flow path that leads to it is in any way caller-controlled. At the very least it would make diagnosing signature mismatch issues (e.g. due to a bug in codegen) difficult and annoying; at worst, it could allow a valid program to be used in malicious ways, so long as a caller can cause the valid program to call a caller-controlled function reference.

The question becomes: how do we enforce this? The simple answer is to use Wasm's approach of type tags - every function table reference has an associated type tag, and the runtime checks the type tag before executing the call (which is how it knows how to trap). The actual checking can't be done by `dynexec`/`dyncall` today, because there is no VM-native notion of function tables but we could emit a table that pairs every function digest with a type tag, and _our_ lowering of `dynexec`/`dyncall` can check the tag associated with the function reference against the tag expected by the callsite before the actual invoke occurs. Yes, this incurs some overhead - but it isn't extreme, and it provides safety that is well worth that overhead IMO.

dialects/hir/src/ops/invoke.rs
```diff
@@ -215,19 +218,184 @@ impl InferTypeOpInterface for ExecFpi {
...
+    }
+
+    fn resolve(&self) -> Option<SymbolRef> {
+        None
+    }
+
+    fn resolve_in_symbol_table(&self, _symbols: &dyn SymbolTable) -> Option<SymbolRef> {
+        None
+    }
+}
```
bitwalker, line 458: There is an issue here that allows the results of indirect calls to bypass unconstrained-advice analysis - because these callbacks always return `None`, the analysis today treats the call as external, then [marks unresolved call results clean](https://github.com/0xMiden/compiler/blob/7832fcd37928818a4930799736610b8e2e3f8864/dialects/hir/src/analyses/advice_taint/propagation.rs#L89-L103). So a private table callee can return raw advice without the finding produced for an equivalent direct call. 

We should probably ensure that the analysis either joins the statically known table callees or conservatively taints unknown indirect results.

To be clear, the issue is really with the analysis, but since this PR is implementing indirect calls, we should probably also ensure that the analysis properly handles them.

codegen/masm/src/linker.rs
```diff
@@ -52,6 +62,32 @@ impl LinkInfo {
...
+    /// Returns true if the component requires an `init` procedure to set up linear memory
+    /// (data segments, global variables, or function tables) before execution.
+    pub fn requires_init(&self) -> bool {
+        self.has_globals() || self.has_data_segments() || self.has_function_tables()
+    }
+
+    /// Get the address of the first page boundary past all statically-allocated memory (global
+    /// variables and function tables), or the end of reserved memory if larger; this is where
+    /// the dynamic heap starts when the program is executed.
+    pub fn heap_base(&self) -> u32 {
```
bitwalker: Checked layout still panics near 4GiB. This function remains infallible. A valid 65,535-page memory plus one 16-byte table links successfully, then panics while rounding `0xffff0010` to the next page. I'd compute and store the heap base within the fallible linker path and return `LayoutOverflow`.

It looks like the related global layout expect paths need the same fix.

codegen/masm/src/lower/component.rs
```diff
@@ -516,6 +511,89 @@ impl MasmComponentBuilder<'_> {
...
+                    )));
+                };
+                // This init procedure lives in the root component module and takes the callee's
+                // MAST root with a cross-module `procref`, which the assembler rejects for
+                // private procedures; since that is a codegen constraint, the required
+                // visibility promotion is applied here rather than expected of IR producers
+                {
+                    let mut callee = callee.borrow_mut();
+                    if callee.visibility() == Visibility::Private {
+                        callee.set_visibility(Visibility::Internal);
```
bitwalker: Two things here:

1. There is no `Internal` visibility in MASM at this point - these procedures will be made public
2. Making the procedures public in a non-public module is fine, because those now-public procedures won't be part of the public interface of the resulting package - _but_ we have to explicitly make sure that the module hierarchy is structured that way. If we make the submodule public, then any public procedure of that submodule, or its public children, will become part of the public interface of the package - and that could very well be undesirable.

For Rust-compiled code, we know that we basically end up with the following structure:

* A root module for the component namespace, which carries the `init` procedure, and the component-level exports
* A submodule tree reflecting the structure of modules in that component. These submodule declarations are currently marked `Public`, though we should probably try and preserve the fact that the core Wasm module is private to the component. The fact that the initial visibility of the top-level submodule is `Public` is hardcoded, but should probably be `Private` instead (as public procedures of private submodules are visible to the parent which declares those submodules).

Anyway, I would suggest we try to resolve this such that the symbols are only visible within the package, not part of its public interface.

frontend/wasm/src/module/module_translation_state.rs
```diff
@@ -104,6 +124,101 @@ impl<'a> ModuleTranslationState<'a> {
...
+                 maximum of {MAX_FUNCTION_TABLE_SLOTS}",
+                table.minimum
+            );
+        }
+
+        // An all-`None` image is fine: every dispatch through such a table fails at runtime on
+        // the zero MAST root of a null slot, matching Wasm's uninitialized-element trap
+        let image = collect_table_image(table_index, defined_idx, module, diagnostics)?;
+
+        // The table symbol is internal to the compiler, so use a hygienic generated name; a Wasm
```
bitwalker: This is not hygenic - in order to be hygenic the generated symbol here needs to be uniqued, such that it doesn't conflict with user-defined symbols.  I was able to trigger an assert by defining a function named  `__indirect_function_table_0`.

codegen/masm/src/legalization.rs
```diff
@@ -110,6 +110,21 @@ pub fn populate_masm_legalization_target(target: &mut ConversionTarget) {
         .add_legal_op::<builtin::Function>()
         .add_legal_op::<builtin::GlobalVariable>()
         .add_legal_op::<builtin::Segment>()
+        .add_legal_op::<builtin::FunctionTable>()
```
bitwalker: This will accept IR that later panics:

* `FunctionTable` is marked unconditionally legal, even outside a module where the linker can never discover it
* `ExecIndirect` lacks legality checks for its 15-felt/no-extension constraints.

Both of those should be dynamic legality checks or verifier-enforced.

frontend/wasm/src/module/module_translation_state.rs
```diff
@@ -104,6 +124,101 @@ impl<'a> ModuleTranslationState<'a> {
...
+        module: &Module,
+        span: SourceSpan,
+        diagnostics: &DiagnosticsHandler,
+    ) -> WasmResult<()> {
+        match self.get_direct_func(func_index)? {
+            CallableFunction::Function { function_ref, .. }
+            | CallableFunction::Intrinsic { function_ref, .. } => self
+                .module_builder
+                .append_function_table_entry(table, index, function_ref, span),
+            CallableFunction::Instruction { .. } => {
```
bitwalker, line 273: We're erasing operation-lowered intrinsic stubs before table construction, so we end up rejecting them here for lacking a procedure body. We need to somehow handle taking a reference against a stub that would be lowered to an operation, probably with some kind of synthetic wrapper just to make the function reference valid

bitwalker (changes requested):
This isn't quite ready to merge (beyond the obvious need to rebase on `next`, which will necessarily have conflicts) see my comments for more info.

frontend/wasm/src/module/module_translation_state.rs
greenhat: The erasure ordering is not the root cause. Building tables before pass 1 would not help — the entry's symbol use would dangle after the erase.

On reachability: this arm cannot be hit through the public SDK surface. The raw externs are private and never address-taken; every path from safe Rust to a funcref slot (fn pointer, closure, vtable) goes through a real Rust function (public wrapper), and rustc materializes those even when they are `#[inline(always)]`. I verified with the exact SDK shape (weak extern + `#[inline(always)]` wrapper + separately-compiled unreachable stub):

```wat
(elem (i32.const 1) func $wrapper_add $wrapper_mul) ;; wrappers, never the stub
(func $wrapper_add (param i32 i32) (result i32)
  local.get 0
  local.get 1
  call $intrinsics::felt::add)                      ;; body = call to the stub
```

Direct calls elsewhere still inline fully. The materialized wrapper is an ordinary `Function`-kind callable, lands in the table fine, and its body's call to the stub gets intrinsic-inlined during translation — so rustc already generates exactly the synthetic wrapper for us in every reachable case. The only way to hit the rejection is to hand-declare an extern against the stub symbols (`intrinsics::felt::*`) and take its address directly. 

Given it is unreachable today, I would keep the diagnostic (rejection) as is.

docs/external/src/appendix/known-limitations.md
greenhat: Done in 3c366457ffa86899a6fd8eb810291ad9ad57e951

dialects/hir/src/ops/invoke.rs
greenhat: Done in 55d3abe57484a79070ddedc9bafbb42870e6f1b1

codegen/masm/src/linker.rs
greenhat: Done in a7e00d95ecc9ff6f5e98fdb65a52059a1f1be445

codegen/masm/src/lower/component.rs
greenhat: Done in fff0509c531a118667dc69332c852c42c667bdab

frontend/wasm/src/module/module_translation_state.rs
greenhat: Done in 0b91cf715cf95a338b29cf2e9a76e8e2b12dac3c

codegen/masm/src/legalization.rs
greenhat: Done in c6e7a39fb35e29afdb3d04d262828eba78f070ec

greenhat:
@bitwalker Thank you! I addressed all the comments. Please do another round.

bitwalker (approved):
This looks pretty good now. I've addressed a handful of related issues I found directly so we can get this landed, since it holds up other work of yours that is waiting on it - but largely the focus of those changes was around bolstering verification, and closing various edge cases that was better addressed here than separately.

PR #1252, opened 2026-07-06
[2/2] feat: note constructors

### LINKED ISSUE #786: Note constructor to use in a tx script
`MyNote` as defined in the https://github.com/0xMiden/compiler/issues/698#issuecomment-3357226906 should have a constructor defined in a way that could be called from a tx script when creating an output note.

greenhat:
Closes #786

**This PR is stacked on #1251 and should be merged after it.**

Notes are now compiled as libraries exporting both the note script and its constructors, so a note can be constructed programmatically (e.g. from a transaction script) without the host precomputing the recipient digest.

- Generated `get_entrypoint_root()` (SDK): `#[note]` impl blocks generate an associated method on the note type returning the MAST root of the crate's `#[note_script]` export. A compiler intrinsic lowered to a MASM `procref` of the lifted note-script export, so the digest is resolved by the assembler.
- `#[note_constructor]` (SDK macro): exports the marked methods of a `#[note]` impl block through the note's WIT interface. Other packages declare the note package as a Miden dependency and call the constructor via the ordinary cross-context call to compute the note's recipient. The caller turns the recipient into an output note through an account procedure (e.g. the basic wallet's `create-note`), because `output_note::create` requires the account-component context since protocol 0.16.
- `#[note]` input structs additionally implement `ToFeltRepr` (breaking; see MIGRATION.md), letting constructors serialize the note inputs the recipient commits to; the note's WIT is also written to `target/generated-wit/` for dependents.
- New generic `hir.procedure_root` op (materializes a procedure's MAST root as a value), repointed at the lifted note-script export during component export lifting and verified at codegen — a missed retarget is a compile error, not a wrong digest.
- Since the constructor and the `procref` live in the note package itself, the recipient commits to the standalone package's script root, so note creation and consumption agree by construction.
- The macros validate the exported surface early (snake-case export names, WIT name/parameter collisions, marker attribute forms), and the frontend diagnoses unknown or malformed note intrinsics instead of panicking.
- New `examples/p2id-tx-script` creates a P2ID note from the recipient computed by the new `build-recipient` constructor of `examples/p2id-note`.

Defining the constructor (`examples/p2id-note`):

```rust
#[note]
impl P2idNote {
    /// Computes the recipient digest of a P2ID note targeted at `target`.
    #[note_constructor]
    pub fn build_recipient(target: AccountId, serial_num: Word) -> Recipient {
        let inputs = P2idNote {
            target_account_id: target,
        };
        let note_script_root = P2idNote::get_entrypoint_root();
        note::build_recipient(serial_num, note_script_root, inputs.to_felt_repr())
    }

    #[note_script]
    pub fn script(self, _arg: Word, account: &mut Wallet) { /* ... */ }
}
```

Calling it from a transaction script (`examples/p2id-tx-script`, with `miden:p2id` declared as a Miden dependency):

```rust
use crate::bindings::miden::p2id::miden_p2id;

#[tx_script]
fn run(arg: Word, account: &mut Wallet) {
    // ... load tag, note_type, target, serial_num, and the asset from the advice provider ...
    let recipient = miden_p2id::build_recipient(target, serial_num.into());
    let note_idx = account.create_note(tag, note_type, recipient);
    account.move_asset_to_note(asset, note_idx);
}
```


greenhat:
~~Alternatively, we can expose only the "getting the root of the note entrypoint" (`P2idNote::get_entrypoint_root()`) from the note and construct the note on the tx script side.~~

This will not work since we want to encapsulate how the note's storage is constructed from the constructor arguments in the note rather than exposing it on the caller side.

```rust
#[note]
impl P2idNote {
    #[note_script]
    pub fn script(self, _arg: Word, account: &mut Wallet) { /* ... */ }
}
```

Calling it from a transaction script:

```rust
use crate::bindings::miden::p2id::miden_p2id;

#[tx_script]
fn run(arg: Word, account: &mut Wallet) {
    // ... load tag, note_type, target, serial_num, and the asset from the advice provider ...
    let note_idx = create_note(target, tag, note_type, serial_num.into());
    account.move_asset_to_note(asset, note_idx);
}

/// Builds the note recipient committing to the note script root, serial number, and note inputs.
fn build_recipient(inputs: AccountId, serial_num: Word) -> Recipient {
    let note_script_root = P2idNote::get_entrypoint_root();
    note::build_recipient(serial_num, note_script_root, inputs.to_felt_repr())
}

pub fn create_note(target: AccountId, tag: Tag, note_type: NoteType, serial_num: Word) -> NoteIdx {
    let recipient = build_recipient(target, serial_num);
    output_note::create(tag, note_type, recipient)
}
```

greenhat:
98b3c62396d64474c4df820c7c60ac6b75c1758e implements `let note_script_root = P2idNote::get_entrypoint_root();`

greenhat:
Rebased and ready.

greenhat:
Since `output_note::create` can only be called from an account, we cannot call it from the note constructor. I changed the note constructor to return `Recipient` instead of the created note index. The tx script will call the `account.create_note` (account's wrapper for the `output_note::create`). I updated the PR description example code.

This got me thinking if the `#[note_constructor]` is the right name for the macro. It still fits, but it's a bit misleading. The alternative might be `#[note_recipient]`. 
/cc @bitwalker @bobbinth

bitwalker (approved):

Issue #1209, opened 2026-06-25
Make MASM advice taint analysis usable on the miden-vm core library


huitseeker:
## Problem

[compiler PR 1135](https://github.com/0xMiden/compiler/pull/1135) introduced MASM-to-HIR lifting and `AdviceTaintAnalysis`. We would like to use that pass from `miden-vm` CI, as in [miden-vm PR 3052](https://github.com/0xMiden/miden-vm/pull/3052), rather than keep a separate MASM lint stack.

We can call the compiler APIs and the pass works on small MASM projects. It does not run on `miden-vm`'s core MASM library because disassembly fails before the taint pass gets a HIR world.

## What we tried

We built a small external tool against compiler commit `53d39713160b4d8651dc7979e8eac5ec1754534e`. It calls:

- [`disassemble_project_target_from_path`](https://github.com/0xMiden/compiler/blob/53d39713160b4d8651dc7979e8eac5ec1754534e/frontend/masm/src/lib.rs#L149-L162)
- [`DisassemblerConfig { infer_missing_signatures: true }`](https://github.com/0xMiden/compiler/blob/53d39713160b4d8651dc7979e8eac5ec1754534e/frontend/masm/src/lib.rs#L41-L48)
- `AnalysisManager::get_analysis::<AdviceTaintAnalysis>()`, matching the compiler's own [`analyze` stage](https://github.com/0xMiden/compiler/blob/53d39713160b4d8651dc7979e8eac5ec1754534e/midenc-compile/src/stages/analyze.rs#L40-L50)

[`AdviceTaintAnalysis`](https://github.com/0xMiden/compiler/blob/53d39713160b4d8651dc7979e8eac5ec1754534e/dialects/hir/src/analyses/advice_taint/mod.rs#L38-L89) exposes the public findings/diagnostic surface we need.

This works on a tiny MASM project where `adv_push` flows into `u32wrapping_add`.

Directly passing `crates/lib/core/asm/miden-project.toml` fails earlier, while parsing `mod.masm`, because the core library root uses module-index syntax:

```masm
pub mod collections
pub mod crypto
...
```

As a workaround, we generated a temporary adapter project:

- keep namespace `miden::core`
- write a synthetic `lint_root.masm`
- copy non-index `.masm` files from `crates/lib/core/asm`
- skip files containing `pub mod`
- run `disassemble_project_target_from_path` on the generated manifest

[...]

PR #1263, opened 2026-07-07
Add MASM lint mode for advice taint analysis

huitseeker:
This PR lets the MASM frontend build a *partial* HIR world for lint use. This is part of #1209.

Strict MASM disassembly stops on the first procedure it cannot lift. That blocks advice taint analysis on large MASM projects, including the Miden VM core library. This change adds a lint path that skips procedures the frontend cannot lift yet. It records the procedure path, source span, and reason. It also skips callers that depend on skipped procedures.

This change also adds support for `mod.masm` module index files, relative calls through declared child modules, source dependency metadata, and `exp.u32`.

This PR does not wire this into `midenc --lint`. That should be a follow-up PR, with a clear policy for skipped procedures and partial analysis.

### DIFF
--- a/frontend/masm/src/lib.rs
+++ b/frontend/masm/src/lib.rs
@@ -285,15 +285,6 @@ pub fn disassemble_project_target_for_lint(
     lift::lift_project_target(inputs, &lift::LiftConfig::lint(config), context)
 }
 
-/// Disassemble pre-resolved project target inputs.
-pub fn disassemble_project_target_input(
-    inputs: project::ProjectTargetInput,
-    config: &DisassemblerConfig,
-    context: Rc<Context>,
-) -> Result<DisassembledWorld> {
-    lift::lift_project_target(inputs, &lift::LiftConfig::strict(config), context)
-}
-
 /// Disassemble pre-resolved project target inputs for linting.
 pub fn disassemble_project_target_input_for_lint(
     inputs: project::ProjectTargetInput,
@@ -349,23 +340,6 @@ pub fn disassemble_project_target_with_dependency_graph(
         &context,
     )?;
     lift::lift_project_target(target, &lift::LiftConfig::strict(config), context)
-}
-
-/// Disassemble a target from a manifest and precomputed dependency graph for linting.
-pub fn disassemble_project_target_with_dependency_graph_for_lint(
-    manifest_path: impl AsRef<Path>,
-    target: Option<&str>,
-    dependency_graph: &miden_project::ProjectDependencyGraph,
-    config: &DisassemblerConfig,
-    context: Rc<Context>,
-) -> Result<DisassembledWorld> {
-    let target = project::resolve_project_target_from_manifest_path_with_dependency_graph(
-        manifest_path.as_ref(),
-        target,
-        dependency_graph,
-        &context,
-    )?;
-    lift::lift_project_target(target, &lift::LiftConfig::lint(config), context)
 }
 
 /// Disassemble a parsed MASM AST module into HIR.
--- a/frontend/masm/src/lib.rs
+++ b/frontend/masm/src/lib.rs
@@ -14,7 +14,7 @@ use std::{collections::BTreeMap, path::Path, rc::Rc, sync::Arc};
 use miden_assembly::{ProjectSourceInputs, ast::ModuleKind};
 use miden_assembly_syntax::{
     ast::{self, Module},
-    debuginfo::{SourceLanguage, SourceManager, Uri},
+    debuginfo::{SourceLanguage, SourceManager, SourceSpan, Uri},
     parser::read_modules_from_root,
 };
 use midenc_hir::{Context, FunctionType, Report, Type, dialects::builtin};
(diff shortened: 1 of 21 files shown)


frontend/masm/src/lift.rs
```diff
@@ -388,6 +627,104 @@ impl ModuleRegistry {
...
+                then_blk, else_blk, ..
+            } => {
+                if let Some(unsupported) = first_non_liftable_instruction(then_blk) {
+                    return Some(unsupported);
+                }
+                if let Some(unsupported) = first_non_liftable_instruction(else_blk) {
+                    return Some(unsupported);
+                }
+            }
+            Op::While { body, .. } | Op::DoWhile { body, .. } | Op::Repeat { body, .. } => {
```
bitwalker: This pre-scan treats `Op::DoWhile` as liftable, but signature validation reaches the unconditional `infer_do_while` `todo!()`, and body lifting contains another `todo!()`. Valid MASM panics instead of recording this procedure as skipped. We should reject the construct during pre-scan or implement both paths.

We should also inspect the condition block for `DoWhile`

frontend/masm/src/lift.rs
```diff
@@ -148,13 +179,105 @@ fn lift_modules(
...
+            Block::new(span, Vec::new()),
+        );
+        stub.set_syscall(procedure.is_syscall());
+        if let Some(signature) = signature {
+            stub.set_signature(signature);
+        }
+        **procedure = stub;
+    }
+
+    linker.link(core::iter::empty(), core::iter::empty())?;
```
bitwalker, line 345: This final linker pass rejects cyclic call graphs before `infer_missing_signatures` reaches the new cycle-to-`SkippedProcedure` handling. A recursive component aborts the entire lint world, including unrelated procedures.

I think we may need to add some support to the linker to allow proceeding in the face of certain errors by collecting them and proceeding with linking anyway (i.e. if we encounter recursion, the linker can obviously resolve the symbol references, even if the resulting call graph cannot be assembled to MAST). In such cases the linker could record the cycle as a deferred error, and proceed with linking anyway. The assembler could then consult the linker for non-fatal errors it gathered, and raise them at that point, while the disassembler could raise diagnostics for those without bailing out of disassembly completely.

The alternative is to detect and stub cyclic SCCs before this, but then we're reimplementing stuff the linker already does for us.

frontend/masm/src/lift.rs
```diff
@@ -195,93 +320,179 @@ impl ModuleRegistry {
...
+                                        None => infer::infer_signature(
+                                            gid,
+                                            &p,
+                                            &self.context,
+                                            &self.linker,
+                                            &self.signatures,
+                                        )?,
+                                    };
+
+                                    if config.lint {
```
bitwalker, line 508: Signature validation does not cover lift-only invariants like local index bounds. An out-of-range `loc_loadw_le` passes here and later escapes through `lift_bodies`, aborting the whole partial world. We probably need to make lifting transactional per procedure or validate every lifting precondition before declaration.

frontend/masm/src/lib.rs
```diff
@@ -221,7 +265,33 @@ pub fn disassemble_project_target_with_sources(
...
+/// Disassemble a project target for linting, skipping procedures that cannot be lifted.
+pub fn disassemble_project_target_for_lint(
+    project: &miden_project::Project,
+    target: Option<&str>,
+    sources: Option<ProjectSourceInputs>,
+    config: &DisassemblerConfig,
+    context: Rc<Context>,
+) -> Result<DisassembledWorld> {
+    let inputs = if let Some(sources) = sources {
+        let metadata = project::collect_dependency_metadata(project, &context)?;
```
bitwalker: This collector loads only dependencies declared directly by the root project. A direct source dependency that calls a transitive source dependency is marked skipped, and that skip propagates into valid root code. This should probably accept a resolved dependency graph/full closure as the existing source-based API does.

bitwalker (changes requested):
Looking good! I found a couple of issues, one of which may require a change in `Linker`, but fairly trivial for the most part. I can get this merged as soon as those are fixed (or we have a plan for fixing them), and will ship a new release of the compiler very soon (basically once this and possibly a couple of other pending PRs land).

frontend/masm/src/lift.rs
huitseeker: Fixed, preflight now rejects the whole do-while procedure.

frontend/masm/src/lift.rs
huitseeker: Tracked in 0xmiden/miden-vm#3457.

frontend/masm/src/lift.rs
huitseeker: Tracked in #1304

frontend/masm/src/lib.rs
huitseeker: Fixed, lint disassembly now takes the parsed sources and resolved dependency graph, then reads metadata from the full dependency closure.

bitwalker (approved):
Looks good, thanks! I've added a few commits to address a couple of minor review findings, rather than do another review cycle just for those, but nothing substantial has changed.

PR #1266, opened 2026-07-08
fix(hir): make operation equivalence hashing consistent with equality

### LINKED ISSUE #1257: Flaky test `rust_sdk_cross_ctx_account_and_note_word` (dependency package not found in the resolver)
https://github.com/0xMiden/compiler/actions/runs/28839808895/job/85531538528?pr=1256

```
   thread 'sdk::rust_sdk_cross_ctx_account_and_note_word' (72393) panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/miden-debug-engine-0.8.1/src/exec/executor.rs:121:25:
    Dependency { name: PackageId("cross-ctx-account-word"), kind: AccountComponent, version: Version { major: 0, minor: 1, patch: 0 }, digest: Word([2146274282055488532, 16159124019805010120, 3800582136282902789, 9192702854356379645]) } not found in resolver
    stack backtrace:
```

greenhat:
Closes #1257

**This PR is stacked on #1265 and should be merged after it.** 

## Problem

`rust_sdk_cross_ctx_account_and_note_word` was flaky on CI: the same account project
occasionally compiled to a different MAST digest within one test run, so a dependent note
script recorded a dependency digest that no longer matched the account package loaded into the
executor's resolver. This is the third distinct source of this symptom, after the linear
operand scheduling fix (9d89a4792) and the predecessor discovery order fix (670579f7b).

## Root cause

CSE keys its candidate map with `OpKey`, whose `Eq`/`Hash` are built on the operation
equivalence infrastructure in `midenc-hir`, and the two violated the Hash/Eq contract (`a == b`
must imply `hash(a) == hash(b)`):

- **Eq** compares operands by value address (`core::ptr::addr_eq`), order-insensitively for
  `Commutative` operations.
- **Hash** fed each operand's **fat pointer** to the hasher — the data address **plus the
  vtable pointer**, which can differ for the same value across codegen units — and always in
  program order.

Equal keys therefore usually hashed differently and landed in different buckets. Whether CSE
deduplicated an equivalent pair depended on the two hashes coinciding modulo the current table
capacity — a function of heap allocation addresses: almost always a miss, occasionally a hit.
The rare hit changed the emitted schedule, the MAST, and the package digest. On a quiet machine
the allocator reproduces the same addresses run after run, which is why it only ever reproduced
under CI heap-layout variance.

## Fix (commit 1)

- `DefaultValueHasher` hashes only the value address (`.addr()`), discarding fat-pointer
  metadata so the hash matches the `addr_eq` comparison.
- `hash_with_options` hashes the operands of `Commutative` operations in the same canonical
  sorted order equality compares them in.

Deduplication is now a pure function of the IR rather than of the heap layout: byte-identical
packages across repeated compilations, compilation orderings, and hosts. Cycle counts drop
where previously luck-dependent deduplications now always apply.

## Hardening (commit 2)

- **`test(transform): pin the OpKey Hash/Eq contract against issue #1257`** — two unit tests near
  `OpKey`, both using a hasher that records the written byte stream (so they compare the hash
  *input*, not a lossy `finish()`):
  - `opkey_hash_is_consistent_with_eq` — operations equal under `OpKey` must hash to the same
    stream: a commutative operand swap, and two distinct `add(a, b)` ops with distinct result
    identities (which `OpKey` ignores). A different-operands negative pair keeps the check
    non-vacuous. Catches program-order operand hashing or result identity leaking into the hash.
  - `default_value_hasher_hashes_thin_address_only` — `DefaultValueHasher` must hash the thin
    value address, catching the exact fat-pointer regression of #1257.

  Both fail deterministically when their regression is reintroduced, independent of allocator
  layout, hash-table capacity, or machine noise — unlike an end-to-end determinism run, which
  only catches the bug if a divergent artifact happens to be observed.

### DIFF
--- a/hir-transform/src/cfg_to_scf/transform.rs
+++ b/hir-transform/src/cfg_to_scf/transform.rs
@@ -1299,27 +1299,43 @@ impl<'a> TransformationContext<'a> {
 ///
 /// This means that both operations are of the same kind, have the same amount of operands and types
 /// and the same attributes and properties. The operands themselves don't have to be equivalent.
+///
+/// Operand identity is irrelevant, but operand types are not: [Self::combine_exit] forwards the
+/// operands of every equivalent return-like operation to a single exit block whose arguments are
+/// typed from the first such operation, so keys must only match when the operand types line up.
+/// Hence the [ValueTypeEquivalence]/[ValueTypeHasher] pair below.
+///
+/// [Self::combine_exit]: TransformationContext::combine_exit
+/// [ValueTypeEquivalence]: midenc_hir::equivalence::ValueTypeEquivalence
+/// [ValueTypeHasher]: midenc_hir::equivalence::ValueTypeHasher
 #[derive(Copy, Clone)]
 struct ReturnLikeOpKey(OperationRef);
 impl Eq for ReturnLikeOpKey {}
 impl PartialEq for ReturnLikeOpKey {
     fn eq(&self, other: &Self) -> bool {
-        use midenc_hir::equivalence::{OperationEquivalenceFlags, ignore_value_equivalence};
+        use midenc_hir::equivalence::{OperationEquivalenceFlags, ValueTypeEquivalence};
         let a = self.0.borrow();
         a.is_equivalent_with_options(
             &other.0.borrow(),
             OperationEquivalenceFlags::IGNORE_LOCATIONS,
-            ignore_value_equivalence,
+            ValueTypeEquivalence,
         )
     }
 }
 impl core::hash::Hash for ReturnLikeOpKey {
     fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
-        use midenc_hir::equivalence::{IgnoreValueEquivalenceOperationHasher, OperationHasher};
+        use midenc_hir::equivalence::{
+            IgnoreValueHasher, OperationEquivalenceFlags, ValueTypeHasher,
+        };
 
-        const HASHER: IgnoreValueEquivalenceOperationHasher = IgnoreValueEquivalenceOperationHasher;
-
-        HASHER.hash_operation(&self.0.borrow(), state);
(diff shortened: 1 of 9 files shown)


bitwalker (changes requested):
I think there are some issues with the link between the `ValueHasher` impls, and the implementation of `is_equivalent`/`is_equivalent_with_options` that needs to be addressed as part of this fix:

* We should introduce a new `ValueEquivalence` trait to parallel `ValueHasher`, and provide a blanket impl for all `Fn(&dyn Value, &dyn Value) -> bool`
* Define `DefaultValueEquivalence` to compare values by their pointer address
* Convert the `check_equivalent` argument of `is_equivalent_with_options` to be an impl of `ValueEquivalence`. Replace the call to `check_equivalent` with `value_equivalence.is_equivalent()`. Remove the `a.ty() != b.ty()` check that precedes it - leave that up to the `ValueEquivalence` implementation`.
* Modify `is_equivalent` to pass an instance of `DefaultValueEquivalence` to `is_equivalent_with_options`
* Add a note to both `ValueHasher` and `ValueEquivalence` traits to warn readers that implementations of one must be mirrored by a semantically equivalent implementation of the other, and that hash equality of two values must agree with `is_equivalent`.

This way, going forward, we can more easily determine when equivalence and hashing semantics differ inappropriately, because we would never expect to see mismatched implementations used in the implementation of `Eq`/`Hash` of some type, e.g. `OpKey`.

Whether we need another implementation of `ValueEquivalence` other than `DefaultValueEquivalence` right now depends on whether there are any other uses of `is_equivalent_with_options` that has a common implementation of `check_equivalent` - I don't recall off the top of my head if we do, but that's easy to confirm.

greenhat:
Nice! Done in 7a14f7a0602fe5d1e3c4eba961d2ee34aaa71749

bitwalker (approved):

PR #1242, opened 2026-07-05
test: implement the protocol batch kernel in Rust

greenhat:
Close #1158 

Port the Miden protocol batch kernel (0xMiden/protocol#2905, plus the expiration running-minimum of 0xMiden/protocol#3019) to Rust compiled with the Miden compiler, to exercise the compiler on a realistic kernel-sized program.

The fixture at `tests/fixtures/batch-kernel` mirrors the MASM kernel module-for-module.

PR #1278, opened 2026-07-14
perf: set release `opt-level = 2` (instead of `s`)

greenhat:
Size-optimized LLVM output leaves execution improvements on the table for release guest builds, while the resulting package-size tradeoff depends on the guest.

On the batch-kernel #1242, isolated opt-`s` versus opt-2 measurements showed:

| Workload | Cycle reduction |
| --- | ---: |
| Batch scenario 1 | 13.6% (34,916 → 30,151) |
| Batch note-erasure scenario | 13.5% (30,553 → 26,436) |
| Tampered-preimage rejection | 11.2% (950 → 844) |
| Consume-before-create rejection | 9.8% (15,969 → 14,411) |

The stripped batch-kernel MAST also decreased by 9.9%, from 112,218 to 101,144 bytes. On this branch, the updated integration-network pins likewise show cycle reductions across the counter, P2ID, custom transaction script, and P2IDE workflows.

This is primarily an execution-speed optimization: package-size effects depend on the guest. The four small example packages measured on this branch grow by 6.9–13.7% under opt-2.

### DIFF
--- a/tests/integration/src/end_to_end/examples/basic_wallet_package_sizes.rs
+++ b/tests/integration/src/end_to_end/examples/basic_wallet_package_sizes.rs
@@ -18,7 +18,7 @@ fn basic_wallet_and_p2id() {
     );
     let account_package = account_test.compile_package();
     assert!(account_package.is_library(), "expected library");
-    expect!["12469"].assert_eq(stripped_mast_size_str(&account_package));
+    expect!["13178"].assert_eq(stripped_mast_size_str(&account_package));
     persist_cargo_miden_dependency("../../examples/basic-wallet", account_package.as_ref());
 
     let mut tx_script_test = CompilerTest::rust_source_cargo_miden(
@@ -28,7 +28,7 @@ fn basic_wallet_and_p2id() {
     );
     let tx_script_package = tx_script_test.compile_package();
     assert!(tx_script_package.is_library(), "expected library");
-    expect!["12981"].assert_eq(stripped_mast_size_str(&tx_script_package));
+    expect!["14526"].assert_eq(stripped_mast_size_str(&tx_script_package));
 
     let mut p2id_test = CompilerTest::rust_source_cargo_miden(
         "../../examples/p2id-note",
@@ -37,7 +37,7 @@ fn basic_wallet_and_p2id() {
     );
     let note_package = p2id_test.compile_package();
     assert!(note_package.is_library(), "expected library");
-    expect!["14097"].assert_eq(stripped_mast_size_str(&note_package));
+    expect!["15353"].assert_eq(stripped_mast_size_str(&note_package));
 
     let mut p2ide_test = CompilerTest::rust_source_cargo_miden(
         "../../examples/p2ide-note",
@@ -46,5 +46,5 @@ fn basic_wallet_and_p2id() {
     );
     let p2ide_package = p2ide_test.compile_package();
     assert!(p2ide_package.is_library(), "expected library");
-    expect!["16360"].assert_eq(stripped_mast_size_str(&p2ide_package));
+    expect!["17785"].assert_eq(stripped_mast_size_str(&p2ide_package));
 }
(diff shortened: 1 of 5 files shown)


tests/integration/src/end_to_end/examples/basic_wallet_package_sizes.rs
```diff
@@ -28,7 +28,7 @@ fn basic_wallet_and_p2id() {
     );
     let tx_script_package = tx_script_test.compile_package();
     assert!(tx_script_package.is_library(), "expected library");
-    expect!["14238"].assert_eq(stripped_mast_size_str(&tx_script_package));
+    expect!["16188"].assert_eq(stripped_mast_size_str(&tx_script_package));
```
bitwalker: It looks like these larger programs actually grow considerably larger, while the smaller programs get smaller, but not to the same degree - are we sure that this is on-balance a better default choice? Saving ~500 cycles on a program < 10k cycles is probably not worth losing ~2k+ cycles on programs > 10k cycles - the more expensive the program is to execute, the more important the savings become - but it seems here that only the smaller programs see the benefit (though that may simply be due to the sample set here).

I'd suggest we try to support deriving the optimzation level from the `OptLevel` setting in `Options` - and choose a default `OptLevel` that maps cleanest to what we want the Cargo default to be.

greenhat: > It looks like these larger programs actually grow considerably larger, while the smaller programs get smaller, but not to the same degree - are we sure that this is on-balance a better default choice? Saving ~500 cycles on a program < 10k cycles is probably not worth losing ~2k+ cycles on programs > 10k cycles - the more expensive the program is to execute, the more important the savings become - but it seems here that only the smaller programs see the benefit (though that may simply be due to the sample set here).

### Size

For the programs that we're tracking, their size increased by 7-13%. I'd call them small (12-19 KB). The batch kernel size, on the other hand, **decreased** by 9.9%, from 112,218 to 101,144 bytes, as I noted in the PR description. So, small programs increase in size, but large programs decrease. My take is that at some point in the program size spectrum, the increase flips to a decrease, but we don't have enough data points to confirm.

### Speed

**All** the programs demonstrate cycle count reduction under `opt-level=2`. For the small programs, it's a 3-7% reduction. For the batch kernel (see the PR description), it's a 13% reduction for the happy path scenario.

Overall, I think it's a good tradeoff to pay a 7-13% size increase for small programs and get a speed increase for all programs **and** size reduction for large programs. It's worth adopting it as our `opt-level` override.

> I'd suggest we try to support deriving the optimization level from the `OptLevel` setting in `Options` - and choose a default `OptLevel` that maps cleanest to what we want the Cargo default to be.

IIUC, the only way for the user to set it is via the `--optimize` option. Should we consider also having it in the miden-project.toml? What should we do if the user sets `opt-level` in their Cargo.toml? Hard error? Currently, we're overriding the `opt-level` option set elsewhere. We're not setting the Cargo default.

bitwalker: Yes, that's correct `--optimize=LEVEL` has to be passed to change it, which is fine.

We can add this as a supported `profile` option in `miden-project.toml`. If someone also sets it in their `Cargo.toml` it'll just get ignored (since our override will take precedence). I don't think we need to try and handle it being set there for the time being - if it ends up being an annoyance, we can always pull that metadata from `Cargo.toml` and check if it conflicts, or just document that it has to be set via one of the two methods mentioned above.

For now though, I think just having the compiler wired up to look at the session `opt_level` for this purpose is sufficient - we can make it more prominent/easy to set in the future if desired.

greenhat: Done in 1a605ca09a9154629961d6ef57348a116dc4bea3

bitwalker (approved):

PR #1279, opened 2026-07-14
perf(wasm): lower aligned accesses through element pointers

greenhat:
Discovered in #1242 

`prepare_addr` already proved natural alignment for Wasm accesses annotated with `align >= 2`, but then discarded that fact: it kept the byte address in a byte-space pointer, so MASM lowering divided the address by four again and routed the access through the dynamic-offset `load_sw`/`store_sw` intrinsics even though the byte offset was known to be zero.

The fact is now carried in the pointer type. For naturally aligned accesses of one-element scalars (`felt`, `i32`, `u32`, including `i64.load32_*` / `i64.store32`), a single checked `divmod` converts the effective byte address to an element address and asserts the remainder is zero, and the resulting `ptr<..., element>` lowers to native `mem_load`/`mem_store` via the backend's existing element-pointer fast path. All other accesses keep the byte-space path. Alignment promises were already enforced with a runtime trap before this change; that behavior is unchanged and now documented as part of the `prepare_addr` contract.

## Measurements

Batch-kernel workloads (`i1158-batch-kernel` branch, where this lowering originated):

| Batch workload | Before | After | Change |
| --- | ---: | ---: | ---: |
| Scenario 1 | 40,486 cycles | 34,916 cycles | -13.8% |
| Note-erasure scenario | 35,735 cycles | 30,553 cycles | -14.5% |
| Tampered-preimage rejection | 1,070 cycles | 950 cycles | -11.2% |
| Consume-before-create rejection | 18,725 cycles | 15,969 cycles | -14.7% |
| Stripped batch-kernel MAST | 123,524 bytes | 112,218 bytes | -9.2% |

Integration-network workloads (this branch):

| Workload | Before | After | Change |
| --- | ---: | ---: | ---: |
| Counter note | 9,058 | 8,430 | -6.9% |
| No-auth authentication procedure | 1,807 | 1,726 | -4.5% |
| RPO-Falcon authentication | 73,748 | 73,624 | -0.2% |
| P2ID note | 7,068 | 6,720 | -4.9% |
| Custom transaction script | 6,815 | 6,528 | -4.2% |
| P2IDE recipient claim | 7,357 | 6,984 | -5.1% |
| P2IDE sender reclaim | 7,819 | 7,426 | -5.0% |

Pinned example package sizes (stripped MAST):

| Package | Before | After | Change |
| --- | ---: | ---: | ---: |
| Basic-wallet account | 12,848 bytes | 12,469 bytes | -2.9% |
| Custom transaction script | 14,238 bytes | 12,981 bytes | -8.8% |
| P2ID note | 14,875 bytes | 14,097 bytes | -5.2% |
| P2IDE note | 17,252 bytes | 16,360 bytes | -5.2% |


dialects/wasm/src/mem.rs
```diff
@@ -1,6 +1,14 @@
...
-use midenc_hir::{Builder, Immediate, PointerType, Report, SourceSpan, Type, ValueRef};
+use midenc_hir::{
+    AddressSpace, Builder, Immediate, PointerType, Report, SourceSpan, Type, ValueRef,
+};
+
+/// The base-2 logarithm of the byte alignment of a Miden memory element.
+const ELEMENT_ALIGNMENT_LOG2: u8 = 2;
+
+/// The size of a Miden memory element in the byte-addressable address space.
+const ELEMENT_SIZE_BYTES: u32 = 1 << ELEMENT_ALIGNMENT_LOG2;
```
bitwalker: I think this unnecessarily obscures the value of this constant, and unintentionally expresses a relationship between size and alignment that is purely incidental (i.e. if we changed to a new field with a different alignment, the size wouldn't necessarily be derivable from the alignment - or vice versa).

I'd just write this as `const ELEMENT_SIZE_BYTES: u32 = 4;`

bitwalker (approved):

dialects/wasm/src/mem.rs
greenhat: Done in https://github.com/0xMiden/compiler/pull/1279/commits/5c962c1fdebd5fac2bb1777134ac6f675cc0186d

Issue #870, opened 2026-01-13
One crate for all SDK user-facing types (`Asset`, `AccountId`, etc.)


greenhat:
As the next step after https://github.com/0xMiden/compiler/pull/865

All end-user-facing types (e.g. `Asset`, `AccountId`, etc.) and their utility methods that might be used in the on-chain and in the off-chain (`miden-base` repo) code should live in a separate new minimalistic crate `miden-sdk-types`. So that everyone uses the same type (e.g. `Asset`) both in the on-chain and off-chain code. 

The `miden-sdk-types` crate should become a dependency for the protocol crates in the `miden-base` repo and provide the types (`Asset`, `AccountId`, etc.) to the protocol crates.

@bobbinth @bitwalker

greenhat:
I've tried to implement it but hit a blocker: missing `enum` (Wasm CM `variant`) type support. So I'll implement it as part of #111 first.

greenhat:
I tried to implement it and I found a few issues that made me question this idea:

### 1. Memory layout of the types in `miden-protocol`
Let's look at the `Asset` type as an example. `Asset` type in the `miden-protocol` in-memory representation is very different from the `(Word, Word)` coming out of the tx kernel and currently used in the Miden SDK (as `Asset(Word, Word)`). The lifetime of the value of this type in an average user's code will most likely look like:

1. Get the value from the tx kernel function `(Word, Word)`;
2. Parse full-fledged `Asset` type `miden-protocol`
3. Store/read/pass it around in the user's code;
4. Convert `Asset` into `(Word, Word)`;
5. Pass the `(Word, Word)` to an account or tx kernel function;

Steps 2 and 4 add approximately 70% overhead in the P2ID note code size. Pretty expensive. This overhead might be less prominent in the case of a large contract with a lot of step 3 actions, but I expect the majority of the code to be rather small.

### 2. Protocol types have parts that are not needed in the on-chain code.

For example, `NoteMetadata` in `miden-protocol` has attachments, while tx kernel functions expect `NoteMetadata` without attachments.

### 3. Some `miden-protocol` types cannot be expressed in WIT

For example, arrays cannot be expressed in WIT and should be changed to something else.

### 4. `miden-protocol`'s `AssetAmount(u64)` vs. `AssetAmount(Felt)` in the Miden SDK;

### 5. All private fields need to be public (for WIT-compatible shape)

The `wit-bindgen` generates bindings that access the fields directly.

[...]

PR #1280, opened 2026-07-16
# feat(sdk): typed fungible asset amounts (`AssetAmount`) + other integer semantics (instead of `Felt`)

### LINKED ISSUE #999: Avoid `Felt` in the SDK API where finite field arithmetic is unwanted
Following the discouraging of using `Felt` to avoid finite field arithmetics in the docs repo at https://github.com/0xMiden/miden-docs/issues/177, and the discussions elsewhere.

### Why 

Finite field arithmetic differs from integer arithmetic (wrapping, division). The asset quantity manipulation is one area where it can lead to unwanted results.

### How

Use integer types in the SDK API where we want to avoid the finite field arithmetics. E.g. make the `Asset::amount(&self) -> u64` and an `Asset` constructor for the fungible asset to accept the amount as `u64` with bounds checking.

@bobbinth what do you think? 

/cc @Keinberger

greenhat:
Closes #999.

The SDK does not send fungible asset amounts as raw `Felt` values. The new `AssetAmount` type contains a `Felt`. The canonical value of this `Felt` is not more than `2^63 - 2^31`. This limit is the maximum fungible amount of the protocol. `AssetAmount` has integer operations only. As a result, the field errors from the issue are not possible. These errors are: a wrap at the field modulus, and a division with the multiplicative inverse.

The protocol 0.16 bindings use raw key words and value words for assets. The kernel does not make assets and does not have balance getter functions. As a result, the guest code gets the typed amount at the points where it reads amounts:

- `Asset::amount()` decodes the amount of a fungible asset and returns an `AssetAmount`. If the asset is not fungible, the function panics. Use `Asset::is_fungible()` in scripts that get assets of the two types. The parity of the metadata limb of the vault key shows the asset type, because `Fungible = 0b01` is the only odd composition value. This check is the check with the smallest cost that kernel-encoded assets permit.
- You can use `AssetAmount` in the signatures of exported component methods. The related WIT type is the `asset-amount` core type. You can also use `AssetAmount` in typed account storage: `StorageValue<AssetAmount>` and `StorageMap<K, AssetAmount>`. To make an `AssetAmount`, use `new(u64)`, `From<u8/u16/u32>` (errors are not possible), or `TryFrom<u64/Felt>` (does validation).
- The `+` and `-` operators do a bounds check. If an overflow or an underflow occurs, the operators panic. See the [implementation](https://github.com/0xMiden/compiler/blob/4f0eb814cbfe4eb77cc03ec4d213800f605a39e6/sdk/base-sys/src/bindings/types.rs?plain=1#L244-L290).

```rust
fn vault_amount(&self, asset_key: Word) -> AssetAmount {
    Asset::new(asset_key, active_account::get_asset(asset_key)).amount()
}

let total = account.vault_amount(key) + asset.amount(); // panics above MAX_U64 instead of wrapping
let rest = total - payment;                             // panics on underflow
```

The other kernel surfaces that had counts or index values as felts also change to integer types:

- Kernel counts are `u32` values. This applies to `tx::get_num_input_notes`, `tx::get_num_output_notes`, `active_account::get_num_procedures`, and the `num_assets` and `num_storage_items` fields of the note info structs. Also, `get_procedure_root` gets its index parameter as a `u32`.
- The attachment search functions return an `Option<u32>`. This type replaces the `AttachmentLocation` struct, which had a felt flag. The attachment index parameters are `u32` values.
- Account nonces use the new `Nonce` type. This type is the `nonce` record from the WIT core-types interface. The interface declared this record before, but the SDK did not use it.
- Block heights use the new `BlockNumber` type (the WIT `block-number` record). You can compare block numbers as integers. Use `TryFrom<Felt>` to do validation of the heights that you read from note storage. The `as_u32` function makes sure that the value obeys the `u32` limit. As a result, a WIT record with a bad value cannot cause an incorrect truncation. The p2ide example now does validation of its timelock height and its reclaim height. Block timestamps are `u32` seconds. Expiration deltas are `u16` values. These types agree with the limits in the kernel documentation.

The arithmetic stays in the field domain. Two times the maximum amount is `2^64 - 2^32 = M - 1`. A compile-time assertion keeps this property safe. As a result, the addition of two correct amounts cannot wrap at the field modulus. The checked add and subtract operations only do an assert on one operand and one comparison. One addition uses 194 cycles on the chain. The `u64::checked_add` version is approximately two times slower. Operators that return a `Result` are approximately 2.3 times slower, because the error value causes a return through memory. The operators panic, but the operators in the off-chain protocol crate return a `Result`. This difference is part of the design. Issue #870 is the point where the two APIs can become the same again. A mock-chain test operates the API against live kernel vault data. The test also gets an `AssetAmount` through the WIT component boundary.

### Why the arithmetic does not use `u64::checked_add`

```rust
pub fn add_as_u64(self, other: Self) -> Self {
    let sum = self
        .as_u64()
        .checked_add(other.as_u64())
        .expect("the sum of two asset amounts must fit in a u64");
    assert!(sum <= Self::MAX_U64, "asset amount addition overflow");
    // The bound check above also guarantees the value is below the field modulus.
    Self {
        inner: Felt::new_unchecked(sum),
    }
}
```

The measurement on a mock chain: one addition uses 378 cycles. The felt version uses 194 cycles. As a result, the u64 version is approximately two times slower. The cause is the u64 emulation on the VM, which operates on felts: one `u32split` for each operand, calls to the `miden::core::math::u64::*` library, and 32-bit limb pairs that go through memory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2871,6 +2871,7 @@ name = "miden-base-sys"
 version = "0.13.1"
 dependencies = [
  "miden-field-repr",
+ "miden-protocol",
  "miden-stdlib-sys",
 ]
 
--- a/examples/auth-component-rpo-falcon512/src/lib.rs
+++ b/examples/auth-component-rpo-falcon512/src/lib.rs
@@ -40,7 +40,7 @@ impl AuthComponent for AuthComponentStorage {
         let input_notes_commit = tx::get_input_notes_commitment();
         let output_notes_commit = tx::get_output_notes_commitment();
 
-        let salt = Word::from([felt!(0), felt!(0), ref_block_num, final_nonce.into()]);
+        let salt = Word::from([felt!(0), felt!(0), ref_block_num.into(), final_nonce.into()]);
 
         let tx_summary = [acct_delta_commit, input_notes_commit, output_notes_commit, salt];
         let msg: Word = hash_words(&tx_summary).into();
--- a/examples/auth-component-rpo-falcon512/src/lib.rs
+++ b/examples/auth-component-rpo-falcon512/src/lib.rs
@@ -40,7 +40,7 @@ impl AuthComponent for AuthComponentStorage {
         let input_notes_commit = tx::get_input_notes_commitment();
         let output_notes_commit = tx::get_output_notes_commitment();
 
-        let salt = Word::from([felt!(0), felt!(0), ref_block_num, final_nonce]);
+        let salt = Word::from([felt!(0), felt!(0), ref_block_num, final_nonce.into()]);
 
         let tx_summary = [acct_delta_commit, input_notes_commit, output_notes_commit, salt];
         let msg: Word = hash_words(&tx_summary).into();
--- a/examples/p2ide-note/src/lib.rs
+++ b/examples/p2ide-note/src/lib.rs
@@ -48,8 +48,8 @@ impl P2ideNote {
         // [target_account_id_suffix, target_account_id_prefix, reclaim_height, timelock_height]
         let target_account_id_suffix = inputs[0];
         let target_account_id_prefix = inputs[1];
-        let reclaim_height = inputs[2];
-        let timelock_height = inputs[3];
+        let reclaim_height = BlockNumber::try_from(inputs[2]).unwrap();
+        let timelock_height = BlockNumber::try_from(inputs[3]).unwrap();
 
         // get block number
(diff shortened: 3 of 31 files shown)


greenhat:
Rebased and ported to protocol v0.16 and ready.

bitwalker (approved):

Issue #1285, opened 2026-07-21
Implement `checked_mod` for i64 as an intrinsic (no VM op) to support Rust `i64 % i64` op


greenhat:
See commented-out test in `tests/integration/src/end_to_end/arithmetic.rs`

ygd58:
I have a draft fix for this ready on `ygd58/compiler@fix-i64-checked-mod`:

- New `checked_mod` proc in `codegen/masm/intrinsics/i64.masm`, mirroring `checked_div`'s abs/sign choreography: validate operands, trap on `i64::MIN % -1` (matching Rust's panicking `%` semantics — the same case `checked_div` already traps on for `/`), otherwise compute the magnitude via `::miden::core::math::u64::mod` on the absolute values and apply the **dividend's** sign (not the divisor's) to the result.
- New `checked_mod_i64` in `codegen/masm/src/emit/int64.rs` (parallel to `checked_div_i64`).
- The missing `I64` arm in `checked_mod`'s dispatch in `codegen/masm/src/emit/binary.rs`.
- Uncomments `test_int_op!(rem, %, i64, i64::MIN..=i64::MAX, 1..=i64::MAX)` in `tests/integration/src/end_to_end/arithmetic.rs`.
- Un-ignores the related `i64_srem` differential-fuzzing regression test in `tests/integration/src/end_to_end/differential/tests/signed.rs` (I'd actually filed this independently as #1318 before you flagged it as a dup of this one — same underlying gap, found via the `tools/fuzza-agent/` differential harness rather than by running it).

Scoped deliberately to just this: `checked_rem_i64` (#1000) and `checked_mul_i64` (#1144) are separate Option-returning `checked_*` gaps I haven't touched.

I don't have a working build of this compiler's toolchain in my current environment, so I've hand-traced the new MASM stack choreography against the working `checked_div`/`checked_add` procs in the same file line-by-line to keep it consistent, but haven't compiled/run it myself yet. Could I get assigned to this so I can open a PR and iterate against real CI?

bitwalker:
@ygd58 If you would like to contribute to the compiler, I'm afraid we would expect you to have the correct toolchain installed locally. Our CI runs a lot of expensive checks that make relying on it for feedback a very poor substitute for iterating locally, and furthermore, I'm not inclined to let someone's agents burn a ton of our CI throughput on things that can be found locally first.

[...]

PR #1286, opened 2026-07-21
test(fuzza): grow differential coverage across six compiler areas

greenhat:
This PR runs the differential fuzzing harness through three coverage-guided campaigns spanning twelve compiler areas, adding 49 differential cases (plus pinned deterministic reproducer and edge-grid twins) for a corpus of 82 cases / 99 tests. Every area was closed with a written unreachability analysis of its remaining cold regions rather than by grinding zero-yield cases.

Beyond the classic area coverage (control flow, arith canonicalization & constant folding, call boundaries, operand scheduling & spills, unsigned/widening integer emitters, signed arithmetic), the later campaigns shift from compile-side coverage to runtime assertion:

- First runtime execution of the entire 128-bit compiler-builtins surface — `__udivti3`/`__umodti3`/`__divti3`/`__modti3`, the dynamic shift libcalls (`__ashlti3`/`__lshrti3`/`__ashrti3`), and u128 popcount/clz/ctz limb legalizations — with zero divergences, doubling as a load test of the u64-division advice-event handling.
- Boundary-value semantics pinned via explicit input grids (`run_case_with_inputs` companion tests): shift-count masking through over-double-width counts, division at divisor 1 / equality / greater-than-dividend, `MIN`/`MAX` wrapping, `checked_shr` on `MIN`, bit-count saturation at zero, zero-length memory ops, zero- and one-trip loops, and sub-word sign boundaries — all matching native Rust semantics exactly.
- Scale robustness: a ~400-op single block, a 64-target `br_table`, 12-level nesting, a 21-frame VM call stack, and 24 data segments all compile in about a second and pass differentially; the operand scheduler is shown to have no reachable fallback or failure arm.

## Bugs found

Each is committed as an `#[ignore]`d reproducer whose doc comment and ignore reason carry the details (exact inputs, bounding, un-ignore condition):

- `sext_shapes` (#1288): silent wrong-result miscompilation in a signed widening mix (`extend*_s` + `i64.mul_wide_s` with a negative-constant multiplicand); passing sibling cases bound the suspect surface.
- `spill_edge` (#1289): compile-time panic in the second spill transform when a spilled value crosses a control-flow edge as a successor argument.
- `i64_srem` (#1285): safe Rust `i64 %` with a dynamic divisor panics the compiler (`checked_mod` has no I64 arm).
- `u64_udiv` / `i64_sdiv`: any executed 64-bit division aborted the differential executor on the miden-core-lib `U64_DIV_EVENT` advice event — fixed in this PR by registering the core-library event handlers in `executor_with_std`; with the handlers in place both tests pass, confirming 64-bit division codegen is correct.

## Knowledge base & tooling

`tools/fuzza-agent/` gains a committed `KNOWLEDGE.md` fact base (verified reachability facts, case-writing tricks, operational gotchas — replacing machine-local scratch as the durable store), an `OUTER-LOOP.md` playbook for running multi-area campaigns, a `cargo make fuzza-probe` task for cheap pre-coverage reachability probes, and matching `AGENT-PROMPT.md`/`README.md` updates (pinned repro twins, compiler-panics-as-findings, probe-first workflow). Known-bug specifics deliberately live only at the ignored tests, so fixes clean up in one place.

The later campaigns extended `KNOWLEDGE.md` with the load-bearing closure arguments (rewrite passes run function-scoped so their region/symbol arms are unreachable; the scheduler has no fallback path; quad-word memory ops have no wasm producer), the float-surface verdicts (every float compute operator fails with a clean unsupported-op diagnostic), and a standing out-of-scope section (linker-stub and float-bit-transport surfaces) so future coverage runs treat those as closed rather than cold.

The differential test registry is split into nine thematic modules under `differential/tests/` — bug reproducers now sit beside the sibling cases that bound them, and `_repro`/`_edges` companions directly below their base cases.


greenhat:
Rebased and ready

bitwalker (approved):

PR #1287, opened 2026-07-21
test(arithmetic): expand integer mul div and rem coverage

greenhat:
Integer multiplication, division, and remainder lacked end-to-end coverage across the fixed-width integer types. Adding signed narrow cases also exposed that the shared test harness encoded i8 and i16 arguments without the Wasm-required i32 sign extension.

Add the missing property tests and route integer operands through the Wasm ABI stack encoder while preserving boolean encoding. Keep signed i64 remainder disabled pending issue #1285 because the VM has no signed i64 remainder operation.

### DIFF
--- a/tests/integration/src/end_to_end/arithmetic.rs
+++ b/tests/integration/src/end_to_end/arithmetic.rs
@@ -14,12 +14,23 @@ use crate::{
     testing::{eval_package, run_masm_vs_rust},
 };
 
+macro_rules! push_wasm_test_arg {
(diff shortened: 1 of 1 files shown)


greenhat:
Rebased and ready.

mooori (approved):
LGTM!

Ranges are chosen to avoid overflow as `test_int_op!` and related macros only handle the happy path. It seems like we don't yet have tests which verify plain (i.e. not checked/overflowing) arith ops behave as expected in the VM when the result overflows?

Issue #1275, opened 2026-07-10
Use `@account_procedure` attribute in the account component interface


greenhat:
The issue in the protocol repo - https://github.com/0xMiden/protocol/issues/3013

greenhat:
Done in #1293

Issue #1190, opened 2026-06-17
Compile transaction script to a library with the `@transaction_script` attribute


greenhat:
https://github.com/0xMiden/protocol/issues/3093

greenhat:
Done in #1293

PR #1292, opened 2026-07-22
feat(sdk): migrate SDK and examples to protocol 0.16

greenhat:
#1190 
#1275

This PR adapts the SDK, frontend, and examples on top of the v0.25 migration branch to the protocol 0.16 transaction-kernel API, pinned at `=0.16.0-alpha.4` (the kernel surface is identical through `0.16.0-beta.1`, but `miden-client 0.16.0-alpha.1` is not yet source-compatible with the beta crates, so the pin follows the client).

The bindings are re-synced to the reshaped kernel across all five layers (externs, stubs, frontend signatures, transform strategies, compile tests): the note asset readers become creation-time readers (`get_assets` -> `get_initial_assets`, `get_assets_info` -> `get_initial_assets_info`), `has_non_fungible_asset` generalizes to `has_asset` by asset id, the `get_initial_*` account-state readers move from `active_account` to `native_account`, and the removed kernel APIs (`get_balance`/`get_initial_balance`, `asset`/`faucet` `create_*`, `has_callbacks`) are dropped together with the emptied `asset` binding module. Breaking changes are recorded in the SDK CHANGELOG with instructions in MIGRATION.md.

Protocol 0.16 recognizes account-interface procedures only via export attributes, so component methods are now explicitly marked with a new `#[account_procedure]` attribute (mirroring `#[auth_script]`, the two being mutually exclusive per component); unmarked methods remain ordinary exports invisible to account-code construction and FPI. Transaction-script entrypoints are tagged `@transaction_script` so `TransactionScript::from_library` can locate them, and since `output_note::create` is now restricted to account-component context, the basic-wallet example exposes a `create_note` component method that the tx-script example calls instead.

Test harnesses are adapted to the new protocol APIs (`NoteScript::from_package` retained on alpha.4, `TransactionScript::from_library` for tx scripts), and cycle/package-size expectations that legitimately moved with the new kernel are refreshed.

Known-failing tests left out of scope: the upstream debug-info items on the VM branch (`write_attachment` linker ambiguity, package-bytes determinism, debug-variable validation, source-location resolution) and the pre-existing `frontend-masm` migration cluster.


greenhat:
@bitwalker I'm finished with this branch and it's ready to review/merge into your branch.

bitwalker (approved):
Thanks!

tests/integration/src/end_to_end/support.rs
```diff
@@ -73,9 +73,9 @@ end
 /// The core library registers the event handlers required to execute core helpers that rely on
 /// the advice provider.
 pub(super) fn default_host_with_core_lib() -> DefaultHost {
-    let core_library = CoreLibrary::default().package();
+    let core_library = CoreLibrary::default();
     let mut host = DefaultHost::default();
-    host.load_library(core_library).expect("failed to load core library into host");
+    host.load_library(&core_library).expect("failed to load core library into host");
```
bitwalker, line 78: We need to create a `HostLibrary` from the `Package` and then set the `handlers` manually - otherwise we don't get any debug info for the library - I'll fix this locally though

PR #1293, opened 2026-07-24
migrate to v0.25 (pre-v0.16 testnet)

bitwalker:
This PR migrates us to the latest VM and protocol, leaving some refactoring/cleanup to be done in following PRs.

This currently builds against branches of `miden-vm` and `miden-debug`, pending release of `miden-vm` v0.25.8 with a slew of critical fixes.

NOTE: Our compiler pipeline is in a bit of a messy state right now, due to a combination of factors, and further refactoring is required to get it back to a presentable/maintainable state. Most notably, the new source provider functionality of the project assembler inverts control, such that the assembler invokes _us_, rather than the other way around. Without doing so, it would be up to us to essentially implement our own project assembler. The reason for this is that the project assembler handles a bunch of details around package dependency resolution, source provenance, etc., that I'd really prefer not to reimplement in the compiler if it can be helped.

To handle this, while still preserving all of the compiler features we support (such as running only parts of the pipeline), we'll need to essentially wrap the project assembler invocation in a `catch_unwind` handler that then checks the error type to handle the various forms of early exit we support. We'd then invoke our usual compiler stages from _within_ the callbacks of the `SourceProvider` trait, selectively, so that we can still break things up into stages and interrupt execution.

NOTE: _All_ project configuration goes in `miden-project.toml` now - `Cargo.toml` only holds the bare minimum project metadata + whatever build configuration it needs; all other metadata is in `miden-project.toml` (e.g. `project-kind`).


greenhat (approved):
Looking good!

djolertrk (approved):
Looks also good for me, even though I have not gone very deeply into it, but I did try to use it for some new debug info features we are about to add!

djolertrk:
btw, If we are about to use the consolidated debug info PR from miden-vm, we will need this https://github.com/0xMiden/compiler/pull/1295/changes/8fbc4efe2a8724c59a6d7b1296a6fb3536e20ddb

PR #1296, opened 2026-07-31
# feat(sdk): typed transaction script arguments for `#[tx_script]`

### LINKED ISSUE #1291: `#[tx_script]` redesign
Currently, the tx script is just a function marked with the `#[tx_script]` attribute.

I suggest we redesign it to be the following:

```rust
#[tx_script]
struct MyTxScript {
    tag: Tag,
    note_type: NoteType,
    recipient: Recipient,
    asset: Asset,
}


#[tx_script]
impl MyTxScript {
    #[tx_script]
    pub fn script(self, account: &mut Wallet) {
        let note_idx = account.create_note(self.tag.into(), self.note_type.into(), self.recipient.into());
        account.move_asset_to_note(self.asset, note_idx);
    }
}
```

Where the `script` is called from a generated `raw_script` wrapper:

```rust
fn raw_script(arg: Word, account: &mut Wallet) {
    let num_felts = adv_push_mapvaln(arg);
    let num_felts_u64 = num_felts.as_canonical_u64();
    assert_eq(Felt::from_u32((num_felts_u64 % 4) as u32), felt!(0));
    let num_words = Felt::new(num_felts_u64 / 4).unwrap();
    let commitment = arg;
    let input = adv_load_preimage(num_words, commitment);
    let tx_script = MyTxScript::read_from_felts(input).unwrap();
    tx_scrip.script(account);
}
```

/cc @bobbinth

greenhat:
Closes #1291

Implements option B of #1291: a transaction script's inputs are a plain data struct passed to the entrypoint by value and unpacked automatically, replacing the raw `Word` argument that scripts had to decode by hand.

```rust
#[derive(FromFeltRepr, ToFeltRepr)]
pub struct TxScriptArgs {
    pub tag: Tag,
    pub note_type: NoteType,
    pub recipient: Recipient,
    pub asset: Asset,
}

#[tx_script]
fn run(args: TxScriptArgs, account: &mut Wallet) {
    let note_idx = account.create_note(args.tag, args.note_type, args.recipient);
    account.move_asset_to_note(args.asset, note_idx);
}
```

The entrypoint is a free function of any name taking a required by-value script-args parameter and an optional `#[account(...)]` reference, in either order. Existing `fn run(arg: Word, ...)` entrypoints keep compiling and behave unchanged, since `Word` decodes as itself.

Transport lives in the `ScriptArgs` trait, implemented for every `FromFeltRepr + ToFeltRepr` type via a blanket impl, with the mode selected at compile time from the new `FromFeltRepr::FIXED_LEN` constant: encodings of at most 4 felts are packed directly into the `TX_SCRIPT_ARGS` word (unused felts are zero and asserted so), while longer or variable-length encodings travel through the advice provider keyed by their Poseidon2 hash, which `adv_load_preimage` verifies in-VM so the host cannot substitute values. Making the mode a compile-time property of the type guarantees the host-side encoder and guest-side decoder cannot disagree on how the args word is interpreted.

`ScriptArgs::decode` is fallible (`ScriptArgsError`): the generated entrypoint wrapper panics on decode errors, failing the transaction, while off-chain code handles them as values. `ScriptArgs::encode` returns `EncodedScriptArgs::{Word, Preimage}` and the caller computes the commitment hash, keeping crypto dependencies out of the SDK.

The trait lives in a new `miden-tx-script-args` crate (re-exported from `miden`) whose off-chain dependency cone is only `miden-field` and `miden-field-repr` — the guest-side transport is a wasm-target-scoped dependency, so hosts encoding script arguments never build the on-chain SDK bindings.


bitwalker:
Sadly this has to be rebased again. I really think we should remove the `Cargo.lock` from these directories - not all of our fixtures have them, and it causes endless rebasing for no reason. A lot of pain, not much gain.

greenhat:
> I really think we should remove the `Cargo.lock` from these directories - not all of our fixtures have them, and it causes endless rebasing for no reason. A lot of pain, not much gain.

We tried removing them before, but I had to add them back due to the protocol crates drift. See https://github.com/0xMiden/compiler/pull/1126

greenhat:
@bitwalker Rebased and ready

greenhat:
Rebased and ready.

sdk/tx-script-args/src/lib.rs
```diff
@@ -0,0 +1,410 @@
...
+fn decode_commitment<T: FromFeltRepr>(_arg: Word) -> ScriptArgsResult<T> {
+    Err(ScriptArgsError::AdviceProviderUnavailable)
+}
+
+/// Decodes a value from a commitment-mode preimage, enforcing the canonical zero padding.
+///
+/// This is the pure half of commitment-mode [`ScriptArgs::decode`]: it runs anywhere, so hosts
+/// can round-trip [`EncodedScriptArgs::Preimage`] bytes in tests and tooling without the VM.
+#[inline(always)]
+pub fn decode_preimage<T: FromFeltRepr>(preimage: &[Felt]) -> ScriptArgsResult<T> {
```
bitwalker, line 171: `decode_preimage` is documented as the pure off-chain half of commitment decoding, but it does not enforce the VM path's whole-word length invariant. For example, `decode_preimage::<Vec<Felt>>(&[2, 5, 6])` returns `Ok([5, 6])`, while the guest rejects the same three-felt advice value with `NonWordMultipleLength`. This lets host validation/tooling accept arguments that deterministically fail in-VM. Please enforce `preimage.len() % WORD_FELTS == 0` here and add a native regression test.

bitwalker (changes requested):
This looks good, just needs a rebase, and one issue was found that we probably should address before merging

sdk/tx-script-args/src/lib.rs
greenhat: Thank you! Fixed in 682aa75246f5dc3ce64bb6fc542a0aad0d34cd7d

greenhat:
@bitwalker Rebased and ready.

bitwalker (approved):

PR #1303, opened 2026-08-03
feat(bench): compare example codegen metrics

djolertrk:
Use debugger's flame graph feature to measure compiler's benchmark, on the small compiler examples for now.

So, it adds example benchmarks for VM cycles, package sizes, and SVG flamegraphs. CI compares PR results against origin/next and posts a summary comment.

To run it locally, just use:

```bash
% cargo make bench
```

or, if you want to point to specific cargo-miden:

```bash
% cargo make bench -- --cargo-miden target/debug/cargo-miden
```

It goes after the https://github.com/0xMiden/compiler/pull/1236 PR.


greenhat (commented):
I love the idea! 

I think we can merge it as is. The contract (basic-wallet, p2id note, etc.) benchmarks will cover only the size, and it's a good starting point. Later, we can add scenarios to run the contracts using the mockchain in `benches`.

djolertrk:
> I love the idea!
> 
> I think we can merge it as is. The contract (basic-wallet, p2id note, etc.) benchmarks will cover only the size, and it's a good starting point. Later, we can add scenarios to run the contracts using the mockchain in `benches`.

Thank you! In order to handle the contract execution, we will need to extend the support for profiling/"flame graphs" in debugger as well. I will play with that.

bitwalker (approved):

PR #1306, opened 2026-08-04
[1/2] Unique the filesystem package cache by build inputs

### LINKED ISSUE #1302: `MIDENC_PACKAGE_CACHE` path should be uniqued by the build inputs

> The `MIDENC_PACKAGE_CACHE` path should only be populated by `midenc` itself during a build, and uniqued by the build inputs (this is the part we don't do today).

greenhat:
Close #1302 

The filesystem package cache at `target/miden/packages` persisted across builds and keyed files by package name alone. The SDK proc macros resolve FPI dependency packages from it through `MIDENC_PACKAGE_CACHE` at expansion time, so when build inputs changed (dependency set, versions, pins, or the compiler itself), a leftover `.masp` from an older build could satisfy the lookup and bake stale procedure roots into generated code — failing only at transaction execution, far from the cause.

The cache path is now `target/miden/packages/<fingerprint>/`, where the fingerprint hashes the compiler version and revision, the build-relevant options (including the inherited `RUSTFLAGS` environment), and the project's recursive manifest closure: root and workspace-root manifests, each path dependency's manifests, resolved dependency schemes (so workspace-inherited pins count), and the content of preassembled `.masp` dependencies. The existing single derivation point in `Session` still feeds both consumers — the registry writes and the env var handed to nested cargo builds — so nothing else changes shape.

Rust sources and lockfiles are deliberately not fingerprinted: the fingerprint covers what changes the set and identity of packages, while content-level changes self-heal — every run rewrites each resolved package into the cache before its consumers expand, and the `include_bytes!` reference the macros emit makes cargo re-expand when package contents change at a stable path. The rationale lives in the module docs.

The FPI expansion also records `option_env!("MIDENC_PACKAGE_CACHE")`, so the fingerprinted cache path itself is a Cargo-tracked input of every consumer: when the fingerprint rotates, cargo re-expands the macro against the new cache even if a stale directory survives on disk. Registry creation additionally prunes stale midenc-owned cache state (sibling fingerprint directories and legacy flat `.masp` files whose expansions predate the env tracking) as defense in depth, and each build holds an advisory lock on a `.build-lock` file inside its fingerprint directory so concurrent builds of one project with different inputs cannot delete each other's live cache mid-build; a cache path that is not fingerprint-named is never used to sweep its parent, and failed removals of owned entries warn with their consequence.

The FPI macro diagnostic for the cache-lookup branch now names the searched cache directory and the expected package file names instead of an empty candidate list and a profile-directory hint that branch never consults (SDK changelog entry included).

### DIFF
--- a/midenc-compile/src/pipeline/frontends/rust.rs
+++ b/midenc-compile/src/pipeline/frontends/rust.rs
@@ -1607,9 +1607,11 @@ pub(crate) mod manifest {
     ///
     /// `MIDENC_PACKAGE_CACHE` is the whole reason `filesystem_cache_dir` is threaded down from
     /// the entry point: it tells the nested `midenc` invocations where to publish the packages
-    /// they compile and where to look for the ones their own dependencies already produced.
-    /// Absent a cache directory the variable is *unset* rather than set to an empty path, which
-    /// is what makes the nested build fall back to its own default.
+    /// they compile and where to look for the ones their own dependencies already produced. The
+    /// directory is fingerprinted by the root build's compiler, options, and manifest closure so
+    /// every participant in that build uses the same isolated cache. Absent a cache directory the
+    /// variable is *unset* rather than set to an empty path, which is what makes the nested build
+    /// fall back to its own default.
     ///
     /// Named — rather than left inline where it was — so that this can be asserted without
     /// spawning `cargo -Z build-std` against the SDK.
--- a/midenc-compile/src/pipeline/frontends/rust.rs
+++ b/midenc-compile/src/pipeline/frontends/rust.rs
@@ -1612,20 +1612,27 @@ pub(crate) mod manifest {
     /// variable is *unset* rather than set to an empty path, which is what makes the nested build
     /// fall back to its own default.
     ///
+    /// The composed rust flags are emitted twice: as `RUSTFLAGS` and, authoritatively, as
+    /// `CARGO_ENCODED_RUSTFLAGS`. Cargo prefers the encoded variable, so an inherited value from
+    /// the caller's environment would otherwise silently replace every mandatory Miden flag —
+    /// `--cfg miden`, the target features, the panic strategy. Setting it explicitly makes the
+    /// inherited value inert; the encoding splits on whitespace, which is exactly how cargo
+    /// itself interprets the plain variable.
+    ///
     /// Named — rather than left inline where it was — so that this can be asserted without
     /// spawning `cargo -Z build-std` against the SDK.
     pub(super) fn cargo_env(
         filesystem_cache_dir: Option<&Path>,
         extra_rust_flags: String,
     ) -> Vec<(&'static str, String)> {
+        let encoded_rust_flags =
+            extra_rust_flags.split_whitespace().collect::<Vec<_>>().join("\x1f");
+        let mut env =
+            vec![("RUSTFLAGS", extra_rust_flags), ("CARGO_ENCODED_RUSTFLAGS", encoded_rust_flags)];
         if let Some(filesystem_cache_dir) = filesystem_cache_dir {
-            vec![
-                ("RUSTFLAGS", extra_rust_flags),
-                ("MIDENC_PACKAGE_CACHE", filesystem_cache_dir.to_string_lossy().into_owned()),
-            ]
-        } else {
-            vec![("RUSTFLAGS", extra_rust_flags)]
+            env.push(("MIDENC_PACKAGE_CACHE", filesystem_cache_dir.to_string_lossy().into_owned()));
         }
+        env
     }
 
     /// Returns the Cargo profile value for a compiler optimization level.
@@ -3175,6 +3182,20 @@ path = "lib.rs"
             without.iter().any(|(key, value)| *key == "RUSTFLAGS" && *value == rustflags),
             "and the rust flags are handed over either way: {without:?}"
         );
+    }
+
+    #[test]
+    fn the_encoded_rustflags_override_inherited_values_with_the_same_flags() {
+        let rustflags = String::from("-C target-feature=+bulk-memory --cfg miden");
+
+        let env = manifest::cargo_env(None, rustflags);
+        let (_, encoded) = env
+            .iter()
+            .find(|(key, _)| *key == "CARGO_ENCODED_RUSTFLAGS")
+            .expect("the encoded variable must be set so an inherited value cannot override it");
+        // 0x1f-separated units, one per whitespace-separated flag — cargo's own split rule for
+        // the plain variable.
+        assert_eq!(*encoded, "-C\x1ftarget-feature=+bulk-memory\x1f--cfg\x1fmiden");
     }
 
     /// A WebAssembly module with a body, for the lowering half of the entry point.
--- a/midenc-compile/src/pipeline/frontends/rust.rs
+++ b/midenc-compile/src/pipeline/frontends/rust.rs
@@ -1653,10 +1653,12 @@ pub(crate) mod manifest {
         explicit: Option<&str>,
     ) -> Vec<String> {
         let mut args: Vec<String> = mandatory.iter().map(|flag| flag.to_string()).collect();
-        let encoded =
-            inherited_encoded.and_then(|value| value.to_str()).filter(|value| !value.is_empty());
-        let plain =
-            inherited_plain.and_then(|value| value.to_str()).filter(|value| !value.is_empty());
+        let encoded = inherited_encoded
+            .and_then(|value| value.to_str())
+            .filter(|value| !value.is_empty());
+        let plain = inherited_plain
+            .and_then(|value| value.to_str())
+            .filter(|value| !value.is_empty());
         if let Some(encoded) = encoded {
             args.extend(encoded.split('\x1f').filter(|arg| !arg.is_empty()).map(str::to_string));
         } else if let Some(plain) = plain {
--- a/midenc-compile/src/pipeline/frontends/rust.rs
+++ b/midenc-compile/src/pipeline/frontends/rust.rs
@@ -1563,29 +1563,32 @@ pub(crate) mod manifest {
         let rustup_toolchain = crate::rust::rustup_toolchain();
         let cargo_build_args = build_cargo_args(cargo_opts, compiler_opts.optimize);
 
-        // Enable memcopy and 128-bit arithmetic ops
-        let mut extra_rust_flags = String::from("-C target-feature=+bulk-memory,+wide-arithmetic");
-        // Propagate the Miden VM target signal to the entire crate graph so Cargo can use it for
-        // cfg-based dependency selection.
-        extra_rust_flags.push_str(" --cfg miden");
-        // Enable errors on missing stub functions
-        extra_rust_flags.push_str(" -C link-args=--fatal-warnings");
-        // Remove the source file paths in the data segment for panics
-        // https://doc.rust-lang.org/beta/unstable-book/compiler-flags/location-detail.html
-        extra_rust_flags.push_str(" -Zlocation-detail=none");
-        // Build with panic=immediate-abort
-        extra_rust_flags.push_str(" -Zunstable-options");
-        extra_rust_flags.push_str(" -Cpanic=immediate-abort");
-        if let Ok(inherited) = std::env::var("RUSTFLAGS")
-            && !inherited.is_empty()
-        {
-            extra_rust_flags.push(' ');
-            extra_rust_flags.push_str(&inherited);
-        }
-        if let Some(explicit) = compiler_opts.rustflags.as_deref() {
-            extra_rust_flags.push(' ');
-            extra_rust_flags.push_str(explicit);
-        }
+        let mandatory_rust_flags = [
+            // Enable memcopy and 128-bit arithmetic ops
+            "-C",
+            "target-feature=+bulk-memory,+wide-arithmetic",
+            // Propagate the Miden VM target signal to the entire crate graph so Cargo can use it
+            // for cfg-based dependency selection.
+            "--cfg",
+            "miden",
+            // Enable errors on missing stub functions
+            "-C",
+            "link-args=--fatal-warnings",
+            // Remove the source file paths in the data segment for panics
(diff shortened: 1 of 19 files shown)


midenc-session/src/package_cache.rs
```diff
@@ -0,0 +1,1337 @@
...
+            log::warn!(
+                target: "package-registry",
+                "failed to open filesystem package cache liveness lock '{}': {err}; continuing without a liveness lock",
+                lock_path.display()
+            );
+            return None;
+        }
+    };
+
+    if let Err(err) = lock.lock_shared() {
```
bitwalker: **[P1] Serialize same-fingerprint writers.** This shared lock prevents pruning but permits multiple builders to publish different bytes to the same package filename. Because package-selecting inputs such as search paths are excluded from the fingerprint, one build can read another build’s FPI roots while linking its own in-memory package. Use immutable generations, include every selecting input, or take an exclusive writer lock.

midenc-session/src/package_cache.rs
```diff
@@ -0,0 +1,1337 @@
...
+///
+/// This is the producer half of the owned-layout contract: the path it builds must satisfy
+/// [`is_owned_filesystem_cache_path`] once a fingerprint component is appended, or the locking
+/// and pruning protocol silently degrades to a debug log. `Session::filesystem_package_cache_dir`
+/// derives through here, and its unit test asserts the coupling.
+pub(crate) fn package_cache_parent(project_dir: &Path) -> PathBuf {
+    project_dir.join("target").join("miden").join("packages")
+}
+
+pub(crate) fn is_owned_filesystem_cache_path(filesystem_cache: &Path) -> bool {
```
bitwalker, line 328: **[P1] A suffix does not prove cache ownership.** The existing public custom-cache constructor reaches this check, but any caller path ending in `miden/packages/<16-hex>` passes and enables recursive deletion of sibling fingerprint directories and flat `.masp` files. Require an ownership marker/capability or restrict pruning to internally derived `Session` cache paths.

midenc-compile/src/pipeline/frontends/rust.rs
```diff
@@ -1606,24 +1606,33 @@ pub(crate) mod manifest {
...
+    /// inherited value inert; the encoding splits on whitespace, which is exactly how cargo
+    /// itself interprets the plain variable.
     ///
     /// Named — rather than left inline where it was — so that this can be asserted without
     /// spawning `cargo -Z build-std` against the SDK.
     pub(super) fn cargo_env(
         filesystem_cache_dir: Option<&Path>,
         extra_rust_flags: String,
     ) -> Vec<(&'static str, String)> {
+        let encoded_rust_flags =
```
bitwalker: **[P1] Preserve encoded Cargo flags.** This replaces inherited `CARGO_ENCODED_RUSTFLAGS` with an encoding derived only from plain `RUSTFLAGS`, silently losing legitimate caller flags. Preserve Cargo’s encoded value using its `0x1f` argument boundaries, merge the mandatory and explicit flags, and emit the combined value.

midenc-session/src/package_cache.rs
```diff
@@ -0,0 +1,1337 @@
...
+                    log::debug!(
+                        target: "package-cache",
+                        "cannot resolve workspace path dependency '{}' while fingerprinting outside a workspace",
+                        dependency.name()
+                    );
+                    self.transcript.field("dependency.path", b"unresolved-workspace");
+                }
+            }
+            DependencyVersionScheme::Workspace { member, .. } => {
+                if let Some(manifest_path) = workspace
```
bitwalker: **[P2] Normalize workspace manifest paths.** `miden-project` accepts workspace members referenced as `dep/miden-project.toml`, but `get_member_by_relative_path` compares against the member directory. That valid spelling therefore records `unresolved-workspace` and skips the member’s manifest closure. Normalize manifest-file paths to their parent before lookup.

sdk/base-macros/src/fpi.rs
```diff
@@ -1188,6 +1188,11 @@ pub(crate) fn augment_foreign_account_bindings(
         #(#trait_items)*
         #active_account_item
         #(#package_includes)*
+        // Record the package cache location in the consumer's dep-info. The value carries the
+        // build-input fingerprint, so Cargo re-expands this macro whenever the fingerprint
+        // rotates — even when a stale cache directory survives on disk. The `include_bytes!`
+        // constants above cover content changes at an unchanged path.
+        const _: Option<&str> = option_env!("MIDENC_PACKAGE_CACHE");
```
bitwalker: **[P2] Qualify generated tracking names.** This code is emitted into consumer scope, where a user-defined `Option` or `option_env!` can shadow these names. Qualify the type and macro through `core` so cache tracking remains hygienic, and add a consumer-scope shadowing test.

bitwalker (changes requested):
Looks pretty good, but Codex found a few issues that I think are worth addressing before we merge

greenhat:
@bitwalker Thank you! I addressed all the review notes. Please do another round.

bitwalker (changes requested):
Having some time to look at this PR in more detail, I'm realizing that I missed a rather large problem with this approach. In general, the `package_cache` module and related machinery is almost entirely unnecessary/over-complicated: During compilation, dependencies are compiled before dependents, and as each Rust-based Miden project is compiled, we set an environment variable with a path to a location on disk that contains all of the packages assembled thus far during that session (accomplished by storing them in the session's package registry). The proc macros that require package metadata for expansion are supposed to resolve packages from this path - leaving dependency resolution details/caching/fingerprinting up to the compiler and assembler.

As I understand it, the primary issue was that we were not creating a sufficiently unique directory for each session - thus allowing build artifacts to potentially leak between sessions unintentionally. Ultimately, that issue can be solved much more simply and directly. Ideally, we could cache artifacts built from source - but I'm much less concerned about that than I am about the excessive complexity of the `package_cache.rs` module, and the fact that it reimplements a lot of the build provenance/fingerprinting logic of the assembler itself.

Consulting with Codex on this, to try and make sure I wasn't missing any finer details, it had the following to say about it, and I agree with its assessment/proposal - I'm pasting it here verbatim rather than summarizing it, so I don't unintentionally leave any useful details out. The direct question it is responding to at the beginning is essentially "This could be solved by simply creating an entirely unique directory for every session, correct?" - but that question was part of a larger one that includes more or less what I stated above.

---

Yes—with one important correction: the directory should be unique per top-level compilation graph, not per nested `Session` object.

Some filesystem exchange is necessary because proc macros run in rustc/Cargo processes and need the assembled package bytes. The fingerprinted, persistent “cache” in this PR is not necessary for correctness.

Why the simpler model works:

- The assembler resolves dependencies depth-first.
- Each dependency is fully assembled and published before the dependent Rust project’s Cargo build starts.
- Every nested Cargo invocation receives the root build’s directory through `MIDENC_PACKAGE_CACHE`.
- The proc macro reads packages from that directory and emits both `option_env!("MIDENC_PACKAGE_CACHE")` and `include_bytes!(...)`.
- The existing [environment-only regression test](https://github.com/0xMiden/compiler/tree/8b200345a7f375c7b3cc4ca7dc2abf1a37bab812/tests/integration/src/sdk/mod.rs:737) already proves that changing only the directory path makes Cargo re-expand an unchanged consumer.

Crucially, the current filesystem cache is not a read-through compiler cache. The registry only [writes packages into it](https://github.com/0xMiden/compiler/tree/8b200345a7f375c7b3cc4ca7dc2abf1a37bab812/midenc-session/src/registry.rs:229); a subsequent compiler session does not load source-built packages from that directory to avoid rebuilding them. The fingerprint machinery primarily makes a persistent staging namespace safe to reuse. That is why it has ended up predicting dependency resolution before the actual resolver runs.

I would replace it with a small build-scoped package exchange:

1. Atomically create one globally unique directory at the beginning of the root compilation.
2. Share one lease/path across the root `Session`, all clones, the registry, and every nested dependency build.
3. Publish each completed package there atomically before spawning its dependent’s Cargo build.
4. Keep it alive until the pipeline and all Cargo/rustc/proc-macro children finish.
5. Delete it on normal completion, or leave harmless crash remnants to normal temporary-directory cleanup.

Keep:

- `MIDENC_PACKAGE_CACHE` propagation.
- `option_env!` tracking.
- `include_bytes!` tracking.
- Atomic package publication.
- Fail-closed handling if the directory cannot be created or written.

Remove:

- The manifest/provenance fingerprint walker.
- Option, environment, workspace, and toolchain hashing.
- Shared-generation locking and serialization.
- Destructive sibling sweeping.
- Permanent lock files and ownership heuristics.

There are two implementation traps:

- Do not generate a new path on every `filesystem_package_cache_dir()` call. A copied empty `OnceLock` across `Session` clones could produce different directories. The build needs a shared `Arc`-owned lease or an eagerly created path.
- Nested dependency `Session`s must continue receiving the root build’s path explicitly. Giving each dependency its own directory would prevent it from seeing already-assembled transitive dependencies.

The main tradeoff is that FPI-using crates become dirty on every top-level build because the environment path changes. That seems acceptable here—and the current implementation rewrites every `.masp` file on every build anyway, so Cargo’s default mtime freshness generally recompiles those consumers already.

Post-build package availability is a separate requirement. The benchmark currently scans the cache after the compiler subprocess exits to obtain dependency forests. That should become an explicit dependency bundle/build output, or an exact reported directory, rather than discovering packages by scanning an internal cache. If the compiler must preserve those files, it can retain the unique generation and report its path without ever reusing it.

Longer term, source-artifact caching should sit behind the assembler/package registry and use the actual resolved provenance and assembled package digests. Cached packages can then be copied or hard-linked into the build-scoped exchange. That avoids reimplementing the assembler’s provenance model.

So my recommendation is to rework this PR around a unique build-generation lease rather than continue patching the 1,470-line `package_cache.rs`. It is both simpler and a better separation of concerns.

---

Let me know your thoughts, especially if you feel there is a subtlety that isn't addressed with this proposed approach.

greenhat:
Agreed. I'm not happy with `package_cache` module. I'm on board with the rework.

One requirement the proposal must keep, though: **the packages have to outlive the compiler process.** In #1298 (stacked on this PR), the contract `build.rs` makes plain `cargo check` / IDE analysis work: it runs a nested `cargo miden build`, then exports `MIDENC_PACKAGE_CACHE` so the *outer* rustc — which runs after the compiler exits — reads the dependency packages. "Delete on normal completion" would delete the directory before its main consumer reads it.

The cleanest fix: let the caller name an export directory (e.g. `cargo miden build --export-packages <dir>`). The compiler keeps its ephemeral internal lease and copies or hard-links the finished packages out on success. The build script derives the directory from `CARGO_MANIFEST_DIR`, so each export replaces the previous one — no generation bookkeeping, no pile-up. A failed build leaves the last good export in place, and the stable path means consumers re-expand only when package contents change (`include_bytes!` tracking), not on every rebuild.

One note: the manifest-closure walk must survive (only the fingerprint transcript goes): it produces the build script's `cargo:rerun-if-changed` watch list and source-dependency count.

I'd like to rework this PR (branch) and #1298 stacked on it into a single branch so that the producer and consumer sit together.

bitwalker:
> One requirement the proposal must keep, though: **the packages have to outlive the compiler process.** In #1298 (stacked on this PR), the contract `build.rs` makes plain `cargo check` / IDE analysis work: it runs a nested `cargo miden build`, then exports `MIDENC_PACKAGE_CACHE` so the _outer_ rustc — which runs after the compiler exits — reads the dependency packages. "Delete on normal completion" would delete the directory before its main consumer reads it.

Agreed, I would suggest we provide a way for a calling process, such as a `build.rs` script, to export `MIDENC_PACKAGE_CACHE` itself, and the compiler would then detect if `MIDENC_PACKAGE_CACHE` has been set when creating the root `Session`, and use that rather than generating its own. It would furthermore set a flag that disables automatic cleanup of the package cache directory.

A caller, e.g. a `build.rs` script, which exports `MIDENC_PACKAGE_CACHE` can then decide where it wants to place the cache, and what to do with it (does it use it as a staging ground, does it reuse the cache as long as the build script isn't re-run, etc.).

I think that provides a better foundation for "caching the cache" so to speak, i.e. only re-running the compiler when Cargo/rustc thinks that the build inputs have changed (e.g. the build script could emit `rerun-if-changed=miden-project.toml`, in addition to the default conditions that force a rebuild).

> The cleanest fix: let the caller name an export directory (e.g. `cargo miden build --export-packages <dir>`)

We could support it as a flag - but I think it's simpler to just have `midenc` detect if that environment variable is set, and use it, rather than generating its own. Not to mention the fact that this is really an internal implementation detail of the Rust build process anyway.

> The build script derives the directory from `CARGO_MANIFEST_DIR`

I would derive from `OUT_DIR` instead, rather than `CARGO_MANIFEST_DIR`, and then export `MIDENC_PACKAGE_CACHE` via `rustc-env=K=V` from the build script. That way fresh rebuilds don't attempt to populate the same directory with artifacts from different builds (though this could maybe be handled by simply removing the directory before using it). I would expect proc macros to have access to environment variables exported from the build script, but testing would confirm for sure.

> One note: the manifest-closure walk must survive (only the fingerprint transcript goes): it produces the build script's `cargo:rerun-if-changed` watch list and source-dependency count.

I would not do this - for one thing, it is still not precise enough, rebuilds won't be triggered if you just change a source file in a source dependency, and because Miden source dependencies are not necessarily _Rust_ sources, the only way you could avoid things being stale is if you watch every file in the source directory of the dependency.

Further, you'll end up having to do basically just as much work as just rebuilding from source every time - because the only way to get _accurate_ provenance is to compute the build provenance of the project dependency graph, which in the case of Rust, requires compiling to MASM anyway.

I would suggest we implement things following an always-rebuild strategy for now, and focus on providing better tools for computing language-specific provenance details, so we can ask Cargo/rustc itself whether a rebuild is warranted, rather than trying to rebuild that ourselves (which we can only do a poor imitation of IMO).

> I'd like to rework this PR (branch) and #1298 stacked on it into a single branch so that the producer and consumer sit together.

Sure, that's fine by me!

greenhat:
Closing in favor of #1328 (has https://github.com/0xMiden/compiler/pull/1306#issuecomment-5256652204 implemented).

PR #1311, opened 2026-08-07
chore: migrate to Miden protocol 0.16.0-rc.4 and VM 0.29

### LINKED ISSUE #1310: VM v0.29 upgrade + v0.16.0-rc.3 protocol upgrade

greenhat:
Closes #1310.

Migrates the compiler and SDK to the protocol 0.16 release-candidate line: miden-protocol/miden-standards `=0.16.0-rc.4`, the VM 0.29 crate family, miden-debug 0.10, miden-field 0.29, and midenc-hir-type 0.10.

The main adaptations: every registry/executor/host/assembler setup now seeds both core-library packages (miden-core plus the new miden-precompiles dependency); the `log_precompile` instruction rename to `log_deferred` is followed through the HIR op, frontend, and codegen; the MASM frontend learns the operand footprints of the new deferred system events; the execution harnesses adopt the typed `AdviceStack` API; and the rpo-falcon512 auth example builds the new 24-felt transaction summary that binds the reference block commitment, expiration delta, and user parameters.

`tests/integration-network` now consumes the released miden-client `0.16.0-rc.1` from crates.io (which pulls the released node proto crates); its mockchain tests are ported to the miden-testing rc.4 transaction-builder API, including the typed tx-script argument and note-constructor suites that landed on `next` mid-migration. The lockfile carries no git dependencies.

An audit of the base-sys bindings against the rc.4 transaction-kernel MASM found every bound procedure in sync; the wrapper-level cleanup it surfaced removes the misleading `output_note::set_*_attachment` aliases (attachments are append-only), makes `add_attachment_from_memory` fail fast on empty or oversized attachments, and gives `note::__entrypoint_root` the same `WordAligned` return area as every other binding.

The cargo-miden test harness no longer caches scaffolded-project templates across runs: the cache seeded each template with a stale copy of the workspace lockfile, which replayed pre-migration dependency graphs into fresh projects and broke `miden-field` builds with mixed `num-bigint` versions. Templates now materialize fresh per test process and resolve from scratch, exactly like a user-run `cargo miden new`.

Fixture and example lockfiles are refreshed so the p3 crates resolve to 0.6.3, matching miden-field 0.29's move to num-bigint 0.5. miden-field's published p3 requirements are still `^0.6` (filed upstream as 0xMiden/miden-vm#3536); nothing in this repository depends on that fix anymore, but generated projects on machines with stale registry caches remain exposed until it ships.

The SDK changelog gains a breaking-change entry: compiled packages link against the updated transaction kernel and core library, so their commitments change, and hosts must use matching protocol `0.16.0-rc.4` crates (for example `miden-testing 0.16.0-rc.4` or `miden-client 0.16.0-rc.1`). Guest code compiles unchanged.

## TODO:

- [x] 0xMiden/node#2421 (node protocol migration) merged.
- [x] 0xMiden/miden-client#2356 merged and miden-client `0.16.0-rc.1` released; `tests/integration-network` switched back to the crates.io client and the git-branch dependency dropped.
- [x] sync tx kernel bindings.


tools/cargo-miden/tests/utils.rs
```diff
@@ -48,71 +45,37 @@ pub(crate) fn project_template_arg(template: &str) -> String {
...
-        match fs::create_dir(&lock_dir) {
-            Ok(()) => break,
-            Err(err) if err.kind() == ErrorKind::AlreadyExists => {
-                if templates_cache_is_ready(&cache_root, &ready_marker) {
-                    return Ok(cache_root);
-                }
-                thread::sleep(Duration::from_millis(100));
+/// Materializes the local stand-in templates for this test process and returns the directory.
+///
+/// The templates are written fresh once per test process, so scaffolded projects always see the
```
bitwalker: I'll have to check to see if this reintroduces the issue where cold builds/test runs take literal _hours_ to finish - but we need to be _extremely_ careful about doing "clean" builds with test cases - every single test that runs with a clean/cold build takes an enormous amount of time, because it pulls in the compiler dependency tree. A single test run could end up producing ~100GB of artifacts that are duplicates.

midenc-session/src/registry.rs
```diff
@@ -148,6 +150,86 @@ impl HybridPackageRegistry {
...
+    /// Seeds the registry with bundled packages that installed packages depend on.
+    ///
+    /// Package manifests record their dependencies with exact digests, so an artifact with the
+    /// right name and version but a different digest can never satisfy one. This walks the
+    /// manifest dependencies of every installed package, including the packages it installs
+    /// along the way, and installs the bundled package that matches a missing dependency
+    /// exactly. When neither the local registry nor a bundled package can satisfy a dependency,
+    /// a warning names the mismatch instead of leaving it to fail later, far from the cause.
+    #[cfg(any(test, feature = "std"))]
+    fn seed_bundled_dependencies(&mut self) -> Result<(), Report> {
```
bitwalker: As noted above, this whole function can be removed

sdk/CHANGELOG.md
```diff
@@ -53,8 +53,27 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   re-exports `miden_field_repr` under its own crate name, so the derives' generated code
   resolves in crates that only depend on `miden` and glob-import it #1291
 
+### Changed
```
bitwalker, line 56: Just a note, but I'd prefer if we added this information in commit bodies, so that we can just lean on the `changelog-prompt` task of the release tooling to aid in generating CHANGELOG entries prior to release. We either have to write the entries by hand, or have an agent summarize the changes that have happened since the last release, but mixing the two isn't likely to work very well.

I'm aiming to make our changelogs more useful in general, so the entries here are the sort of thing we're after for sure, but I also don't want CHANGELOG conflicts to become a constant pain like they are in the miden-vm repo, so that's why I landed on the `changelog-prompt` approach. I'm open to suggestions on this front though. I can say from experience working in the VM repo that including CHANGELOG entries in every PR leads to constant merge conflicts, so that's the main thing I'm trying to avoid (that and the useless CHANGELOG entries generated by tools like `release-plz`).

midenc-session/src/registry.rs
```diff
@@ -98,6 +98,8 @@ impl HybridPackageRegistry {
             }
         }
 
+        registry.seed_bundled_dependencies()?;
```
bitwalker: This can be removed and replaced by simply adding a line above here to add the `miden-precompiles` package via `CoreLibrary::precompiles_package`.

The project assembler already does dependency resolution and verification, so we don't need to duplicate that here.

bitwalker: I should note that I'm also modifying `miden-core` to statically link `miden-precompiles`, so the separation is going away in a 0.29.2 release anyway

bitwalker (commented):
This is basically ready to merge, just one tweak I'd like to make first

sdk/CHANGELOG.md
greenhat: Sure. I'm on board with generating it from the commits. I'll drop CHANGELOG updates in future PRs.

greenhat:
@bitwalker Done. Please do another round.

bitwalker (approved):

Issue #1155, opened 2026-06-01
Add tests for `saturating` arithmetics


mooori:
These are not yet tested in `end_to_end/arithmetic.rs`.

bitwalker:
Done in #1354

Issue #1144, opened 2026-05-22
native/MASM divergence for `i64::checked_mul`


mooori:
Reproduction requires #1140 

```sh
cargo nextest run -p midenc-integration-tests \
    --run-ignored all \
    -E 'test(/checked_mul_i64_edge_case/)'
```

```
    thread 'end_to_end::arithmetic::checked_mul_i64_edge_cases' (188951) panicked at tests/integration/src/end_to_end/arithmetic.rs:1082:13:
    Found minimal(shrinked) failing case: (-4611686018427387904, 2)
    Failure: Reason("assertion failed: `(left == right)` \n  left: `[0, 0, 0, 0, 0, 0, 0, 128, 1]`, \n right: `[0, 0, 0, 0, 0, 0, 0, 128, 0]`: VM output mismatch at tests/integration/src/end_to_end/arithmetic.rs:1075")
```

mooori:
The difference is in the last element which indicates overflow. It is off in Wasm due to a miscompilation of the overflow check for `i64::checked_mul` in LLVM's Wasm backend when `wide-arithmetic` is enabled: https://github.com/llvm/llvm-project/issues/199910. This propagates to the generated Masm.

PR #1315, opened 2026-08-09
test(arithmetic): add saturating arithmetic coverage

UfukNode:
Adds coverage for Rust's saturating integer arithmetic helpers.

The arithmetic integration tests already cover checked, overflowing, and wrapping variants, but saturating add/sub/mul/div were not covered. This adds the same style of VM-vs-Rust checks for the supported integer widths.

`saturating_mul_i64` is included as ignored because it currently hits the existing i64 multiplication divergence tracked in #1144.

Refs #1155.

Testing:
- `cargo fmt --check`
- `git diff --check`
- `cargo test -p midenc-integration-tests saturating_ -- --nocapture`

### DIFF
--- a/tests/integration/src/end_to_end/arithmetic.rs
+++ b/tests/integration/src/end_to_end/arithmetic.rs
@@ -217,6 +217,23 @@ macro_rules! test_unary_op_total {
     };
 }
 
+macro_rules! test_saturating_arith {
+    ($fn_name:ident, $strategy:ident, $($op_ty:tt),+ $(,)?) => {
+        $(
+            concat_idents::concat_idents!(test_name = $fn_name, _, $op_ty {
+                #[test]
+                fn test_name() {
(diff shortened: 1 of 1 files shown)


tests/integration/src/end_to_end/arithmetic.rs
```diff
@@ -217,6 +217,23 @@ macro_rules! test_unary_op_total {
     };
 }
 
+macro_rules! test_saturating_arith {
```
bitwalker, line 220: I would rename this macro to `test_binary_fn` to mirror the underlying function it delegates to, since it is not specific to saturating arithmetic.

UfukNode: Thanks good point. Renamed it to test_binary_fn.

mooori (commented):
I think `u128` and `i128` should be covered too.

PR #1323, opened 2026-08-11
test(wasm): cover remaining i32 binary operations

### LINKED ISSUE #1212: Test interpreted Wasm op against compiled wasm op
Add tests which assert that the following produce the same result:

- Wasm instruction executed/interpreted by Wasm runtime/interpreter
- vs compiling that Wasm instruction and executing it on Miden VM

This enables direct testing of Wasm instruction translation/compilation.

#1207 has set it up for `i32` and added tests for a few `i32` arith instructions.

TODO

- [ ] remaining `i32` arithmetics
- [ ] `i64` arithmetics
- [ ] load/store?

crazywriter1:
## Summary
- Extend the Wasm interpreter-vs-Miden VM test harness coverage for remaining `i32` binary operations.
- Add coverage for arithmetic, unsigned division/remainder, bitwise operations, shifts, and rotations.
- Exercise Wasm shift-count masking and zero-divisor traps through existing edge-case strategies.

Closes #1212

## Test plan
- [x] `CARGO_BUILD_JOBS=1 cargo test -p midenc-integration-tests wasm_translation::i32`


tests/integration/src/end_to_end/wasm_translation/i32.rs
```diff
@@ -108,16 +108,81 @@ fn i32_add() {
...
+
 #[test]
 fn i32_div_s() {
     test_i32_wasm_op_binary("i32.div_s", NumericStrategy::<i32>::div_signed_checked());
 }
 
+#[test]
+fn i32_div_u() {
+    // The values remain bit-identical i32 parameters at the Wasm boundary; `i32.div_u`
+    // interprets those bits as unsigned. This strategy deliberately includes zero divisors.
```
mooori: Signed values *can* be used in place of unsigned values, however a signed strategy does most likely not cover all critical values that should be tested for an unsigned op.

I think a strategy that fits this test perfectly already exists. If not, the right approach would be adding a strategy that fits the test.

mooori (changes requested):
Tests of unsigned ops should not use signed strategies, see the inline comment for details.

tests/integration/src/end_to_end/wasm_translation/i32.rs
```diff
@@ -17,9 +17,10 @@ use crate::{
 /// Proptest harness for a binary `(i32, i32) -> i32` Wasm operation.
 ///
 /// The `wat_op` is executed in a function exported as `entrypoint`.
-fn test_i32_wasm_op_binary<S>(wat_op: &str, strategy: S)
+fn test_i32_wasm_op_binary<S, T>(wat_op: &str, strategy: S)
 where
-    S: Strategy<Value = (i32, i32)>,
+    S: Strategy<Value = (T, T)>,
```
mooori: With the strategy being over `T` instead of a concrete type, I would avoid conversions with `as` and rather have a trait method for every required conversion. E.g. `as_u32()` and `as_i32`.

tests/integration/src/end_to_end/wasm_translation/i32.rs
```diff
@@ -103,21 +106,100 @@ where
     }
 }
 
+/// Converts a generated input into the bit pattern passed to a Wasm `i32` parameter.
+trait WasmI32Input {
```
mooori: IMO the name of the trait and its method are confusing.

tests/integration/src/end_to_end/wasm_translation/i32.rs
```diff
@@ -103,21 +106,100 @@ where
...
+}
+
+#[test]
+fn i32_shr_s() {
+    test_i32_wasm_op_binary("i32.shr_s", NumericStrategy::<i32>::shr_signed_checked());
+}
+
+#[test]
+fn i32_shr_u() {
+    test_i32_wasm_op_binary("i32.shr_u", NumericStrategy::<i32>::shr_signed_checked());
```
mooori: That's still an unsigned op using a signed strategy.

mooori (changes requested):

tests/integration/src/end_to_end/wasm_translation/i32.rs
crazywriter1: Right. There's no unsigned shift strategy yet, only `shr_signed_checked()` and `shr_signed_checked_u32_shift()`, so I'll add `shr_unsigned_checked()` and use it here.

`shl`/`rotl`/`rotr` also use the signed one. They're sign-agnostic so I left them but I can switch them too if you want it consistent.

tests/integration/src/end_to_end/wasm_translation/i32.rs
crazywriter1: Done. The trait now exposes `as_i32()` and `as_u32()`, so nothing in the harness casts. The interpreter gets `a.as_i32()` and the VM stack gets `u64::from(a.as_u32())`. The only `as` left are inside the two trait impls.

bitwalker (approved):

PR #1324, opened 2026-08-11
test(arithmetic): add shift and rotate coverage

### LINKED ISSUE #1153: Increase arithmetic op test coverage for `shift` and `rotate`
These tests should be added to `tests/integration/src/end_to_end/arithmetic.rs`

### `rotate`

- [ ] `rotate_left`
- [ ] `rotate_right`

Note that Rust accepts `rotate_count > integer_width`.

### `shift`

Plain and wrapping `shr`, `shl` are already covered.

TODO:

- [ ] `checked_sh*`
- [ ] `overflowing_sh*`
- [ ] `unbounded_sh*`

erhnysr:
Closes #1153.

Adds the coverage the issue asked for in `tests/integration/src/end_to_end/arithmetic.rs`, following the existing `test_int_op!` / `test_binary_fn` conventions:

**rotate**
- `rotate_left` / `rotate_right`

**shift**
- `checked_shl` / `checked_shr`
- `overflowing_shl` / `overflowing_shr`
- `unbounded_shl` / `unbounded_shr`

76 new tests across the integer widths.

## On `rotate_count > width`

The issue called this out specifically. The rotate tests take the shift amount as `any::<u32>()`, so counts at or above the operand width aren't an edge case here — they're the common case, since the generator draws across the full `u32` range while the widths are 8 to 128. Each is checked against the corresponding Rust intrinsic, which reduces `count % width`, so the compiled MASM has to agree with that reduction rather than merely avoid a panic.

No divergence between Miden VM and native Rust semantics turned up in any op or width, including the at-or-above-width counts.

## Tests

`cargo test -p midenc-integration-tests arithmetic` — 364 passed, 0 failed, 6 ignored. The 6 ignored are pre-existing `#[ignore]`s unrelated to this change. The filter also picks up the `intrinsics::*_arithmetic` tests, so neighbouring coverage is included in that run.

I didn't run the full integration suite locally — 740 tests with per-test compilation is impractical on this machine, so I scoped it to arithmetic and left the rest to CI.

## Overlap with #1315

#1315 (issue #1155, saturating arithmetic) touches the same file but different regions — a `test_saturating_arith!` macro around L217 and invocations just after `checked_rem_i64` around L897. My additions sit after `test_checked_arith` and after the `wrapping_shr_i128` block, well below both. The two diffs apply independently; whichever lands first, the other should still apply cleanly.

### DIFF
--- a/tests/integration/src/end_to_end/arithmetic.rs
+++ b/tests/integration/src/end_to_end/arithmetic.rs
@@ -954,6 +954,21 @@ where
     T: ToBytes + ToMidenRepr + FromMidenRepr + PrimInt + Arbitrary + std::fmt::Debug,
     U: ToMidenRepr + PrimInt + Arbitrary,
 {
+    test_binary_fn_with_runner(op, fn_name, strategy, TestRunner::default());
+}
+
+/// Like [`test_binary_fn`], but runs `strategy` with the caller-provided `runner`. Pass a
+/// [`NumericStrategy::test_runner`] when the strategy is `NumericStrategy`-generated, so its higher
+/// case count reliably exercises the edge cases.
+fn test_binary_fn_with_runner<T, U>(
+    op: fn(T, U) -> T,
+    fn_name: &str,
+    strategy: impl Strategy<Value = (T, U)>,
+    mut runner: TestRunner,
+) where
+    T: ToBytes + ToMidenRepr + FromMidenRepr + PrimInt + Arbitrary + std::fmt::Debug,
+    U: ToMidenRepr + PrimInt + Arbitrary,
+{
     // The return value of `type_name` isn't stable, but it's good enough for this test.
     let lhs_ty_name = type_name::<T>();
     let rhs_ty_name = type_name::<U>();
@@ -970,7 +985,7 @@ where
     let mut test = CompilerTest::rust_fn_body(&main_fn, None);
     let package = test.compile_package();
 
-    let res = TestRunner::default().run(&strategy, move |(a, b)| {
+    let res = runner.run(&strategy, move |(a, b)| {
         let rust_out = op(a, b);
 
         // Write the operation result to 20 * PAGE_SIZE.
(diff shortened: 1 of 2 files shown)


tests/integration/src/end_to_end/arithmetic.rs
```diff
@@ -1062,6 +1062,126 @@ fn test_checked_arith<T>(
     }
 }
 
+/// Like [`test_checked_arith`], but for shift/rotate-style operations whose shift amount is a
+/// `u32` regardless of the operand width (e.g. `checked_shl`, `checked_shr`).
+fn test_checked_shift<T, U>(
```
mooori: I think this helper should be more general to enable later re-use. Can it become `test_checked_binary_fn` modeled after `test_binary_fn`?

tests/integration/src/end_to_end/arithmetic.rs
```diff
@@ -1062,6 +1062,126 @@ fn test_checked_arith<T>(
...
+            panic!("Found minimal(shrinked) failing case: {value:?}\nFailure: {reason:?}");
+        }
+        Ok(_) => (),
+        _ => panic!("Unexpected test result: {:?}", res),
+    }
+}
+
+/// Like [`test_overflowing_arith`], but for shift-style operations whose shift amount is a `u32`
+/// regardless of the operand width (e.g. `overflowing_shl`, `overflowing_shr`).
+fn test_overflowing_shift<T, U>(
```
mooori: I think this helper should be more general to enable later re-use. Can it become `test_overflowing_binary_fn` modeled after `test_binary_fn`?

mooori (changes requested):
Using strategies like `(any::<T>(), any::<T>())` is unlikely to hit the critical edge cases of ops. The strategies should be generated with `NumericStrategy`. For example there is already `NumericStrategy::shr_signed_checked_u32_shift`. Other ops might not yet have a matching `NumericStrategy`, in that case please go ahead and add it.

PR #1328, opened 2026-08-12
Per-build package cache with caller adoption, WIT-in-package, and the contract build script

### LINKED ISSUE #345: Use account's Miden package WIT file in a note script project

### LINKED ISSUE #1215: The `#[account(...)]` macro is red in LSP until someone manually builds the account project dependency
DX is suboptimal. For example, in the new project template, the LSP will throw an error until someone manually builds the account project or runs the integration test suite that will materialize packages for all dependencies on the disk.

I'm thinking of launching `cargo miden build --release` on the dependency project from the `#[account(...)]` macro as the last resort, but it's ugly as hell.

### LINKED ISSUE #1298: Explore using `build.rs` in the contract to compile dependencies

### LINKED ISSUE #1300: SDK macros: dependency `.masp` resolution does not enforce version and digest pins consistently
## Summary

The SDK macros resolve each Miden path dependency to a compiled `.masp` package at macro-expansion time (`sdk/base-macros/src/dependency_package.rs`). They read the dependency's embedded WIT from that package, and they bake its FPI procedure roots into the generated code. The resolver does not enforce the manifest pins consistently across its search paths. It can select a package that the manifest forbids, or a different package than the assembler links later. The failure then surfaces at link or run time, far from the cause, with nothing that points back to the wrong artifact.

Surfaced by the #1248 (introduced in #1293)

### 1. The `MIDENC_PACKAGE_CACHE` branch does not check the version pin

`sdk/base-macros/src/dependency_package.rs:362-372`:

```rust
        let mut id_mismatches = Vec::new();
        for candidate in candidates {
            let package = read_package(&candidate)?;
            if package_id_matches(&package, &package_stems) {
                return Ok(ResolvedDependencyPackage { path: candidate, package });
            }
            id_mismatches.push((candidate, package.name.to_string()));
        }
```

The general search path checks the package id **and** the version pin (`:426-437` calls `package_version_matches`). The cache branch checks only the id. This is the primary path in driven builds: midenc sets `MIDENC_PACKAGE_CACHE` for nested builds (`midenc-compile/src/pipeline/frontends/rust.rs:1605-1626`), and the cache is a shared publish-and-lookup directory that persists artifacts across builds. It can hold stale or newer versions under the same id; freshest-first adoption then violates the pin.

Failure scenario: the manifest pins `^2.0`; the cache holds a fresh `0.1.0` → the macro accepts `0.1.0` and generates bindings and FPI roots for the wrong ABI.

The doc on `resolve_dependency_package` (`:325-331`) promises the version check for "every candidate located by searching". The doc does not match this branch.

### 2. `Digest` pins and the digest half of `Exact` pins are ignored

`sdk/base-macros/src/dependency_package.rs:466-481`:

```rust
/// Digest pins are enforced by the assembler at link time; checking them here would require
/// computing content digests during macro expansion, so they are accepted as-is.
fn package_version_matches(
    package: &Package,
    version: Option<&miden_project::VersionRequirement>,
) -> bool {
    match version {
        Some(miden_project::VersionRequirement::Semantic(requirement)) => {
            requirement.inner().matches(&package.version)
        }
        Some(miden_project::VersionRequirement::Exact(exact)) => exact.version == package.version,
        Some(miden_project::VersionRequirement::Digest(_)) | None => true,
    }
}
```

`VersionRequirement::Exact(Version)` is documented upstream (miden-package-registry) as "an exact assembled package version, **including both semantic version and digest**". The code compares only the semver. `Digest(_)` is accepted unconditionally. The comment justifies this with the cost of computing digests, but `miden_mast_package::Package::digest() -> Word` is available on the already-deserialized package, so the check is cheap.

Failure scenario: the manifest digest-pins artifact A. A fresher artifact B with the same id and version is found first. The macro emits B's WIT and procedure roots. The assembler later resolves the pin to A. The generated FPI calls carry roots that A does not contain, and the transaction fails at run time.

### 3. The explicit `.masp`-file path skips the pins

`sdk/base-macros/src/dependency_package.rs:337-342`:

```rust
    if root.is_file() {
        return Ok(ResolvedDependencyPackage {
            path: root.to_path_buf(),
            package: read_package(root)?,
        });
    }
```

A dependency that points directly at a `.masp` file is read without any checks. Skipping the **id** check is documented and correct — the manifest key need not equal the prebuilt package's id. But a `version`/digest pin is explicit user intent, independent of file naming. Check it here too.

### 4. Candidate-loop robustness

- One corrupt or old-format `.masp` anywhere in the search aborts the whole resolution: `read_package(&candidate)?` at `:426` (and `:364` in the cache branch). The ambient directories (`CARGO_TARGET_DIR`, `OUT_DIR` ancestors, cwd targets) make this easy to hit. Record the deserialize failure as a rejected candidate and continue; keep the hard error only when nothing resolves.
- The comment at `:394-396` says candidates are "ordered freshest-first **within each directory class**". The code pools private and shared matches into one freshness sort (`:399-403`), so a fresher shared (workspace) artifact can shadow the dependency's own artifact. Align the comment or the code.

## Proposed fix

1. Extract one verifier, e.g. `fn verify_candidate(package, stems, requirement) -> Result<(), RejectReason>`, that checks the id, the semver pin, and the digest (for `Digest` and `Exact` via `Package::digest()`). Apply it to all five candidate sources: explicit file, cache, private, shared, ambient.
2. In the cache branch, record version/digest mismatches in the rejected list, the same way id mismatches are recorded.
3. Treat a failed deserialize as a rejected candidate ("failed to deserialize: ..."), not a hard error.
4. Correct the `resolve_dependency_package` doc and the freshest-first comment.
5. Tests: a version pin rejected in the cache branch; a digest pin selecting between two same-id/same-version artifacts; a corrupt artifact skipped when a valid one exists later in the search order.

### LINKED ISSUE #1302: `MIDENC_PACKAGE_CACHE` path should be uniqued by the build inputs

greenhat:
Close #345
Close #1298
Close #1302
Close #1215
Close #1300

This branch combines the package-cache producer side (#1306), the WIT-in-package consumer side (#1248), and the contract build script (#1298) into one delivery, reworked around the per-build cache design agreed in the #1306 review thread.

## Per-build package cache

The compiler derives its filesystem package cache once per top-level compilation. When the calling process already exported `MIDENC_PACKAGE_CACHE`, the compiler adopts that directory and leaves it in place — the caller owns its location and lifetime, which is how packages stay readable after the compiler exits. Otherwise it creates a lease: a globally unique directory under the project's `target/miden/packages/`, shared with every nested build, deleted when the compilation ends. Uniqueness per build makes staleness impossible by construction, so the fingerprint derivation, permanent lock files, and stale-cache pruning of the earlier iterations are gone, along with the manifest-closure walk and the `cargo miden package-cache` query they served (`midenc-session` drops ~1,600 lines of cache machinery for a ~200-line lease module). The cache handle is `Arc`-shared across session clones, created eagerly when the root build constructs its package registry, and retained by the registry itself, so the directory outlives whichever of the two drops last; nested dependency sessions never derive their own and receive the root's path as before. Creation failures fail closed. A killed build leaves an inert lease remnant that `cargo clean` removes; nothing sweeps, by design.

## Compiler-recorded resolution (`miden-deps/`)

The compiler records its dependency resolution in the package cache before each nested cargo build spawns — at which point every dependency is already resolved and published: `miden-deps/<consumer>.deps.toml` maps each declared dependency to the artifact the resolver selected, and the root consumer's `miden-deps/<consumer>.watch` lists the source inputs of every resolved dependency. The SDK macros consume the map instead of re-deriving resolution — workspace, workspace-path, and git dependencies resolve during expansion exactly as the compiler resolved them, and the `.masp` name-probing survives only as a fallback for hand-assembled caches. A dependency whose package embeds no component WIT — a link-only MASM library — is skipped rather than rejected; only a macro that actually references it reports the missing WIT, at the reference site.

## Contract `build.rs` (#1298)

Templates and repository examples ship a `build.rs` that makes plain `cargo check`/`cargo build` and IDE analysis resolve dependency packages: outside a midenc-driven build it stages package-cache generations under its own `OUT_DIR`, fills the next generation with a nested `cargo miden build --release` that adopts it through `MIDENC_PACKAGE_CACHE`, and re-exports the same variable to macro expansion via `rustc-env`. Inside a midenc-driven build it does nothing (recursion guard).

The nested build runs whenever cargo re-runs the script, and the script watches the project manifests plus the compiler-recorded dependency inputs, so editing a dependency re-stages its package on the next check. Only a fully successful nested build publishes its generation — the export pointer is the atomic switch — so a failed build keeps exporting the previous generation whole, never a mix of new and old packages, and fails the outer build when no good generation exists yet. One documented caveat: cargo keys build-script output by crate name and version, so two same-named projects sharing one `CARGO_TARGET_DIR` reuse each other's staging — per-checkout target directories avoid it.

The benchmark harness uses the same adoption mechanism: it hands the compiler a runner-owned cache directory through the environment, which old compiler versions simply ignore (their packages persist in the legacy project cache the runner falls back to), and each profile run loads exactly one package generation, never a combination. During the transition, the workflow's baseline invocation passes `--skip-failed-builds`, since the `next`-built compiler cannot build the examples whose macros require embedded dependency WIT; the flag becomes a no-op once this lands.

## WIT-in-package (#1248)

The component WIT generated by `#[component]` is embedded in the compiled `.masp` (a `wit` custom section), and the SDK macros read dependency WIT from the dependency's compiled package instead of `target/generated-wit/`. `#[note]` embeds its export-only public WIT the same way, which is what lets a transaction script import a note's constructors now that notes are consumed as dependencies. The `wit = "..."` manifest keys become a fallback for packages without embedded WIT, and conflict with embedded WIT is an error. Macro-side dependency resolution reads only from `MIDENC_PACKAGE_CACHE` (or an explicit `.masp` manifest path); the ambient `target/miden/<profile>` searches are removed. FPI expansions emit `option_env!("MIDENC_PACKAGE_CACHE")` and `include_bytes!(<package>)`, which is the invalidation contract the cache relies on: a midenc-driven build re-expands consumers because the lease path rotates, and a stable adopted directory re-expands them only on content changes.

See the root `CHANGELOG.md` (`[Unreleased]`) and `sdk/sdk/MIGRATION.md` for the user-facing summary and migration steps.


examples/basic-wallet/build.rs
```diff
@@ -0,0 +1,122 @@
...
+    env, fs,
+    path::{Path, PathBuf},
+    process::{Command, Output},
+};
+
+fn main() {
+    // The manifests are the one precise trigger: they declare the dependency set this
+    // script stages packages for. Naming any watch disables cargo's watch-everything
+    // default, which would otherwise re-run the nested build after every source edit.
+    println!("cargo:rerun-if-changed=miden-project.toml");
```
bitwalker: **[P1] Track dependency inputs before reusing staged packages**

These directives disable Cargo's default watch-everything behavior, but the replacement trigger set omits dependency source and WIT inputs. After a successful check, editing a path dependency does not rerun this script, so later macro expansion can keep using stale WIT and procedure roots until the consumer manifest is touched or the target directory is cleaned. Please have the compiler/resolver emit the resolved input or dep-info set and translate that into `rerun-if-changed` entries, or use another provenance mechanism that guarantees dependency edits restage the packages.

examples/basic-wallet/build.rs
```diff
@@ -0,0 +1,122 @@
...
+    if env::var_os("MIDENC_PACKAGE_CACHE").is_some() {
+        return;
+    }
+
+    let manifest_dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+    // The staged cache lives in this script's OUT_DIR, so it belongs to this crate and
+    // build configuration and is removed by `cargo clean`. The directory is not cleared
+    // between runs: a failed nested build then leaves the previously staged packages
+    // usable, because analyzing against the last built packages beats failing the check.
+    let cache_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap()).join("miden-packages");
```
bitwalker: **[P1] Publish an adopted cache generation only after full success**

This reuses and mutates the last successful directory package by package, and the code below continues after a nested build failure. If dependency A is replaced and compiling B then fails, the outer build can consume new A together with old B; removed dependencies also remain available. Please stage into a fresh sibling generation and export it only after the entire nested build succeeds, otherwise fail the outer build.

sdk/base-macros/src/dependency_package.rs
```diff
@@ -0,0 +1,793 @@
...
+                        error_span,
+                        format!(
+                            "failed to canonicalize dependency '{}' path '{}': {err}",
+                            dependency.name(),
+                            absolute_path.display()
+                        ),
+                    )
+                })?;
+                let resolved =
+                    resolve_dependency_package(dependency.name().as_ref(), &dependency_root)?;
```
bitwalker: **[P1] Resolve macros through compiler-selected artifacts**

Macro resolution currently implements a stricter and incompatible model than compiler resolution. A valid path such as `../dep/miden-project.toml` canonicalizes to a file and is then treated as a `.masp`, while workspace, workspace-path, and Git dependencies are skipped even though the compiler resolves and publishes them. This loop also requires WIT from every path dependency, including link-only MASM libraries that no SDK macro references. Please consume the exact artifact mapping selected by the compiler independently of source scheme, and load component WIT only for dependencies actually referenced by a macro.

midenc-session/src/lib.rs
```diff
@@ -352,40 +364,86 @@ impl Session {
...
-        registry::HybridPackageRegistry::new_with_filesystem_cache(
-            &self.options,
-            self.filesystem_package_cache_dir(),
-        )
-        .map(Box::new)
+        #[cfg(feature = "std")]
+        let filesystem_cache = self.filesystem_package_cache_dir()?;
+        #[cfg(not(feature = "std"))]
+        let filesystem_cache = None;
+        registry::HybridPackageRegistry::new_with_filesystem_cache(&self.options, filesystem_cache)
```
bitwalker: **[P2] Keep the lease alive for the returned registry**

Only the lease's `PathBuf` is copied into the registry, so the public registry can outlive the last `Session` owner. I reproduced obtaining a registry, dropping the session (which deletes the directory), and then publishing through the still-live registry, which recreates and uses the path without a live unique lease. The registry returned here should retain a shared lease guard rather than only the path.

sdk/base-macros/src/dependency_package.rs
```diff
@@ -0,0 +1,793 @@
...
+    /// the package between them. The host process may also outlive many builds (rust-analyzer
+    /// keeps proc-macro servers running), so entries are never trusted by path alone.
+    static PACKAGE_READS: core::cell::RefCell<PackageReadCache> =
+        core::cell::RefCell::new(std::collections::HashMap::new());
+}
+
+/// Reads and deserializes a compiled Miden package, reusing the previous read of an unchanged
+/// file.
+fn read_package(package_path: &Path) -> Result<Arc<Package>, Error> {
+    let fingerprint = fs::metadata(package_path)
```
bitwalker: **[P2] Do not use mtime and length as content identity**

The adopted cache deliberately keeps stable paths that can be atomically replaced. A same-length replacement with a colliding or coarse modification timestamp will make a long-lived rust-analyzer proc-macro process return the old `Package`. Please key this memoization by the bytes/content identity, or avoid process-wide caching for mutable adopted staging paths.

examples/basic-wallet/build.rs
```diff
@@ -0,0 +1,122 @@
...
+    // Re-evaluate when the build mode or the tool selection changes.
+    println!("cargo:rerun-if-env-changed=MIDENC_PACKAGE_CACHE");
+    println!("cargo:rerun-if-env-changed=CARGO_MIDEN");
+    // These inputs shape the compiled packages.
+    println!("cargo:rerun-if-env-changed=RUSTFLAGS");
+    println!("cargo:rerun-if-env-changed=RUSTUP_TOOLCHAIN");
+
+    // Inside a midenc-driven build the compiler owns the package cache, macro expansion
+    // already sees the variable, and a nested build would recurse into this script forever.
+    if env::var_os("MIDENC_PACKAGE_CACHE").is_some() {
```
bitwalker: **[P2] Normalize an empty cache variable consistently**

`PackageCacheLease` treats an empty `MIDENC_PACKAGE_CACHE` as unset, while this presence check treats it as a compiler-driven build and returns; the macro-side lookup also converts the empty value to `PathBuf("")`. Consequently `MIDENC_PACKAGE_CACHE='' cargo check` skips staging and can search the current directory. Please require a non-empty value consistently at every boundary.

midenc-compile/src/pipeline/frontends/rust.rs
```diff
@@ -1607,24 +1610,64 @@ pub(crate) mod manifest {
...
+    /// explicit `--rustflags`, preserving the pre-existing override order — nothing a caller
+    /// passes is dropped.
+    pub(super) fn merge_rust_flags(
+        mandatory: &[&str],
+        inherited_encoded: Option<&std::ffi::OsStr>,
+        inherited_plain: Option<&std::ffi::OsStr>,
+        explicit: Option<&str>,
+    ) -> Vec<String> {
+        let mut args: Vec<String> = mandatory.iter().map(|flag| flag.to_string()).collect();
+        let encoded = inherited_encoded
```
bitwalker: **[P2] Preserve Cargo's present-empty encoded-flags precedence**

Cargo treats the presence of `CARGO_ENCODED_RUSTFLAGS` as authoritative even when its value is empty. Filtering the empty value here makes it look absent and falls back to `RUSTFLAGS`; for example, encoded flags set to `""` plus `RUSTFLAGS="-C opt-level=3"` inherits a flag that Cargo itself suppresses. Please distinguish absent from present-empty and fall back to plain flags only when the encoded variable is absent.

benches/src/lib.rs
```diff
@@ -191,11 +222,14 @@ impl BenchmarkRunner {
             .with_context(|| format!("failed to parse {}", inputs_path.display()))?;
         let mut executor = Executor::from_config(config);
         let packages_dir = project_dir.join("target/miden/packages");
-        let mut dependencies = fs::read_dir(&packages_dir)
-            .with_context(|| format!("failed to read {}", packages_dir.display()))?
-            .map(|entry| entry.map(|entry| entry.path()))
-            .collect::<std::io::Result<Vec<_>>>()?;
-        dependencies.retain(|path| path.extension().is_some_and(|ext| ext == "masp"));
+        let mut dependencies = collect_dependency_packages(&packages_dir)?;
```
bitwalker: **[P2] Do not combine current and stale package generations**

This collects legacy cache generations and then combines them with the current exported generation below. Packages from stale entries are not necessarily inert: installing two packages with the same name and version but different MAST digests is rejected, so a retained or concurrently produced generation can fail profiling before execution. Please load only the exact generation produced for the current compile and isolate any legacy fallback from it.

bitwalker (changes requested):
Nice work! Codex found some issues, including a few I noticed during my own manual review, and I think all of them are probably important to fix - the P1 issues primarily, but the P2 issues are only slightly less significant

midenc-compile/src/pipeline/frontends/rust.rs
```diff
@@ -1072,6 +1069,188 @@ impl RustProjectFrontend {
...
+                manifest_path,
+                project_root,
+                workspace_root,
+                library_path,
+                ..
+            }) => {
+                if matches!(origin, ProjectSourceOrigin::Git { .. }) {
+                    continue;
+                }
+                add(manifest_path);
```
bitwalker: **[P1] Derive invalidation from Cargo's `dep-info`**

This hand-built watch set still omits inputs that can change the compiled dependency package, including the dependency's `Cargo.lock`, `build.rs`, and `.cargo/config.toml`. Because the outer build script disables Cargo's default tracking and then trusts this list as complete, changing one of those inputs can leave plain `cargo check` expanding against the previous `.masp`. Please consume the dependency build's Cargo `dep-info` (or equivalent compiler-produced provenance) instead of trying to predict the complete input closure here; that will necessarily track more of the inputs Cargo actually used.

examples/basic-wallet/build.rs
```diff
@@ -0,0 +1,215 @@
...
+/// is resolved through the `cargo` that drives this build. The nested build gets its own
+/// cargo target directory: the outer cargo holds a lock on this build's target directory
+/// while build scripts run, and a nested build against the same directory would deadlock.
+fn run_cargo_miden_build(manifest_dir: &Path, cache_dir: &Path) -> Output {
+    let mut command = match env::var_os("CARGO_MIDEN") {
+        Some(cargo_miden) => Command::new(cargo_miden),
+        None => Command::new(env::var_os("CARGO").unwrap_or_else(|| "cargo".into())),
+    };
+    command
+        .args(["miden", "build", "--release"])
```
bitwalker: **[P1] Honor the outer Cargo feature selection while staging**

This always builds the root with its default features, regardless of whether the outer command was `cargo check --features ...`, `--no-default-features`, or `--all-features`. A project that is valid only with an enabled feature can therefore have its nested staging build fail; if a previous generation exists, analysis can instead continue against packages staged for the wrong feature graph. Please forward the effective feature selection, or preferably expose a dependency-only staging operation so this script does not need to compile the root a second time.

sdk/base-macros/src/dependency_package.rs
```diff
@@ -0,0 +1,1136 @@
...
+        skipped: Vec::new(),
+        artifact_map_path: artifact_map.as_ref().map(|map| map.map_path.clone()),
+    };
+
+    for dependency in package.dependencies() {
+        let name = dependency.name().as_ref();
+        // Registry dependencies are MASM base libraries (`miden-core`, `miden-protocol`)
+        // consumed at link time only; no SDK macro references them, so their (large)
+        // packages are never deserialized here.
+        if matches!(dependency.scheme(), miden_project::DependencyVersionScheme::Registry(_)) {
```
bitwalker: **[P1] Include registry-resolved components in the artifact exchange**

`DependencyVersionScheme::Registry` is generic and can resolve an assembled component package; it is not limited to `miden-core` and `miden-protocol`. Skipping every registry dependency here, while the compiler also omits registry nodes from the artifact map, means a valid registry component can be selected and linked by the compiler but disappear during `#[account(...)]`/`#[component(...)]` resolution, ultimately producing the false diagnostic that it was not declared. Please publish or map the selected registry artifact like the other schemes, retaining a fast path only for packages known to be link-only.

midenc-session/src/package_lease.rs
```diff
@@ -0,0 +1,216 @@
...
+            target: "package-cache",
+            "adopted the caller-provided package cache '{}'",
+            dir.display()
+        );
+        Ok(Self::Adopted(dir))
+    }
+
+    /// Mints a unique lease directory under [`package_cache_parent`] of `project_dir`.
+    fn lease(project_dir: &Path) -> Result<Self, String> {
+        let parent = package_cache_parent(project_dir);
```
bitwalker: **[P2] Put build-owned writes under the configured target directory**

This always creates the lease under `<project>/target`, and the generated build script similarly forces nested Cargo output under `<manifest>/target/miden/build-script`. Consequently a build from a read-only checkout still fails even when the caller supplies a writable `--target-dir`/`CARGO_TARGET_DIR`. The cache path is now derived once and explicitly threaded to every participant, so it no longer needs to ignore the configured target directory to maintain agreement.

benches/src/lib.rs
```diff
@@ -190,12 +243,25 @@ impl BenchmarkRunner {
...
-            .collect::<std::io::Result<Vec<_>>>()?;
-        dependencies.retain(|path| path.extension().is_some_and(|ext| ext == "masp"));
+        // Load exactly one package generation. The current compiler adopts the runner-owned
+        // cache directory (see `Self::compile`), so when that directory holds packages it is
+        // the generation this run produced and the only one loaded. The legacy project-cache
+        // scan serves old compiler versions only — combining generations could pair a stale
+        // package with the current one under the same name and version but a different
+        // digest, which package installation rejects.
+        let export_dir = self.package_cache_dir(project_dir);
+        let mut dependencies: Vec<PathBuf> = if export_dir.is_dir() {
```
bitwalker, line 258: **[P2] Verify the export belongs to this compiler invocation**

This export directory is stable across runner invocations and is never cleared. If a current compiler populates it and a later run uses an older compiler that ignores `MIDENC_PACKAGE_CACHE`, that run writes fresh packages to the legacy cache, but this non-empty stale export suppresses the fallback and is profiled instead. Removed dependencies can likewise survive in the export. Please clear/version it before each compile, or require a marker produced by the current invocation before treating it as this run's generation.

handoff.md
```diff
@@ -0,0 +1,186 @@
+# Handoff: PR #1328 (per-build package cache + WIT-in-package + contract build.rs)
+
+Untracked scratch file — NEVER commit it (same rule for `review*.md`, `pr.md`, `plan.md`).
```
bitwalker: **[P2] Remove internal review scratch files**

This commit also adds `handoff.md`, `oracle-review.md`, and `review_claude.md`—about 560 lines of internal process notes, stale SHAs, and deferred findings. This line explicitly says these scratch files must never be committed. Please remove all three before merging.

bitwalker (changes requested):
This is pretty close, but there are still some issues remaining. I would especially note the comment about dependency watch lists and the recommendation to consume Cargo's `dep-info` for this, since that will necessarily be more comprehensive coverage of things that should trigger a rebuild

PR #1350, opened 2026-08-24
feat(bench): benchmark MockChain transactions

djolertrk:
It depends on the https://github.com/0xMiden/miden-debug/pull/127


examples/storage-example/src/lib.rs
```diff
@@ -26,19 +21,30 @@ struct MyAccount {
     asset_qty_map: StorageMap<Word, AssetAmount>,
 }
 
-impl foo::Guest for MyAccount {
```
greenhat, line 29: I'd like to leave the storage example as is since it is intended to demonstrate the "escape hatch" of providing your own WIT file for power users.

greenhat (changes requested):
Nice! 
Please make a task in the Makefile.toml to run the benchmark suitable for agent consumption (stdout only, no report file, no github comments, etc.).
As a side note, could you please skip posting github comment if there are no changes in size or cycles?

greenhat (approved):

Issue #1351, opened 2026-08-26
Set `@source_name` for procedures with divergent source/linkage name


mooori:
Names in Wasm's name section need not be unique. For debug builds of `examples/` with `panic-abort` rustc produces duplicate names in the Wasm name section, see #1341.

#1342 deduplicates names, which can lead to divergent source/linkage names. In that case, the new `@source_name` attribute should be added to the corresponding procedure. See [vm3716](https://github.com/0xMiden/miden-vm/pull/3716) for details.

Blocked until the compiler uses vm with that feature, probably `>= 0.30`.

PR #1357, opened 2026-08-27
fix: handle duplicate func names in Wasm's name section

### LINKED ISSUE #1341: Frontend assumes names from Wasm's `name` section are unique
However [according to the spec](https://webassembly.github.io/spec/core/appendix/custom.html#name-section)

> Names need not be unique.

So far names turned out to be unique in our examples and fixtures. When switching the default panic strategy from `immediate-abort` to `abort` (#1339), `Felt` related formatting code produces duplicate function names in the `name` section for multiple examples/fixtures.

### Reproduction

```sh
cargo make midenc # build bin/midenc
cd examples/counter-contract
RUSTFLAGS="-C panic=abort" ../../bin/midenc miden-project.toml

thread 'main' (185908) panicked at hir/src/ir/operation.rs:209:17:
Unable to insert builtin.function in symbol table of builtin.module: symbol _RNvXsq_NtCs9z35J0Io6i7_11miden_field10wasm_midenNtB5_16FeltFromIntErrorNtNtCs9YlYMtCqyny_4core3fmt5Debug3fmt is already registered to another operation _RNvXsq_NtCs9z35J0Io6i7_11miden_field10wasm_midenNtB5_16FeltFromIntErrorNtNtCs9YlYMtCqyny_4core3fmt5Debug3fmt.
```

### Why this was hidden so far

It's formatting related code which until recently wasn't supported for the Miden target and got eliminated due to `panic=immediate-abort`. With `panic=abort` it stays around.

`Felt` is special but I think regardless we must not assume that names in the `name` section are unique.

### Proposed fix

Try to derive a unique HIR symbol, e.g. by combining the name with the unique function index.

### LINKED ISSUE #1360: Wasm frontend ignores function export names
The frontend currently uses the export table only to decide whether a function is public or private. The (optional) name-section name is then used as HIR symbol and the export name is not exposed.

### Symptom and reproduction

```wat
;; export-name.wat

(module $export_name_test.wasm
  ;; WAT identifier: $foo_1
  ;; Name-section name: "foo_2"
  ;; Export name: "foo_3"
  (func $foo_1 (@name "foo_2") (result i32)
    i32.const 42
  )
  (export "foo_3" (func $foo_1))
)
```

Generate HIR with

```sh
midenc export-name.wat \
  --entrypoint=foo_3 \
  --emit=hir=- \
  -Canalyze-only
```

The generated HIR contains:

```
builtin.function public extern("C") @foo_2()
```

There is no function named `@foo_3`, even though `foo_3` is the Wasm export name. So the following fails:

```sh
❯ midenc export-name.wat \
    --entrypoint=export_name_test::foo_3 \
    -o export-name.mas
[ERROR linker]     | failed to resolve target "root_ns:root@1.0.0"::export_name_test::foo_3
Error:   x undefined symbol reference
  help: maybe you are missing an import?
```

(Succeeds with `--entrypoint=export_name_test::foo_2`)

### Possible fixes

 - Create a public forwarding function or alias for every function export.
 - Use the export name directly when possible, with aliases for additional exports.

mooori:
Closes #1341
Closes #1360 

Both are related to function names and fixes for each must work well in combination.

### Summary

- Creates unique linkage names for functions whose name is duplicated in the Wasm name section. The linkage names are tracked separately to not modify the Wasm name section.
- Makes export name the linkage for exported functions. Previously the export name was not taken into account for linkage names.
- Debug info: Functions whose name is not unique in Wasm's name section are left out of the name map. This forces them to be resolved via (unique) `low_pc` instead of (non-unique) `source_name`.

### Related follow-up work

- #1351
  - depends on `miden-vm >= 0.30`

# Note on testing

Duplicate name-section names are easy to reproduce with `miden-field`, see #1341. Without `miden-field` it's tricky. Since `miden-field` may change, I avoided it in the test and went for synthetic DWARF instead.


frontend/wasm/src/module/mod.rs
```diff
@@ -314,13 +331,215 @@ impl Module {
...
+        // Collect and validate function exports
+        let mut exported_as: FxHashMap<FuncIndex, Symbol> = FxHashMap::default();
+        let mut export_names: FxHashSet<Symbol> = FxHashSet::default();
+        for (export_name, entity) in &self.exports {
+            let EntityIndex::Function(func_idx) = entity else {
+                continue;
+            };
+            let export_sym = Symbol::intern(export_name.as_str());
+
+            if exported_as.insert(*func_idx, export_sym).is_some() {
```
bitwalker, line 407: Please preserve multiple export aliases for a function instead of rejecting them. This occurs in ordinary optimized Rust: compiling two exported functions `foo` and `bar` that both return `42` with `rustc -C opt-level=3 --target wasm32-unknown-unknown` produces two exports referring to function index 0. This check rejects that output. It also regresses component startup targets with multiple export aliases: I verified that a component whose startup target is exported as both `aliased-start` and `other-start` translates on the base commit but fails here. Choose a primary linkage name and preserve the additional exports through aliases or forwarding functions, with a regression test for this case.

frontend/wasm/src/module/mod.rs
```diff
@@ -314,13 +331,215 @@ impl Module {
...
+                    );
+                }
+            }
+        }
+
+        // Reject collisions between export names and global variable names
+        let mut global_names: FxHashSet<Symbol> = FxHashSet::default();
+        for global_idx in self.globals.keys() {
+            let name = self.global_name(global_idx);
+            if export_names.contains(&name) {
```
bitwalker, line 433: Please disambiguate the internal global linkage name instead of rejecting the public function export. For example, this valid module translates on the base commit but fails here:

```wat
(module $counter_module
  (global $counter (mut i32) (i32.const 0))
  (func $read_counter (export "counter") (result i32)
    global.get $counter))
```

The global's name-section name and the function's export name are allowed to coincide; the private global's source name should not constrain the public interface. Extend the source/linkage distinction to globals so the public export keeps its name and references to the renamed global remain consistent, and cover this with a regression test.

bitwalker (changes requested):
Looks good, but there are two relatively significant issues that need to be addressed before this can be merged, see comments for details

frontend/wasm/src/module/mod.rs
mooori: Following our sync on Monday, my understanding is that for now we don't expect Wasm producers to generate collisions between `global` and `export` names. The code in this PR already emits a diagnostic in that case and as discussed I've opened #1390 if we need this later.

PR #1365, opened 2026-08-31
feat: experimental Rust MIR frontend

greenhat:
The scope of this PR is deliberately small so it is easier to review: it translates a single `add` function end to end and nothing more. It lays the bedrock — the crate layout, the driver, the translator skeleton, and the test approach — on which the next stages build (see `frontend/rust-mir/roadmap.md`).

## What this adds

A new experimental crate `midenc-frontend-rust-mir` that reads Rust MIR and builds HIR from it directly, without the WebAssembly path. The frontend translates every local, non-generic function; every unsupported MIR construct is a hard error that names the construct. Tests evaluate the translated function with the HIR interpreter (`add(1, 2) == 3`), both in the raw form and after SSA recovery.

The branch also fixes a real bug in the shared `Local2Reg` pass, found by this frontend's output pattern: promoting chained local-to-local copies left references to erased values, and the verifier did not catch it. The fix comes with a regression test that fails without it.

## Trade-offs

**How the MIR is obtained.** The frontend links against rustc and runs the compiler in-process through `rustc_public`, the stable-MIR API, behind a standalone driver. rustc stops after metadata (`--emit=metadata`), so the MIR stays in memory and no MIR text is parsed. The alternative — a rustc codegen-backend dylib, which would give us the compiler's own mono-item collector for whole-program discovery — was deliberately not taken: it inverts the control flow (cargo would drive midenc instead of midenc calling a frontend), couples us to the unstable `CodegenBackend` trait, and adds a toolchain-parity dylib to distribute. The translator consumes only `rustc_public` types, so this decision is revisitable at the whole-program stage without rewriting the translator; the roadmap records the decision point.

**No SSA construction in the frontend.** Each MIR local becomes a local slot (definitions store, uses load, no phi nodes), and the existing `local2reg` pass recovers SSA. This keeps the translator trivial at the cost of depending on that pass — which is what surfaced the `Local2Reg` bug above.


djolertrk (commented):
Nice @greenhat! How structural CF is being represented at MIR level? or it supports unstructured one only?

greenhat:
Rebased.

PR #1372, opened 2026-09-04
feat: calling procedure roots stored in a storage slot (`hir.dyncall`)

### LINKED ISSUE #1352: Let account component store one of its public procedure roots in a storage slot
from https://github.com/0xMiden/protocol/pull/3527#discussion_r3817951211

> Could we not do this already via the component metadata? For example, we could have a new storage schema type - something like `miden::protocol::stored_procedure` and then we'd be able to get all storage slots that have this type?
> 
> In TOML it could look something like this:
> 
> ```toml
> [[storage.slots]]
> name = "authority_procedure"
> type = "miden::protocol::stored_procedure"
> ```

greenhat:
Closes #1352.

An account component can now keep the MAST root of another component's procedure in a storage slot and call through it. The target is chosen when the account is deployed or its storage is updated, not when the caller is compiled, so a component can dispatch to an authority, a policy or a hook that is swapped without rebuilding the caller. The call runs in a fresh context through the VM's `dyncall` instruction, exactly like a compile-time `call` to a sibling component.

## Usage

A slot is declared with the new `StoredProcedure<fn(..) -> R>` value type. The `fn` type is the procedure's signature; the macro generates a call trait per slot, so the stored root is invoked like a method:

```rust
use miden::{component, component_storage, Felt, StorageValue, StoredProcedure, Word};

#[component_storage]
struct DispatcherStorage {
    #[storage(description = "root of a procedure incrementing a counter under a key")]
    increment: StorageValue<StoredProcedure<fn(key: Word) -> Felt>>,
    #[storage(description = "root of a procedure adding a delta to a counter under a key")]
    add: StorageValue<StoredProcedure<fn(key: Word, delta: Felt) -> Felt>>,
}

#[component]
trait Dispatcher {
    #[account_procedure]
    fn dispatch(&mut self, key: Word, delta: Felt) -> Felt;
}

#[component]
impl Dispatcher for DispatcherStorage {
    fn dispatch(&mut self, key: Word, delta: Felt) -> Felt {
        let after_increment = self.increment.get().call(key);
        self.add.get().call(key, delta) + after_increment
    }
}
```

The host writes the roots; the guest has no constructor for them. With the protocol's storage API a slot is seeded from a package's lifted export:

```rust
init_storage.insert_value(StorageValueName::from_slot_name(&slot), lifted_export_root(&counter_package, "increment-count"));
```

Calling a slot that was never populated traps with a dedicated error code. Signatures may use Miden core types and WIT primitives, with at most 12 flat argument values and 12 argument field elements (11 when the result is returned through a pointer). The target may be a Rust component or a hand-written MASM procedure, as long as its stack interface matches the declared signature; the end-to-end tests cover both, including a dispatch to `rbac::has_role` from the standards library and slots with `bool`, six-`u64`, unit and `Word` signatures.

## How it works

`#[component_storage]` rewrites each `StoredProcedure` slot into a marker type, emits a `<Field>Call` trait, and generates one hidden inline WIT world per storage struct whose imports are named `dyncall-<field>` and take the root `word` as their first parameter. The frontend recognises that prefix and lowers the import to the new IR op instead of linking it, so nothing about the target has to be known at compile time. Generated WIT names are escaped, so parameters named like WIT keywords work. The `dyncall-` prefix is reserved: dependency interfaces, component methods and note constructors using it are rejected by the macros. `MIGRATION.md` documents the new surface.

## IR changes

New `hir.dyncall` op, the dynamic counterpart of `hir.call`: the callee is a MAST root given as four field-element operands, followed by the arguments, under the component-model calling convention:

```
%result = hir.dyncall [%r0, %r1, %r2, %r3](%a, %b) : extern("component-model") (felt, u32) -> (u32);
```

The MASM lowering spills the root to a reserved word just above the guest address space (a `ReservedCell` registry now names that word and the heap-info word next to it), schedules the arguments, and issues `dyncall` on the reserved address. Because the root shares the 16-element operand stack window with the arguments, legalization bounds root plus arguments at 16 elements, mirroring the index-plus-arguments rule of `hir.exec_indirect`.

## Not in this PR

- The component metadata does not carry the slot's signature yet; the slot is emitted as a plain `word` until the protocol registers a stored-procedure type.


greenhat:
Rebased.

PR #1379, opened 2026-09-08
docs: update internal docs, add simple AGENTS.md

bitwalker:
Largely just some simple documentation tweaks. There are two bigger/questionable changes here:

* The introduction of an `AGENTS.md` - I'm uncertain whether it is a good idea or not, I haven't ever needed one myself, but both for external contributions, and as a way to avoid repeating ourselves internally, it seems like a really simple one might be worth exploring. Let me know what you think. @djolertrk suggested it, so I'm tagging him as well in case he has any feedback.
* The introduction of an explicit AI tool usage policy. This is largely meant to dissuade the slew of drive by PRs we get, but also as a way to encode our own guidelines around how we use these tools. Feel free to propose changes or argue against what I've drafted here - but I think it strikes the right balance (it is inspired by LLVM's policy, just vastly simplified to the core principles).


greenhat (changes requested):
The docs are looking good, but the AGENTS.md and CLAUDE.md files are __always__ loaded into the context. Meaning that my agents will always be loading them into the context as well, and I very much don't want that.

CONTRIBUTING.md
```diff
@@ -63,52 +67,86 @@ i.e. this branches state:
...
+publishing a partial baseline.
+```
+
+Contributors should ensure their PRs are comprised of no more than a single commit, but if necessary, multiple commits are permitted so long as they adhere to the following rules:
+
+- Each commit is logically distinct from the others
+- Each commit is valid on its own (i.e. tests pass and lints are clean)
+- No "oops typo" or "addressing review feedback" commits.
+
+Contributors are encouraged to squash commits and force push on their PR branches to uphold those rules. If you do not do so, a maintainer may choose to manually merge your PR by rewriting your commits themselves, or by squash-merging the PR and collapsing the entire branch into a single commit. If a maintainer does not have time to do this however, then your PR may linger unmerged until such time as someone gets around to doing it for you.
```
mooori, line 111: nit: Especially with new external contributors, I'd prefer squash-merging instead of them force-pushing. That makes it easier to do only partial reviews after comments have been addressed.

CONTRIBUTING.md
```diff
@@ -63,52 +67,86 @@ i.e. this branches state:
...
+- No "oops typo" or "addressing review feedback" commits.
+
+Contributors are encouraged to squash commits and force push on their PR branches to uphold those rules. If you do not do so, a maintainer may choose to manually merge your PR by rewriting your commits themselves, or by squash-merging the PR and collapsing the entire branch into a single commit. If a maintainer does not have time to do this however, then your PR may linger unmerged until such time as someone gets around to doing it for you. 
+
+TIP: If you want your PR merged quickly, then keeping up good commit hygiene makes it much more likely that will happen.
 
 ### Code Style and Documentation
 
-- For documentation in the codebase, we follow the [rustdoc](https://doc.rust-lang.org/rust-by-example/meta/doc.html) convention with no more than 100 characters per line.
+We provide `cargo make` tasks for all of our primary checks:
```
mooori, line 117: nit: might also mention `test` tasks

CONTRIBUTING.md
bitwalker: For larger PRs, yes, we do not want changes squashed/force-pushed until the end of review. I would argue this doesn't even apply to new contributors, because we should not want them even submitting such PRs to begin with (and that's maybe something that should be included in these guidelines as well).

For small PRs, it is better to view a complete diff IMO, otherwise it becomes difficult to reason about what parts of the code that the diff shows as "unchanged" are actually part of a previous commit in the same PR, and what is in the base branch. I pretty much universally do my reviews on the full diff for PRs under 1k lines for this reason. That is of course my preference though.

If both you and @greenhat agree though, we can change this. I'd rather have us all generally be on the same page about how contributors interact with the project.

greenhat: I'd like to be able to see the code changes that address my comments.

docs/testing-guide.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Best Practices
+
+* DO test using multiple methodologies, when doing so tests unique properties that any single methodology would fail to capture (e.g. a transformation pass might have unit tests for helper functions used by the pass internally, as well as lit tests that exercise the pass against specific IR shapes).
+* DO add tests when changing existing code that lacks adequate coverage
+* DO migrate old assertion-based tests which build IR manually to lit tests, if making changes to those tests. This makes those tests easier to maintain going forward.
+* DON'T change existing test expectations or semantics without justifying those changes explicitly in your pull request.
+* DON'T write tests directly in the SDK crates under `sdk/` - testing of these crates requires compilation to Miden Assembly, and thus require you to write an integration test that builds against the SDK and calls the functions you wish to test.
```
greenhat, line 48:
```suggestion
```

We have a lot of tests in the crates `sdk/` folder (for macros, `field-repr`, etc.).

greenhat (approved):

docs/testing-guide.md
bitwalker: Maybe we should reword this to be specifically the crates that only get compiled to Miden Assembly? The others you mentioned don't fall into that category (they are build-time deps, and so run on the host).

What I want to avoid is agents writing tests in a `#[cfg(test)]` module in a crate that is not designed to ever compile to/run on the host architecture (and so executing such tests would either be misleading, i.e. false-positives/false-negatives, or fail because they reference linker stubs that are unreachable). That basically covers all of the SDK crates that aren't macro crates.

greenhat: > Maybe we should reword this to be specifically the crates that only get compiled to Miden Assembly? 

This might work:

```suggestion
* DON'T write tests in the SDK crates under `sdk/` targeting Miden - testing of these crates requires compilation to Miden Assembly, and thus require you to write an integration test that builds against the SDK and calls the functions you wish to test.
```

>The others you mentioned don't fall into that category (they are build-time deps, and so run on the host).

Some of them are crates for the offchain code (`field-repr` family, `wasm-metadata`).

> What I want to avoid is agents writing tests in a `#[cfg(test)]` module in a crate that is not designed to ever compile to/run on the host architecture (and so executing such tests would either be misleading, i.e. false-positives/false-negatives, or fail because they reference linker stubs that are unreachable). 

Hmm, I've never had this issue.

PR #1384, opened 2026-09-09
agents: add commit hygiene agent skill, update AGENTS.md

greenhat:


AGENTS.md
```diff
@@ -1,7 +1,20 @@
-This file is your map to authoritative, versioned guidance for contributing to this project.
+## Testing.
 
-Follow each link below when its task condition applies.
+After finishing the implementation (code changes) always build and test the workspace with `cargo make test-all` and fix the errors.
```
bitwalker, line 3: I often explicitly _don't_ want this.

AGENTS.md
```diff
@@ -1,7 +1,20 @@
...
 
-- AI-assisted contributions: [AI Tool Policy](docs/ai-tool-policy.md)
-- Submitting pull requests or issues: [Contribution Guidelines](CONTRIBUTING.md)
-- Writing tests: [Testing Guidelines](docs/testing-guide.md)
+When running `cargo test` or `nextest` set `CARGO_TARGET_DIR` env var to the workspace's `target` folder.
+
+After you are finished run the following tasks when applicable:
```
bitwalker, line 7: Same here - I'd want this to be something done before an LLM submits a PR, not something it does as soon as it finishes making any individual change. There isn't any particular downside to running these things, but it does take a lot of time, so I generally avoid running any of these until I'm preparing to clean up my branch and submit a PR.

That's obviously personal preference/workflow-specific - but I'd suggest we only mention these tasks as the preferred way to check for the specific things those tasks check for; and leave it up to the discretion of the AI model whether it wants to do them or not.

AGENTS.md
```diff
@@ -1,7 +1,20 @@
...
+- `cargo make format-rust` to ensure Rust code is formatted
+- `cargo make clippy` to ensure code passes the set of lints we have enabled, pass `--fix` to automatically fix issues when possible
+- `cargo make unused` to identify any unused dependencies/items that should be removed
+- `cargo make rustdocs` to ensure Rust-based documentation comments build
+- `cargo make docs` to ensure our Docusaurus-based documentation pages build
+
+# Migration instructions on Miden SDK changes
+
+Any user-facing changes in the crates in the `sdk` folder should be analyzed and migration instruction should be provided if needed in sdk/sdk/MIGRATION.md.
```
bitwalker, line 16: I don't want this to be handled by external contributors, and especially LLM-generated contributions, unless we're the ones doing it.

I wonder if maybe it makes sense to mention this alongside the note on files not to touch, and specify _unless explicitly asked to do so_. That way it doesn't interfere with user instructions to make such changes, but also tells the model not to do it unless asked.

AGENTS.md
```diff
@@ -1,7 +1,20 @@
-This file is your map to authoritative, versioned guidance for contributing to this project.
+## Testing.
 
-Follow each link below when its task condition applies.
+After finishing the implementation (code changes) always build and test the workspace with `cargo make test-all` and fix the errors.
 
-- AI-assisted contributions: [AI Tool Policy](docs/ai-tool-policy.md)
-- Submitting pull requests or issues: [Contribution Guidelines](CONTRIBUTING.md)
-- Writing tests: [Testing Guidelines](docs/testing-guide.md)
+When running `cargo test` or `nextest` set `CARGO_TARGET_DIR` env var to the workspace's `target` folder.
```
bitwalker, line 5: That'd already be the default, no?

bitwalker (commented):
I'm leaning more towards not adding the agent config stuff at all, based on the discussion in the other PR - but figured I'd follow up with my notes on your suggestions here, which I think are largely fine.

Let me know how you'd like to proceed on this front, we can always revisit later too

AGENTS.md
greenhat: Then it should not be in the AGENTS.md. We should keep here __only__ instructions that all of us use all (most of) the time.

AGENTS.md
greenhat: I agree that we should only put instructions here that we all use all the time.

FWIW, I have had this instruction forever, and I have never seen it fired after an individual change. I see it happens only after the agent finishes what I asked.

AGENTS.md
greenhat: > I don't want this to be handled by external contributors, and especially LLM-generated contributions, unless we're the ones doing it.

Yeah, I have not thought about external contributions here.

> I wonder if maybe it makes sense to mention this alongside the note on files not to touch, and specify _unless explicitly asked to do so_. That way it doesn't interfere with user instructions to make such changes, but also tells the model not to do it unless asked.

I have this instruction in my AGENTS.md so I'd prefer my agent to do it without me asking for it. So this instruction goes out as well (see the "all have" rule above).

AGENTS.md
greenhat: Oh, right. It was implemented in the testing infra lately.

PR #1329, opened 2026-08-12
feat: fix debug expression contract

djolertrk:
Addresses the https://github.com/0xMiden/compiler/issues/1317.

Depends on https://github.com/0xMiden/miden-vm/pull/3605

PR #1388, opened 2026-09-11
migrate to vm 0.33, protocol 0.17.0-rc.5

bitwalker:
This PR starts the migration to the latest VM and protocol crates - I can't complete the migration yet unfortunately, as the `miden-client` crates haven't been migrated/released yet. Once they are, we can finish up this PR and release 0.11.0 of the compiler.

@greenhat I'll leave it to you to pick up where I left off, as well as identify any issues with the migration here.


sdk/base-sys/src/bindings/note.rs
```diff
@@ -240,76 +214,57 @@ pub fn compute_storage_commitment(storage: &[Felt]) -> Word {
...
-            attachments_commitment[3],
-            ptr as *mut Felt,
-        )
-    };
-    assert_attachment_count(num_attachments);
-    unsafe {
-        commitments.set_len(num_attachments);
-    }
-    commitments
+    load_attachment_words(attachments_commitment, MAX_ATTACHMENTS_PER_NOTE)
```
mooori, line 223: The method is called `write_...` but only calls `load_attachment_words`. Looks like `load_attachment_words` still writes, though the naming is confusing

sdk/base-sys/src/bindings/note.rs
```diff
@@ -240,76 +214,57 @@ pub fn compute_storage_commitment(storage: &[Felt]) -> Word {
...
-            attachment_commitment[3],
-            ptr as *mut Felt,
-        )
-    };
-    assert_attachment_word_count(num_words);
-    unsafe {
-        attachment.set_len(num_words);
-    }
-    attachment
+    load_attachment_words(attachment_commitment, MAX_ATTACHMENT_WORDS)
```
mooori, line 235: The method is called `write_...` but only calls `load_attachment_words`. Looks like `load_attachment_words` still writes, though the naming is confusing

tests/integration-network/src/mockchain/note_asset_bindings.rs
```diff
@@ -0,0 +1,281 @@
...
+/// read of the wrong note observable. The note script mirrors the per-note asset *count* (one) in
+/// its `EXPECTED_NUM_ASSETS` constant.
+const NOTE_ASSET_AMOUNTS: [u64; 2] = [100_000, 25_000];
+
+/// Note script asserting the note asset and identity bindings against the note the host built.
+///
+/// Everything the script compares against is either fixed by the host (the issuing faucet, passed
+/// through note storage, and the asset count) or read through a second, independent binding, so a
+/// binding that returns the wrong value aborts the transaction.
+const NOTE_ASSET_BINDINGS_NOTE_SOURCE: &str = r#"
```
mooori, line 60: For storing it in a raw string the source code is quite complex. Could this at some point be converted to Rust in `tests/fixture`?

tests/integration/src/sdk/base/asset.rs
```diff
@@ -0,0 +1,146 @@
...
+        .build();
+
+    let mut test = CompilerTestBuilder::rust_source_cargo_miden(
+        cargo_proj.root(),
+        WasmTranslationConfig::default(),
+        [],
+    )
+    .build();
+
+    test.compile_package();
```
mooori, line 69: Out of scope for this PR, but in general I'm wondering: Would it make sense to convert tests which only verify that compilation succeeds to lit tests? Or are there advantages in keeping it as is?

mooori (approved):
LGTM. Not an in-depth review given the size of the diff.

midenc-session/src/libs.rs
```diff
@@ -184,7 +168,7 @@ impl LinkLibrary {
 #[cfg(feature = "std")]
 pub(crate) fn load_package_from_path(path: &Path) -> Result<Arc<Package>, Report> {
     let bytes = std::fs::read(path).into_diagnostic()?;
-    miden_mast_package::Package::read_from_bytes_unchecked(&bytes)
+    miden_mast_package::Package::read_from_bytes_trusted(&bytes)
```
huitseeker: Could we use the validating reader here? In miden-mast-package 0.29, `read_from_bytes_unchecked` skipped MAST rehashing but still decoded the manifest with `PackageManifest::read_from_safe`, which rejected procedure exports whose node was not a procedure root or whose digest did not match. In 0.33, `read_from_bytes_trusted` also skips those manifest cross-checks, and its contract limits it to packages written and retained by the same trusted system. This helper reads explicit link paths, search-path packages, and `--sysroot/lib/*.masp`, so malformed external packages that the old code rejected are now accepted. Please use `Package::read_from_bytes` here and add a loader-level malformed-export regression test.

tests/integration/src/sdk/base/asset.rs
greenhat: This would mean putting each test case in `test/fixtures` as a full cargo project. But they differ only in a few lines of code. It does not look good. If there is a way to make it a lit test without duplicating the common code, I'm open to it.

tests/integration-network/src/mockchain/note_asset_bindings.rs
greenhat: Good catch! Done in 844f203a45440d75e05b082e53dc49cc121968d7

sdk/base-sys/src/bindings/note.rs
greenhat: Good catch! Done in fa93d43027c8cf8b3da78331f95fde79abd7a707

greenhat:
@djolertrk Could you please take a look at the changes (fix) related to debug vars location in this PR?

djolertrk:
> @djolertrk Could you please take a look at the changes (fix) related to debug vars location in this PR?

Looking into it.

Will do the review of the whole change once again.

djolertrk:
> > @djolertrk Could you please take a look at the changes (fix) related to debug vars location in this PR?
> 
> Looking into it.
> 
> Will do the review of the whole change once again.

Basically I fixed most of it in the https://github.com/0xMiden/compiler/pull/1329.

codegen/masm/src/lower/lowering.rs
```diff
@@ -1630,148 +1634,443 @@ impl HirLowering for arith::Split {
...
-                    global_index: *global_index,
+        [] | [ExpressionOp::StackValue] => stack_position.map(DebugVarLocation::Stack),
+        [ExpressionOp::WasmLocal(index)]
+        | [ExpressionOp::WasmLocal(index), ExpressionOp::StackValue] => Some(
+            local_offset(*index)
+                .map(DebugVarLocation::Local)
+                .unwrap_or(DebugVarLocation::Unavailable),
+        ),
+        [ExpressionOp::FrameBase { base, byte_offset }]
+        | [ExpressionOp::FrameBase { base, byte_offset }, ExpressionOp::StackValue] => {
```
djolertrk, line 1711: `FrameBase + StackValue` should return the computed address, not dereference it. I reproduced base=100, offset=4 returning memory[104]=99 instead of 104. This predates the offset fixes; #1329 handles it, so we can bring that in afterward.

codegen/masm/src/lower/component.rs
```diff
@@ -1661,7 +1647,18 @@ impl MasmFunctionBuilder {
                 stack.push(arg as ValueRef);
             }
         }
+        // Taken from the same offsets `locaddr` is emitted with, so that debug locations and
+        // generated code cannot disagree about where a local lives.
+        let local_offsets = function
+            .iter_locals()
+            .map(|local| local.absolute_offset() as u32)
```
djolertrk: `absolute_offset()` sums all preceding local sizes, so collecting every offset here is O(n²). Could we build this table in one cumulative pass, sharing the layout calculation with `locaddr`?

djolertrk (changes requested):
Left two comments only. The remaining expression semantics fix overlaps with #1329, which I’ll rebase on top of this once this is merged.

codegen/masm/src/lower/lowering.rs
greenhat: Sounds good!

codegen/masm/src/lower/component.rs
greenhat: Done.

greenhat:
@djolertrk Thank you! I addressed your notes. Please do another round.

djolertrk (approved):
Great, lgtm! thanks!

greenhat (approved):

PR #1393, opened 2026-09-16
fix: make `cargo miden build` a dumb wrapper around `midenc` and honor `--manifest-path` in `miden build`

### LINKED ISSUE #1391: `miden build --manifest-path` is silently ignored; the project is always resolved from the cwd
## Summary

`miden build` accepts `--manifest-path`, but ignores it. Project resolution always looks for `miden-project.toml` relative to the current working directory, so building a contract from anywhere other than its own directory fails — even when `--manifest-path` points directly at a valid manifest.

Because the flag is accepted and then silently dropped, the failure surfaces as a confusing "file not found" error naming a path the user never asked for.

## How I ran into it

I was building a small counter app on testnet — the contract from `miden new`, plus a tiny web UI that calls into the Rust integration crate to read and increment the value. I wanted the integration binaries to compile the contracts themselves, which is what the scaffolded `integration/src/helpers.rs` does via `build_project_in_dir()`. That helper shells out to:

```
miden build --release -o <out.masp> --manifest-path <dir>/Cargo.toml
```

Running `cargo test` in the generated project, the shipped example test failed with an error pointing at the *integration* directory rather than the contract I had named:

```
Error:   x failed to load Miden project from
  | /…/counter-project/integration/miden-project.toml
  `-> No such file or directory (os error 2)
```

That path is just `$PWD/miden-project.toml`. The `--manifest-path` argument had no effect.

## Reproduction

Toolchain `0.16.0`, from a generated project. Run all three from `integration/`:

```console
$ cd counter-project/integration

# A — what the scaffold passes (Cargo.toml)
$ miden build --release --manifest-path ../contracts/counter-account/Cargo.toml
Error:   x failed to load Miden project from /…/counter-project/integration/miden-project.toml
  `-> No such file or directory (os error 2)

# B — the correct manifest type, absolute-resolvable path
$ miden build --release --manifest-path ../contracts/counter-account/miden-project.toml
Error:   x failed to load Miden project from /…/counter-project/integration/miden-project.toml
  `-> No such file or directory (os error 2)

# C — add --working-dir and it builds
$ miden build --release --working-dir ../contracts/counter-account \
    --manifest-path ../contracts/counter-account/miden-project.toml
    Finished `release` profile [optimized] target(s) in 1.36s
Compiled ../contracts/counter-account/target/miden/release/counter-account.masp
```

A and B fail **identically**, which is the tell: the value of `--manifest-path` is never read. Only `--working-dir` affects resolution.

Building from inside the contract directory (`cd contracts/counter-account && miden build --release`) also works, again because it changes the cwd.

## Expected

Either:

1. `--manifest-path <path>` loads that manifest and resolves the project relative to it, independent of the cwd (this is how `cargo --manifest-path` behaves, which is what the flag name implies); or
2. if `--working-dir` is genuinely required alongside it, reject the combination with a clear error instead of silently falling back to the cwd.

Silently ignoring a supplied path is the worst of the three, because the error message names a path the user never supplied and sends them looking in the wrong place.

## Impact

This is what breaks `build_project_in_dir()` in the [project template](https://github.com/0xMiden/project-template/blob/main/integration/src/helpers.rs#L74-L90), so the example test that ships with `miden new` does not pass out of the box. Filing that half separately against `project-template`.

## Environment

```
midenup            1.0.1
Miden toolchain    0.16.0 (testnet channel)
cargo              1.96.0-nightly (e84cb639e 2026-03-21)
Platform           Linux x86_64 (WSL2)
```

## Possibly related

- #1362 — `miden new` generates a project `miden build` rejects on 0.16.0 (different symptom, also project-manifest loading)
- 0xMiden/miden-vm#3641 — Windows: project manifest path is dropped, so the workspace cannot be built (closed; same class of bug, different platform/path)

greenhat:
Close #1391

This PR makes `cargo miden build` a one-call wrapper around the `midenc` driver and fixes the project locator in `midenc`, which is what `miden build` runs under `midenup`.

The root cause was that `cargo miden build` was not dumb enough. It had its own copy of the driver: it parsed the arguments, took the `Cargo.toml` in the current directory as the project, built the session and wrote the package itself. `midenc` took the `miden-project.toml` in the current directory instead and never read `--manifest-path`. On top of that, the session treated the two manifests differently and inferred the entrypoint only for a `Cargo.toml` input. So the wrapper worked, our tests (which go through the wrapper) were green, and we missed two bugs in `miden build`: `--manifest-path` was silently dropped (#1391), and every Rust program failed with `requested target type is executable, but root module provided to assembler is library` (e.g. `examples/fibonacci`, `cd examples/fibonacci && midenc --release`).

This PR:
- adds `Midenc::exec` to `midenc-driver`, the library entry point that runs one compilation without installing a logger and returns the path of the written package (or nothing after `--stop-after`). `cargo miden build` is now one call to it with the arguments forwarded verbatim, a leading `--` included, and prints nothing itself since the driver prints the `Compiled ...` line;
- adds `Options::resolve_input` in `midenc-session`, the one rule for what a request builds, shared by the driver and `Compiler::into_session`: the input file as given; else `--manifest-path`, which must be an existing `Cargo.toml` or `miden-project.toml` (anything else is a usage error naming the flag); else the `miden-project.toml` in the working directory, or the `Cargo.toml` there when no Miden manifest exists beside it. Given both an input and `--manifest-path`, they must name the same file;
- gates the entrypoint inference on the selected target being rooted at a Rust source, instead of on which manifest file was named;
- one rule for which executable a project builds, `ProjectManifest::selected_executable`, used by the session (entrypoint), project preparation (what gets assembled) and the nested cargo build (its pre-flight check): `--target`, else `--name`, else the executable named after the package, else the sole executable, else an ambiguity error. Before, preparation picked by `--name` or the package name and ignored `--target`, so `--target other` derived `other::entrypoint` and assembled the package-named executable;
- refuses a source file input given together with `--manifest-path` with its own message (an input file next to `--manifest-path` makes sense only when it is that same manifest);
- moves the Cargo-to-Miden manifest mapping, the manifest name predicates and the workspace manifest predicate into `midenc-session`, so the session, project preparation and the nested cargo build share them (the nested build carried its own copy with a dead default);
- reports a Miden workspace root as such, with a hint to select a member, instead of `missing field package`;
- `cargo miden build --help` prints `midenc`'s help;
- `cargo miden` reads `MIDENC_TRACE` and `MIDENC_TRACE_TIMING` through the new `midenc_log::midenc_logger`, the same as `midenc` and `hir-opt`. `CARGO_MIDEN_LOG` is gone;
- updates the `midenc` and `cargo-miden` docs pages.

A relative `--manifest-path` is resolved against the directory the compiler is run from, the same as an input file and the same as `cargo` treats the flag. `--working-dir` does not move it, but it does move the default locator.

I put `miden-project.toml` before `Cargo.toml` in the default locator since the Miden manifest is the project's own. A Cargo workspace root has no Miden manifest, so it still gets the `unable to determine package from Cargo workspace root` diagnostic, and a MASM package that sits beside a virtual Cargo workspace (host-side tests) is no longer rejected.

Tests:
- `midenc-driver/tests/masm_root.rs` drives the locator on a MASM-only project through `Midenc::exec`: no arguments at all, the exact command from the issue (`-o` plus `--manifest-path` from a directory without a project), `--working-dir`, `--stop-after`, and an input file with a matching or a mismatching `--manifest-path`;
- `tools/cargo-miden/tests/miden_build.rs` builds a scaffolded Rust program through the driver with no arguments (the `fibonacci` case);
- `tests/lit/cargo-miden/build_forwarding.shtest` pins the forwarding itself: `cargo miden build --help` is `midenc`'s help, and a `--` delimiter reaches `midenc`;
- `tools/cargo-miden/tests/build_forwarding.rs` builds a scaffolded Rust program with `--manifest-path` from another directory, through the library and through the binary, and checks the wrapper prints the `Compiled` line once;
- unit tests in `midenc-session` (locator rules, `--manifest-path` validation) and `midenc-compile` (entrypoint inference per target root, workspace roots, and the selection rule run through both the session and project preparation for each `--target` / `--name` combination, so the two cannot pick different executables in silence).

Left for follow-up PRs:
- a build named by `--manifest-path` still anchors the default `--target-dir` (and the nested cargo toolchain lookup via `rustup`) on the invocation directory, not on the project directory the way `cargo` does.

Note: the first two commits carry the fix and the unification, the rest are refinements of the locator, docs and tests. Reviewing the whole diff is probably easier than going by commits.


docs/external/src/usage/midenc.md
```diff
@@ -41,19 +41,23 @@ midenc --help
...
   -p, --package <SPEC>
           Package(s) to build
 
       --manifest-path <PATH>
-          Path to the package/project manifest
+          Path to the project manifest to build
 
-          If unspecified, the compiler will create a virtual manifest for the given input
+          Either a `miden-project.toml`, or the `Cargo.toml` beside it. This names the project when
+          no input file is given; given both, they must name the same file. Absent both, the project
```
mooori: > given both, they must name the same file

I'm a bit confused assuming "both" here means `FILE` and `manifest-path`. I'd expect `FILE` to be a source file to compile, then it shouldn't be the same as `manifest-path`

midenc-compile/src/lib.rs
```diff
@@ -38,8 +38,13 @@ pub type CompilerResult<T> = Result<T, Report>;
...
 pub struct CompilerStopped(&'static str);
 
-/// Run the compiler using the provided [midenc_session::Session]
-pub fn compile(context: Rc<Context>) -> CompilerResult<()> {
+/// Run the compiler using the provided [midenc_session::Session], and report what it wrote.
+///
+/// The path is `Some` only when the assembled package was written to a file of its own — the very
+/// file the "Compiled …" line names, unless `--quiet` suppressed that line. A package emitted to
+/// standard output, a session that asked for no `masp` output, and a run stopped before assembly
+/// all leave nothing to name.
```
mooori:
```suggestion
/// The path is `Some` only when the assembled package was written to a file of its own. A package emitted to
/// standard output, a session that asked for no `masp` output, and a run stopped before assembly
/// all don't write a file.
```

nit: I feel like no need to describe what logs are printed here

midenc-compile/src/pipeline/prepare.rs
```diff
@@ -640,7 +641,18 @@ pub fn prepare_project(
     // build compiled. `Session::new` now reads the manifest's *AST* for the three facts it needs
     // before a session exists (its own documentation lists them) and builds no project at all.
     let project = Project::load(&manifest_path, source_manager).map_err(|err| {
-        err.wrap_err(format!("failed to load Miden project from {}", manifest_path.display()))
+        // A workspace root is a manifest the loader cannot load *because* it selects no package,
+        // so on the error path it gets the diagnostic that belongs to it rather than the generic
+        // "failed to load" wrapper.
```
mooori: nit: the comment kinda duplicates the error message

midenc-compile/src/pipeline/prepare.rs
```diff
@@ -678,20 +690,25 @@ pub fn prepare_project(
...
+/// a *positional* locator's name is refused and where a Cargo workspace root that selects no
+/// package is rejected. A `--manifest-path` is validated as a flag where it is given, by
+/// `Options::resolve_input`.
 fn normalize_locator(input: &InputFile) -> CompilerResult<PathBuf> {
     let file_name = input.file_name();
     match file_name.file_name() {
-        Some(name) if name.eq_ignore_ascii_case("Cargo.toml") => {
+        // Recognized through `midenc_session::is_cargo_manifest`, the same predicate
+        // `project_manifest_path` maps with: a locator this arm accepts is exactly one that
+        // mapping rewrites.
```
mooori:
```suggestion
```

nit: IMO it's clearer without comment

mooori (commented):
It seems like `Session::new` and `prepare_project` can select different executables for the same build. Should tests in `prepare.rs` always create a session *and* prepare the project to ensure they are consistent?

See for instance:

```diff
diff --git a/midenc-compile/src/pipeline/prepare.rs b/midenc-compile/src/pipeline/prepare.rs
index 3b61eff45..476e676ea 100644
--- a/midenc-compile/src/pipeline/prepare.rs
+++ b/midenc-compile/src/pipeline/prepare.rs
@@ -1332,6 +1332,10 @@ namespace = "miden:base/transaction-script@1.0.0"
             "miden-project.toml",
             MIXED_EXECUTABLES_MANIFEST,
         );
+        let mut registry = registry_with_rust();
+        registry
+            .register(crate::pipeline::frontends::MASM_FRONTEND)
+            .expect("the MASM registration should register");

         for (target, expected) in
             [("helper", None), ("prepare_fixture", Some("prepare_fixture::entrypoint"))]
@@ -1352,6 +1356,20 @@ namespace = "miden:base/transaction-script@1.0.0"
                 expected,
                 "`--target {target}` selects the executable whose root is asked about"
             );
+
+            let prepared = prepare_project(
+                &input(&manifest),
+                &session.options,
+                &registry,
+                session.source_manager.as_ref(),
+            )
+            .expect("the executable selected by the session should prepare");
+
+            assert_eq!(
+                prepared.target.name.inner().as_ref(),
+                target,
+                "preparation must select the same executable as entrypoint inference"
+            );
         }
     }
```

```sh
❯ cargo test --locked -p midenc-compile --lib the_entrypoint_follows_the_selected_executables_root -- --nocapture

running 1 test

thread 'pipeline::prepare::tests::the_entrypoint_follows_the_selected_executables_root' (51572) panicked at midenc-compile/src/pipeline/prepare.rs:1368:13:
assertion `left == right` failed: preparation must select the same executable as entrypoint inference
  left: "prepare_fixture"
 right: "helper"
test pipeline::prepare::tests::the_entrypoint_follows_the_selected_executables_root ... FAILED

failures:

failures:
    pipeline::prepare::tests::the_entrypoint_follows_the_selected_executables_root

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 302 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p midenc-compile --lib`
```

midenc-driver/tests/masm_root.rs
```diff
@@ -0,0 +1,218 @@
+//! A Miden Assembly project as the root of a `midenc` build, and the options that locate it.
```
mooori, line 1: Wondering if some of these tests could be expressed more easily as lit tests

tools/cargo-miden/tests/build_forwarding.rs
mooori, line 1: Wondering if some of these tests could be expressed more easily as lit tests

greenhat:
> It seems like `Session::new` and `prepare_project` can select different executables for the same build. Should tests in `prepare.rs` always create a session _and_ prepare the project to ensure they are consistent?

Good catch! Thanks! I unified the selection. `ProjectManifest::selected_executable` now takes the options and is used by the session (entrypoint), `prepare_project` (what gets assembled) and the nested cargo build check. The rule is `--target`, else `--name`, else the bin named after the package, else the sole bin, else ambiguous. Before, `prepare_project` ignored `--target`, so your case built the package-named bin with the entrypoint of the other one. I used your test shape: the mixed test and the two new ones run both the session and `prepare_project` for each `--target`/`--name` combination. Done in https://github.com/0xMiden/compiler/pull/1393/commits/e8da36258

docs/external/src/usage/midenc.md
greenhat: Yes, `FILE` can be a manifest too (`midenc Cargo.toml` is what `tests/support` passes), so "both" only makes sense when it's that same manifest. I reworded the help and a source file next to `--manifest-path` is now refused with its own message instead of "name different files". Done in https://github.com/0xMiden/compiler/pull/1393/commits/dacc2f856

midenc-compile/src/pipeline/prepare.rs
greenhat: Removed.

midenc-compile/src/pipeline/prepare.rs
greenhat: Done.

tools/cargo-miden/tests/build_forwarding.rs
greenhat: Good point! The two forwarding checks (`--help` is `midenc`'s help, `--` reaches `midenc`) are now a lit test in `tests/lit/cargo-miden/build_forwarding.shtest` and the Rust versions are gone. I kept the `--manifest-path` one as a Rust test since it scaffolds a program with `cargo miden new` and the Rust harness owns that setup (template path, `TEST` env, cwd lock). Done in https://github.com/0xMiden/compiler/pull/1393/commits/f88595965

midenc-driver/tests/masm_root.rs
greenhat: I'd rather keep these as Rust tests. They assert on what `Midenc::exec` returns (the package path, and `None` after `--stop-after`), which is the contract `cargo miden build` and the SDK build-script support rely on, and a lit test only sees stdout and the exit code. The forwarding checks that are about output moved to lit (see the other thread).

greenhat:
@mooori Thank you! I addressed all the notes. Please do another round.

mooori (approved):

Issue #1373, opened 2026-09-04
Reconsider location of (`cargo miden new`) template code.


Keinberger:
## Background

Opening this because the project template moved into this repo, which I ran into while updating everything to v0.16. I suppose the motivation behind that was "cargo miden new" depending on that code directly, so moving it to the compiler repo makes the release coherent.

## Downstream use of project template

**Submodule of `0xMiden/agentic-template`:**
That repo pins a specific `project-template` commit and composes it with a frontend template into one agent-facing workspace, with parent-level hooks and skills that reference the submodule's paths.

**It is a repo people clone:**
I think there is real value in the code behind `cargo miden new` being visible as a standalone open-source template rather than only reachable through a CLI flag. Not every dev may start from `cargo miden new`, and some are not aware of the tooling yet.

## Complications with the new code placement 

A git submodule cannot point at a subdirectory of another repo, so `agentic-template` cannot consume `extra/templates/project` directly.

**If the code stays here, the workarounds I can see workaround reolve around creating scripts to import the compiler template code:**

- A script in the standalone repo that pulls `project/` from the `templates/v*` release, with a version check using the sh.
- Or `agentic-template` doing that directly at setup, either the same pull or `cargo miden new`, instead of pinning a submodule.

Both work, but then the standalone repo only holds generated output, which will drift. And the second one loses the pinned commit, so agent runs against it are no longer reproducible.

## Ask

Would you be open to moving the template code back to the standalone repo?

Is keeping it standalone actually complex on your side, or would that be fine?

For me the current placement is the harder option for the `agentic-template` integration.

FYI, I opened a PR to migrate the `project-template` code for v16. Until we have finalized an approach going forward with what to do with the template code, would you mind approving the PR so that I can merge it and release v16 of the `agentic-template`? Just so that that is unblocked for now. @greenhat 

---

Tagging @greenhat and @bitwalker for request for comment.

greenhat:
Sorry for the late reply!

Moving the template back to the standalone repo would significantly complicate the compiler release workflow. Having the template in the compiler repo allows us to fully automate the compiler release, every step of it. It would not be feasible to fully automate it if the template is in a standalone repo.

I think the workaround where `agentic-template` calls `cargo miden new` is the most resilient.

Also, I want to bring up the idea of extracting skills that currently reside in the template into a separate toolchain component, as we started discussing in https://github.com/0xMiden/project-template/pull/52#issuecomment-5439693007

[...]

PR #1394, opened 2026-09-16
chore: sync project template skills with v0.16

Keinberger:
## Overview

This PR mirrors the merged project-template v0.16 skill migration into the compiler-bundled project template used by `cargo miden new`. The standalone project-template copy was updated in 0xMiden/project-template#64, and this keeps the embedded compiler scaffold from lagging behind it.

## project template code placement

This also relates to #1373, where I raised the question of whether the project template code should keep living in compiler or move elsewhere. Linking it here because this sync is another concrete example of the maintenance cost we should discuss.


greenhat (approved):
Thank you!

